PR #321, opened 2018-06-30
Created DevExtreme.MVC NPM Pack.

bbakermmc:
Packages for DevExtreme for use in MVC project.  Including JZIP for built in exports.


hikalkan:
Hi @bbakermmc 

Thank you for your pull request. I should want some changes:

* Please remove `aliases` and `clean` parts from the mapping file. Because these are only needed in the application's mapping file. Should not be repeated for each package.
* Package names should match the depending package name. Like `@abp/bootstrap` for the `bootstrap` package. Your package name is `@abp/devextreme.mvc` but there is no package named `devextreme.mvc` in the npmjs registry. I suggest you to split this into two packages: `@abp/devextreme` and `@abp/devextreme-aspnet-data`.
* Another missing part is the bundle contributors which are needed to add related styles & scripts into razor pages/views.

See https://github.com/volosoft/abp/blob/master/docs/AspNetCore/Client-Side-Package-Management.md#package-contribution

Thanks.

bbakermmc:
Whats the point of a pack if you're making it a 1 to 1 to the npm package?  Thats just dumb.  Set all the core stuff in 1 json file and be done with it.  Otherwise a Pack should be defined out as a package to implement some kind of functionality.  IE: In DevExtreme.Mvc is the collection of packages required.  For DevExtreme.Angular its a different set of file.  Otherwise in my module couldn't I just make my own packge.json and implement whatever i want, is there a way to do that instead?

For the bundling do you want a bundle.config in the pack folder.  As usual the documentation is a little vague and really only talks about implementing things in a module.  I also don't believe in implementing bundles for other people.  I don't know what "theme" they want to use.  Best Practice they should be re bundling all these into one bundle anyways.  Also DevExtreme with Jquery needs to be defined in the header, so am I suppose to make that bundle or what?  You cant always force the users with a package, sometimes if you want to use something you need to know how it works.  DevExtreme is one of them since it can be customized, maybe you dont want client side exporting, so dont bundle jzip.  Or maybe you want to dark compact version of the theme, so you need to bundle those two files instead of the default theme.

Your bundle thing needs a remove option also, I didnt see mention of one.  I would need to remove Jquery from the main bundle, to bundle it with DevExtrme since it needs to go in the header.

Does clean run for every pack?  I would want to make sure that no other pack implements another version or maps to the same directory

bbakermmc:
I only created 1 theme bundle, you can just add more to the DevExtremeStyle bundle if people want to use other ones or custom ones.

hikalkan:
@bbakermmc 
> Whats the point of a pack if you're making it a 1 to 1 to the npm package? Thats just dumb.

There are 3 reasons we create such @abp/* packages for each npm package:

1. It ensures that all ABP developers uses the same version of the depended package, so we don't have conflicts between different modules depend on the same library.
2. @abp/* package has a mapping file that knows which files should be copied from the `node_modules` folder to the `libs` folder. If we directly depend on the original package, we should manually do it. Creating @abp/* package makes it one-time and automatic for everyone.
3. @abp/* package properly defines dependencies to other @abp/* packages. So, ensures that dependencies also will be properly copied to the `libs` folder.
4. Creating bundle contributors knows how to add library scripts/styles to the file (by also ensure about dependencies).

So, all these benefits are valid even for a single package.

Why we want single package? Because developers may want to use one of the packages instead of all. If you want, you can create a meta-package that depends on multiple packages. One of the sample is shared-theme package: https://github.com/volosoft/abp/blob/master/npm/packs/aspnetcore.mvc.ui.theme.shared/package.json which only defines dependencies.

So, I think you got the point :)

>Your bundle thing needs a remove option also, I didnt see mention of one. I would need to remove Jquery from the main bundle, to bundle it with DevExtrme since it needs to go in the header.

Do you mention about rutime bundling & minification? If so, you can write a bundle contributor and add to an existing bundle and then you can manipulate the bundle files, so you can remove any item. But removing jquery will probably break the page.

>Does clean run for every pack? I would want to make sure that no other pack implements another version or maps to the same directory

It deletes the libs folder completely and re-generates when you run the gulp task. Two different package don't write into same folder normally, because every package writes to its own folder in the libs folder.

One reason we want a single @abp/* package depends on a single npm package is to ensure that different @abp/* packages do not include the same library and try to copy it to the same folder.

So, you can definitely create @abp meta-packages which depends on multiple @abp packages.

Note: For the tui-editor package (https://github.com/volosoft/abp/blob/master/npm/packs/tui-editor/abp.resourcemapping.js), we broke the rule. But we will fix this soon. Otherwise, another package depends on (for example) squire-rte will conflict on versioning and copying files.

bbakermmc:
Are the duplicates only checked per bundle?  There is where I see part of the issue, if you make a bundle that needs jquery in the header, and one before the end of html,  wont jquery be duplicated?

hikalkan:
Duplicates are prevented in a page even they are added into separate bundles.

hikalkan:
Duplicates are prevented in a page even if they are added into separate bundles. I mean if you add JQuery to two different bundle and add these bundles to the same page, JQuery is added only once. Second bundle will not include the JQuery.

hikalkan:
As I can see, this PR adds 4 @abp NPM packages and 8 bundle contributors:

------------------------------------------------------

#### NPM Packages

- @abp/devextreme-aspnet-data
- @abp/devextreme
- @abp/globalize
- @abp/jszip

#### Bundle Contributors

##### DevExtremeStyleContributor
- /libs/devextreme/dist/css/dx.common.css

##### DevExtremeLightCompactStyleContributor
[DependsOn(typeof(DevExtremeStyleContributor))]
- /libs/devextreme/dist/css/dx.light.compact.css

##### DevExtremeLightStyleContributor
[DependsOn(typeof(DevExtremeStyleContributor))]
- /libs/devextreme/dist/css/dx.light.css

##### DevExtremeScriptContributor
[DependsOn(typeof(JsZipScriptContributor))]
- /libs/devextreme/dist/js/dx.all.js

##### DevExtremeGlobalizedScriptContributor
[DependsOn(typeof(JsZipScriptContributor))]
[DependsOn(typeof(GlobalizeScriptContributor))]
- /libs/devextreme/dist/js/dx.all.js

##### DevExtremeAspNetDataScriptContributor
[DependsOn(typeof(DevExtremeScriptContributor))]
- devextreme-aspnet-data/js/dx.aspnet.data.js
- devextreme/aspnet.js

##### GlobalizeScriptContributor
[DependsOn(typeof(JQueryScriptContributor))]
- /libs/cldrjs/cldr.js
- /libs/cldrjs/event.js
- /libs/cldrjs/supplemental.js
- /libs/globalize/globalize.js
- /libs/globalize/globalize/message.js
- /libs/globalize/globalize/number.js
- /libs/globalize/globalize/currency.js
- /libs/globalize/globalize/date.js

##### JsZipScriptContributor
[DependsOn(typeof(JQueryScriptContributor))]
- /libs/jszip/jszip.js"

------------------------------------------------------

### Improvement Request

I had a few comments on the code review. Please fix them.

In addition, what do you think about not creating multiple devextreme contributors, but creating a configuration?

I believe there should be only 1 style contributor and 1 script contributor for devexpress library. Otherwise, if two different modules add different styles in the page, it will conflict.

Instead of creating `DevExtremeStyleContributor`, `DevExtremeLightCompactStyleContributor` and `DevExtremeLightStyleContributor`, we should only define a single `DevExtremeStyleContributor` like this:

````C#
public class DevExtremeStyleContributor : BundleContributor
{
    public override void ConfigureBundle(BundleConfigurationContext context)
    {
        context.Files.AddIfNotContains("/libs/devextreme/dist/css/dx.common.css");

        var options = context.ServiceProvider.GetRequiredService<IOptions<DevExtremeBundleOptions>>().Value;

        switch (options.StyleType)
        {
            case DevExtremeStyleType.LightCompact:
                context.Files.AddIfNotContains("/libs/devextreme/dist/css/dx.light.compact.css");
                break;
            case DevExtremeStyleType.Light:
                context.Files.AddIfNotContains("/libs/devextreme/dist/css/dx.light.css");
                break;
            default:
                throw new AbpException("Unknown DevExtremeStyleType: " + options.StyleType);
        }
    }
}

public class DevExtremeBundleOptions
{
    public DevExtremeStyleType StyleType { get; set; } = DevExtremeStyleType.Light;
}

public enum DevExtremeStyleType
{
    Light,
    LightCompact
}
````

Then a developer can choice styling globally like that:

````C#
services.Configure<DevExtremeBundleOptions>(options =>
{
    options.StyleType = DevExtremeStyleType.LightCompact;
});
````

Thus, all devexpress instances in the application will use the same styling.

**Comment by bbakermmc** [comment]
Again, there are multiple "themes" for devextreme, we run a custom one, we also run the standard and compact on different pages in the same module.

**Comment by bbakermmc** [comment]
Making 1 DeveExtreme "Script" isnt too bad, I guess you wouldnt use globalize if you want to only include it at a page level?  have you updated the "themeing" tutorial yet?  Is there a way to override these at a page level? or just not include and add them as needed with the built in bundle tags?

```c#
using System.Collections.Generic;
using Microsoft.Extensions.DependencyInjection;
using Microsoft.Extensions.Options;
using Volo.Abp.AspNetCore.Mvc.UI.Bundling;
using Volo.Abp.AspNetCore.Mvc.UI.Packages.DevExtremeAspNetData;
using Volo.Abp.AspNetCore.Mvc.UI.Packages.Globalize;
using Volo.Abp.AspNetCore.Mvc.UI.Packages.JsZip;
using Volo.Abp.Modularity;

namespace Volo.Abp.AspNetCore.Mvc.UI.Packages.DevExtreme
{
    [DependsOn(typeof(JsZipScriptContributor))]
    [DependsOn(typeof(GlobalizeScriptContributor))]
    [DependsOn(typeof(DevExtremeAspNetDataScriptContributor))]
    public class DevExtremeScriptContributor : BundleContributor
    {
        public override void ConfigureBundle(BundleConfigurationContext context)
        {
            context.Files.AddIfNotContains("/libs/devextreme/dist/js/dx.all.js");

            var options = context.ServiceProvider.GetRequiredService<IOptions<DevExtremeScriptBundleOptions>>()
                .Value;

            if (options.ScriptType.IncludeData)
            {
                context.Files.AddIfNotContains("/libs/devextreme-aspnet-data/js/dx.aspnet.data.js");
            }
            else if (options.ScriptType.IncludeMvcHelpers)
            {
                context.Files.AddIfNotContains("/libs/devextreme-aspnet-data/js/dx.aspnet.data.js");
                context.Files.AddIfNotContains("/libs/devextreme/aspnet.js");
            }
            else
            {
                throw new AbpException("Unknown DevExtremeScriptTypes: " + options.ScriptType);
            }

            if (options.AllowClientSideExporting) context.Files.AddIfNotContains("/libs/jszip/jszip.js");

            if (options.UseGlobalize)
            {
                context.Files.AddIfNotContains("/libs/cldr/cldr.js");
                context.Files.AddIfNotContains("/libs/cldr/event.js");
                context.Files.AddIfNotContains("/libs/cldr/supplemental.js");
                context.Files.AddIfNotContains("/libs/cldr/unresolved.js");
                context.Files.AddIfNotContains("/libs/globalize/globalize.js");
                context.Files.AddIfNotContains("/libs/globalize/message.js");
                context.Files.AddIfNotContains("/libs/globalize/number.js");
                context.Files.AddIfNotContains("/libs/globalize/currency.js");
                context.Files.AddIfNotContains("/libs/globalize/date.js");
            }
        }
    }

    /// <summary>
    ///     Define the default bundle for DevExtreme
    /// </summary>
    public class DevExtremeScriptBundleOptions
    {
        /// <summary>
        ///     Sets what js files to use for bundling
        /// </summary>
        public DevExtremeScriptTypes ScriptType { get; set; } = new DevExtremeScriptTypes();

        /// <summary>
        ///     Enables the dxGrids to export to the client (Includes JsZip)
        /// </summary>
        public bool AllowClientSideExporting { get; set; } = true;

        /// <summary>
        ///     Allows dx controls to globalize the output to users region (Includes Globalize)
        /// </summary>
        public bool UseGlobalize { get; set; } = true;
    }

    public class DevExtremeScriptTypes
    {
        public bool IncludeMvcHelpers { get; }
        public bool IncludeData { get; }
        
        /// <summary>
        ///     Default ctor that enables just the base DevExtreme js files, not including the data js files.
        /// </summary>
        public DevExtremeScriptTypes()
        {
            IncludeData = false;
            IncludeMvcHelpers = false;
        }

        /// <summary>
        ///     Override to include js data files and/or MvcHelpers
        /// </summary>
        /// <param name="includeData"></param>
        /// <param name="includeMvcHelpers"></param>
        public DevExtremeScriptTypes(bool includeData, bool includeMvcHelpers)
        {
            IncludeData = includeData;
            IncludeMvcHelpers = includeMvcHelpers;
        }

        /// <summary>
        ///     Shortcut to enable MvcHelpers
        /// </summary>
        /// <param name="includeMvcHelpers"></param>
        public DevExtremeScriptTypes(bool includeMvcHelpers)
        {
            if (!includeMvcHelpers) return;
            IncludeData = true;
            IncludeMvcHelpers = true;
        }
    }
}
```

PR #451, opened 2018-09-27
some comments as i went thru this tutorial

274188A:
ok, all good up until we add the razor page Index.cshtml (RazorPage) .  I think we are missing steps to say have your class IndexModel class inherit from 'BookStorePageModelBase'.  At lease i think that is what i assume is needed.  If this is correct we need to add this step in.  if not I'm not understanding correctly.


274188A:
for clarity I am talking about "ASP.NET Core MVC Tutorial - Part I"

PR #454, opened 2018-09-27
re-worded the part where we instruct the user to use `BookStorePageMo…

274188A:
…delBase` as base

### DIFF
--- a/docs/Tutorials/AspNetCore-Mvc/Part-I.md
+++ b/docs/Tutorials/AspNetCore-Mvc/Part-I.md
@@ -308,7 +308,7 @@ Open the `Index.cshtml` and change the content as shown below:
 <h2>Books</h2>
 ````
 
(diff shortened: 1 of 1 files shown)


docs/Tutorials/AspNetCore-Mvc/Part-I.md
```diff
@@ -308,7 +308,7 @@ Open the `Index.cshtml` and change the content as shown below:
 <h2>Books</h2>
 ````
 
-* This page **inherits** from the `BookStorePageBase` class which comes with the startup template and provides some shared properties/methods used by all pages.
+* Change the default inhertitance of the Razor View Page Model so it **inherits** from the `BookStorePageBase` class (instead of `PageModel`).  The `BookStorePageBase` class which comes with the startup template and provides some shared properties/methods used by all pages.
```
hikalkan: A small typing error: 'inhertitance'.

PR #538, opened 2018-10-10
Remove macOS directory attrib file

bruno-garcia:
"In the Apple macOS operating system, .DS_Store is a file that stores custom attributes of its containing folder, such as the position of icons or the choice of a background image."


hikalkan:
Thanks @bruno-garcia

PR #550, opened 2018-10-12
Document translation to Chinese.

maliming:
Document translation to Chinese.


maliming:
@hikalkan 
At present, the existing documents of abp have been basically translated.

hikalkan:
Thank you very much for your efforts :)

PR #556, opened 2018-10-23
Cross-platform path and project reference path.

maliming:
- Splicing path multi-platform support (win, osx, linux)
- Modify the path referenced between projects.


hikalkan:
Thank you @maliming :)

PR #594, opened 2018-11-15
Replace ServiceConfigurationContext with IServiceCollection

iAmBipinPaul:


hikalkan:
Thank you for fixing this. It seems a copy/paste error while we are introducing ServiceConfigurationContext.

PR #595, opened 2018-11-15
Remove wwwroot

iAmBipinPaul:
The path to the index.js file is `pages/books/index.js`  instead of `wwwroot/pages/books/index.js`.


hikalkan:
Thanks @iAmBipinPaul

PR #603, opened 2018-11-20
update SMSMessage Constructor

Nokecy:
SMSMessage  Properties is Read only


hikalkan:
Yes, it is read only. However, you won't need to set it. You will just fill items inside the dictionary after creating the entity. Example:

````c#
smsMessage.Properties["foo"] = "bar";
````

PR #607, opened 2018-11-25
incorrect file name

marcelo-maciel:
index.cshtml instead of index.html


hikalkan:
Thanks @marcelo-maciel for fixing this.

PR #612, opened 2018-11-28
Adjusting menu off screem problem

marcelo-maciel:
Before and after below.

image


hikalkan:
Thanks @marcelo-maciel for the fix.

PR #613, opened 2018-11-28
Translate the Localization Text Files

maliming:


hikalkan:
Thank you @maliming :)

PR #615, opened 2018-11-30
Adding Portuguese Translation

marcelo-maciel:
Adding Portuguese translation


hikalkan:
Thanks a lot @marcelo-maciel

hikalkan:
It would be very appreciated if you can also translate documents of modules. See all localization files:

https://abp.io/documents/abp/latest/Contribution/Localization-Text-Files

Especially, account and identity modules are used by the startup template.

Thanks a lot.

PR #628, opened 2018-12-10
Translate modules/template to PT-BR

marcelo-maciel:
Translate modules/template to PT-BR


hikalkan:
Thanks a lot @marcelo-maciel

PR #640, opened 2018-12-11
Update Index.cshtml

hitaspdotnet:


hikalkan:
`PersonInput` is a property in the PageModel class and its type is `PersonModel`. So, no error on current sample.

PR #642, opened 2018-12-12
Translate Background Jobs and improve other documents

maliming:
Translate Background Jobs and improve other documents


hikalkan:
Thanks a lot @maliming. You are awesome!

PR #646, opened 2018-12-13
Avoiding break of style on pages with big texts

marcelo-maciel:
Before:
image

After:
image


armgnunlu:
Thank you! @marcelo-maciel

PR #652, opened 2018-12-14
fix Duplicate key (AuditLogId、AuditLogId1)

Nokecy:
image


hikalkan:
Thanks for fixing. I will check & merge it.

PR #673, opened 2018-12-20
Add indentityUser create method extraProperties property and Fix Localization problem on My Profile modal

Nokecy:


hikalkan:
But please send the localization fix as a separated PR.

We don't like to expose ExtraProperties directly to the clients, because it may be a security risk. We may store sensitive data in the ExtraProperties and we only want some of them to available to clients.

hikalkan:
So, I'm closing this now. But please send the localization fix as a separated PR.
Thank you for your time.

Nokecy:
@hikalkan  Can I know how to use ExtraProperties?   thanks！

hikalkan:
It's just a dictionary. Get an entity from database, set a new value to dictionary and save again. That's all.

PR #699, opened 2018-12-28
Contributor misspelled.

maliming:
Built and tested on the local computer all passed.


hikalkan:
Thanks a lot @maliming for this fix.

Issue #708, opened 2019-01-03
Create a separated module for Identity AspNetCore integration


hikalkan:
Package name is `Volo.Abp.Identity.AspNetCore`.

To fix existing applications;

1. Add `Volo.Abp.Identity.AspNetCore` package to your web layer (that is hosting your application).
2. Add `AbpIdentityAspNetCoreModule` dependency to your module class (like ``[DependsOn(typeof(AbpIdentityAspNetCoreModule))]``)

PR #714, opened 2019-01-04
Apply #708 for Account Module.

maliming:
Apply #708 for Account Module.

http://ci.volosoft.com:5480/blue/organizations/jenkins/abp/detail/abp/630/pipeline/#step-19-log-614


hikalkan:
Thanks @maliming

PR #717, opened 2019-01-06
Add CacheOptions implement per cache item and global defaults. 

sailucheng:
- [1 ]  add Cache options  for per cache item and global default 
- [2 ]  fixed cache item default name.  there was a bug here.


hikalkan:
Thanks a lot @sailucheng for your contribution. I will review it soon.

PR #725, opened 2019-01-10
Fix switch language cookie expires error

Nokecy:


hikalkan:
Thanks @Nokecy for your contribution, I will review it.

framework/src/Volo.Abp.AspNetCore.Mvc/Volo/Abp/AspNetCore/Mvc/Localization/AbpLanguagesController.cs
```diff
@@ -1,18 +1,42 @@
...
-        public IActionResult Switch(string culture, string uiCulture = "") //TODO: Implement return URL
+        public IActionResult Switch(string culture, string uiCulture = "", string returnUrl = "")
         {
-            //TODO: Check allowed languages and so on...
+            if (!GlobalizationHelper.IsValidCultureCode(culture))
+            {
+                throw new AbpException("Unknown language: " + culture + ". It must be a valid culture!");
+            }
+
+            if (!GlobalizationHelper.IsValidCultureCode(uiCulture))
```
hikalkan: uiCulture can be empty, it's allowed but this check throws ex in this case.

framework/src/Volo.Abp.AspNetCore.Mvc/Volo/Abp/AspNetCore/Mvc/Localization/AbpLanguagesController.cs
```diff
@@ -1,18 +1,42 @@
...
+            {
+                throw new AbpException("Unknown language: " + uiCulture + ". It must be a valid culture!");
+            }
+
+            string cookieValue = CookieRequestCultureProvider.MakeCookieValue(new RequestCulture(culture, uiCulture));
+
+            Response.Cookies.Append(CookieRequestCultureProvider.DefaultCookieName, cookieValue, new CookieOptions
+            {
+                Expires = Clock.Now.AddYears(2),
+                HttpOnly = true
```
hikalkan: No need to `HttpOnly`. Because;

1. It's not a secret. Clients can safely read this.
2. We may need this in the client side too.

framework/src/Volo.Abp.AspNetCore.Mvc/Volo/Abp/AspNetCore/Mvc/Localization/AbpLanguagesController.cs
```diff
@@ -1,18 +1,42 @@
...
+            });
+
+            if (!string.IsNullOrWhiteSpace(returnUrl))
+            {
+                return Redirect(returnUrl);
```
hikalkan: Return URL should be checked for security. Example: https://github.com/abpframework/abp/blob/master/modules/account/src/Volo.Abp.Account.Web/Pages/Account/Login.cshtml.cs#L70

Nokecy:
see #732

PR #730, opened 2019-01-12
fixed an error to restore defaults after setting Global values twice

Nokecy:


Nokecy:
After calling the SettingManager.SetGlobalAsync method twice, the value becomes the default value.

hikalkan:
Thank you @Nokecy for finding the bug. However, this solution is a workaround. Instead, I fixed it by https://github.com/abpframework/abp/commit/fcbd984d3642b0fca64ed1bead31892c2b00fcdb

I will already work on the SettingManager to refactor & add more tests.

PR #752, opened 2019-01-23
 Fix IdentityServer Cors Policy

Goxiaoy:
We need to query database to add cors header according to IdentityServerOptions.Cors config


modules/identityserver/src/Volo.Abp.IdentityServer.Domain/Volo/Abp/IdentityServer/Clients/IClientCorsOriginRepository.cs
```diff
@@ -0,0 +1,19 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Volo.Abp.Domain.Repositories;
+using Volo.Abp.IdentityServer.Clients;
+
+namespace Volo.Abp.IdentityServer
+{
+    public interface IClientCorsOriginRepository:IBasicRepository<ClientCorsOrigin>
```
hikalkan: We should not define repositories for non-aggregate root entities because they are not collections in the MongoDB, hence will not work.

modules/identityserver/src/Volo.Abp.IdentityServer.MongoDB/Volo/Abp/IdentityServer/MongoDB/MongoClientCorsOriginRepository.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Volo.Abp.IdentityServer.Clients;
+
+namespace Volo.Abp.IdentityServer.MongoDB
+{
+    public class MongoClientCorsOriginRepository : MongoDbRepository<IAbpIdentityServerMongoDbContext, ClientCorsOrigin>, IClientCorsOriginRepository
+    {
```
hikalkan: This source code does not compile at all.

hikalkan:
I am rejecting this PR, sorry. But I will implement it myself. 
Thanks for your contribution. I didn't know that we should implement ICorsPolicyService.

hikalkan:
Follow this: https://github.com/abpframework/abp/issues/826

PR #766, opened 2019-01-28
update docs-nav.json

Y2zz:


Y2zz:
@hikalkan I added the missing link in the document navigation

PR #771, opened 2019-01-30
Translate Microservice Architecture document

maliming:


ismcagdas:
Thanks a lot @maliming 😄

PR #819, opened 2019-02-18
Translating PT-BR of #818

marcelo-maciel:
Translating PT-BR of #818


hikalkan:
You're very fast! Thanks :)

PR #855, opened 2019-03-03
use Pomelo.EntityFrameworkCore.MySql

Y2zz:
close #854 
solved #853

PR #856, opened 2019-03-04
change mysql ef provider to pomelo

Eilyyyy:
pomelo provider is better


hikalkan:
merged https://github.com/abpframework/abp/pull/855

PR #877, opened 2019-03-14
Fixed add new tags bug

Nongzhsh:
Fixed an bug where adding a new tag was invalid when updating a post


hikalkan:
Thanks for fixing this.

PR #880, opened 2019-03-16
Fix for [Serializable] Attribute in ApplicationFeatureConfigurationDto and added unit tests

brunobertechini:
In Microservices Demo, when running PublicWebSite service or BackendAdmin app there is an error in the logs that ApplicationFeatureConfigurationDto is not serializable.

This class was added without the attribute.

I also have added one very simple approach to check all classes in the Mvc.Contracts project to be serializable (we can extend later for other projects/patterns)


hikalkan:
Thanks @brunobertechini for pointing this.

I fixed it by https://github.com/abpframework/abp/commit/287725d4d4d4b7a8580b80931af207aed7bcf3b3

>I also have added one very simple approach to check all classes in the Mvc.Contracts project to be serializable (we can extend later for other projects/patterns)

I didn't like this implementation, so didn't merge your PR, sorry :) We may find another way in the future if needed.

PR #932, opened 2019-03-29
fix #931 The dictionary configuration key uses the string name of the variable.

### LINKED ISSUE #931: RemoteServiceConfiguration Key Set Issue
``` public string BaseUrl
        {
            get => this.GetOrDefault(nameof(BaseUrl));
            set => this[BaseUrl] = value;
     }
```
when set the dictionnary's key must be **nameof(BaseUrl)**

maliming:


hikalkan:
Thanks a lot :)

PR #952, opened 2019-04-03
Docking with third-party OAuth services for ease of rewriting CurrentUser

ddrsql:


hikalkan:
Hi,

Thank you for your contribution. I merged it. However, after a deeper investigation, I had to rollback it partially with commit https://github.com/abpframework/abp/commit/f3be1105086a6fea8aae921c51ec384149747256

This is by design to be compatible with AspNet Core's design.

AspNet Core's `IAuthorizationService` takes `ClaimsPrincipal` as a parameter and this principal eventually passed to ABP's PermissionValueProviders. We shouldn't ignore this parameter.

ddrsql:
Thank you for accepted, but 'Client、Role、User PermissionValueProvider ' is Overwritten by historical version.

PR #967, opened 2019-04-05
Add blog unit tests.

maliming:


maliming:
Missing test entity method. I will make up

maliming:
Entity unit tests done.

PR #997, opened 2019-04-12
Improved Chinese translation

Webdiyer:
Improved Chinese translation for Bundling-Minification.md doc file


maliming:
Thanks @Webdiyer

PR #1002, opened 2019-04-13
fix #935 Basic connection string management for tenants.

### LINKED ISSUE #935: Basic connection string management for tenants
This will manage the `AbpTenantConnectionStrings` table in practical.

image

### Tenant Management Page

* Add a new action to rows in the tenant management UI datatable, named 'Connection strings'.
* Open a modal when the action is clicked.

### The Modal

image

* "Use the shared database" is checked if this tenant has no 'Default' connection string in the database. And also the "default connection string" section is hidden.
* If user unchecks "Use the shared database", then  "default connection string" section becomes visible and can enter a string.

In this time, we will not implement conn string validation. Also, we don't implement module based conn string management. It will be done later.

### Notes

* Use Tenant aggr. root methods to manage conn strings, add new methods if needed.

maliming:

### DIFF
--- a/modules/tenant-management/src/Volo.Abp.TenantManagement.Application.Contracts/Volo/Abp/TenantManagement/ITenantAppService.cs
+++ b/modules/tenant-management/src/Volo.Abp.TenantManagement.Application.Contracts/Volo/Abp/TenantManagement/ITenantAppService.cs
@@ -1,10 +1,16 @@
 ﻿using System;
+using System.Threading.Tasks;
 using Volo.Abp.Application.Services;
 
 namespace Volo.Abp.TenantManagement
 {
     public interface ITenantAppService : IAsyncCrudAppService<TenantDto, Guid, GetTenantsInput, TenantCreateDto, TenantUpdateDto>
     {
-        //TODO: Manage connection strings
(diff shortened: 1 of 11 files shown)


hikalkan:
Thank you for the PR. I made a few refactoring on it: https://github.com/abpframework/abp/commit/21d7b0c9e86ee3027b61120f8aea95ac6f1b6654

* Call repository's Update/UpdateAsync method after updating an entity. Because, not all database APIs support change tracking & auto update. See TenantAppService changes.
* Move tenant/index.js to under Pages folder. (this was old usage, new Pages folder is supported)
* No need to DispayName attributes on TenantInfoModel since the property names matche to the values in the localization .json file. (it looks with DisplayName prefix, then without it).
* Changes ITenantAppService method names (like SetDefaultConnectionStringAsync to UpdateDefaultConnectionStringAsync) for consistency.
* Refactored Tenant.cs to make methods with 'Default' keyword just as shortcuts. GitHub changes is a bit tricky, see the source code directly.
* GetDefaultConnectionStringAsync should not be [HttpPost]
* Extracted ConnectionStringsModal.js. We don't prefer to write js inside cshtml

Some of these not well documented, sorry. Will document soon.

maliming:
Thank you very much, I have learned.

PR #1003, opened 2019-04-15
fix #953 Add support for FluentValidation.

### LINKED ISSUE #953: Add support for Fluent Validation....
- Basically https://www.nuget.org/packages/Abp.FluentValidation as described [here](https://aspnetboilerplate.com/Pages/Documents/Validating-Data-Transfer-Objects?searchKey=validation#fluent-validation)

maliming:
So far, localized messages can be found at: https://fluentvalidation.net/localization#istringlocalizer


framework/test/Volo.Abp.FluentValidation.Tests/Volo/Abp/FluentValidation/ApplicationService_FluentValidation_Tests.cs
```diff
@@ -0,0 +1,167 @@
...
+            }
+
+            public Task<string> MyMethodAsync(MyMethodInput input)
+            {
+                return Task.FromResult(input.MyStringValue + input.MyMethodInput2.MyStringValue2 +
+                                       input.MyMethodInput3.MyStringValue3);
+            }
+        }
+
+        public class MyMethodInput : AbstractValidator<MyMethodInput>
```
hikalkan: We may want to separate validators from the DTO. In this case, we should investigate validators on app startup and use them for validation.

hikalkan:
Thank you for the implementation. I will merge it then will refactor if needed.

PR #1008, opened 2019-04-16
Update BookStore sample porject based on the latest startup template

Webdiyer:


hikalkan:
Great! Thank you for the contribution.

PR #1048, opened 2019-04-25
UrlEncode post url within Redirect method in blogging module

Webdiyer:
Error "InvalidOperationException: Invalid non-ASCII or control character in header" when url contains non ASCII characters, see AspNetCore issue [4919](https://github.com/aspnet/AspNetCore/issues/4919)


Webdiyer:
May {blog.ShortName} should be url encoded either.

PR #1068, opened 2019-04-26
Fix spelling mistakes

acjh:


hikalkan:
Thanks @acjh you are perfect :)

PR #1069, opened 2019-04-28
Add EntityFrameworkCore Sqlite support.

Eilyyyy:
add sqlite support.


maliming:
Thank you for your contribution , can you add unit tests?

Eilyyyy:
It seems that there is no unit test for entityframework provider project like sqlserver, mysql.

PR #1108, opened 2019-05-09
Fix #365 Throwing/hiding exception on audit logging should be optional 

### LINKED ISSUE #365: Throwing/hiding exception on audit logging should be optional
`AuditingStore` currently hides the exception. But for some companies, it should be critical and should throws exception and rollback the main transaction.

demirmusa:
Optional parameter throwExceptions has been added to audit logging.
#365

### DIFF
--- a/abp_io/src/Volo.Utils.SolutionTemplating/Volo/Utils/SolutionTemplating/Building/Steps/SwitchEntityFrameworkCoreToMongoDbStep.cs
+++ b/abp_io/src/Volo.Utils.SolutionTemplating/Volo/Utils/SolutionTemplating/Building/Steps/SwitchEntityFrameworkCoreToMongoDbStep.cs
@@ -21,9 +21,9 @@ namespace Volo.Utils.SolutionTemplating.Building.Steps
             var lines = file.GetLines();
             for (var i = 0; i < lines.Length; i++)
             {
-                if (lines[i].Contains("ProjectReference") && lines[i].Contains("MyCompanyName.MyProjectName.EntityFrameworkCore"))
+                if (lines[i].Contains("ProjectReference") && lines[i].Contains("MyCompanyName.MyProjectName.EntityFrameworkCore.DbMigrations"))
                 {
-                    lines[i] = lines[i].Replace("EntityFrameworkCore", "MongoDB");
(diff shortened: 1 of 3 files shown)


modules/audit-logging/src/Volo.Abp.AuditLogging.Domain/Volo/Abp/AuditLogging/AuditingStore.cs
```diff
@@ -43,6 +43,9 @@ public async Task SaveAsync(AuditLogInfo auditInfo)
             catch (Exception ex)
             {
                 Logger.LogException(ex, LogLevel.Error);
+
+                if (throwExceptions)
```
hikalkan: Instead of throwing a new exception, just throw current one, like:

````csharp
throw
````

However, it is better to not use try-catch if throwExceptions is false.

File: framework/src/Volo.Abp.Auditing/Volo/Abp/Auditing/AuditingManager.cs [hunk 2]
```diff
@@ -110,23 +110,23 @@ protected virtual void MergeEntityChanges(AuditLogInfo auditLog)
             }
         }
 
-        protected virtual async Task SaveAsync(DisposableSaveHandle saveHandle)
+        protected virtual async Task SaveAsync(DisposableSaveHandle saveHandle, bool throwExceptions = false)
```
hikalkan: Instead of `bool throwExceptions = false`, `bool hideErrors = true` would be better.

modules/audit-logging/src/Volo.Abp.AuditLogging.Domain/Volo/Abp/AuditLogging/AuditingStore.cs
demirmusa: shouldn't it log exception if hideErrors =false ?

hikalkan: No need to logging, because this exception will be caught by another code (exception filter) and logged there.

PR #1111, opened 2019-05-09
Fix #365 Throwing/hiding exception on audit logging should be optional

### LINKED ISSUE #365: Throwing/hiding exception on audit logging should be optional

demirmusa:
Optional parameter HideErrors  added to audit logging


hikalkan:

PR #1120, opened 2019-05-14
Check Cli has a new version.

mehmetuken:
I try insert cli new version check and suggest to update.

### DIFF
--- a/framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/NuGet/NuGetService.cs
+++ b/framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/NuGet/NuGetService.cs
@@ -31,7 +31,7 @@ namespace Volo.Abp.Cli.NuGet
                 client.Timeout = TimeSpan.FromSeconds(30);
                 
                 var responseMessage = await client.GetAsync(
-                    $"https://api.nuget.org/v3-flatcontainer/${packageId.ToLowerInvariant()}/index.json",
+                    $"https://api.nuget.org/v3-flatcontainer/{packageId.ToLowerInvariant()}/index.json",
                     CancellationTokenProvider.Token);
                 if (!responseMessage.IsSuccessStatusCode)
                 {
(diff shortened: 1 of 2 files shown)


hikalkan:
Great! Thanks a lot.

framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/NuGet/NuGetService.cs
```diff
@@ -0,0 +1,55 @@
...
+        }
+
+        public async Task<string> GetLatestVersionAsync(string packageId)
+        {
+            using (var client = new HttpClient())
+            {
+                client.Timeout = TimeSpan.FromSeconds(30);
+                
+                var responseMessage = await client.GetAsync(
+                    $"https://api.nuget.org/v3-flatcontainer/{packageId.ToLowerInvariant()}/index.json",
```
hikalkan: This endpoint seems returning all versions including previews/betas. Is there a way of filtering only stable versions?

mehmetuken: Not filter by endpoint but i examined to nuget client is using SemanticVersion parser. Parsing version after now is prerelease here i implement this my code.
[https://github.com/NuGet/NuGet.Client/blob/3803820961f4d61c06d07b179dab1d0439ec0d91/src/NuGet.Core/NuGet.Versioning/SemanticVersion.cs#L216-L233](url)

hikalkan: OK, then we can do it in a similar way: Check if version contains `beta`, `preview` or `alpha` :)
It would be good if you update PR, otherwise I can handle it after merge.
It is good to add a parameter like `bool includePreviews=false` to the ``GetLatestVersionAsync`` method. Thanks.

mehmetuken: OK i handle it :)

hikalkan: Thanks a lot :)

PR #1126, opened 2019-05-16
Cli onfigure entity framework after installing a module

yekalkan:


framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/Commands/AddModuleCommand.cs
```diff
@@ -93,6 +97,7 @@ protected virtual string GetUsageInfo()
             sb.AppendLine("Examples:");
             sb.AppendLine("  abp add-module Volo.Blogging                      Adds the module to the current soluton.");
             sb.AppendLine("  abp add-module Volo.Blogging -s Acme.BookStore    Adds the module to the given soluton.");
+            sb.AppendLine("  abp add-module Volo.Blogging -s Acme.BookStore --SkipDbMigrations false    Adds the module to the given soluton but doesn't add-migration.");
```
hikalkan: Also add options above.

hikalkan: `SkipDbMigrations` should be written as `skip-db-migrations`

framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/Commands/AddModuleCommand.cs
```diff
@@ -104,6 +109,7 @@ public static class Solution
             {
                 public const string Short = "s";
                 public const string Long = "solution";
+                public const string SkipDbMigrations = "skip-db-migrations";
```
hikalkan: SkipDbMigrations should be its own class

PR #1131, opened 2019-05-17
Automatically install CLI dependencies if not installed before

### LINKED ISSUE #1121: Automatically install CLI dependencies if not installed before
Dependencies of the CLI:

* Yarn
* Gulp (should be globally installed).

yekalkan:
Resolves #1121

### DIFF
--- a/framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/CliService.cs
+++ b/framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/CliService.cs
@@ -103,13 +103,8 @@ namespace Volo.Abp.Cli
                     {
                         output = stdOut.ReadToEnd();
                         output += stdErr.ReadToEnd();
-
-                        stdErr.Close();
-                        stdOut.Close();
                     }
                 }
-
-                process.Close();
             }
 
             return output;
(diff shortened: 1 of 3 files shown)


framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/CliService.cs
```diff
@@ -63,6 +67,68 @@ public async Task RunAsync(string[] args)
...
+                process.Start();
+
+                using (var stdOut = process.StandardOutput)
+                {
+                    using (var stdErr = process.StandardError)
+                    {
+                        output = stdOut.ReadToEnd();
+                        output += stdErr.ReadToEnd();
+
+                        stdErr.Close();
```
hikalkan: No need to Close calls.

framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/CliService.cs
```diff
@@ -63,6 +67,68 @@ public async Task RunAsync(string[] args)
...
+            }
+        }
+
+        private string GetInstalledNpmPackages()
+        {
+            var output = "";
+
+            using (var process = new Process())
+            {
+                process.StartInfo = new ProcessStartInfo("CMD.exe")
```
hikalkan: CMD.exe can not work on a non-Windows OS. Also fix CmdHelper if you find a solution.

Issue #1134, opened 2019-05-19
CrudAppService/AsynCrudAppService Create methods should care multitenancy


hikalkan:
Current code is like that:

````charp
public virtual async Task<TEntityDto> CreateAsync(TCreateInput input)
{
    await CheckCreatePolicyAsync();

    var entity = MapToEntity(input);
    await Repository.InsertAsync(entity, autoSave: true);

    return MapToEntityDto(entity);
}
````

New code will be like:

````csharp
public virtual async Task<TEntityDto> CreateAsync(TCreateInput input)
{
    await CheckCreatePolicyAsync();

    var entity = MapToEntity(input);
    if (entity is IMultiTenant)
    {
        TryToSetTenantId(entity);
    }

    await Repository.InsertAsync(entity, autoSave: true);

    return MapToEntityDto(entity);
}
````

TryToSetTenantId uses reflection to set the TenantId (only if TenantId property has public setter), because IMultiTenant has no setter.

hikalkan:
Resolved by https://github.com/abpframework/abp/pull/1136

Webdiyer:
Awesome! thanks a lot.

PR #1136, opened 2019-05-19
#1134 Create methods should care multitenancy.

mehmetuken:
#1134

### DIFF
--- a/framework/src/Volo.Abp.Ddd.Application/Volo/Abp/Application/Services/AsyncCrudAppService.cs
+++ b/framework/src/Volo.Abp.Ddd.Application/Volo/Abp/Application/Services/AsyncCrudAppService.cs
@@ -1,10 +1,10 @@
 ﻿using System.Linq;
 using System.Threading.Tasks;
-using Microsoft.AspNetCore.Authorization;
 using Volo.Abp.Application.Dtos;
 using Volo.Abp.Domain.Entities;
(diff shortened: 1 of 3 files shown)


hikalkan:
Thanks a lot :)

maliming:
hi @mehmetuken 
Should we check if the current entity's TenantId has a value and then assign it?

hikalkan:
With current design, it is impossible that entity has a tenant id because we are calling TryToSetTenantId just after the MapToEntity. One exception can be if input has TenantId:

````csharp
var entity = MapToEntity(input);

if(entity is IMultiTenant)
{
    TryToSetTenantId(entity);
}
````

Actually `null` tenant id value is even a valid value (we may want to create a host entity with tenantid is null while working with a tenant user where currenttenant.id is not null - and in this case we are overriding it).

So, a good improvement can be like this:

````csharp
var entity = MapToEntity(input);
if(entity is IMultiTenant && !HasTenantIdProperty(input))
{
    TryToSetTenantId(entity);
}
````

This solves all problems. Can you implement this and send a PR please.

maliming:
He may use the automapper to configure TenantId. Not necessarily in the input class.
Of course this may not be common.

hikalkan:
Using automapper is not good since automapping configuration code can not access to Dependency Injection services. We can ignore it.

PR #1137, opened 2019-05-20
Check entity has a tenant id value.

mehmetuken:
#1134

### DIFF
--- a/framework/src/Volo.Abp.Ddd.Application/Volo/Abp/Application/Services/CrudAppServiceBase.cs
+++ b/framework/src/Volo.Abp.Ddd.Application/Volo/Abp/Application/Services/CrudAppServiceBase.cs
@@ -154,9 +154,14 @@ namespace Volo.Abp.Application.Services
         
         protected virtual void TryToSetTenantId(TEntity entity)
         {
+            if (entity is IMultiTenant tenantEntity && tenantEntity.TenantId.HasValue)
+            {
+                return;
+            }
+            
             var tenantId = CurrentTenant.Id;
 
-            if (tenantId == null)
+            if (!tenantId.HasValue)
             {
                 return;
             }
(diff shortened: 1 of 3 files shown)


framework/src/Volo.Abp.Ddd.Application/Volo/Abp/Application/Services/CrudAppServiceBase.cs
```diff
@@ -154,9 +154,14 @@ protected virtual void MapToEntity(TUpdateInput updateInput, TEntity entity)
         
         protected virtual void TryToSetTenantId(TEntity entity)
         {
+            if (entity is IMultiTenant tenantEntity && tenantEntity.TenantId.HasValue)
```
hikalkan: If entity is not multi-tenant, this if doesn't work. Can you implement this based on my comment: https://github.com/abpframework/abp/pull/1136#issuecomment-493872615

mehmetuken: `return entity.GetType().GetProperty(nameof(IMultiTenant.TenantId)) != null` like this ?

hikalkan: I mean no need to check tenantEntity.TenantId.HasValue at all.
Just create a HasTenantIdProperty method as explained in https://github.com/abpframework/abp/pull/1136#issuecomment-493872615
Thanks.

mehmetuken: I hope I understand correctly :) otherwise im closing this pr.

hikalkan:
Thanks a lot @mehmetuken

PR #1151, opened 2019-05-24
resolve #1147 Use serialization/deserialization for memorydb.

### LINKED ISSUE #1147: Use serialization/deserialization for memorydb
MemoryDb can be useful for tests, especially MongoDb integration tests are very slow.
But, to make MemoryDb more realistic, we should serialize entities on insert/update, deserialize on get.

We can use binary serialization for simplicity.

maliming:
asp net core built-in binary serialization needs to mark the entity as `[Serializable]`
Consider other better serialization components


hikalkan:
You're right. Marking entity serializable is not possible always. Consider to change to Json serialization please. BTW, make it changeable (like IRabbitMqSerializer we've done)

maliming:
Already updated.

hikalkan:
Excellent implementation! Thanks 👍

Issue #806, opened 2019-02-13
Use Polly for dynamic c# proxy calls.


hikalkan:

gentledepp:
Perfect!
I just wanted to add this to your list as Polly seems to be the de-facto standard for this scenario by now.
It is even used in xamarin apps

maliming:
Fixed by https://github.com/abpframework/abp/pull/1170

PR #1170, opened 2019-05-29
Use HttpClientFactory and Polly

elfmon:
#806


framework/src/Volo.Abp.Http.Client/Volo/Abp/Http/Client/DynamicProxying/DynamicHttpProxyInterceptor.cs
```diff
@@ -128,41 +128,41 @@ private async Task<T> MakeRequestAndGetResultAsync<T>(IAbpMethodInvocation invoc
...
-            {
-                var clientConfig = ClientOptions.HttpClientProxies.GetOrDefault(typeof(TService)) ?? throw new AbpException($"Could not get DynamicHttpClientProxyConfig for {typeof(TService).FullName}.");
-                var remoteServiceConfig = RemoteServiceOptions.RemoteServices.GetConfigurationOrDefault(clientConfig.RemoteServiceName);
+            var clientConfig = ClientOptions.HttpClientProxies.GetOrDefault(typeof(TService)) ?? throw new AbpException($"Could not get DynamicHttpClientProxyConfig for {typeof(TService).FullName}.");
+            var remoteServiceConfig = RemoteServiceOptions.RemoteServices.GetConfigurationOrDefault(clientConfig.RemoteServiceName);
 
-                var action = await ApiDescriptionFinder.FindActionAsync(remoteServiceConfig.BaseUrl, typeof(TService), invocation.Method);
-                var apiVersion = GetApiVersionInfo(action);
-                var url = remoteServiceConfig.BaseUrl + UrlBuilder.GenerateUrlWithParameters(action, invocation.ArgumentsDictionary, apiVersion);
+            var client = HttpClientFactory.Create(clientConfig.RemoteServiceName);
```
hikalkan: Shouldn't we dispose the client after usage?

elfmon: No.  You should not dispose of your client. To be more general, you should not dispose of anything retrieved via a DI container, which in ASP.NET Core is by default the service collection.   
References: https://stackoverflow.com/questions/50912160/should-httpclient-instances-created-by-httpclientfactory-be-disposed

Issue #1124, opened 2019-05-15
Align module & class names


hikalkan:
For example, `AbpAuditLoggingDomainModule` uses the `Abp` prefix, but `BackgroundJobsDomainModule` does not use it.
We will add prefix for standard Abp modules.

wayneboyles:
#1175 has been submitted for review.  Please let me know if there's anything you see that I missed and I'll be happy to continue to work on it.

tuyndv:
Dear @hikalkan,

Please don't use Abp prefix. Because these classes are already under Abp namespace.

Thanks

PR #1175, opened 2019-05-29
PR For Align module & class names #1124

wayneboyles:
Renamed all of the ABP Modules in the Framework solution.  Builds just fine.
Updated all of the Module names in the Application Modules solutions.  All Build just fine.
Updated all of the Module names in the ABP IO Solution.  Builds.
Updated all of the Module names in the Docs Solution.  Builds.
Updated all of the Module names in the Template Solutions.  Both Build.

The Sample projects are referencing Version 0.15 from NuGet so I haven't been able to update the module references in those projects.

There were two invalid project references in one of the template projects.  I fixed the local reference and the project builds just fine.


hikalkan:
OMG! Thank you for your huge contribution. We will review it soon :)

hikalkan:
Moved to 0.19 milestone.

xdmushui:
I don't think there's anything wrong with the framework module name having the 'Abp' prefix.

natiki:
I agree with @xdmushui. It is **much clearer** to have Abp as a prefix to all ABP modules.... For instance:

image

I know without having to look that all the modules there are ABP except for the highlighted one which I then had to go and look up. Please don't remove the Abp prefix. This is the same reason I want to be able to put all ABP data into its own [schema](https://github.com/abpframework/abp/issues/1429) - so that it is obvious what is ABP artifacts.

vnetonline:
I agree with this thread about prefix is good to have so we can see at a glance what is related to about frameqork

Issue #1155, opened 2019-05-27
Is here  a bug?


yilianhuaixiao:
https://github.com/abpframework/abp/blob/21f8edaf13b0a00cf10471dd8d53b1c3c6dd6f92/framework/src/Volo.Abp.Core/Microsoft/Extensions/DependencyInjection/ServiceCollectionCommonExtensions.cs#L48

Is there  a bug?  
 Shoudn't it be
```
  i.GetTypeInfo().IsGenericType==false
```

maliming:
It doesn't look like a bug. What problem have you encountered?

yilianhuaixiao:
if factoryInterface is a generic type definition ,  factoryInterface.GenericTypeArguments  may  return blank array.  factoryInterface.GenericTypeArguments[0]  may throw exception.
In  ServiceCollectionExtensions.cs , i saw the  follows use.
`services.AddSingleton<IServiceProviderFactory<ContainerBuilder>>(new AutofacServiceProviderFactory(configurationAction))`

maliming:
You are right, I will try to fix it.

maliming:
@yilianhuaixiao
Can you share the code for the above problem?

yilianhuaixiao:
OK. I pushed a request first time in my github life.

maliming:
I mean, did you use the IServiceProviderFactory interface or did your other code cause this problem?

Please share the relevant code.

yilianhuaixiao:
PersonAppService_Tests.cs cannot pass the test. Is this the reason?

maliming:
But the construction and testing in the CI have passed.
Did your local test fail? If yes, please share the failure information (including the stack)

yilianhuaixiao:
After debugging I get it. the code is right.

Microsoft.Extensions.DependencyInjection.IServiceProviderFactory`1[Autofac.ContainerBuilder]  

IsGenericType  is true

PR #1178, opened 2019-05-30
fixed issue#1155

yilianhuaixiao:
fixed the   #1155


maliming:
> @yilianhuaixiao
Can you share the code for the above problem?

https://github.com/abpframework/abp/issues/1155#issuecomment-496124843

yilianhuaixiao:
see  PersonAppService_Tests.cs   Cannot pass test. maybe the bug caused

PR #1193, opened 2019-06-03
Restore localization files in BookStore sample

Xeevis:
dc214152e010530e8e25b7ad4e2ff8f52c4b748c made the BookStore sample use localization files from latest template instead of sample originals. Seeing those strings are still in use and aren't localized anymore, I assume this wasn't entirely intended change? Thanks for the review.


hikalkan:
Thank you very much for your contribution.

PR #1221, opened 2019-06-11
Vietnam translations

nguyendev:


maliming:
hi @nguyendev 
Here is the latest resource information.
https://github.com/abpframework/abp/blob/master/docs/en/Contribution/Localization-Text-Files.md

PR #1294, opened 2019-06-17
Dapper Integration.

maliming:


hikalkan:
First implementation depends on EF Core. We work to remove this dependency in further releases.

PR #1295, opened 2019-06-17
Improve CLI version checking

### LINKED ISSUE #1290: Improve CLI version checking
Related to #1113

Currently version check is working with latest NuGet package and compares equality to assembly file version, this brings up problems in some scenarios.

**Like when using newer version MyGet nightly build where 0.17.0 is said to be newer than 0.18.0:**

image

My proposal is to make the following changes:

- [x] Attempt to determine exact installed version by parsing the `dotnet tool list -g` output, with fallback to existing [GetCliVersion()](https://github.com/abpframework/abp/blob/578cadddefbf4a2eb2515aad2697c50029a6e6e7/framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/CliService.cs#L151)
- [x] Compare versions using the official [NuGet.Versioning](https://www.nuget.org/packages/NuGet.Versioning) SemVer implementation to determine the newer one
- [x] If installed version is a preview, compare it against MyGet nightly builds, those are always latest previews
- [x] Preview versions cannot be updated with `dotnet tool update`. `Update` unlike `Install` currently lacks the required --version support https://github.com/dotnet/sdk/issues/2551, workaround is to uninstall && install. Should be reflected in the WRN.

Thoughts? I'm willing to give this one a shot if accepted.

Xeevis:
Newly uses 4 update channels that get selected automatically depending on the context
- **Development** (If not installed through `dotnet tool install`, won't check for updates thus speeds up startup when developing in Visual Studio)
- **Stable** (If current version number is {major}.{minor}.{patch} (eg 0.18.0) will check NuGet feed for newest stable version)
- **Prerelease** (If current version number includes hyphen {major}.{minor}.{patch}-{identifier} (eg 0.18.0-beta) will check NuGet feed for newest version including stable and prerelease versions.
- **Nightly** (If current version number includes hyphen followed by "preview" identifier, will check MyGet feed for latest nightly builds)

New version notices are tailored to respective update channels and give precise instructions how to update the specific tool that was invoked. Reason why I opted to `--tool-path` instead of `-g` in them is because tools can be installed in various locations at various versions.

`-g` would work only for tools placed in default `%USERPROFILE%/.dotnet/tools` 
`--tool-path` works simply & universally for both default and custom location tools

But it should be possible to determine if tool is in default location and provide instructions using `-g` if desired.

- [x] Tested to work on Linux
- [x] Fix for VoloNugetPackagesVersionUpdater, was appending Task object instead of version string

Feel free to ask questions or suggest edits.

image
(Reduced CLI output and using made up versions for demonstration purposes)

resolves #1290

### DIFF
--- a/framework/src/Volo.Abp.Cli.Core/Volo.Abp.Cli.Core.csproj
+++ b/framework/src/Volo.Abp.Cli.Core/Volo.Abp.Cli.Core.csproj
@@ -15,6 +15,7 @@
     <PackageReference Include="Ionic.Zip" Version="1.9.1.8" />
     <PackageReference Include="HtmlAgilityPack.NetCore" Version="1.5.0.1" />
     <PackageReference Include="Microsoft.AspNetCore.Hosting.Abstractions" Version="2.2.0" />
+    <PackageReference Include="NuGet.Versioning" Version="5.1.0" />
     <PackageReference Include="System.Text.Encoding.CodePages" Version="4.5.1" />
     <PackageReference Include="System.Security.Permissions" Version="4.5.0" />
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="3.1.0" />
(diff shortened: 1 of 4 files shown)


hikalkan:
Thanks a lot for your contribution :)

PR #1308, opened 2019-06-19
Add missing malihu-custom-scrollbar-plugin

Webdiyer:


maliming:
Can you briefly introduce it?

hikalkan:
Yes, please, why did you add this to identity module, but not added to others? Also, why we add such a library to every page instead of adding it to the global bundle? Finally, I am not sure that is needed for the basic theme.

Webdiyer:
Because permission management modal requires this script, in file permission-management\src\Volo.Abp.PermissionManagement.Web\Pages\AbpPermissionManagement\permission-management-modal.js
line 192 and 195

maliming:
Maybe @armgnunlu knows the reason.
https://github.com/abpframework/abp/commit/190cd8d465629a85faecf75b829b997010e83f62#diff-560b52073d087d414ffe45232ce24808R28

Webdiyer:
The content of the permission management modal need to be scrollable, otherwise the modal dialog will be toooo long if there are too many permissions.

hikalkan:
>Because permission management modal requires this script

Didn't know that. If so, we need to depend on this library from the shared theme module. I will check this. Thanks.

hikalkan:
Created https://github.com/abpframework/abp/issues/1309
Closing this since we don't want to add it to every page. Thanks @Webdiyer for finding that :)

PR #1336, opened 2019-06-21
We don't need "b.Property<string>(nameof(AppUser.MyProperty))" any more.

gdlcf88:


hikalkan:
We already don't need to specify table column/field name for EF Core, because it is the default value for a property if not specified.

So, why we have specified it for some pre-build modules ([example](https://github.com/abpframework/abp/blob/dev/modules/identity/src/Volo.Abp.Identity.EntityFrameworkCore/Volo/Abp/Identity/EntityFrameworkCore/IdentityDbContextModelBuilderExtensions.cs#L28))? We found that, in some advanced mappings (especially if you cross use multiple module in same database and reference each other - I can't remember the exact case now, but..) we had some problems with default naming. It was creating PropertyName_1 columns, and specifying the column name solved the problem. That's why we had specified column names for reusable modules. But, for applications you will not need it.

I think, for the startup template, we don't need such a code. I didn't understand what problem it solves.

gdlcf88:
As a beginner of Abp vNext, I added some customized properties in AppUser.cs.
For exmaple:
```
    /* This entity shares the same table/collection ("AbpUsers" by default) with the
     * IdentityUser entity of the Identity module.
     *
     * - You can define your custom properties into this class.
     * - You never create or delete this entity, because it is Identity module's job.
     * - You can query users from database with this entity.
     * - You can update values of your custom properties.
     */
    public class AppUser : FullAuditedAggregateRoot<Guid>, IUser
    {
        #region Base properties

        /* These properties are shared with the IdentityUser entity of the Identity module.
         * Do not change these properties through this class. Instead, use Identity module
         * services (like IdentityUserManager) to change them.
         * So, this properties are designed as read only!
         */

        public virtual Guid? TenantId { get; private set; }

        public virtual string UserName { get; private set; }

        public virtual string Name { get; private set; }

        public virtual string Surname { get; private set; }

        public virtual string Email { get; private set; }

        public virtual bool EmailConfirmed { get; private set; }

        public virtual string PhoneNumber { get; private set; }

        public virtual bool PhoneNumberConfirmed { get; private set; }

        #endregion

        /* Add your own properties here. Example:
         *
         * public virtual string MyProperty { get; set; }
         */

        public virtual int MyNumber { get; protected set; }

        private AppUser()
        {
            
        }
    }
```

"MyNumber" is a customized property.
However, the Add-Migration command ignore "MyNumber", unless I add the code to `ConfigureCustomUserProperties<TUser>()` method:
```
        public static void ConfigureCustomUserProperties<TUser>(this EntityTypeBuilder<TUser> b)
            where TUser: class, IUser
        {
            //b.Property<string>(nameof(AppUser.MyProperty))...
            b.Property<int>(nameof(AppUser.MyNumber));
        }
```

This is no trouble for development. But as a beginner, when the Add-Migration ignore my property, I've been looking for a solution for a long time. So I try to optimize it.

I thought you knew it, so I didn't give any description. I'm sorry about that.

hikalkan:
Adding such a code into this class makes it harder to understand for developers.

Also, it doesn't solve all problems. What if I want to customize the mapping (like adding Required or max length)?

We will create a document for this, so developers will know to add it.

So, closing this PR. If you like it, continue to use this code in your template :)

Thanks for your contribution.

PR #1337, opened 2019-06-22
fix #1335 Port replacement for migration and unit test projects.

### LINKED ISSUE #1335: Login doesn't work on tiered template
If you download a tiered template from https://abp.io/get-started, login doesn't work on main app.
To fix this, replace the URL in IdentityServerClientRedirectUris table with the URL of main web app.

maliming:
Cli and template files should be updated.


hikalkan:
OK, will publish on monday.

PR #1346, opened 2019-06-23
Fixed all references in documentation part I of PageBase and removed Base.

StevenOwenNell:


hikalkan:
Thanks a lot.

docs/en/Tutorials/AspNetCore-Mvc/Part-I.md
```diff
@@ -337,13 +337,13 @@ Open the `Index.cshtml` and change the content as shown below:
...
-@inherits Acme.BookStore.Web.Pages.BookStorePageBase
+@inherits Acme.BookStore.Web.Pages.BookStorePage
 @model IndexModel
 
 <h2>Books</h2>
 ````
 
-* This code changes the default inheritance of the Razor View Page Model so it **inherits** from the `BookStorePageBase` class (instead of `PageModel`).  The `BookStorePageBase` class which comes with the startup template and provides some shared properties/methods used by all pages.
+* This code changes the default inheritance of the Razor View Page Model so it **inherits** from the `BookStorePage` class (instead of `PageModel`).  The `BookStorePage` class which comes with the startup template and provides some shared properties/methods used by all pages.
 * Ensure that the `IndexModel` (*Index.cshtml.cs)* has the `Acme.BookStore.Pages.Books` namespace, or update it in the `Index.cshtml`.
```
cesardv, line 347: This commit was a while ago, but shouldn't this say "has the Acme.BookStore.*Web*.Pages.Books namespace..."?

hikalkan: @maliming can you check and fix the documentation if necessary.
Thanks @cesardv

maliming: hi @cesardv 

We no longer use `@inherits` in the page.
https://github.com/abpframework/abp/pull/4247

PR #1348, opened 2019-06-23
Correction to documentation

StevenOwenNell:
Removed Prepended 'Base' where necessary
Made javascript consistent
Added further description


hikalkan:
Hi @StevenOwenNell thank you for the contribution.

I am only not sure about the datatables parameters. Are you sure that it is needed. I think it seems unnecessary and makes it dirty for such a basic tutorial.

StevenOwenNell:
Hi @hikalkan - I put it there as the javascript was inconsistent with the code you were changing from, or am I mistaken?

hikalkan:
You're right. I will check if it is really needed.

Issue #1332, opened 2019-06-21
Move connection string names to constants for modules.


Fisher-Joe:
Add ConnectionName in AbpXXXManagementConsts.cs and use it at IXXManagementDbContext and XXManagementDbContext

hikalkan:
Thanks for the suggestion, will do it. It is appreciated if you send a PR.

Fisher-Joe:
Ok, I do it.

maliming:
fixed by https://github.com/abpframework/abp/pull/1351

PR #1351, opened 2019-06-24
Feature/modify modules constants

Fisher-Joe:
Move connection string names to constants for modules. 
#1332 is done.


hikalkan:
Thanks, you're great :)

maliming:
There is no problem with the code.
Just don't know if the constant should be placed in the Domain or in the Domain.Shared layer.

Fisher-Joe:
@maliming
I see,do I have to re PR?

maliming:
@Fisher-Joe 
No, it just makes the PR clearer.

Fisher-Joe:
@maliming 
Ok, I will pay attention next time.

PR #1369, opened 2019-06-25
 fix(multi-tenancy): correctly set TenantId on create in AppService

### LINKED ISSUE #1360: CreatorId and TenantId not autofilled on creation in tenants
I followed the [tutorial](https://docs.abp.io/en/abp/latest/Tutorials/AspNetCore-Mvc/Part-I) and created my first own `FullAuditedAggregateRoot` entity named `Parkingplace`. The entity also implements `IMultiTenant` to provide multi tenant support.

```cs
public class Parkingplace : FullAuditedAggregateRoot<Guid>, IMultiTenant
    {
        public const int MaxParkingplaceName = 128;

        public Guid? TenantId { get; set; }

        public string Name { get; protected set; }

        public ParkingplaceType Type { get; protected set; }

        public Guid OwnerId { get; protected set; }

        protected Parkingplace()
        {
        }

        public Parkingplace(string name, ParkingplaceType type)
        {
            Name = name;
            Type = type;
        }
    }
```

```cs
        public static void ConfigureParkSucht(this ModelBuilder builder)
        {
            Check.NotNull(builder, nameof(builder));

            /* Configure your own tables/entities inside here */
            builder.Entity<Parkingplace>(b =>
            {
                b.ToTable(ParkSuchtConsts.DbTablePrefix + "Parkingplaces", ParkSuchtConsts.DbSchema);
                b.ConfigureFullAuditedAggregateRoot();
                b.ConfigureMultiTenant();

                b.Property(parkingplace => parkingplace.Name)
                    .IsRequired()
                    .HasMaxLength(Parkingplace.MaxParkingplaceName);
                b.HasIndex(parkingplace => parkingplace.Name);
            });
        }
```

I went ahead and created the matching Dtos and an AppService for the entity. But I have the problem that the `CreatorId` and `TenantId` are not set automatically when I create the entity in a tenant.

If I use the host (default) tenant the `CreatorId` ist set correctly. The `TenantId` remains `null` which is expected.

image

Do I need to configure the entity somewhere else?

Silthus:
Fixes #1360

### DIFF
--- a/framework/src/Volo.Abp.Ddd.Application/Volo/Abp/Application/Services/AsyncCrudAppService.cs
+++ b/framework/src/Volo.Abp.Ddd.Application/Volo/Abp/Application/Services/AsyncCrudAppService.cs
@@ -109,7 +109,7 @@ namespace Volo.Abp.Application.Services
 
             var entity = MapToEntity(input);
 
-            if (entity is IMultiTenant && !HasTenantIdProperty(entity))
+            if (entity is IMultiTenant && HasTenantIdProperty(entity))
             {
                 TryToSetTenantId(entity);
             }
--- a/framework/src/Volo.Abp.Ddd.Application/Volo/Abp/Application/Services/AsyncCrudAppService.cs
+++ b/framework/src/Volo.Abp.Ddd.Application/Volo/Abp/Application/Services/AsyncCrudAppService.cs
@@ -109,10 +109,7 @@ namespace Volo.Abp.Application.Services
 
             var entity = MapToEntity(input);
 
-            if (entity is IMultiTenant && HasTenantIdProperty(entity))
-            {
-                TryToSetTenantId(entity);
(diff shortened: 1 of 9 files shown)


maliming:
hi @Silthus 
Thank you for your contribution, can you modify it together?
https://github.com/abpframework/abp/blob/42f37c5ff01ad853a5425d15539d4222cd0dab69/framework/src/Volo.Abp.Ddd.Application/Volo/Abp/Application/Services/CrudAppService.cs#L109

Silthus:
> Thank you for your contribution, can you modify it together?

Oh sorry did not see that.

Is there a reason for the check being duplicated in both services? Can I move the `if (entity is IMultiTenant && !HasTenantIdProperty(entity)) ` line directly into the `TrySetTenantId` method of the `BaseAppService`?

PR #1379, opened 2019-06-26
Resolve #1114 Use the new zip class library (SharpZipLib)

### LINKED ISSUE #1114: Change Ionic.Zip on CLI
Since it does not support .net core natively.

maliming:
Already tested under mac, the path problem has been resolved.


maliming:
It seems that Volo needs to be modified at the same time.

Http://ci.volosoft.com:5480/blue/organizations/jenkins/volo/detail/volo/2990/

hikalkan:
Thanks @maliming

maliming:
After the merge I will update the code of the volo project.

PR #1401, opened 2019-06-28
Fix inconsistent REST API urls.

maliming:


modules/identity/src/Volo.Abp.Identity.HttpApi/Volo/Abp/Identity/IdentityUserLookupController.cs
```diff
@@ -9,7 +9,7 @@ namespace Volo.Abp.Identity
     [RemoteService]
     [Area("identity")]
     [ControllerName("UserLookup")]
-    [Route("api/identity/user-lookup")]
+    [Route("api/identity/user-lookups")]
```
hikalkan: ``api/identity/users/lookup`` would be better. Lookups is not so natural. Endpoints become:

* ``api/identity/users/lookup/{id}``
* ``api/identity/users/lookup/by-username/{userName}``

What do you think?

modules/identity/src/Volo.Abp.Identity.HttpApi/Volo/Abp/Identity/ProfileController.cs
```diff
@@ -7,6 +7,7 @@ namespace Volo.Abp.Identity
     [RemoteService]
     [Area("identity")]
     [ControllerName("Profile")]
+    [Route("/api/identity/profiles")]
```
hikalkan: `/api/identity/my-profile` would be better since this service is not managing profiles, but only manages it for the current user.

modules/identity/src/Volo.Abp.Identity.HttpApi/Volo/Abp/Identity/ProfileController.cs
```diff
@@ -15,16 +16,21 @@ public ProfileController(IProfileAppService profileAppService)
...
         }
 
+        [HttpPut]
         public Task<ProfileDto> UpdateAsync(UpdateProfileDto input)
         {
             return _profileAppService.UpdateAsync(input);
         }
 
+        [HttpPost]
+        [Route("ChangePassword")]
```
hikalkan: `change-password` would be good.

modules/tenant-management/src/Volo.Abp.TenantManagement.HttpApi/Volo/Abp/TenantManagement/TenantController.cs
```diff
@@ -18,41 +19,55 @@ public TenantController(ITenantAppService service)
...
 
+        [HttpGet]
+        [Route("{id}/DefaultConnectionString")]
         public Task<string> GetDefaultConnectionStringAsync(Guid id)
         {
             return _service.GetDefaultConnectionStringAsync(id);
         }
 
+        [HttpPut]
+        [Route("{id}/DefaultConnectionString")]
```
hikalkan: `{id}/default-connection-string` would be better. same for GetDefaultConnectionStringAsync and DeleteDefaultConnectionStringAsync methods.
Want to use kebap-case wherever possible.

maliming:
Ok, I will update it.

hikalkan:
Thanks a lot.

PR #1402, opened 2019-06-28
Resolve #1397 ContentLayout supports html lang.

### LINKED ISSUE #1397: Docs should change <Title> and <html lang="..."> according to selected language
Currently both the document title and language are set to English at all times and won't be liked by search engines.

maliming:


framework/src/Volo.Abp.AspNetCore.Mvc.UI.Theme.Basic/Themes/Basic/Layouts/Application.cshtml
```diff
@@ -25,11 +25,17 @@
...
+    if (PageLayout.Content.Lang != null)
+    {
+        pageLang = $" lang={PageLayout.Content.Lang}";
+    }
 }
 
 <!DOCTYPE html>
 
-<html lang="en">
+<@($"html{pageLang}")>
```
hikalkan:
```suggestion
<html @pageLang>
```

It is not good to move fundamental html tag to a dynamic string which may break VS editor (it may not recognize the html opening tag). I suggest the above syntax.

Also, put language code inside " and ". Result will be like:

````xml
<html lang="fr">
````
````

Comment (maliming) [reply framework/src/Volo.Abp.AspNetCore.Mvc.UI.Theme.Basic/Themes/Basic/Layouts/Application.cshtml]: Ok, I update it, just want to remove the extra space if lang is empty.

Comment (hikalkan) [reply framework/src/Volo.Abp.AspNetCore.Mvc.UI.Theme.Basic/Themes/Basic/Layouts/Application.cshtml]: Then you can do this:

````xml
<html@pageLang>
````

Comment (Xeevis) [reply framework/src/Volo.Abp.AspNetCore.Mvc.UI.Theme.Basic/Themes/Basic/Layouts/Application.cshtml]: There is much simpler solution, by the time layout is rendered culture is already defined in the CultureInfo so it can be written like this and it will work universally with no other code required. Even if it were null (which shouldn't happen) Razor is smart enough to not render tags that are null. This can be done in all 3 basic theme layouts.

    <html lang="@CultureInfo.CurrentUICulture.Name">

Requires using in each layout file or even better can be specified once in the related _ViewImports.cshtml

    @using System.Globalization

File: framework/src/Volo.Abp.AspNetCore.Mvc.UI.Theme.Basic/Themes/Basic/Layouts/Application.cshtml [hunk 2]
```diff
@@ -25,17 +25,11 @@
...
-    if (PageLayout.Content.Lang != null)
-    {
-        pageLang = $" lang={PageLayout.Content.Lang}";
-    }
 }
 
 <!DOCTYPE html>
 
-<@($"html{pageLang}")>
+<html lang="@CultureInfo.CurrentCulture.Name">
```
Xeevis: It should be **CurrentUICulture**. While for the most time they are identical, there may be rare cases when they can differ. **CurrentCulture** in .NET controls number, date, currency formatting. While **CurrentUICulture** represents language of the text, which is same function as `<html lang="">` attribute.

Possible use case: User may like to stick with default English language text, but may also prefer his native formatting to better understand dates and numbers which is allowed by string.Format();`, but document should still me marked "en".

maliming: I see that there is more priority in using CurrentCulture in the framework.

Xeevis: I don't understand. In the framework CurrentUICulture is used to determine which resources to use. There is no ambiguity, CurrentCulture is not meant to determine document language and I don't see it being used for anything bar being assigned in the framework or working as fallback if CurrentUICulture is null (which I believe can't happen).

maliming: @hikalkan What do you think?

PR #1406, opened 2019-06-29
Make sure the database is created correctly

kelely:
When using mysql, if the database does not exist, it will not be executing migrate. So we need to make sure the database has been created before executing migrate.


maliming:
Don't call `EnsureCreated()` before `Migrate()`. `EnsureCreated()` bypasses Migrations to create the schema, which causes `Migrate()` to fail.

kelely:
Ohhhhh, then in the case of MySQL, you need to manually create an empty database.

maliming:
This problem should be solved by the mysql database provider you are using.

PR #1428, opened 2019-07-03
Add build-all.sh

CdLian:
I think Linux and macOS users need this script.

### DIFF
--- a/build-all.sh
+++ b/build-all.sh
@@ -1,0 +1,46 @@
+#!/bin/bash
+
+# COMMON PATHS
+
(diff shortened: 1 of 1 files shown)


hikalkan:
Powershell can run on MacOS and Linux if you install it.
It is easy for me to accept this PR, and thank you for this, but we don't want to maintain both scripts in parallel. What do you think?

CdLian:
Oh, thank you, Linux and macOS can run PowerShell scripts, the Shell script is not necessary. Now I will close the PR.

PR #1432, opened 2019-07-04
Fixed Namespace Issue

wayneboyles:
`Volo.Abp.UI.Navigation` was split into two namespaces - Volo.Abp.**Ui** and Volo.Abp.**UI** .  Updated the `Volo.Abp.Ui` namespaces to match the base namespace of `Volo.Abp.UI`


hikalkan:
Thanks a lot for fixing this.

maliming:
By the way, when we modify the content of the core framework, we also need to check whether there are problems with other projects that reference the core framework.

Issue #1455, opened 2019-07-08
Mailkit integration


yekalkan:
- [x] Volo.Abp.MailKit package
- [ ] Integrate to MVC startup templates (not module templates).

Issue #1398, opened 2019-06-28
SmtpEmailSender error when using gmail


Goxiaoy:
When I try to use SmtpEmailSender to send an e-mail with html body, then i got an error
```
[Error] Syntax error, command unrecognized. The server response was: 
System.Net.Mail.SmtpException: Syntax error, command unrecognized. The server response was: 
   at System.Net.Mail.SmtpConnection.ConnectAndHandshakeAsyncResult.End(IAsyncResult result)
   at System.Net.Mail.SmtpTransport.EndGetConnection(IAsyncResult result)
   at System.Net.Mail.SmtpClient.ConnectCallback(IAsyncResult result)
```

So I have to replace ISmtpEmailSender with my own sender

```
     context.Services.Replace(
                ServiceDescriptor.Transient<ISmtpEmailSender, MySmtpEmailSender>());
```

```
    public class MySmtpEmailSender:SmtpEmailSender
    {

        public MySmtpEmailSender(ISmtpEmailSenderConfiguration smtpConfiguration, IBackgroundJobManager backgroundJobManager) : base(smtpConfiguration, backgroundJobManager)
        {
        }

        protected override async Task SendEmailAsync(MailMessage mail)
        {
            AlternateView textView;
            AlternateView htmlView;
            mail = Fix(mail, out textView, out htmlView);
            try
            {
                using (var smtpClient = await BuildClientAsync())
                {
                    await smtpClient.SendMailAsync(mail);
                }
            }
            catch (Exception e)
            {
                throw;
            }
            finally
            {
                textView?.Dispose();
                htmlView?.Dispose();
                mail.Dispose();
            }

        }

        protected override void SendEmail(MailMessage mail)
        {
            AsyncHelper.RunSync(() => SendEmailAsync(mail));
        }


        protected MailMessage Fix(MailMessage message, out AlternateView textView, out AlternateView htmlView)
        {
            htmlView = null;
            textView = null;
            //force set encoding to utf8
            message.BodyEncoding=Encoding.UTF8;
            if (message.IsBodyHtml)
            {
                var mimeType = new ContentType("text/html");
                htmlView = AlternateView.CreateAlternateViewFromString(message.Body, mimeType);
                message.AlternateViews.Add(htmlView);
            }
            else
            {
                const string mediaType = "text/plain";
                textView = AlternateView.CreateAlternateViewFromString(message.Body, message.BodyEncoding, mediaType);
                textView.TransferEncoding = TransferEncoding.SevenBit;
[...]

PR #1463, opened 2019-07-09
Mail kit integration

yekalkan:
resolves https://github.com/abpframework/abp/issues/1455
resolves https://github.com/abpframework/abp/issues/1398


framework/src/Volo.Abp.MailKit/Volo/Abp/MailKit/AbpMailKitOptions.cs
```diff
@@ -2,8 +2,13 @@
...
 namespace Volo.Abp.MailKit
 {
-    public class AbpMailKitConfiguration : IAbpMailKitConfiguration
+    public class AbpMailKitOptions
     {
         public SecureSocketOptions? SecureSocketOption { get; set; }
+
+        public AbpMailKitOptions()
+        {
+            SecureSocketOption = null;
```
hikalkan: no need to set null since it is the default value. Remove ctor.

PR #1486, opened 2019-07-15
Use Serilog.Sinks.Async to output file logs.

### LINKED ISSUE #1479: Asynchronous Serilog file logging using Serilog.Sinks.Async
>By default, the file sink will flush each event written through it to disk. To improve write performance, specifying buffered: true will permit the underlying stream to buffer writes.
>
>The Serilog.Sinks.Async package can be used to wrap the file sink and perform all disk access on a background worker thread.

In the context of Web applications where requests generate lots of entries even on `Information` level and where blocking logging is not necessary, it might be beneficial to general ABP performance, to utilize asynchronous logging by default, especially in environments where intensive I/O operations can cause bottleneck.

Usage of [Serilog.Sinks.Async](https://github.com/serilog/serilog-sinks-async)
>Use this sink to reduce the overhead of logging calls by delegating work to a background thread. This is especially suited to non-batching sinks like the File and RollingFile sinks that may be affected by I/O bottlenecks.

    Install-Package Serilog.Sinks.Async

Replace

    .WriteTo.File("Logs/logs.txt")

With

    .WriteTo.Async(a => a.File("Logs/logs.txt")

maliming:
Resolve #1479

I only add this feature to the MVC template. For other projects, most of them are test projects. Adding this feature is not very meaningful.

We should explain this in the documentation in the future.


hikalkan:
That's great @maliming :)

PR #1495, opened 2019-07-17
Solve the problem that the tenant information cannot be obtained correctly in the route.

hubo0831:
解决了以下问题：
1. 路由租户信息使用Url解码(UrlDecode)；
2. 查询串租户信息使用Url解码(UrlDecode)；
3. 在有路径租户信息时，由于路由中间件未先注册，导致路由租户获取不到租户信息，因此要先注册终端点路由中间件(UseEndpointRouting)


maliming:
Please use English, Thank you.

hubo0831:
resolve some problems:
1. use WebUtility.UrlDecode decode route value; (we may use chinese name)
2. use WebUtility.UrlDecode decode query string value;
3. if use RouteTenantResolveContributor, register EndpointRoutingMiddleware before MultiTenancyMiddleware., or IRoutingFeature RouteTenantResolveContributor retrieve is null

maliming:
hi @hubo0831  Can you give a specific example?

hubo0831:
`internal sealed class EndpointRoutingMiddleware
{
	public EndpointRoutingMiddleware(MatcherFactory matcherFactory, CompositeEndpointDataSource endpointDataSource, ILogger<EndpointRoutingMiddleware> logger, RequestDelegate next)
	{
	}

	public async Task Invoke(HttpContext httpContext)
	{
		EndpointSelectorContext feature = new EndpointSelectorContext();
		await(await InitializeAsync()).MatchAsync(httpContext, feature);
		if (feature.Endpoint != null)
		{
			SetFeatures(httpContext, feature);
			Log.MatchSuccess(_logger, feature);
		}
		else
		{
			Log.MatchFailure(_logger);
		}
		await _next(httpContext);
	}
	private static void SetFeatures(HttpContext httpContext, EndpointSelectorContext context)
	{
		httpContext.Features.Set((IRoutingFeature)context);
		httpContext.Features.Set((IRouteValuesFeature)context);
		httpContext.Features.Set((IEndpointFeature)context);
	}
}`
.net asp.net core 2.2 default use EndpointRoutingMiddleware, IRoutingFeature is set after route match

hubo0831:
sorry, i have no sample code

hubo0831:
I use abp multitenant , use route resolve multitenant,but i can't resolve tenant information from route,i use chinese name, tenant information must UrlDecoded

maliming:
First we should not call `app.UseEndpointRouting()` in the `UseMultiTenancy `method;
If you need it, you should call it yourself before calling the `UseMultiTenancy `method.

If your tenant name has a non-English string, you can customize the `TenantResolveContributor `and replace the built-in `TenantResolveContributor`.

maliming:
Of course, we should also consider the case where the tenant name is in a non-English string.

hubo0831:
I already do so,  but I consider abp framework should solve some common problems. Route include tenant information is very common.

PR #1511, opened 2019-07-20
Remove redundant type attributes

Xeevis:
Type attributes on `<script>` and `<link ref="stylesheet">` are redundant because currently set values are the default, which leads to warnings in [the W3C Markup Validation Service](https://validator.w3.org)

image

---

Quote from [HTML 5.2 W3C Recommendation (§ 4.12.1. The script element)](https://www.w3.org/TR/html52/semantics-scripting.html#element-attrdef-script-type)

>The type attribute allows customization of the type of script represented:
>
>Omitting the attribute, or setting it to a JavaScript MIME type, means that the script is a classic script, to be interpreted according to the JavaScript Script top-level production. Classic scripts are affected by the charset, async, and defer attributes. Authors should omit the attribute, instead of redundantly giving a JavaScript MIME type.

Quote from [HTML 5.2 W3C Recommendation (§ 4.8.6.11. Link type "stylesheet")](https://www.w3.org/TR/html52/links.html#link-type-stylesheet)

>The default type for resources given by the stylesheet keyword is text/css.


hikalkan:
Thanks @Xeevis for make it clean.

PR #1523, opened 2019-07-24
Pie chart example

yekalkan:


samples/DashboardDemo/src/DashboardDemo.Application.Contracts/CountersWidgetInputDto.cs
```diff
@@ -2,10 +2,7 @@
 
 namespace DashboardDemo
 {
-    public class CountersWidgetInputDto
+    public class CountersWidgetInputDto : WidgetFilter
```
hikalkan: Remove WidgetFilter and duplicate StartDate  and EndDate, it is not a problem to duplicate here.

hikalkan: Do it in the dev branch, I've merged it.

PR #1524, opened 2019-07-24
Fix bug of AddIfNotContains extension method in AbpTagHelperAttributeListExtensions.cs

Webdiyer:


maliming:
Thanks @Webdiyer

maliming:
Did not cause any problems, just code errors.

PR #1525, opened 2019-07-24
Bootstrap daterangepicker

yekalkan:
Used in dashboard demo


hikalkan:
Should change the momentjs dependency here

PR #1545, opened 2019-07-29
Resolve #1536 Use minified version of the script/style.

### LINKED ISSUE #1536: Use minified version of the script/style instead of minification process
* This will increase bundling performance
* Also some scripts has problems with minification (https://github.com/abpframework/abp/issues/1279), so we can directly use the minified version.

This works only if there is a min.css/min.js in the same folder with the normal script file.

maliming:
When the js/css file ends with min or there is a file ending with min with the same name, splicing these files instead of Minify them again.

If it is a js file, we are making sure that the file ends with a semicolon.


maliming:
The order of the files is important, so you can't process the normal files and min files separately and then merge them.

hikalkan:
Thanks @maliming

PR #1565, opened 2019-08-01
Feature/navigation multilevel dropdown

thediaval:


npm/ng-packs/packages/core/src/lib/directives/ellipsis.directive.ts
```diff
@@ -0,0 +1,39 @@
...
+
+@Directive({
+  selector: '[abpEllipsis]',
+})
+export class EllipsisDirective implements AfterContentInit {
+  @Input('abpEllipsis')
+  witdh: string;
+
+  @HostBinding('title')
+  @Input()
```
mehmet-erim: I guess, this input is unnecessary. Can you test it?

Issue #1587, opened 2019-08-05
Add Access-Control-Expose-Headers for ABP error response


hikalkan:
@mehmet-erim says "I can't handle "_AbpErrorFormat" header". It seems we should add Access-Control-Expose-Headers to allow it. See https://stackoverflow.com/questions/51971342/cant-get-headers-of-response-in-angular-5?answertab=active#tab-top

For the simplest solution, we can modify this Access-Control-Expose-Headers header (or add if it does not exists) to add _AbpErrorFormat to the list. We can do this where we return the error response.

maliming:
ok I will try.

PR #1591, opened 2019-08-06
Add the AbpExposeHeaders extension method.

maliming:
Resolves https://github.com/abpframework/abp/issues/1587


framework/src/Volo.Abp.AspNetCore/Microsoft/AspNetCore/Cors/AbpCorsPolicyBuilderExtensions.cs
```diff
@@ -0,0 +1,12 @@
+﻿using Microsoft.AspNetCore.Cors.Infrastructure;
+
+namespace Microsoft.AspNetCore.Cors
+{
+    public static class AbpCorsPolicyBuilderExtensions
+    {
+        public static CorsPolicyBuilder WithAbpExposeHeaders(this CorsPolicyBuilder corsPolicyBuilder)
```
hikalkan: `WithAbpExposedHeaders` would be better.

hikalkan:
Also, please update the startup template (https://github.com/abpframework/abp/tree/dev/templates/app/aspnet-core/src) where we are using Cors. It should be in 2 or 3 projects.

PR #1616, opened 2019-08-12
fix #1615 Add a prefix to AutoMap*Attribute.

### LINKED ISSUE #1615: We should add a prefix to AutoMap*Attribute to prevent conflicts with AutoMaper.
from https://github.com/aspnetboilerplate/aspnetboilerplate/issues/4786

maliming:


hikalkan:
I should think on that. If AutoMapper already has these attributes, we may consider to remove ours.

maliming:
AutoMapper only provides one AutoMapAttribute, and we have multiple.

hikalkan:
I decided to remove them: https://github.com/abpframework/abp/issues/1706

>AutoMapper only provides one AutoMapAttribute, and we have multiple.

Yes, but AutoMapper's single attribute can do the same with attribute properties

PR #1631, opened 2019-08-14
Fix #1613 the namespaces of the module template AddEmbedded method.

### LINKED ISSUE #1613: the project at "templates" is wrong namespace when AddEmbedded()
this issues is not appear at hostingEnvironment.IsDevelopment()==true 
because is will use  real file but not Embedded file.
example as MyCompanyName.MyProjectName.Web.MyProjectNameWebModule.cs:
```csharp
namespace MyCompanyName.MyProjectName.Web
{
    [DependsOn(
        typeof(MyProjectNameHttpApiModule),
        typeof(AbpAspNetCoreMvcUiThemeSharedModule),
        typeof(AbpAutoMapperModule)
        )]
public class MyProjectNameWebModule : AbpModule
    {

        {
            Configure<VirtualFileSystemOptions>(options =>
            {
                //wrong namespace
                options.FileSets.AddEmbedded<MyProjectNameWebModule>("MyCompanyName.MyProjectName");
                //correct namespace
                //options.FileSets.AddEmbedded<MyProjectNameWebModule>("MyCompanyName.MyProjectName.Web");
            });
   }
}
```
as above, could that be right?

yinchang0626:


hikalkan:
Thanks @yinchang0626 
@maliming can you check if this is necessary and merge if it is.

maliming:
@hikalkan 
I saw two code styles in the modules.

The first is the AddEmbedded method and passes the baseNamespace parameter.eg: MyCompanyName.MyProjectName

The second is that AddEmbedded does not pass any parameters.

For example, when calling AddVirtualJson, use a different path because the baseNamespace above is different.

Which style should I use in the template?

maliming:
If multiple projects have their own **Localization** folder. The folder structure also needs to be modified:

Such as:
```
\Localization\ApplicationContracts\
\Localization\DomainShared\
```

Does our template need to create this structure in advance?

hikalkan:
>The first is the AddEmbedded method and passes the baseNamespace parameter.eg: MyCompanyName.MyProjectName

Passing namespace to AddEmbedded is needed when the project has a root namespace. If root namespace is empty, then you don't pass a base namespace.
The parameter to AddEmbedded indicated the root namespace matches to the root folder.

I can handle this PR, no problem ;)

PR #1634, opened 2019-08-15
Resolve #1632 MongoDB test problems & optimization.

### LINKED ISSUE #1632: MongoDB test problems & optimization
We are using Mongo2Go library to create MongoDB databases in automated tests which creates temporary MongoDB instances.

However, I see that it may reuse the same instance that is a problem especially when we run tests in parallel.

Also, I think that we can share same instance with different database names which is faster. So, we will use the code below for MongoDB tests anymore.

### OLD STYLE

````csharp
using System;
using Mongo2Go;
using Volo.Abp;
using Volo.Abp.Data;
using Volo.Abp.Modularity;

namespace MyCompanyName.MyProjectName.MongoDB
{
    [DependsOn(
        typeof(MyProjectNameTestBaseModule),
        typeof(MyProjectNameMongoDbModule)
        )]
    public class MyProjectNameMongoDbTestModule : AbpModule
    {
        private MongoDbRunner _mongoDbRunner;

        {
            _mongoDbRunner = MongoDbRunner.Start();

            Configure<DbConnectionOptions>(options =>
            {
                options.ConnectionStrings.Default = _mongoDbRunner.ConnectionString.EnsureEndsWith('/') + "MyProjectName";
            });
        }

        public override void OnApplicationShutdown(ApplicationShutdownContext context)
        {
            _mongoDbRunner.Dispose();
        }
    }
}
````

### NEW STYLE

````csharp
using System;
using Mongo2Go;
using Volo.Abp.Data;
using Volo.Abp.Modularity;

namespace MyCompanyName.MyProjectName.MongoDB
{
    [DependsOn(
        typeof(MyProjectNameTestBaseModule),
        typeof(MyProjectNameMongoDbModule)
        )]
    public class MyProjectNameMongoDbTestModule : AbpModule
    {
        private static readonly MongoDbRunner MongoDbRunner = MongoDbRunner.Start();

        {
            var connectionString = MongoDbRunner.ConnectionString.EnsureEndsWith('/') +
                                   "MyProjectName_" +
                                   Guid.NewGuid().ToString("N");

            Configure<DbConnectionOptions>(options =>
            {
                options.ConnectionStrings.Default = connectionString;
            });
        }
    }
}
````

maliming:


hikalkan:
Great! Thanks.

PR #1646, opened 2019-08-16
 Fix typo of mongodb connection string in sample project

Rwing:


maliming:
Can you modify it together?
https://github.com/abpframework/abp/blob/dev/samples/BookStore-Angular-MongoDb/aspnet-core/src/Acme.BookStore.HttpApi.Host/appsettings.json#L7

PR #1675, opened 2019-08-23
Fix: IEventBus only has PublishAsync method - replaced Publish

dcagnetta:


hikalkan:
Thank you for noticing this. However, we can't call an async method like that. We should use await keyword (probably not possible, because this is a sync method) or use AsyncHelper.

@maliming can you check this sample and ensure that it is properly working.

Thanks.

dcagnetta:
Please could this be removed. I created another PR with the corrected changes above: https://github.com/abpframework/abp/pull/1683

PR #1683, opened 2019-08-26
Fix: IEventBus only has PublishAsync method - replaced Publish

dcagnetta:
This PR replaces correct async methods as well


maliming:
hi @dcagnetta 

I can't download the code for review, can you invite me to become a contributor?

image

PR #1709, opened 2019-09-03
Resolve #1708 Use AddMaps instead of AddProfile in startup templates.

### LINKED ISSUE #1708: Use AddMaps instead of AddProfile in startup templates
Because AddMaps also registers AutoMap attributes.

see https://github.com/abpframework/abp/issues/1705

maliming:


maliming:
Do modules and samples also need to be changed?

hikalkan:
Please use #1708 inside the PR comment (not only title). Otherwise, GitHub does not link PR to the issue correctly.

PR #1718, opened 2019-09-05
Delayed initialization of the language.

### LINKED ISSUE #1235: UseAbpRequestLocalization needs to a db connection on startup
We want to minimize startup delay. Microsoft's [request localization middleware](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/localization?view=aspnetcore-2.2#implement-a-strategy-to-select-the-languageculture-for-each-request) requires to know the list of the languages & the default language on initialization. So, we need to connect to db since user's language and the list of the languages are stored in the database.

We may write our own middleware or find a way of lazy loading these settings to reduce application startup delay.

maliming:
Resolve #1235


maliming:
Found a problem, working on it.

PR #1736, opened 2019-09-09
Add IIdentityStoreAutoSaveChanges.

### LINKED ISSUE #1732: `EntityCreatedEto<UserEto>` can't get all the user's values.
Hi, 
I just created a class to handle `EntityCreatedEto<UserEto>` event, which is generally created according to the [BlogUserSynchronizer](https://github.com/abpframework/abp/blob/master/modules/blogging/src/Volo.Blogging.Domain/Volo/Blogging/Users/BlogUserSynchronizer.cs) class, except that I handled the "Created" event instead of the "Updated" event.

However, only the `Id`, `UserName`, and `Email` values existed in the `UserDto`, and the others are null:

image

I also debugged the "Updated" event, it can get all the values.
Is this a BUG, or is it designed like this.

maliming:
Resolve #1732


hikalkan:
Good solution. Thanks.

PR #1738, opened 2019-09-10
Modify an error description

abphub:
Inheritance interface is not “ITenantResolver”，modified “ITenantResolveContributor”.
See lines 118 and 134.


maliming:
hi @abphub 
Can you correct the English documents together? : )

PR #1806, opened 2019-09-26
I found that .NET Core 3.0 works with Swashbuckle.AspNetCore -V 5.0.0-rc2

brittos:
https://docs.microsoft.com/en-us/aspnet/core/tutorials/getting-started-with-swashbuckle?view=aspnetcore-3.0&tabs=visual-studio


Xeevis:
rc-4 is out

PR #1814, opened 2019-09-27
Fix case of misleading CLI update advisory when custom path is used for installation

### LINKED ISSUE #1769: CLI update advisory needs to be aware of the executing tool's location
Commit by @ebicoglu d372b9cefaba6f59eeee79ec51471ff89b7207e4 introduced regression in a way user is advised to update to a new version. New code assumes tool is always in a default (global) location which is:

OS | Path
-- | --
Linux/macOS | $HOME/.dotnet/tools
Windows | %USERPROFILE%\\.dotnet\\tools

This however won't be the case if tool is installed with `--tool-path` switch. To demonstrate how would that be problematic:
> PS> dotnet tool install --tool-path "C:\AbpTool" Volo.Abp.Cli  
> PS> C:\AbpTool\abp
> [21:56:55 WRN] ABP CLI has a newer stable version ...
> [21:56:55 WRN] Update Command:
> [21:56:55 WRN] dotnet tool update -g Volo.Abp.Cli
> PS> dotnet tool update -g Volo.Abp.Cli
> Tool 'volo.abp.cli' is not currently installed.

This is the reason why `-g` wasn't used in the advisory, with `--tool-path` it works universally, no matter where is the tool installed. It's possible to use `-g`, but it's necessary to first determine if executing tool is indeed in it's default location.

Xeevis:
Tested on Windows and Ubuntu linux to work properly. 

Resolves #1769

### DIFF
--- a/framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/CliService.cs
+++ b/framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/CliService.cs
@@ -169,8 +169,16 @@ namespace Volo.Abp.Cli
             }
         }
 
+        private static bool IsGlobalTool(string toolPath)
+        {
+            var globalPaths = new[] { @"%USERPROFILE%\.dotnet\tools\", "%HOME%/.dotnet/tools/", };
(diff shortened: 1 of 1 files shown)


yekalkan:
Thank you @Xeevis

PR #1841, opened 2019-10-05
Prevent possible null reference exception

eduherminio:
* Prevent possible null reference exception in `SelectItems.GetItems()`
* Other minor changes


maliming:
@eduherminio  Thanks!

PR #1913, opened 2019-10-14
Added IDistributedCache<TCacheItem, TCacheKey>

### LINKED ISSUE #1900: Introduce IDistributedCache<TCacheItem, TCacheKey>
`IDistributedCache<TCacheItem>` assumes the cache key as string. `TCacheKey` generic argument make it easier to use the cache.
(notice that we don't want to remove `IDistributedCache<TCacheItem>`, but want to add `IDistributedCache<TCacheItem, TCacheKey>` as an alternative approach)
It will call `.ToString()` of the key object by default.

gterdem:
IDistributedCache<TCacheItem, TCacheKey> implementation.  resolve #1900


framework/test/Volo.Abp.Caching.Tests/Volo/Abp/Caching/PersonCacheItem.cs
```diff
@@ -17,7 +17,14 @@ public PersonCacheItem(string name)
             Name = name;
         }
     }
+
+    public class DummyObjectAsCacheKey
```
hikalkan: Override ToString and return a merged value of the properties. Otherwise all keys will be same (you know default ToString returns the class name - same for all objects)

framework/src/Volo.Abp.Caching/Volo/Abp/Caching/DistributedCache.cs
```diff
@@ -439,4 +439,327 @@ protected virtual void SetDefaultOptions()
             DefaultCacheOptions = GetDefaultCacheEntryOptions();
         }
     }
+    /// <summary>
+    /// Represents a distributed cache of <typeparamref name="TCacheItem" /> type.
+    /// Uses a generic cache key type of <typeparamref name="TCacheKey" /> type.
+    /// </summary>
+    /// <typeparam name="TCacheItem">The type of cache item being cached.</typeparam>
+    /// <typeparam name="TCacheKey">The type of cache key being used.</typeparam>
+    public class DistributedCache<TCacheItem, TCacheKey> : IDistributedCache<TCacheItem, TCacheKey>
```
hikalkan: Duplicating the whole class code is not a good idea, because all the code of these two classes will change together in the future.

Actually `DistributedCache<TCacheItem>` is a special type of `DistributedCache<TCacheItem, TCacheKey>`. So, why not inherit `DistributedCache<TCacheItem>` from `DistributedCache<TCacheItem, string>`? In this way, no code duplication occurs.
Same is possible for the interfaces.

framework/test/Volo.Abp.Caching.Tests/Volo/Abp/Caching/PersonCacheItem.cs
gterdem: Yeah, i missed the part about default ToString method of  the object. 
Although, I wanted to ask; does the generic object's ToString need to be overriden? Wouldn't it cause all the object's ToString method modified and should be declared in some sort of extension class (I think it already exists somewhere)?

Or do you think the complex object to be used as key should be marked with an interface like below?
```javascript
public class DummyObjectAsCacheKey: ICanBeCacheKey
```
I thought handling it inside the 
```javascript
protected virtual string NormalizeKey(TCacheKey key)
```
is more effective (but might be less efficient since checking type of key in every call).

I'll be sending new PR and let me know if I can improve.

framework/src/Volo.Abp.Caching/Volo/Abp/Caching/DistributedCache.cs
```diff
@@ -52,23 +79,76 @@ public class DistributedCache<TCacheItem> : IDistributedCache<TCacheItem>
...
 
             SetDefaultOptions();
         }
+        protected virtual string NormalizeKey(TCacheKey key)
+        {            
+            Type type = key.GetType();
+            string keyValue = key.ToString();
+
+            // If complex type of object, override the key value with property concatenation
+            if (!typeof(IComparable).IsAssignableFrom(type) || type.IsPrimitive || type.IsValueType)
```
hikalkan: We should not calculate the key value. We just call key.ToString() and use it as the key value to let the developer override ToString() and return a the value for a complex object.
We document it, so developers know what to do to be able to use a complex object as the cache key.

Issue #1176, opened 2019-05-29
Align Option Class Names


wayneboyles:
Similar to Issue #1124 the Option class names are not consistent.  For example we have `AbpMvcOptions` but then others like `VirtualFileSystemOptions`

Personally I think all of the options should be prefixed with 'Abp' so we'd end up with `AbpVirtualFileSystemOptions` - this makes it real clear we're configuring the Framework.

hikalkan:
I agree on that.

PR #1919, opened 2019-10-15
Align option class names

yekalkan:
resolves https://github.com/abpframework/abp/issues/1176


tuyndv:
Please don't use Abp prefix. Because these classes are already under Abp namespace.

hikalkan:
We needed to add Abp prefix for some option classes those have potential to make developers confusing (since they have same/similar names of asp.net core's options). Then we have a situation where some of them have Abp, while some of them haven't.

I am still not sure about this (it is also a breaking change, while it may be reasonable since v1.0 has not been released yet).

tuyndv:
Thanks @hikalkan

PR #1942, opened 2019-10-18
Update for resetting user password in UserManagement module.

gdlcf88:


hikalkan:
Can you implement this also for the angular UI? We would not like to introduce fragmentation between different kind of UIs.

KidoPlay:
I think we need to check old password before update it to a new one.
Better security.

gdlcf88:
Resetting user's password without original password check in UserManagement is reasonable, in my submission.

hikalkan:
>Resetting user's password without original password check in UserManagement is reasonable, in my submission.

Yes, because it is for admin users and they normally don't know the passwords of all users.

hikalkan:
@yekalkan can you test & merge this?

wocar:
Great! Hopefully this can be merged soon 🥇

PR #1955, opened 2019-10-20
Change customize property example for AppUser.

gdlcf88:
https://github.com/abpframework/abp/blob/master/docs/en/Entities.md

> All properties have protected setters. This is to prevent the entity from arbitrary changes from outside of the entity. For exmple, it would be dangerous to set TotalItemCount without adding a new product to the order. It's value is maintained by the AddProduct method.


hikalkan:
Thank you, but the example should be simple, otherwise some developers may think all property setters should be protected (believe me, it happens) :)

PR #1956, opened 2019-10-20
Complete some permission checks.

gdlcf88:


gdlcf88:
I am also updating codes for Angular UI.

hikalkan:
Thank you for that, as a last minute bugfix before v1.0 :)

mehmet-erim:
Thanks

PR #1965, opened 2019-10-21
Remove unnecessary UseAuthentication method.

### LINKED ISSUE #1958: Bearer Access Token Authorize Failed
1、Add [Authorize] to TestController in HttpApi project
```
    [Authorize]
    [Route("api/test")]
    public class TestController : POController
    {
    }
```
2、Get Access Token from IDS4
image

3、Try Call api/test using the access token
image

4、Authorize Failed and redirect to Login Page

5、Modify Web Module like this:
```
            app.UseAuthentication();
            //app.UseAuthorization();   //should after  app.UseIdentityServer();

            app.UseJwtTokenMiddleware();

            if (MultiTenancyConsts.IsEnabled)
            {
                app.UseMultiTenancy();
            }
            app.UseIdentityServer();
            app.UseAuthorization();  //move to here
```
  Problem solved !  Is this a bug?

maliming:
Resolve #1958

### DIFF
--- a/samples/BookStore-Angular-MongoDb/aspnet-core/src/Acme.BookStore.HttpApi.Host/BookStoreHttpApiHostModule.cs
+++ b/samples/BookStore-Angular-MongoDb/aspnet-core/src/Acme.BookStore.HttpApi.Host/BookStoreHttpApiHostModule.cs
@@ -153,8 +153,7 @@ namespace Acme.BookStore
             app.UseVirtualFiles();
             app.UseRouting();
             app.UseCors(DefaultCorsPolicyName);
-            app.UseAuthentication();
-            app.UseAuthorization();
+            app.UseIdentityServer();
             app.UseJwtTokenMiddleware();
 
             if (MultiTenancyConsts.IsEnabled)
@@ -162,7 +161,7 @@ namespace Acme.BookStore
                 app.UseMultiTenancy();
             }
 
-            app.UseIdentityServer();
+            app.UseAuthorization();
             app.UseAbpRequestLocalization();
 
             app.UseSwagger();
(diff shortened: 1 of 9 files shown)


samples/BookStore-Angular-MongoDb/aspnet-core/src/Acme.BookStore.HttpApi.Host/BookStoreHttpApiHostModule.cs
```diff
@@ -153,16 +153,15 @@ public override void OnApplicationInitialization(ApplicationInitializationContex
...
+            app.UseIdentityServer();
             app.UseJwtTokenMiddleware();
 
             if (MultiTenancyConsts.IsEnabled)
             {
                 app.UseMultiTenancy();
             }
 
-            app.UseIdentityServer();
+            app.UseAuthorization();
```
hikalkan: We can not do this, because;

* `UseMultiTenancy` requires `UseAuthentication` comes before it, because it gets tenantid from the current claims.
* `UseIdentityServer` can not come before `UseMultiTenancy`, otherwise identity server can not run multi-tenant.

So, the only solution is to duplicate authorization middleware (one added by us, one added by IDS4).

maliming: Another problem is that `UseJwtTokenMiddleware `needs to be before `UseAuthorization`. But it is now fixed by Https://github.com/abpframework/abp/commit/f767232d42cc7076b87d5f52911b02eb15b9fc44.

PR #2000, opened 2019-10-24
Added DataTimeValueConverters function.

maliming:
For the `Owned entity type` we need to manually configure its value conversion. It is difficult to determine the relationship through reflection.

Another problem is that in unit tests, Can't configure `AbpClockOptions `in the test method class, because `OnModelCreating `will only execute once, the configuration may conflict. Currently `AbpClockOptions `is configured in `AbpEntityFrameworkCoreTestModule`.

@hikalkan Any suggestions?


hikalkan:
>For the Owned entity type we need to manually configure its value conversion. It is difficult to determine the relationship through reflection.

We can document this and try to improve in the future.

>Another problem is that in unit tests, Can't configure AbpClockOptions in the test method class, because OnModelCreating will only execute once, the configuration may conflict. Currently AbpClockOptions is configured in AbpEntityFrameworkCoreTestModule.

This is a restriction of EF Core, you're right. As a workaround set Clock in the derived DbContext's constructor (directly create a Clock object).

maliming:
> Set Clock in the derived DbContext's constructor (directly create a Clock object).

How to implement it in unit testing? I don't understand.

PR #2032, opened 2019-10-29
code optimization

yilianhuaixiao:
Now that the following code：
```
var serviceProviderFactory = services.GetSingletonInstanceOrNull<IServiceProviderFactory<TContainerBuilder>>();
if (serviceProviderFactory == null)
		    {
			    throw new AbpException($"Could not find {typeof(IServiceProviderFactory<TContainerBuilder>).FullName} in {services}.");
		    }
```

requires  'IServiceProviderFactory<TContainerBuilder>' Service Type,
For intuition and efficiency, my code may be better. And strickly saying, they are not of equal value.  If registered service type 'XxxServiceProviderFactory<TContainerBuilder>', the former code may `throw AbpException($"Could not find {typeof(IServiceProviderFactory<TContainerBuilder>).FullName} in {services}.")`, my code won't.
It will just return 'services.BuildServiceProvider()'.

### DIFF
--- a/framework/src/Volo.Abp.Core/Microsoft/Extensions/DependencyInjection/ServiceCollectionCommonExtensions.cs
+++ b/framework/src/Volo.Abp.Core/Microsoft/Extensions/DependencyInjection/ServiceCollectionCommonExtensions.cs
@@ -42,13 +42,9 @@ namespace Microsoft.Extensions.DependencyInjection
 
 		    foreach (var service in services)
 		    {
-			    var factoryInterface = service.ImplementationInstance?.GetType()
-				    .GetTypeInfo()
-				    .GetInterfaces()
-				    .FirstOrDefault(i => i.GetTypeInfo().IsGenericType &&
-				                         i.GetGenericTypeDefinition() == typeof(IServiceProviderFactory<>));
-
-			    if (factoryInterface == null)
+			    var factoryInterface = service.ServiceType;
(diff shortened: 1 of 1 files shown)


hikalkan:
I am not sure if this produces the same result (I didn't spend time much :). I will consider this later since it has not an important enhancement.

framework/src/Volo.Abp.Core/Microsoft/Extensions/DependencyInjection/ServiceCollectionCommonExtensions.cs
```diff
@@ -42,13 +42,9 @@ public static IServiceProvider BuildServiceProviderFromFactory([NotNull] this IS
 
 		    foreach (var service in services)
 		    {
-			    var factoryInterface = service.ImplementationInstance?.GetType()
-				    .GetTypeInfo()
-				    .GetInterfaces()
-				    .FirstOrDefault(i => i.GetTypeInfo().IsGenericType &&
-				                         i.GetGenericTypeDefinition() == typeof(IServiceProviderFactory<>));
-
-			    if (factoryInterface == null)
```
maniya81, line 51: You can't use var when reassigning to an existing variable

if (factoryInterface == null)
    factoryInterface = service.ServiceType;

if (factoryInterface == null || 
    !factoryInterface.IsGenericType ||
    factoryInterface.GetGenericTypeDefinition() != typeof(IServiceProviderFactory<>))
{
    // handle error or return
}

Issue #2078, opened 2019-11-05
User menu missing icons.


wakuflair:
![image](https://user-images.githubusercontent.com/3471718/68174664-c1627a80-ffb9-11e9-96ba-3e44957235ee.png)

It's similar to #1697, I think I can send a PR.

PR #2079, opened 2019-11-05
Resolved https://github.com/abpframework/abp/issues/2078

wakuflair:


framework/src/Volo.Abp.AspNetCore.Mvc.UI.Theme.Basic/Themes/Basic/Components/Toolbar/UserMenu/Default.cshtml
```diff
@@ -22,6 +22,13 @@
                 var disabled = menuItem.IsDisabled ? "disabled" : string.Empty;
 
                 <a class="dropdown-item @cssClass @disabled" href="@(menuItem.Url ?? "#")" target="@menuItem.Target" @Html.Raw(elementId)>
+                    @if (menuItem.Icon != null)
+                    {
+                        if (menuItem.Icon.StartsWith("fa"))
```
maliming: Why does the icon need to start with `fa`?

wakuflair: I thought it should be the same with other menus.

e.g:

https://github.com/abpframework/abp/blob/dev/framework/src/Volo.Abp.AspNetCore.Mvc.UI.Theme.Basic/Themes/Basic/Components/Menu/Default.cshtml

PR #2096, opened 2019-11-06
Update PostgreSQL-Integration.md

xyfy:
Add some necessary steps and provide more information of connectionstring.


hikalkan:
@maliming can you check this if these information fits into this document (and aligns with other documents).

Issue #1456, opened 2019-07-08
Remove CrudAppService and rename AsyncCrudAppService to CrudAppService


hikalkan:
No need to support an sync version of this base class.
This is part of reducing sync over async calls.
Related to the work https://github.com/abpframework/abp/issues/1403

PR #2102, opened 2019-11-06
Updated docs with CrudAppService rename.

olicooper:
Relates to issue https://github.com/abpframework/abp/issues/1456


maliming:
Thanks @olicooper 👍

PR #2112, opened 2019-11-07
Update Part-I.md

xyfy:
fixed the layout component 

If this is not done here, the delete pop-up box in `Part II: Create, Update and Delete books` will not appear


mehmet-erim:
See the [`app-routing.module.ts`](https://github.com/abpframework/abp/blob/dev/samples/BookStore-Angular-MongoDb/angular/src/app/app-routing.module.ts#L37). `BooksModule` wrapped with the `ApplicationLayoutComponent`

PR #2117, opened 2019-11-07
Add missing DataTables options for tenant-management module.

gdlcf88:


hikalkan:
@ebicoglu can you check if this is necessary.

gdlcf88:
Paging runs incorrectly without the options `paging: true` and `serverSide: true`.

PR #2130, opened 2019-11-08
fix(theme-shared):fix #2107,when use Esc key to close modal ,should trigger destroy to avoid raise an error

### LINKED ISSUE #2107: (ng.theme.shared)  get error when modal close
If two kinds or more of abp-modal in a page,
for  example UsersComponent :it has edit modal and permission modal.
first open edit then close and then open permission
you will receive a "ERROR TypeError: Cannot read property 'nativeElement' of undefined" when closing permission modal by press Esc Key.
image

yinchang0626:
fix #2107 
when use Esc key to close modal ,should trigger destroy to avoid raise an error

### DIFF
--- a/npm/ng-packs/packages/theme-shared/src/lib/components/modal/modal.component.ts
+++ b/npm/ng-packs/packages/theme-shared/src/lib/components/modal/modal.component.ts
@@ -122,10 +122,12 @@ export class ModalComponent implements OnDestroy {
           this.isConfirmationOpen = false;
           if (status === Toaster.Status.confirm) {
             this.visible = false;
+            this.destroy$.next();
           }
         });
     } else {
       this.visible = false;
+      this.destroy$.next();
     }
   }
 


mehmet-erim:
You should add in the `visible` setter. It's enough and correct way

After your changes, 46th - 49th lines should be like this:
```
else {
 this.renderer.removeClass(document.body, 'modal-open');
 this.disappear.emit();
this.destroy.next();
this.destroy.complete();
}
```

yinchang0626:
@mehmet-erim 
thank for helping, i think you are right
I tried  that way you say to fix,
when i test that code:
move code to visible setter is no problem,
but it appear error in this:
```text
else {
 this.renderer.removeClass(document.body, 'modal-open');
 this.disappear.emit();
 this.destroy$.next();
 this.destroy$.complete();
}
```
and work in this
```text
else {
 this.renderer.removeClass(document.body, 'modal-open');
 this.disappear.emit();
 this.destroy$.next();
}
```
I dont know why after call complete()
takeUntil(this.destroy$) will still triger the keyup event
if you has suggest ,pls let me know,
I will change it to
```text
else {
 this.renderer.removeClass(document.body, 'modal-open');
 this.disappear.emit();
 this.destroy$.next();
}
```

PR #2169, opened 2019-11-14
Add AbpMinifyModule.

### LINKED ISSUE #2008: Minify the script returned by ServiceProxyScript.

maliming:
Resolve #2008


hikalkan:
One improvement suggestion:

Can you automatically minify ServiceProxyScript and ApplicationConfigurationScript on production? What do you think?

Send a separate PR since I merged this.

Thanks.

maliming:
I want to include `MinifyGeneratedScript` in `AbpAspNetCoreMvcOptions`. The default is `true`.

After that, the script is minify in `AbpServiceProxyScriptController `and `AbpApplicationConfigurationScriptController `according to this option.

In the template.
```
Configure<AbpAspNetCoreMvcOptions>(options =>
{
	options.ConventionalControllers.Create(typeof(XXXApplicationModule).Assembly);
	options.MinifyGeneratedScript = context.Services.GetHostingEnvironment().IsProduction();
});
```

hikalkan:
That's fine, but we can do it (`options.MinifyGeneratedScript = context.Services.GetHostingEnvironment().IsProduction()`) in the framework?

To allow it to be overrided by the developer, we can make an enum (Enabled/Disabled/Auto) or make it `bool?` (nullable) and if it is null/auto then we decide based on the environment.

PR #2178, opened 2019-11-15
Fix UseVirtualFiles middleware return type

Xeevis:
All IApplicationBuilder extension methods from [ASP.NET Core built-in middleware](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.builder.iapplicationbuilder?view=aspnetcore-3.0#extension-methods) always return IApplicationBuilder. Most ABP middleware does too. `UseVirtualFiles` seems to be the only exception to the rule and it's stopping use of fluent design. As a result this code throws error: 
`Operator '.' cannot be applied to operand of type 'void'`

```csharp
public override void OnApplicationInitialization(ApplicationInitializationContext context)
{
    context.GetApplicationBuilder()
        .UseCorrelationId()
        .UseVirtualFiles() // <-- VOID
        .UseAbpRequestLocalization()
        .UseRouting()
        .UseMiddleware<FakeAuthenticationMiddleware>()
        .UseAuthorization()
        .UseAuditing()
        .UseUnitOfWork()
        .UseMvcWithDefaultRouteAndArea();
}
```


maliming:
Thanks @Xeevis

PR #2190, opened 2019-11-17
Resolved #1508  Adding cli namespace normalizer.

### LINKED ISSUE #1508: abp cli command new: Should prevent user from using special chars in project name.
I noticed this morning that the `abp new`  command does not prevent users from using special chars in the project name. (i.e. `abp new Acme.Book-Store`)

image

mehmetuken:
Reference: https://stackoverflow.com/questions/773557/which-characters-are-allowed-in-a-vs-project-name
Reference: https://github.com/dotnet/templating/blob/dcf5adbd4b5665887f22ad19592dedfb9f1e1b68/src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/ValueForms/DefaultSafeNamespaceValueFormModel.cs#L35

### DIFF
--- a/framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/Args/CommandLineArgumentParser.cs
+++ b/framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/Args/CommandLineArgumentParser.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using Volo.Abp.Cli.Utils;
 using Volo.Abp.DependencyInjection;
 
 namespace Volo.Abp.Cli.Args
@@ -40,6 +41,7 @@ namespace Volo.Abp.Cli.Args
 
             if (!argumentList.Any())
             {
+                target = NamespaceHelper.NormalizeNamespace(target);
                 return new CommandLineArgs(command, target);
             }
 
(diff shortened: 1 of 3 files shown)


framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/Utils/NamespaceHelper.cs
```diff
@@ -0,0 +1,21 @@
...
+    public static class NamespaceHelper
+    {
+        public static string NormalizeNamespace([CanBeNull] string value)
+        {
+            if (string.IsNullOrEmpty(value))
+            {
+                return value;
+            }
+
+            value = Regex.Replace(value, @"(^\s+|\s+$)", "");
```
yekalkan: I think "String.Trim()" does the same job and much easier to understand than a regex.

framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/Args/CommandLineArgumentParser.cs
```diff
@@ -40,6 +41,7 @@ public CommandLineArgs Parse(string[] args)
 
             if (!argumentList.Any())
             {
+                target = NamespaceHelper.NormalizeNamespace(target);
```
yekalkan: This normalizing should be done only when 'target' must fit the project name regex. If you make this normalization here, any command which is not related to project name regex will be effected. 

So this normalization has to be moved to commands which takes a project name as 'target'.

PR #2196, opened 2019-11-18
Updating pt-BR account localization

thiag0coelho:


maliming:
@thiag0coelho  Thanks! 👍

PR #2197, opened 2019-11-18
feature(theme-basic) resolve #2193 hide language dropdown when only one language

### LINKED ISSUE #2193: Angular template - hide language dropdown if only 1 option
While creating a application, I only support 1 language like so

            Configure<AbpLocalizationOptions>(options =>
            {
                options.Languages.Add(new LanguageInfo("pt-BR", "pt-BR", "Português"));
            });

And then I expect the language dropdown to be hidden, because there is only one option

yinchang0626:
resolve #2193


mehmet-erim:
Thanks @yinchang0626

Issue #2235, opened 2019-11-22
Microservice PublicWebSite oidc fails parsing UserInfoEndpoint response


itryan:
In the MicroserviceDemo, when AuthServer creates new user without role, the oidc seems unable to complete /signin-oidc request.

1. on AuthServer, register new user "test", which is created without roles
2. on PublicWebSite, login but get following error

```
System.Collections.Generic.KeyNotFoundException: The given key was not present in the dictionary.
   at System.Text.Json.JsonElement.GetProperty(String propertyName)
   at Volo.Abp.AspNetCore.Authentication.OAuth.Claims.MultipleClaimAction.Run(JsonElement userData, ClaimsIdentity identity, String issuer)
   at Microsoft.AspNetCore.Authentication.OpenIdConnect.OpenIdConnectHandler.GetUserInformationAsync(OpenIdConnectMessage message, JwtSecurityToken jwt, ClaimsPrincipal principal, AuthenticationProperties properties)
   at Microsoft.AspNetCore.Authentication.OpenIdConnect.OpenIdConnectHandler.HandleRemoteAuthenticateAsync()
```

The missing claim key is `role`, AuthServer UserInfoEndpoint returns json without `role` property, but oidc tries to read this value and exception occurs. Should the missing claim key be ignored instead of raising KeyNotFoundException?

https://github.com/abpframework/abp/blob/b273f29ebc8fa722a3044dfab461703a9573df40/samples/MicroserviceDemo/applications/PublicWebSite.Host/PublicWebSiteHostModule.cs#L72

https://github.com/abpframework/abp/blob/ea549b4305e9bc3291eb06b3191f165edf2d8228/framework/src/Volo.Abp.AspNetCore.Authentication.OAuth/Microsoft/AspNetCore/Authentication/OAuth/Claims/AbpClaimActionCollectionExtensions.cs#L39

https://github.com/abpframework/abp/blob/654eeede81a733382066bfd00ab19b6b65ceec0e/framework/src/Volo.Abp.AspNetCore.Authentication.OAuth/Volo/Abp/AspNetCore/Authentication/OAuth/Claims/MultipleClaimAction.cs#L17

hikalkan:
Fixed by https://github.com/abpframework/abp/pull/2238

PR #2238, opened 2019-11-23
fix: Microservice PublicWebSite oidc fails parsing UserInfoEndpoint response

itryan:
fix: Microservice PublicWebSite oidc fails parsing UserInfoEndpoint response

#2235


hikalkan:
Thanks @itryan

PR #2240, opened 2019-11-23
Adding an additional parameter for the SettingPageGroup's constructor.

wakuflair:


hikalkan:
Thanks. I will review this.

hikalkan:
SettingPageGroups are static, I mean you create them hard-coded. So, adding a hard-coded parameter seems no meaning to me. You can already do whatever you need in your component.
What I am missing?

wakuflair:
@hikalkan 
https://github.com/abpframework/abp/blob/42f37c5ff01ad853a5425d15539d4222cd0dab69/modules/setting-management/src/Volo.Abp.SettingManagement.Web/Pages/SettingManagement/Index.cshtml#L31

Here I want to pass an additional parameter to the Widget, so that I I can reuse the widget and execute different logic by checking the parameter.
For now I couldn't find a way to do that.


> SettingPageGroups are static, I mean you create them hard-coded

Sorry I don't quite understand what you said.
You mean I can create a derived class from `SettingPageGroups`?

PR #2249, opened 2019-11-25
Set the message of errorInfo to the message of AbpValidationException.

### LINKED ISSUE #2244: AbpValidationException doesn't output the message
When the exception is thrown, the message supplied in to the constructor is not output to the user.
```json
{
  "error": {
    "code": null,
    "message": "Your request is not valid!",
    "details": "The following errors were detected during validation.\r\n",
    "validationErrors": []
  }
}
```
```csharp
throw new AbpValidationException("Your query is malformed.");
```

maliming:
Resolve #2244

I think `BusinessException `should inherit `AbpException`, So we can directly assign the message of AbpException to the message of errorInfo.


hikalkan:
@maliming,

````csharp
            if (exception is AbpValidationException)
            {
                errorInfo.Message = exception.Message;
            }
````

This is good for cases like https://github.com/abpframework/abp/issues/2244, but as you know ABP framework also throws this exception. If we directly get exception.Message as the error message, then it can not localize the message with the current implementation! You can see `errorInfo.Message = L["ValidationErrorMessage"];` which is a localized message.

PR #2254, opened 2019-11-26
updated Volo.Abp.Autofac to 1.0.2

274188A:


maliming:
hi @274188A 
You can't remove it, have you tested the build project?

maliming:
I updated the package references. see https://github.com/abpframework/abp/pull/2255

PR #2307, opened 2019-12-04
Fixed mongo repository expression error in AuditLogging module.

maliming:
MongoDB.Driver's `GroupBy` method cannot translate `DateTime.Property`.
eg `GroupBy(t => ExecutionTime.Date)`


hikalkan:
Thanks @maliming. But this will have a performance problem since it brings all audit logs into memory and group in the code.
Can you search on that how we can implement this in MongoDB. We can use Mongo's own API (instead of LINQ), no problem.

maliming:
I will check it again.

hikalkan:
Should I merge now?

maliming:
@hikalkan Yes.

hikalkan:
Thanks @maliming

PR #2319, opened 2019-12-06
Upgrade to Npgsql.EntityFrameworkCore.PostgreSQL 3.1.0

gdlcf88:
Should merge after @maliming 's https://github.com/abpframework/abp/pull/2309


maliming:
https://github.com/abpframework/abp/pull/2309/files#diff-768d39fe271092c4441d354382639627R20-R22

PR #2357, opened 2019-12-10
fix inconsistent tenantId format

Goxiaoy:
In module `FeatureManager`, file `TenantFeatureManagerExtensions`  `ToString() ` is used instead of `ToString("N")`
Incosistent format will cause `TenantFeatureValueProvider` always get the incorrect value

```
public static class TenantFeatureManagerExtensions
    {
        public static Task<string> GetOrNullForTenantAsync(this IFeatureManager featureManager, [NotNull] string name, Guid tenantId, bool fallback = true)
        {
            return featureManager.GetOrNullAsync(name, TenantFeatureValueProvider.ProviderName, tenantId.ToString(), fallback);
        }

        public static Task<List<FeatureNameValue>> GetAllForTenantAsync(this IFeatureManager featureManager, Guid tenantId, bool fallback = true)
        {
            return featureManager.GetAllAsync(TenantFeatureValueProvider.ProviderName, tenantId.ToString(), fallback);
        }

        public static Task SetForTenantAsync(this IFeatureManager featureManager, Guid tenantId, [NotNull] string name, [CanBeNull] string value, bool forceToSet = false)
        {
            return featureManager.SetAsync(name, value, TenantFeatureValueProvider.ProviderName, tenantId.ToString(), forceToSet);
        }
    }
```

### DIFF
--- a/framework/src/Volo.Abp.Features/Volo/Abp/Features/TenantFeatureValueProvider.cs
+++ b/framework/src/Volo.Abp.Features/Volo/Abp/Features/TenantFeatureValueProvider.cs
@@ -19,7 +19,7 @@ namespace Volo.Abp.Features
 
         public override async Task<string> GetOrNullAsync(FeatureDefinition feature)
         {
-            return await FeatureStore.GetOrNullAsync(feature.Name, Name, CurrentTenant.Id?.ToString("N"));
+            return await FeatureStore.GetOrNullAsync(feature.Name, Name, CurrentTenant.Id?.ToString());
         }
     }
 }


maliming:
@hikalkan 
The feature-management module also has inconsistencies,
But I didn't find the code using `ToString (" N ")` in the setting module.

hikalkan:
@Goxiaoy you only changed in one place. What about others?
@maliming can you check all and fix the problem completely if there is a problem.
We will probably release v1.1.1 tomorrow.

Goxiaoy:
I am going to close this pr. I have not noticed that there are many places using this format and I think  it’s better for your guys to decide which one is recommended.

PR #2381, opened 2019-12-12
Implement AsyncBackgroundJob.

### LINKED ISSUE #2374: BackgroundJob Async Execute
The BackgroundJob class only has a sync Execute method, it would be great if you could add an async replacement.

maliming:
Resolve #2374


jortygu:
why delete the AsyncBackgroundJob?

maliming:
@jortygu 

`AsyncBackgroundJob` was not removed.

https://github.com/abpframework/abp/blob/dev/framework/src/Volo.Abp.BackgroundJobs.Abstractions/Volo/Abp/BackgroundJobs/AsyncBackgroundJob.cs#L7

PR #2414, opened 2019-12-18
resolve #2385 fix mongo.exe dispose

### LINKED ISSUE #2385: The mongodb.exe process to continue running after the unit test is over.
https://github.com/abpframework/abp/issues/1632#issuecomment-563386192

image

mehmetuken:
using xunit collection fixtures

ref: https://xunit.net/docs/shared-context

### DIFF
--- a/framework/test/Volo.Abp.MongoDB.Tests/Volo/Abp/MongoDB/AbpMongoDbTestModule.cs
+++ b/framework/test/Volo.Abp.MongoDB.Tests/Volo/Abp/MongoDB/AbpMongoDbTestModule.cs
@@ -16,7 +16,7 @@ namespace Volo.Abp.MongoDB
     {
         public override void ConfigureServices(ServiceConfigurationContext context)
         {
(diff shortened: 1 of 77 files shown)


maliming:
This is great 👍 , I will test it as soon as possible.

hikalkan:
Thanks @mehmetuken 

I also want to review this in the next milestone. First impression: I didn't like to add [Collection(MongoTestCollection.Name)] to every test class. If this is the only solution, we should be sure that this will not be a problem in the future.

mehmetuken:
I was search this solution i found only way any tests share same context we collected all tests single collection. Because other way we don't now which test is finish or runing. The way collected tests and we share the same context. 

xunit document say:

>Collection Fixtures
 When to use: when you want to create a single test context and share it among tests in several test classes, and have it cleaned up after all the tests in the test classes have finished.

hikalkan:
Thank you @mehmetuken I finally merged it. It is the only way to fix this issue.

PR #2423, opened 2019-12-19
export some  component and service in identity account module

yinchang0626:
please export  `ng-packs` angular module's  component and service   for replace and reuse,thanks


yinchang0626:
What should I do for actions/labeler? It get me a Cross.

acjh:
> What should I do for actions/labeler? It get me a Cross.

Nothing. It's a limitation of *labeler*: https://github.com/actions/labeler/issues/12

PR #2425, opened 2019-12-19
feat/add-route-action

### LINKED ISSUE #2186: Ng routes for lazy-loaded modules not appearing in menu
I have a number of modules that I want to appear in the menu, but the menu items are only populated from routes defined in `app-routing.module.ts`. If I move the route `data` attribute to a lazy-loaded sub-module the menu item no-longer shows up. I have tried using `addAbpRoutes` from the `@abp/ng.core` package within the constructor of the `app-module` and the sub-modules with no success.

On the basic Angular application supplied, perform the following steps to reproduce:

In `app-routing.module.ts`, replace the Home route with the following:
```typescript
{
    path: '',
    loadChildren: () => import('./home/home.module').then(m => m.HomeModule),
},
```
In `home-routing.module.ts`, replace the main route with the following:
```typescript
{
    path: '',
    component: ApplicationLayoutComponent,
    data: {
      routes: {
        name: '::Menu:Home',
      } as ABP.Route,
    },
    children: [{ path: '', component: HomeComponent }],
},
```

mehmet-erim:
Added an action to `@abp/ng.core` package named `AddRoute`.

Usage:

```ts
this.store.dispatch(
      new AddRoute({
        name: 'My new page',
        iconClass: 'fa fa-dashboard',
        path: 'page',
        invisible: false,
        order: 2,
        requiredPolicy: 'MyProjectName::MyNewPage',
        parentName: 'AbpIdentity::Menu:IdentityManagement',
      } as Omit<ABP.Route, 'children'>),
    );
```
Looks like:

Screen Shot 2019-12-19 at 17 29 55

Resolves #2186


olicooper:
This is great news, thanks for your hard work!

PR #2440, opened 2019-12-22
add pagesize/pageindex in PagedResultRequestDto for the mvvm framewor…

easyfast:
…k like Vue or Angular.
In these framework, the component of  pagination customary use these property


hikalkan:
Thank you but we can not accept this change. It creates duplication (two property for the same purpose). Handle it in the client side please.

PR #2441, opened 2019-12-22
When using FluentValidation for validation, members in ValidationResult are missing

kelely:
When using FluentValidation for validation, if validation fails, members in ValidationResult are always null


maliming:
hi @kelely

kelely:
hi @maliming 
I adjusted the unit tests for validating members in ValidationResult.

PR #2464, opened 2019-12-25
Re-design interception & sync/async APIs

hikalkan:
* Resolves #2026
* Resolves #2121
* Resolves #2035
* Resolves #2075

PR #2466, opened 2019-12-26
Remove the sync method for background job module.

maliming:
https://github.com/abpframework/abp/pull/2464#issuecomment-568943613


hikalkan:
Background job is OK, but we shouldn't change the tenant store. Why? Let me explain;

First of all, `IConnectionStringResolver.Resolve` method is sync by its nature. We can not change it, unfortunately.

For RemoteTenantStore.cs;

* When we use `AsyncHelper` inside the `Resolve` method, `AsyncHelper` is always called, in every request. But if we move `AsyncHelper` usage to the `RemoteTenantStore` (like my implementation), it is called only if it can not find the connection string in the cache.

For TenantStore.cs;

* If we remove the sync `Find` method, we force to `AsyncHelper` usage unnecessarily while we could implement it truely sync. Native sync is better than async over sync usage.

So, please cancel the tenantstore changes for now, until we find a good way to make `IConnectionStringResolver.Resolve` async.

maliming:
I have cancelled related changes to tenant.

hikalkan:
Thanks ;)

PR #2470, opened 2019-12-26
Cancel value generation for aggregate roots and entities.

maliming:
When they use Guid as the primary key type


hikalkan:
Perfect @maliming :)

Issue #2468, opened 2019-12-26
Localize exception message on LimitedResultRequestDto


hikalkan:
This requires to add a new localization resource to the `Volo.Abp.Ddd.Application.Contracts` package.

PR #2473, opened 2019-12-26
Localize exception message on limited result request dto

yekalkan:
resolves https://github.com/abpframework/abp/issues/2468


framework/src/Volo.Abp.Ddd.Application.Contracts/Volo/Abp/Application/Dtos/LimitedResultRequestDto.cs
```diff
@@ -30,9 +32,13 @@ public class LimitedResultRequestDto : ILimitedResultRequest, IValidatableObject
 
         public virtual IEnumerable<ValidationResult> Validate(ValidationContext validationContext)
         {
+            var l = validationContext.GetService(typeof(IStringLocalizer<AbpDddResource>)) as IStringLocalizer<AbpDddResource>;
```
hikalkan: Move this into the `if (MaxResultCount > MaxMaxResultCount)` to not call `GetService` unnecessarily if the request is already valid.

hikalkan: I've done it: https://github.com/abpframework/abp/commit/c010c567930c3f61c31b61f9491d14c21894e6b1

PR #2489, opened 2019-12-28
Move AbpIntegratedTest to Volo.Abp.Testing namespace

recepdmr:
abp proxy helper class code refactoring.

### DIFF
--- a/framework/src/Volo.Abp.TestBase/Volo/Abp/Testing/AbpIntegratedTest.cs
+++ b/framework/src/Volo.Abp.TestBase/Volo/Abp/Testing/AbpIntegratedTest.cs
@@ -2,10 +2,8 @@
 using Microsoft.Extensions.DependencyInjection;
 using Volo.Abp.Modularity;
 
-namespace Volo.Abp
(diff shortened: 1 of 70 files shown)


framework/src/Volo.Abp.TestBase/Volo/Abp/Testing/AbpIntegratedTest.cs
```diff
@@ -2,10 +2,8 @@
 using Microsoft.Extensions.DependencyInjection;
 using Volo.Abp.Modularity;
 
-namespace Volo.Abp
+namespace Volo.Abp.Testing
```
hikalkan: When you change this namespace, all the projects/solutions using the AbpIntegratedTest will break. So, if we make this change, we should change all modules/templates/samples in this repository. Otherwise, I can not accept this PR :)

recepdmr: I'm sorry, I'm not testing ,I will testing and send a new pull request.

PR #2493, opened 2019-12-29
Update repository method to WithDetails in docs

### LINKED ISSUE #2488: There is no suitable method  `IncludeDetails` to override in EfCoreRepository
I have a repository inherited from `EfCoreRepository` I wrote an extension method  in  `PmsCloudEfCoreQueryableExtensions.cs` like this : 
``` C#
  public static IQueryable<RateGroup> IncludeDetails(this IQueryable<RateGroup> queryable, bool include = true)
        {
            if (!include)
            {
                return queryable;
            }

            return queryable
                .Include(x => x.Mpe);


        }
```

then I tried to override `IncludeDetails` method  in the repository as the [documentaion](https://docs.abp.io/en/abp/latest/Best-Practices/Entity-Framework-Core-Integration#repository-implementation) said.
. but I got this error:

image

acjh:
Resolves #2488 

Repository's `IncludeDetails` was refactored to `WithDetails` in https://github.com/abpframework/abp/commit/649f80e8de2231f6e3d6ad645697832d76cad905 (released in version 0.3.7).

Existing source code that corresponds to the example in docs:
https://github.com/abpframework/abp/blob/2707a1ee4b93712e94d73eb3ba0a385cfa3853a1/modules/identity/src/Volo.Abp.Identity.EntityFrameworkCore/Volo/Abp/Identity/EntityFrameworkCore/EfCoreIdentityUserRepository.cs#L146-L149

### DIFF
--- a/docs/en/Best-Practices/Entity-Framework-Core-Integration.md
+++ b/docs/en/Best-Practices/Entity-Framework-Core-Integration.md
@@ -174,12 +174,12 @@ public static IQueryable<IdentityUser> IncludeDetails(
 
 * **Do** use the `IncludeDetails` extension method in the repository methods just like used in the example code above (see FindByNormalizedUserNameAsync).
 
-- **Do** override `IncludeDetails` method of the repository for aggregates root which have **sub collections**. Example:
+- **Do** override `WithDetails` method of the repository for aggregates root which have **sub collections**. Example:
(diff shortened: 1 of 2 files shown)


hikalkan:
Thank you @acjh :)

Issue #2497, opened 2019-12-29
Call ConfigureAwait(false) on await usages


hikalkan:

javiercampos:
I've made the above PR (#2500) for all Volo.Abp.* solutions in the repository (including framework and modules). I've built and ran the tests and everything is ok except for 1 test which doesn't look related to this (expained on the PR)

mehmetuken:
We don't need this because aspnet core doesn't have SynchronizationContext. 
ref: https://blog.stephencleary.com/2017/03/aspnetcore-synchronization-context.html

And i merge #2414 after don't hanging.

https://github.com/mehmetuken/abp/commit/b4bce66ae6218e19c00aa361e2faefb3ed7e0463/checks?check_suite_id=376506937
https://github.com/mehmetuken/abp/commit/40e8a65f98ec58bb2f042fe6d43b27343a092784/checks?check_suite_id=376313708

javiercampos:
@mehmetuken : yes, aspnet core doesn't have one, but some test runners (and possibly other environments) do, and to me at least, this seems to be causing issues with the sync contexts on both xunit and nunit (which they use for parallelization, even if you disable it), on integration tests ( see #2075 ) when the hardware is not powerful enough (or it's busy)

mehmetuken:
I understand but hanging problem still gone #2464 and #2414 pr for me. You can try which one ci runner using your merge this two change after.

javiercampos:
Yes, as I said, most of the hangs probably came from the use of the `AsyncEx` library (this got fixed once you removed the sync code), *still*, it's a good etiquette to use `ConfigureAwait(false)` on library code and make no assumptions on who is calling it and how. Capturing the synchronization context (if one exists) should be done by whoever needs it captured, and the rest of the `await` calls should refrain from doing so... this is true for basically all library code: it should be the consumer the one that should capture the context, and not the library: whether that context exists or not, it should not be the library the one to decide so.

Not calling `ConfigureAwait(false)` in awaitables when you don't need to capture the context for a specific reason is just wrong code, and if it works, it does so "because of the environment", which is never a good thing for a library to be reliant on.

There's absolutely zero harm on calling `ConfigureAwait(false)` in ABP, and there's a possible tangible benefit on doing so, so I'd start doing so rather sooner than later. That's why I asked if I should make this PR (and was told so here: https://github.com/abpframework/abp/issues/2075#issuecomment-569507256 )

mehmetuken:
[...]

PR #2500, opened 2019-12-29
Call ConfigureAwait(false) on all library related await calls

javiercampos:
Related to #2497 

Only added the `ConfigureAwait(false)` calls on library solutions (all Volo.Abp.* solutions).

All tests pass except for 1 but doesn't seem related to this. The failing one is:
`Volo.Abp.Emailing.EmailTemplateStore_Tests.Should_Get_Registered_Template_With_Layout`

Which fails with:

```
   Shouldly.ShouldAssertException : template.Content
    should contain (case insensitive comparison)
"<body>
    Please confirm your email address by clicking the link below."
    but was actually
"<!DOCTYPE html>
<html lang="en" xmlns="http://www.w3.org/1999/xhtml">
```

Doesn't seem related to any of these commits and should probably be fixed separately


javiercampos:
All commits are 1 per solution (except the fixing one, which the analyzer couldn't figure out correctly)

hikalkan:
That's a great contribution. Thank you for your huge effort.

maliming:
@mehmetuken mentioned a https://github.com/Fody/ConfigureAwait class library. It can save a lot of code. I haven't tried it, but from the introduction, it really is.

javiercampos:
@maliming adding Fody to the build pipeline might be painful... when it works, it works fine: when it doesn't, you'll want to pull all your hairs off... I'd GREATLY recommend **AGAINST** it for a library. I find it more than ok for app code (I've used it in the past for many things), but not for a library :-)

hikalkan:
@maliming  Fody seems good, yes, but I also think to not add a new dependency for now. I will merge this PR and delay the Fody decision.

PR #2503, opened 2019-12-30
Remove related EF Core global filters if multitenancy is disabled

### LINKED ISSUE #949: Remove related EF Core global filters if multitenancy is disabled
Based on https://github.com/abpframework/abp/issues/914

maliming:
Resolve #949


hikalkan:
@maliming there is a misunderstanding here. 
This PR doesn't remove the EF Core global filters, but only returns false in the expression, so it turns off the logical filter. However, the filtering conditions still present in the generated SQL.

What I mean is to completely skip [this code](https://github.com/abpframework/abp/blob/76d742683aeb2dc71275fde9c4e55de0a164feb7/framework/src/Volo.Abp.EntityFrameworkCore/Volo/Abp/EntityFrameworkCore/AbpDbContext.cs#L557-L561) if multi-tenancy is disabled.

But this is not easy to implement since we may not resolve dependencies in that time. This is why I haven't implement it yet :)

Issue #2310, opened 2019-12-04
Document how to switch to different DBMS for EF Core for the app startup template


hikalkan:
- [x] MySQL
- [x] PostgreSQL
- [x] SQLite
- [x] Others?

gterdem:
Here is a list of [efcore database providers](https://docs.microsoft.com/en-us/ef/core/providers/index?tabs=dotnet-core-cli).

`CosmosDb `can be added as an other.
`MariaDb `uses MySql Provider, maybe requires a few tweaks.

- MySql
- PostgresSql
- SQLite
documentations are ready.

hikalkan:
Moved CosmosDB to another issue and closing this one: https://github.com/abpframework/abp/issues/2532
Thanks.

PR #2531, opened 2020-01-02
Pr/2310

gterdem:
updates #2310

### DIFF
--- a/docs/en/docs-nav.json
+++ b/docs/en/docs-nav.json
@@ -277,6 +277,10 @@
                 {
                   "text": "Switch to PostgreSQL",
                   "path": "Entity-Framework-Core-PostgreSQL.md"
+                },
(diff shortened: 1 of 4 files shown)


docs/en/docs-nav.json
```diff
@@ -277,6 +277,10 @@
                 {
                   "text": "Switch to PostgreSQL",
                   "path": "Entity-Framework-Core-PostgreSQL.md"
+                },
+				{
```
hikalkan: Can you fix the indent. We don't use tabs, we use "two space" instead of a tab (for JSON files).

hikalkan: OK, I've fixed: https://github.com/abpframework/abp/commit/525387654af80129f0836af8ca665b39c5e028ec

PR #2550, opened 2020-01-04
Dynamic C# Api client-side validation.

### LINKED ISSUE #2526: Validate DTOs before server calls on dynamic C# API client
Implement same validation in HTTP Client proxies (using data annotations on DTOs). So, we can get validation exception even before calling the remote service.

mehmetuken:
Resolve #2526


maliming:
hi @mehmetuken 
I haven't researched it in depth, but I think if adding **ValidationInterceptor** to it can solve this issue?

https://github.com/abpframework/abp/blob/88ded4ef643f2eb8786574794b45b84e0d66e4f9/framework/src/Volo.Abp.Http.Client/Microsoft/Extensions/DependencyInjection/ServiceCollectionDynamicHttpClientProxyExtensions.cs#L165

mehmetuken:
Thank you @maliming. It's very clear way. I'm new interceptor concept.

hikalkan:
Thanks to both of you ;)

PR #2573, opened 2020-01-07
feat(account): add the enableLocalLogin condition to auth-wrapper

mehmet-erim:


hikalkan:
@mehmet-erim you can merge this since they both approved.

PR #2589, opened 2020-01-08
Grammatical issue with Identity en localization

olicooper:


maliming:
@olicooper  Thanks! 👍

Issue #2583, opened 2020-01-08
Serilog.Extensions.Hosting


Gosha24:
Hello . In your framework you are using package Serilog.Extensions.Hosting., while in Serilog.Extensions.Hosting documentation it is written: ASP.NET Core applications should consider using Serilog.AspNetCore instead, which bundles this package and includes other ASP.NET Core-specific features.
Can you comment on this?

maliming:
It looks like we need to replace it with `Serilog.AspNetCore`.

johnnba:
![image](https://user-images.githubusercontent.com/2950822/72130611-08ebf500-33b5-11ea-8481-df11cb1acb9d.png)
@Gosha24 @maliming 
actually, if we want to show all logs on console, we should referrence the `Serilog.Aspnetcore` currently . then  we can add the console logger configuration. so it's a good suggestion.

 ```
  Log.Logger = new LoggerConfiguration()
#if DEBUG
                .MinimumLevel.Debug()
#else
                .MinimumLevel.Information()
#endif
                .MinimumLevel.Override("Microsoft", LogEventLevel.Debug)
                .Enrich.FromLogContext()
                .WriteTo.Async(c => c.File("Logs/logs.txt"))
                .WriteTo.Console()
                .CreateLogger();
```

maliming:
Fixed by https://github.com/abpframework/abp/pull/2597

PR #2597, opened 2020-01-09
Serilog enrichments.

### LINKED ISSUE #831: Volo.Abp.AspNetCore.Serilog Package
We want to add some extra properties to logs (e.g. UserId, TenantId, ClientId, CorrelationId... etc.). However, if we create the logger like the [suggested way](https://github.com/serilog/serilog-aspnetcore#instructions), we can not add an Log Enricher which can access to these properties, because in the Program.cs, before WebHost build, we can not resolve any service from DI.

As I've searched, the only way seems to LogContext.PushProperty or LogContext.Push (to push enricher) [like this](https://github.com/serilog/serilog-aspnetcore/issues/38#issuecomment-370669529).

So, we need to have a custom AspNet Core middleware that adds properties/enrichers to the LogContext, used like `app.UseSerilogEnrichers()`.

If there is no better and easier way, it would be good to have a package like `Volo.Abp.AspNetCore.Serilog` which depends on the [Serilog.AspNetCore](https://www.nuget.org/packages/Serilog.AspNetCore) package. It can include the middleware explained above. In addition, it can auto-register Serilog logger factory to remove need to add `.UseSerilog()` here:

````csharp
        WebHost.CreateDefaultBuilder(args)
	    .UseStartup<Startup>()
            .UseSerilog() // <-- Add this line
            .Build();
````

mehmetuken:
Resolve #831
* Adding Serilog enrichments middleware. Adding this properties to LogContext TenantId, UserId, ClientId, CorrelationId.

Using WriteTo.File output template
`{Timestamp:yyyy-MM-dd HH:mm:ss.fff zzz} [{Level:u3}] {Message:lj} {Properties}{NewLine}{Exception}`
or yours using Sinks.MSSqlServer, Sinks.Elasticsearch auto save properties.

Change property names set AbpAspNetCoreSerilogEnrichersOptions.

* Change Serilog.Extensions.Hosting packages to Serilog.AspNetCore
Serilog.Extensions.Hosting says:
> ASP.NET Core applications should consider using Serilog.AspNetCore instead, which bundles this package and includes other ASP.NET Core-specific features.

### DIFF
--- a/framework/Volo.Abp.sln
+++ b/framework/Volo.Abp.sln
@@ -1,4 +1,4 @@
-Microsoft Visual Studio Solution File, Format Version 12.00
+﻿Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio Version 16
 VisualStudioVersion = 16.0.28922.388
 MinimumVisualStudioVersion = 10.0.40219.1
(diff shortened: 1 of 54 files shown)


maliming:
Thanks @mehmetuken  awesome! 💯 

Resolved #2583 issue at the same time.

hikalkan:
Great @mehmetuken I will review this.

PR #2598, opened 2020-01-09
implement hangfire job execution disable

Goxiaoy:
Provide a solution for disable hangfire running backround jobs according to AbpBackgroundJobOptions.

I want to enqueue some background jobs in my program but do not want to execute it in the same program. But according to [hangfire/#225](https://github.com/HangfireIO/Hangfire/issues/225) , there has no way to start/stop job execution gracefully now. So this would be a workaround。


hikalkan:
Thanks @Goxiaoy I will review this.

hikalkan:
I didn't like the `context.Services.BuildServiceProvider()` usage. Delaying this PR to the next version to re-evaluate and think on an alternative solution.

Goxiaoy:
> I didn't like the `context.Services.BuildServiceProvider()` usage. Delaying this PR to the next version to re-evaluate and think on an alternative solution.

I do not like it either. But I am just curious why there is a module called `Volo.Abp.HangFire`, I thought  that HangFire can only be used for background jobs.  I want to config the option `AbpHangfireOptions` according to `AbpBackgroundJobOptions`, but it's hard to implement it without break the current dependence design.

hikalkan:
Agree. This is why I moved to the next milestone. Should think on it more :)

realLiangshiwei:
We have implemented this : https://github.com/abpframework/abp/pull/3382

douglasg14b:
Is there a guide or docs on how that PR solves for this? How do we disable hangfire job execution?

PR #2599, opened 2020-01-09
#215 role name should be unchangeable

### LINKED ISSUE #215: Role name should be unchangeable
* We should provide a display name so we can change it and not allow to change the unique name.
* In addition, if we want to make the role name changeable, we should handle that event and update permissions.

mperk:
Resolve #215 
When the role name is changed, the event handler update permissions


hikalkan:
Thank you @mperk :) We'll review this.

maliming:
hi @mperk 
I think even if the role name can be changed, we should add `DisplayName `to the `IdentityRole`. This way we can provide a description for the role. what do you think?

modules/identity/src/Volo.Abp.PermissionManagement.Domain.Identity/Volo/Abp/PermissionManagement/Identity/RoleUpdateEventHandler.cs
```diff
@@ -0,0 +1,42 @@
...
+
+        public async Task HandleEventAsync(IdentityRoleNameChangedEvent eventData)
+        {
+            var role = await RoleRepository.FindAsync(eventData.IdentityRole.Id, false).ConfigureAwait(false);
+            if (role == null)
+            {
+                return;
+            }
+
+            var permissionGrantsInRole = await PermissionGrantRepository.GetListAsync("R", eventData.OldName).ConfigureAwait(false);
```
maliming: Can replace `"R"` with  `RolePermissionValueProvider.ProviderName` constants.

modules/identity/src/Volo.Abp.PermissionManagement.Domain.Identity/Volo/Abp/PermissionManagement/Identity/RoleUpdateEventHandler.cs
```diff
@@ -0,0 +1,42 @@
...
+            foreach (var permissionGrant in permissionGrantsInRole)
+            {
+                await PermissionManager.UpdateProviderKeyAsync(permissionGrant, eventData.IdentityRole.Name).ConfigureAwait(false);
```
maliming: I'm not sure that the `IPermissionManager` should add the` UpdateProviderKeyAsync` method.

Maybe it can use `IPermissionGrantRepository` directly. so some changes below can be undone.

modules/identity/src/Volo.Abp.PermissionManagement.Domain.Identity/Volo/Abp/PermissionManagement/Identity/RoleUpdateEventHandler.cs
mperk: Ok, I fixed.

modules/identity/src/Volo.Abp.PermissionManagement.Domain.Identity/Volo/Abp/PermissionManagement/Identity/RoleUpdateEventHandler.cs
mperk: ok. I can move in the repo.

mperk:
> hi @mperk
> I think even if the role name can be changed, we should add `DisplayName `to the `IdentityRole`. This way we can provide a description for the role. what do you think?

I am not sure. When the system is using, the role names may be confused after a while. 
This senario can be a problem: 
Ex. a role before change: 
Name: admin
NormalizedName: ADMIN
DisplayName: admin

after change:
Name: admin
NormalizedName: ADMIN
DisplayName: readonly admin

hikalkan:
Thank you @mperk for your contribution.

modules/identity/src/Volo.Abp.PermissionManagement.Domain.Identity/Volo/Abp/PermissionManagement/Identity/RoleUpdateEventHandler.cs
maliming: It seems that we have ignored the above discussion. : )

PR #2619, opened 2020-01-12
refactory and enhance CrudAppService 

yinchang0626:
Hi everyone: 
   Practically,we are very often using `CrudAppService` in begin,so on this pull request,I enhance some function for more flexibility and keep no breaking change.

## 1. `Replace Inheritance With Delegation` : 
add class [`PagedAndSortedOperation` ](https://github.com/yinchang0626/abp/blob/feature/Enhance_CrudAppService/framework/src/Volo.Abp.Ddd.Application/Volo/Abp/Application/PagedAndSortedOperation.cs),
 `Usage:` 
1. Inheritanced from CrudAppService still can override ApplySorting,ApplyPaging to replace method, 
2. Others AppService can reuse it by DI to list data with filter.sort.page...
## 2. `Extract Superclass`:
Original [`CrudAppService`](https://github.com/yinchang0626/abp/blob/eb14d8f01a8b80631755fc7391513162ecf736e0/framework/src/Volo.Abp.Ddd.Application/Volo/Abp/Application/Services/CrudAppService.cs#L62) move to [`BaseCrudAppService`](https://github.com/yinchang0626/abp/blob/feature/Enhance_CrudAppService/framework/src/Volo.Abp.Ddd.Application/Volo/Abp/Application/Services/BaseCrudAppService.cs) except GetAsync(TKey id) ,`BaseCrudAppService` is designed for `TEntity` and `TEntity<TKey>` but not limit in `TEntity<TKey>`.
Because in face legacy db schema,usually must be orm some entity with `Composite Key`,I know this doesn't make sense,but we need to handel  it without modifying primary key although it is a `aggregate root`.
So, I create a [`EntityWithCompositeKeyCrudAppService`](https://github.com/yinchang0626/abp/blob/feature/Enhance_CrudAppService/framework/src/Volo.Abp.Ddd.Application/Volo/Abp/Application/Services/EntityWithCompositeKeyCrudAppService.cs)(base on `BaseCrudAppService`) that without `GetAsync(TKey id)` but have `GetFindAsync(TKey key)` for Get an Entity By Key(like `DbSet.Find`).

 `Usage:` Just like Original `CrudAppService`,let AppService Inheritance `EntityWithCompositeKeyCrudAppService<TEntity, TGetOutputDto, TGetListOutputDto, TKey, TGetListInput, TCreateInput, TUpdateInput>`
note `TKey` is  Composite Key Type not for Entity<>
then you can do Create,Update,Delete,Find(get one by key) for IEntity with Composite Key

#### Known issues:
When use `CompositeKey`,avoid to implement `ISoftDelete`,I doesn't  test soft delete this entity and create the same `CompositeKey` will correct or not.

## 3. Modify `GetListAsync`
Change return type from `TGetListOutputDto`  to `TGetOutputDto`.
According to [best practices](https://github.com/abpframework/abp/blob/dev/docs/en/Best-Practices/Application-Services.md#getting-a-list-of-entities) ,so return a list of detailed DTOs.
and mark original `CrudAppService<TGetOutputDto, TGetListOutputDto, in TKey, in TGetListInput, in TCreateInput, in TUpdateInput>` as `Obsolete` because of no longer need `TGetListOutputDto` Type

If this Pull Request is accepted,I can write some unittest or doc in the future.
Thanks

### DIFF
--- a/framework/src/Volo.Abp.Ddd.Application.Contracts/Volo/Abp/Application/Services/IBaseCrudAppService.cs
+++ b/framework/src/Volo.Abp.Ddd.Application.Contracts/Volo/Abp/Application/Services/IBaseCrudAppService.cs
@@ -1,0 +1,20 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+using Volo.Abp.Application.Dtos;
+
+namespace Volo.Abp.Application.Services
+{
+    public interface IBaseCrudAppService<TGetOutputDto, TGetListOutputDto,in TKey,in TGetListInput,in TCreateInput,in TUpdateInput>
+        : IApplicationService
+        where TGetOutputDto : IEntityDto
+        where TGetListOutputDto : IEntityDto
+    {
+        Task<PagedResultDto<TGetOutputDto>> GetListWithDetailsAsync(TGetListInput input);
+        Task<TGetOutputDto> CreateAsync(TCreateInput input);
+        Task DeleteAsync(TKey id);
+        Task<PagedResultDto<TGetListOutputDto>> GetListAsync(TGetListInput input);
+        Task<TGetOutputDto> UpdateAsync(TKey id, TUpdateInput input);
+    }
+}
--- a/framework/src/Volo.Abp.Ddd.Application.Contracts/Volo/Abp/Application/Services/IBaseCrudAppService.cs
+++ b/framework/src/Volo.Abp.Ddd.Application.Contracts/Volo/Abp/Application/Services/IBaseCrudAppService.cs
@@ -12,9 +12,9 @@ namespace Volo.Abp.Application.Services
         where TGetListOutputDto : IEntityDto
     {
         Task<PagedResultDto<TGetOutputDto>> GetListWithDetailsAsync(TGetListInput input);
-        Task<TGetOutputDto> CreateAsync(TCreateInput input);
-        Task DeleteAsync(TKey id);
         Task<PagedResultDto<TGetListOutputDto>> GetListAsync(TGetListInput input);
+        Task<TGetOutputDto> CreateAsync(TCreateInput input);
         Task<TGetOutputDto> UpdateAsync(TKey id, TUpdateInput input);
+        Task DeleteAsync(TKey id);
     }
 }
--- a/framework/src/Volo.Abp.Ddd.Application.Contracts/Volo/Abp/Application/Services/IBaseCrudAppService.cs
+++ b/framework/src/Volo.Abp.Ddd.Application.Contracts/Volo/Abp/Application/Services/IBaseCrudAppService.cs
@@ -6,13 +6,11 @@ using Volo.Abp.Application.Dtos;
 
 namespace Volo.Abp.Application.Services
 {
-    public interface IBaseCrudAppService<TGetOutputDto, TGetListOutputDto,in TKey,in TGetListInput,in TCreateInput,in TUpdateInput>
+    public interface IBaseCrudAppService<TGetOutputDto,in TKey,in TGetListInput,in TCreateInput,in TUpdateInput>
         : IApplicationService
         where TGetOutputDto : IEntityDto
-        where TGetListOutputDto : IEntityDto
     {
-        Task<PagedResultDto<TGetOutputDto>> GetListWithDetailsAsync(TGetListInput input);
-        Task<PagedResultDto<TGetListOutputDto>> GetListAsync(TGetListInput input);
+        Task<PagedResultDto<TGetOutputDto>> GetListAsync(TGetListInput input);
         Task<TGetOutputDto> CreateAsync(TCreateInput input);
         Task<TGetOutputDto> UpdateAsync(TKey id, TUpdateInput input);
         Task DeleteAsync(TKey id);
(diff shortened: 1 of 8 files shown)


maliming:
hi @yinchang0626 

I don't see the need to introduce these changes.

When you have requirements for custom sorting, paging, etc, you can customize the `XXCrudAppService `or `YYCrudAppService `classes, which inherit `CrudAppService `and override the methods of the base class. This is simple and easy to understand. Replacing the `IPagedAndSortedOperation` component is global.

Based on the name of `IPagedAndSortedOperation `and its methods, I can hardly see that they have any relationship.

Regarding `GetListWithDetailsAsync`, based on best practices, I think we can use `WithDetails `by default, and provide an option to enable or disable it if necessary.

yinchang0626:
hi @maliming 
  As you say ,inherit `CrudAppService` is very easy to override `ApplySorting`,`ApplyPaging`.
But actually,in `YYCrudAppService` ,we often need to create more methods of list data will like:
``` csharp
public class YYCrudAppService : CrudAppService
{
        public virtual async Task<PagedResultDto<AnotherListOutputDto>> GetAnotherListAsync(AnotherListInput input)
        {
            await CheckGetAnotherListPolicyAsync().ConfigureAwait(false);

            var query = createAnotherFilteredQuery(input);
            #region repeat code
            var totalCount = await AsyncQueryableExecuter.CountAsync(query).ConfigureAwait(false);

            query = ApplySorting(query, input);
            query = ApplyPaging(query, input);

            var entities = await AsyncQueryableExecuter.ToListAsync(query).ConfigureAwait(false);
           #region repeat code
            return new PagedResultDto<AnotherListOutputDto>(
                totalCount,
                entities.Select(AnotherListOutputDto).ToList()
            );
        }
}
```
And not to mention if only inherit from AppService,We Will need to write(copy) the same `ApplySorting`,`ApplyPaging` code.  but they are not need to change everywhere
and if we change will like:
```csharp
public class YYCrudAppService : CrudAppService//or AppService
{
        private IPagedAndSortedOperation _pagedAndSortedOperation;

        public virtual async Task<PagedResultDto<AnotherListOutputDto>> GetAnotherListAsync(AnotherListInput input)
        {
            await CheckGetListPolicyAsync().ConfigureAwait(false);

        //you can call  method like below in evey AppService and not limit in CrudAppService(YYCrudAppService can not inherit `CrudAppService`)
            var result = await PagedAndSortedOperation.ListAsync(
                input,
                (x) => createAnotherFilteredQuery(input))

            return CreatePagedResultDto<TGetListOutputDto>(result);
        }
}
```

Since we already have `PagedAndSortedResultRequestDto`,`IPagedAndSortedResultRequest`,`IPagedResultRequest`,`ILimitedResultRequest`..., For DRY SoC ,I think if want to do a list method,should focus on input ,query map to dto ,and output,default page operation can be handle by other class.
I agree maybe name of `IPagedAndSortedOperation ` is not clear and  should it be a global component or not .but I hope `CrudAppService` is more flexibility,more just only think about domain knowledge ,thanks

yinchang0626:
@maliming 
About `GetListWithDetailsAsync`:
I agree you say ,in one GetListAsync that use `WithDetails` by default and provide an option to  change behavior,
but maybe will let it to `breaking change`.At lease below method signatures TGetListOutputDto may be will be TGetListOutputDto or TGetOutputDto :
```csharp
 Task<PagedResultDto<TGetListOutputDto>> GetListAsync(TGetListInput input)
```
do you have some advised or I can try to do it without consider `breaking change`  :)

maliming:
I think it can call `WithDetails` of `Repository ` in the `CreateFilteredQuery `method.
```
/// <summary>
/// This method should create <see cref="IQueryable{TEntity}"/> based on given input.
/// It should filter query if needed, but should not do sorting or paging.
/// Sorting should be done in <see cref="ApplySorting"/> and paging should be done in <see cref="ApplyPaging"/>
/// methods.
/// </summary>
/// <param name="input">The input.</param>
protected virtual IQueryable<TEntity> CreateFilteredQuery(TGetListInput input)
{
	return Repository.WithDetails();
}
```

yinchang0626:
Yes,no problem, GetListAsync will call WithDetails of Repository when result with Detail
but question is when return dto ,if provide an option to decide what kind of dto type. 
method signatures like below, it can return both of TGetOutputDto and  TGetListOutputDto by client:
```csharp
 Task<PagedResultDto<TWhatKindOfDto>> GetListAsync(TGetListInput input,bool withDetails)
type of TWhatKindOfDto :
if(withDetails==true)
   TWhatKindOfDto is TGetOutputDto 
if(withDetails==false)
   TWhatKindOfDto is TGetListOutputDto
```


I think need a breaking change or not to do this feature ,
or just follow Repository.WithDetails(), and dto become only one TGetOutputDto and no longer need TGetListOutputDto, but in same simple usage scenarios(ex: just for options of select ) ,It still join others table in query.

hikalkan:
It is a bit long to read, I added to milestore 2.1 to review it deeply and decide to merge or cancel (you can continue to discuss if needed :))

johnnba:
@yinchang0626   i agree the  Composite Key don't make sense with ddd; but it's more practical in the real project;  so i'm appreciated it; 
and i have one more demand eventually. in the common crudappservice, we also need to find or get the  dto lists directly call the sql sentence  in some occasionally case  such as sql view or procedure.  
so will you also enhance the findDtoBySql/GetListBySql async function directly by sql  in the crud app service?
i see it  has the pratical value worth the enhancement in the crud app service.

yinchang0626:
@johnnba 
Thank for your reply.
Just like you said,I am working on legacy db with abp module,
I think support  `sql view or procedure` to list data is more simple than do it in a new `DomainService` or `Specification` now .
But we know abp is a framework base on `DDD` .
I can't decide what feature  can be include in abp.
Maybe @hikalkan can tell us does abp need those feature  of  legacy db? ,and then how to implement (should it in CrudAppService or Repository or others...)
If necessary,I can contribution what I do,thanks.

hikalkan:
Hi all,

I reviewed this. The idea is good and the implementation is also good. However, I need to think more on the implementation details, so moved this to the next milestone. Sorry for the delay.

hikalkan:
I re-implemented this by https://github.com/abpframework/abp/issues/3162
I hope it satisfies your requirements.

PR #2623, opened 2020-01-13
correct chinese localization

RayMMond:
Correct chinese translations in localization resources.


maliming:
Thank you for correcting them.

PR #2670, opened 2020-01-17
EnablePostSignOutAutoRedirect = true

### LINKED ISSUE #2519: EnablePostSignOutAutoRedirect = true
In the identity server, we have to set EnablePostSignOutAutoRedirect = true.

Currently, when we set PostLogoutRedirectUris to an URL, we are not redirected after logged out !

PS: I suppose it is the issue, can you confirm that ?

gterdem:
Resolves #2519 after when updated account logout is merged.


framework/src/Volo.Abp.AspNetCore.Mvc/Volo/Abp/AspNetCore/Mvc/Authentication/ChallengeAccountController.cs
```diff
@@ -42,7 +42,12 @@ public async Task<ActionResult> Logout(string returnUrl = "", string returnUrlHa
         {
             await HttpContext.SignOutAsync().ConfigureAwait(false);
 
-            return RedirectSafely(returnUrl, returnUrlHash);
+            if (HttpContext.User.Identity.AuthenticationType == "Identity.Application")
```
maliming:
```
"Identity.Application"
```

hi @gterdem 
This seems to be Identity's cookie scheme. Do we assume that developers are using the identity module?

gterdem: hi @maliming,
Yes. Since it's redirecting to local domain. But I may be missing any other scenarios where developer is not using identity module.

hikalkan:
Please delete LogoutController.cs if not needed.

gterdem:
Done via #2691.

PR #2675, opened 2020-01-18
fix templates/module something wrong

### LINKED ISSUE #2665: dotnet build command not worked on generated module
If I generate module online (on abp.io webpage), dotnet build compiles everything successfully (with some warnings), but if the module is generated via abp CLI (2.0.0), there are errors (dotnet restore command succeeded).

Errors:
image

The module was created with command: abp new MyMod1 -t module

Online generated module was without UI.

yinchang0626:
resolve #2665 
By the way fix `MyProject` to `MyProjectName` wrong name,so cli didn't rename.
and maybe when start `HttpApi.Host` should be use `MyProjectName_Module` database not `Default`


hikalkan:
Great! Thank you for your contribution.

PR #2683, opened 2020-01-20
framework & module & template & sample to target .NET Standard 2.0

### LINKED ISSUE #2668: Use ABP in a mobile(Xamarin) application (.Net Standard 2.0)
I would like to use ABP to share domain, application, and data access classes across the various user interfaces: desktop, web, and mobile. I can't use ABP in Xamarin because some projects (modules) are set to .Net Standard 2.1 and Xamarin only supports .Net Standard 2.0:
-Volo.Abp.EntityFrameworkCore
-Volo.Abp.EntityFrameworkCore.Sqlite

Would a downgrade to .Net Standard 2.0 be possible to standardize and make it compatible with Xamarin?

I didn't find an example of using ABP with Xamarin as there was before in Boilerplate. Is it possible ABP + Xamarin?

Any guidance on this?

maliming:
Resolve #2668

### DIFF
--- a/framework/src/Volo.Abp.Dapper/Volo.Abp.Dapper.csproj
+++ b/framework/src/Volo.Abp.Dapper/Volo.Abp.Dapper.csproj
@@ -3,7 +3,7 @@
   <Import Project="..\..\..\common.props" />
 
   <PropertyGroup>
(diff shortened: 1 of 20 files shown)


hikalkan:
Great! Thanks @maliming

PR #2695, opened 2020-01-22
Remove the entity default index of the identity server module.

### LINKED ISSUE #2451: SQL Server Warning on Clustered Indexes
When I generated a SQL Server 2019 script for the default tables and run it, I get the following warnings:
* Warning! The maximum key length for a clustered index is 900 bytes. The index 'PK_IdentityServerApiSecrets' has maximum length of 8516 bytes. For some combination of large values, the insert/update operation will fail.
* Warning! The maximum key length for a clustered index is 900 bytes. The index 'PK_IdentityServerClientClaims' has maximum length of 1016 bytes. For some combination of large values, the insert/update operation will fail.
* Warning! The maximum key length for a clustered index is 900 bytes. The index 'PK_IdentityServerClientPostLogoutRedirectUris' has maximum length of 4016 bytes. For some combination of large values, the insert/update operation will fail.
* Warning! The maximum key length for a clustered index is 900 bytes. The index 'PK_IdentityServerClientRedirectUris' has maximum length of 4016 bytes. For some combination of large values, the insert/update operation will fail.
* Warning! The maximum key length for a clustered index is 900 bytes. The index 'PK_IdentityServerClientSecrets' has maximum length of 8516 bytes. For some combination of large values, the insert/update operation will fail.

maliming:
Resolve #2451
based on https://github.com/IdentityServer/IdentityServer4/blob/master/src/EntityFramework.Storage/src/Extensions/ModelBuilderExtensions.cs

**Need to add or recreate migrations.**

### DIFF
--- a/modules/identityserver/src/Volo.Abp.IdentityServer.EntityFrameworkCore/Volo/Abp/IdentityServer/EntityFrameworkCore/IdentityServerDbContextModelCreatingExtensions.cs
+++ b/modules/identityserver/src/Volo.Abp.IdentityServer.EntityFrameworkCore/Volo/Abp/IdentityServer/EntityFrameworkCore/IdentityServerDbContextModelCreatingExtensions.cs
@@ -63,16 +63,12 @@ namespace Volo.Abp.IdentityServer.EntityFrameworkCore
             {
                 grantType.ToTable(options.TablePrefix + "ClientGrantTypes", options.Schema);
 
-                grantType.HasKey(x => new { x.ClientId, x.GrantType });
-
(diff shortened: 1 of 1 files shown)


hikalkan:
@maliming you are removing the primary keys (they are composite PK here) of the tables, not their indexes.

As I see the IDS4 implementation, they defined Id properties to the entities. But our entities has no Id property. We are implementing DDD and these entities are not aggregate roots, so we didn't define an unnecessary Id.

I don't know how to solve the problem. Maybe we need to decrease the total length of the primary keys, or define an Id property for these entities (which will require the code change on the entities and other services). This would be a big breaking change. We can go with the first option if we need.

PR #2713, opened 2020-01-25
fix docs wrong code that causes to all test failed

yunusd:


hikalkan:
Thank you @yunusd for your contribution.

But, which tests fail, can you point?

yunusd:
> Thank you @yunusd for your contribution.
> 
> But, which tests fail, can you point?

Current documentation is pointing that Book entity has Id prop while don't. Abviously this causing compile time error. When we try to seed Book data without Id in BookStoreTestDataSeedContributor.cs all tests failing. So test section of documentation can't be done with current docs.

Screenshot from 2020-01-27 10-35-33

acjh:
Caused by https://github.com/abpframework/abp/commit/8f2de58e4412533290fcf61ef407a6b429bfcabb (0.20.0)

Follows https://github.com/abpframework/abp/commit/1b35606bdab4464520d385b355082ec422241e23 (0.22.0) for `Book`, `BookStoreTestDataSeedContributor`
Follows https://github.com/abpframework/abp/commit/c5cf2ae24a353fbe948dec2a1ffa430df7b0053d (1.0.0) for `Book`

hikalkan:
Thank you @yunusd :)

Issue #2615, opened 2020-01-10
partial docs


hikalkan:
I'm writing the FluentValidation documentation which includes such a section:

![image](https://user-images.githubusercontent.com/1210527/72153722-c11e9b00-33bf-11ea-9242-316fba911061.png)

This section is repeated for most of the ABP modules. Instead of copy/paste for all the documentation, such a feature would be nice:

* Allow to include (embed) a document into another document.
* Allow to pass parameters to the inner document. We can reuse the parameter system introduced before. In the partial (inner) document, we can use the parameters for conditional rendering or use their values.

PR #2722, opened 2020-01-28
Partial docs

yekalkan:
resolves https://github.com/abpframework/abp/issues/2615


modules/docs/src/Volo.Docs.Domain.Shared/Volo/Docs/Projects/ProjectConsts.cs
```diff
@@ -7,6 +7,7 @@ public static class ProjectConsts
         public const int MaxDefaultDocumentNameLength = 128;
         public const int MaxNavigationDocumentNameLength = 128;
         public const int MaxParametersDocumentNameLength = 128;
+        public const int MaxPartialTemplatesDocumentNameLength = 128;
```
hikalkan: This is not used?

PR #2727, opened 2020-01-29
Updated abp.io document for commercial abp.io

AkinSabriCam:


framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/ProjectBuilding/Templates/App/AppTemplateBase.cs
```diff
@@ -10,7 +10,11 @@ public abstract class AppTemplateBase : TemplateInfo
         protected AppTemplateBase(string templateName)
             : base(templateName, DatabaseProvider.EntityFrameworkCore, UiFramework.Mvc)
         {
-            DocumentUrl = "https://docs.abp.io/en/abp/latest/Startup-Templates/Application";
+            if(templateName=="app-pro")
+                DocumentUrl= "https://docs.abp.io/en//commercial/latest";
+            else
+                DocumentUrl = "https://docs.abp.io/en/abp/latest/Startup-Templates/Application";
+
```
maliming: hi 

I think we should modify it in a derived class of `Template`. Like `ModuleProTemplate`.

https://github.com/abpframework/abp/blob/f901f909192aed7fad7d7635ebd8de3b3c6de26c/framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/ProjectBuilding/Templates/Module/ModuleProTemplate.cs#L16

AkinSabriCam: Hi thank you for your feedback.
If I modify it in ModuleProTemplate , I would change just pro doc link of moduleTemplate thats why I changed it in AppTemplateBase
am i wrong? waiting  for your feedback

maliming: hi @AkinSabriCam 
I mean setting `DocumentUrl` in the constructor of the `AppProTemplate `class.

AkinSabriCam: Hi @maliming 
I got it now. 
I will modify it.  Thank you

Issue #2026, opened 2019-10-28
Slow performance on Linux


stay2be:
Hi folks,

first let me thank you for this great project. I've learned a lot while working with this framework and the design principles behind it.

I've run into the following problem when deploying a project based on abp to an Azure App Service Linux instance. I'm not sure if this is related to abp or maybe to Identity Server 4.

**About the environment:**

- Standard template generated via: https://www.abp.io/get-started (direct download)
- Switch from MSSQL to Postgres: https://docs.abp.io/en/abp/latest/Best-Practices/PostgreSQL-Integration
- Deployment to Azure App Service Linux:
  - Url: https://abp-test-app.azurewebsites.net/
  - Database: Managed Postgres on Digital Ocean
  - App Service Plan: B1
- Gitlab repo: https://gitlab.com/abp-test/testapp

**Findings:**

- All non Identity Server endpoints working fine:
  - https://abp-test-app.azurewebsites.net/swagger/index.html
  - https://abp-test-app.azurewebsites.net/api/test (including db calls)
- Endpoints, where Identity Server 4 seams to be contacted are slow or not responding:
  - https://abp-test-app.azurewebsites.net/.well-known/openid-configuration
  - https://abp-test-app.azurewebsites.net/api/abp/application-configuration
  - https://abp-test-app.azurewebsites.net/Account/Login
  - Generating tokens:
    - https://abp-test-app.azurewebsites.net/connect/token
```
POST https://abp-test-app.azurewebsites.net/connect/token
Content-Type: application/x-www-form-urlencoded

grant_type:password&client_id:TestApp_App&client_secret:1q2w3e*&username:admin&password:1q2w3E*&scope:TestApp
```

Going into the logs yields that following:
- After restarting the app in Azure the call to /connect/token succeeds
- A second some seconds after the first call after yields no response

```
// First request
2019-10-28 09:52:14.338 +00:00 [INF] Request starting HTTP/1.1 POST http://abp-test-app.azurewebsites.net/connect/token application/x-www-form-urlencoded 113
2019-10-28 09:52:14.663 +00:00 [INF] Invoking IdentityServer endpoint: IdentityServer4.Endpoints.TokenEndpoint for /connect/token
[...]

PR #2807, opened 2020-02-11
Soft Delete enhancements #139

### LINKED ISSUE #139: Soft Delete enhancements
- [ ] Provide a way of undelete.
- [ ] Provide a way of hard delete.

Even we can consider to add an enum to all entities (like Active, InActive, Deleted) or define this enum property for SoftDelete

gterdem:
Resolves #139 

Added `HardDelete` and `HardDeleteAsync` methods to `RepositoryExtensions`.
Added HardDelete Tests for EfcoreRepository and MongoDbRepository.


framework/src/Volo.Abp.Ddd.Domain/Volo/Abp/Domain/Entities/EntityHelper.cs
```diff
@@ -109,5 +112,26 @@ public static Type FindPrimaryKeyType([NotNull] Type entityType)
...
         }
+
+        public static object GetEntityId(object entity)
+        {
+            if (!IsEntity(entity.GetType()))
+            {
+                throw new AbpException(entity.GetType() + " is not an Entity !");
+            }
+
+            return ReflectionHelper.GetValueByPath(entity, entity.GetType(), "Id");
```
maliming: hi
`entity` that implement the `IEntity `interface do not have an `Id` property.

framework/src/Volo.Abp.Ddd.Domain/Volo/Abp/Domain/Repositories/RepositoryExtensions.cs
```diff
@@ -41,5 +43,59 @@ public static class RepositoryExtensions
...
+            }
+        }
+        public static void HardDelete<TEntity, TPrimaryKey>(this IRepository<TEntity, TPrimaryKey> repository, Expression<Func<TEntity, bool>> predicate)
+            where TEntity : class, IEntity<TPrimaryKey>, ISoftDelete
+        {
+            foreach (var entity in repository.Where(predicate).ToList())
+            {
+                repository.HardDelete(entity);
+            }
+        }
```
maliming: hi  

Repositories are no longer providing synchronization methods.
https://github.com/abpframework/abp/issues/2026#issuecomment-568681191

framework/src/Volo.Abp.Ddd.Domain/Volo/Abp/Domain/Entities/EntityHelper.cs
gterdem: Hello maliming,
So you mean if an entity is inherited from `Entity` (or `IEntity`) it won't pass this check which will be causing problems. Do I get it right?

framework/src/Volo.Abp.Ddd.Domain/Volo/Abp/Domain/Repositories/RepositoryExtensions.cs
gterdem: Yeah makes sense. Will remove sync methods in next commit.

framework/src/Volo.Abp.Ddd.Domain/Volo/Abp/Domain/Entities/EntityHelper.cs
maliming: The Entity will pass the check, but it does not have an `Id `property. Only entities that inherit the `IEntity<IdKeyType>` interface will have an `Id` property.

gterdem: Ok got it now..
So it should be something like;
```
if (!ReflectionHelper.IsAssignableToGenericType(entity.GetType(), typeof(IEntity<>)))
{
     throw new AbpException(entity.GetType() + " is not an Entity with an Id property !");
}
```
Seems good to go by this?

maliming: I think the `GetKeys `method of entity can be used. This way we can delete the entities that implement `IEntity `and `IEntity<T>`.

https://github.com/abpframework/abp/blob/654eeede81a733382066bfd00ab19b6b65ceec0e/framework/src/Volo.Abp.Ddd.Domain/Volo/Abp/Domain/Entities/Entity.cs#L18

gterdem: I am not sure if it is viable. If an entity has no `Id` property, it can be many-to-many entity without Id property (juntion table) which we wouldn't want to hard delete them all right?

I may have lack of knowledge about use cases of `Entity` and `IEntity` though.

maliming: There are some documents we can refer to.
https://docs.abp.io/en/abp/latest/Entities#entities-with-composite-keys

hikalkan:
I merged by making some improvements and refactoring: https://github.com/abpframework/abp/commit/fc8d34eae94bb628ea170321fe8236fa4eb73b9e

PR #2808, opened 2020-02-11
Add multi tenant migrations to the migrator console application

### LINKED ISSUE #1371: Add multi tenant migrations to the migrator console application
In the `aspnetboilerplate` framework running the `Migrator` app would migrate the host database and all tenant databases that had a separate connection string.

Currently the migrator tool only migrates the host database and ignores tenants with own databases and connection strings.

Also the connection string is currently not stored encrypted.

gterdem:
Resolves #1371.

### DIFF
--- a/templates/app/aspnet-core/src/MyCompanyName.MyProjectName.Domain/Data/MyProjectNameDbMigrationService.cs
+++ b/templates/app/aspnet-core/src/MyCompanyName.MyProjectName.Domain/Data/MyProjectNameDbMigrationService.cs
@@ -1,8 +1,12 @@
-﻿using System.Threading.Tasks;
+﻿using System;
+using System.Linq;
+using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Logging.Abstractions;
 using Volo.Abp.Data;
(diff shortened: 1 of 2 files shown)


leonkosak:
@hikalkan is somehow possible to have one migrator application and arguments for seed-ing host and tenant databases?

hikalkan:
Yes, possible. Currently, it migrates all, you can easily make it parametric. The source code is in your solution.

leonkosak:
@hikalkan This solution is great, but the better solution would be that IMyProjectNameDbSchemaMigrator would be "split":

- IMyProjectNameHostDbSchemaMigrator
- IMyProjectNameTenantSchemaMigrator

Of course, this also means two different OnModelCreating methods, but this should not be the "deal-breaker" if SQL tables for specific modules should be in host and tenant databases.
For such situation, the developer should be aware that ConfigureModuleName() method should be defined in OnModelCreating method for host and for tenant.

What do you think about this solution?

hikalkan:
see https://github.com/abpframework/abp/issues/2993

leonkosak:
I started, but then I realized that in my case, this is not useful enough to finalize development.

PR #2826, opened 2020-02-14
Cache docs in local database & topic labels.

### LINKED ISSUE #1994: Documentation: topic labels
Since your documentation is in active development, I would like to suggest to mark recently added documents with [new] label in the documentation tree. This way we can track changes and learn new framework features as soon as you update the documentation. 

Devexpress documentation for example:
image

### LINKED ISSUE #2805: Cache documents in the local database for the documents module
* Create a database table (and corresponding entity) to store copies of the documents.
* Create an application service method (authorized) to read all the documents from the data source and store in the database. This is needed initially save all the docs. It is also useful to refresh the cached docs.
* Continue to the current logic to get the docs from the original source (with the caching system - this time works with database instead of redis cache. so the database record can be a LastCachedTime field for example).
* Refresh the cache whenever you get the doc from the original source.

This caching system will then be used to develop https://github.com/abpframework/abp/issues/1621 and https://github.com/abpframework/abp/issues/1994

maliming:
Resolve #2805 , resolve #1994 

1. Store `Document `and `DocumentContributors `as entities into the database. Documents have different language codes and versions. Rename `IDocumentStore `to `IDocumentSource`.

2. Determine whether the file is `Upd`or `New` according to the file's commits.
- **Question: A new document may be modified multiple times in a short time. Although it has multiple commits, it is still a new document.**

- **Question: How many days have a document changed that should show the Upd label.** I specify 30 days by default. https://github.com/abpframework/abp/pull/2826/files#diff-7bf979fec9bcb043de25e87457696494R127

3. Added file last modified time and contributor information.

image

4. We can pull single or all documents in project management. Will replace documents in the database.

image

image


gterdem:
> Added file last modified time and contributor information.

I want to suggest enhancement about this topic; 
What do you think about new changes have different background color on UI for the duration of upd label so it could be easier to notice?

maliming:
**There are also performance issues to be optimized.**

maliming:
> What do you think about new changes have different background color on UI for the duration of upd label so it could be easier to notice?



image

Ice2burn:
If you edit a typo after 30 days topic will be marked as UPD?

hikalkan:
We were discussing it today :) We'll try to find a proper way (like the amount of the edit). If you have a good suggestion, it is appreciated.

Ice2burn:
I was thinking about settings this up manually in Navigation Document. New json section "labels" or parameter "label" for example. And this problem will be closed for json-based documentation. Hovewer there is a chance that developer will forget to mark upper level nodes, but it's not your problem :)

In terms of github-based documentation, I need a bit of research to see what info available. I think New and Upd labels should appear only after version change. So checking edit date is not enough.

hikalkan:
Following with https://github.com/abpframework/abp/issues/2843

PR #2867, opened 2020-02-20
#2294 Make `Http.Client` projects usable by traditional netfx projects

### LINKED ISSUE #2294: Volo.Abp.Http.Client.IdentityModel must specify Microsoft.NET.Sdk.Web sdk ?
I use abp as my services framework. and call the api by the **Volo.Abp.Http.Client.IdentityModel** on the WPF application.
So we develop the wpf application to the new machine, it can not run because it lacks of asp.netcore runtimes.
Then we find the cause is Volo.Abp.Http.Client.IdentityModel. It depend on the **Microsoft.NET.Sdk.Web**,  So need to install aspnetcore runtime to run it successful. 

Can replace the Sdk.Web with Sdk ? 

Thanks in advance.

NecatiMeral:
Resolve #2294
To support .NET-Framework based projects I've updated the corresponding projects to target netstandard2.0. The `Volo.Abp.Http.Client.IdentityModel` project hasn't been touched because of the `IHttpContextAccessor` dependency and I think it should be used by DotnetCore (Hosts) only.


framework/src/Volo.Abp.Http.Client/Volo.Abp.Http.Client.csproj
```diff
@@ -1,18 +1,16 @@
...
-    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <TargetFramework>netstandard2.0</TargetFramework>
     <AssemblyName>Volo.Abp.Http.Client</AssemblyName>
     <PackageId>Volo.Abp.Http.Client</PackageId>
     <AssetTargetFallback>$(AssetTargetFallback);portable-net45+win8+wp8+wpa81;</AssetTargetFallback>
     <GenerateAssemblyConfigurationAttribute>false</GenerateAssemblyConfigurationAttribute>
     <GenerateAssemblyCompanyAttribute>false</GenerateAssemblyCompanyAttribute>
     <GenerateAssemblyProductAttribute>false</GenerateAssemblyProductAttribute>
-    <IsPackable>true</IsPackable>
-    <OutputType>Library</OutputType>
```
maliming: hi 
Why remove these?

NecatiMeral: I removed these to be compliant with other netstandard projects in the framework directory.

NecatiMeral:
I've found a missing detail in my customization. The `IdentityModelRemoteServiceHttpClientAuthenticator` does in fact handle two different authentication actions:

1. Relay the existing bearer-token in the current request
2. Apply the `IIdentityModelAuthenticationService`

So the implementation details for these actions must be splitted. I'll edit my PR asap.

NecatiMeral:
Hi @maliming,

I've updated my PR to properly work on netfx and netcore frameworks.
Tested with a client service like [ClientDemoService](https://github.com/abpframework/abp/blob/dev/samples/MicroserviceDemo/applications/ConsoleClientDemo/ClientDemoService.cs) with identity-server and users by running a netcore backend with netcore and netfx clients.

maliming:
hi  @NecatiMeral 

I merged some conflicts.
I'm not sure whether the cross-platform code conforms to the design style of abp. Because abp is all based on net core. Of course you can use this code in your project to replace the default implementation of abp. Including `Volo.Abp.Http.Client.IdentityModel. Relay` project.

NecatiMeral:
hi @maliming 

Thanks for your efforts. Our preferred solution in the long-term view would be to use the upstream modules / packages instead of customizing every bit just for our needs. I think there are multiple developers who are requiring netstandard-clients for different scenarios (mobile, winforms, ...).

I know that abp was created because of the huge changes in netcore3.1. For the backend aspect it's totally fine, but maybe abp should consider this compatibility for clients. the effort to maintain netstandard compatibility on client-side is pretty low.

We can't use abp if the built-in-clients cannot be used in netstandard-projects because we don't want to manually generate our clients for each api version (with [AutoRest](https://github.com/Azure/autorest) for example).

hikalkan:
I simplified this implementation: https://github.com/abpframework/abp/pull/2951
@NecatiMeral can you check if it has any problem?
I removed some classes. I especially didn't like the EnvironmentHelper which is a little ugly.
What do you think?

hikalkan:
I also didn't find the .Relay project naming useful. I will change it as .Web.

NecatiMeral:
Thanks for your feedback. I'll review the changes tomorrow.

NecatiMeral:
hi @hikalkan,

thank you for the review / updates. Looks good to me!

Issue #2837, opened 2020-02-16
ABP CLI Replace Packages with Previews


hikalkan:
We re-started to release nightly previews. However, it is tedious to replace all NuGet/NPM packages with the previews and return to stables when its released.

So, add a new command to the ABP CLI to replace all NuGet/NPM packages with the previews and vice verse. Example commands:

````
abp --switch-to-preview
abp --switch-to-stable
````

If anyone has a better command name, it is appreciated :)

**Comment by acjh** [comment]
How about:

```sh
abp sources switch --preview
abp sources switch --stable
```

Inspired by `nuget sources`: https://docs.microsoft.com/en-us/nuget/reference/cli-reference/cli-ref-sources

hikalkan:
Good suggestion. But "source" has a single meaning for NuGet, while it may mean "the source of the modules" for abp. We have/will another command to replace a module reference by its source code, so a root "sources" command would not be suitable. Maybe "abp package-sources -p" and "abp package-sources -s" (long versions are --preview and --stable).

acjh:
> replace a module reference by its source code

Do you mean replace a remote reference with a local reference?

hikalkan:
You know, startup project is using the modules as NuGet packages. For example, for the Identity module, you don't have the source code. We will try to add a command to CLI to remove Identity Nuget packages, download and add [its source code](https://github.com/abpframework/abp/tree/dev/modules/identity) directly into the current solution, so you can freely modify it.

PR #2868, opened 2020-02-20
ABP CLI Replace Packages with Previews

yekalkan:
resolves https://github.com/abpframework/abp/issues/2837


framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/ProjectModification/PackageSourceSwitcher.cs
```diff
@@ -0,0 +1,179 @@
...
+
+            var versionAttr = doc.CreateAttribute("Version");
+            versionAttr.Value = newVersion;
+            newNode.Attributes.Append(versionAttr);
+            return newNode;
+        }
+
+        private string GetFormattedDate()
+        {
+            var formattedDate = "";
```
soyhanbeyazit: Is there any specific reason to have a method like this? AFAIK, `DateTime.Now.ToString("yyyyMMdd")` should be enough to get an output like this `20200221`

hikalkan:
Does this cover NPM packages?

yekalkan:
@hikalkan No

hikalkan:
>@hikalkan No

then it should do.

maliming:
hi @yekalkan Has this PR been completed?

yekalkan:
> hi @yekalkan Has this PR been completed?

No. I'm waiting for npm-preview packages.

PR #2880, opened 2020-02-21
Fixed the bug that the controller was not logged.

### LINKED ISSUE #2819: Audit Log system doesn't log controller actions by default
If a request only consists of a controller action call and not calls an app service, then audit log system doesn't log it.
You can add [Audited] to the controller to make it logged, but this should not be needed when I check the documentation.

maliming:
Resolve #2819

I filtered non-controller requests in the  `AbpAuditingMiddleware`.
https://github.com/abpframework/abp/pull/2880/files#diff-d157839becffb19531fffdc9b4346149R67-R73

Does the framework need to audit requests for razor pages?
If yes, I need to remove the filtering code.


hikalkan:
>Does the framework need to audit requests for razor pages?

Yes, we want to write audit log for all kind of requests.

maliming:
Filtering code has been removed.

PR #2901, opened 2020-02-24
Add document full text search function (elastic search).

maliming:
We should design a search page, which can also be used for support forums.

This is a very simple search results page I made.

image

image


maliming:
hi @armgnunlu 
Can you help design a search results page? We can use it in docs and qa modules.

hikalkan:
I moved this to the next milestone to review it deeper and work on the UI design.

modules/docs/app/VoloDocs.Web/appsettings.json
```diff
@@ -1,4 +1,8 @@
 ﻿{
   "ConnectionString": "Server=localhost;Database=VoloDocs;Trusted_Connection=True;MultipleActiveResultSets=true",
-  "LogoUrl": "/assets/images/Logo.png"
+  "LogoUrl": "/assets/images/Logo.png",
+
+  "ElasticSearch": {
+    "Url": "http://172.25.134.11:9200"
```
yekalkan: We can change `http://172.25.134.11:9200` to `http://localhost:9200`

yekalkan:
also, see https://github.com/abpframework/abp/issues/3009. We can think about it after this is merged.

maliming:
@yekalkan pr has been updated.

ebicoglu:
we need to add Elastic Search usage in https://docs.abp.io/en/abp/latest/Modules/Docs
Right now the project is not running without Elastic Search setup.

PR #2912, opened 2020-02-25
Make AbpResourceOwnerPasswordValidator return detailed reason.

maliming:


hikalkan:
It think it is not a good idea to localize error messages received by machines (not by humans) and written to logs. Becuase, the developer/person reading the logs may not know the language of the client throws the exception.

Error logs should be in English unless its shown to end users.

Thanks.

maliming:
This PR is for https://github.com/volosoft/volo/issues/1248

hikalkan:
Thanks @maliming you're great! :)

PR #2921, opened 2020-02-26
Fix scopes may be null bug.

maliming:
`IdentityScopes` and  `apiScopes` may be null.


hikalkan:
If so, we should also check null for https://github.com/abpframework/abp/pull/2921/files#diff-b38b99661c592c1e1baf3ff3d8d09f25R109 ?

maliming:
You are right.  I will update it.

hikalkan:
Thanks.

PR #2937, opened 2020-02-28
Use `IHttpClientFactory` instead of manually creating a `HttpClient`

NecatiMeral:
This enables customization of the created client or it's handlers from outside of this module.

`IHttpClientFactory` provides different endpoints to customize the created clients. One of them would be to implement a custom `IHttpMessageHandlerBuilderFilter` and register it in the ioc framework.

My use case is a validation of self-signed certificates that aren't known to the clients or injecting a [WinHttpHandler](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.winhttphandler?view=dotnet-plat-ext-3.1) into the httpclient automatically to be used by .net-fx projects which use the `Volo.Abp.IdentityModel` package.


maliming:
hi @NecatiMeral 

Since there is no `AddHttpClient `method to register `IHttpClientFactory`. It will cause dependency injection problems.

PR #2953, opened 2020-03-03
Revert "Use `IHttpClientFactory` instead of manually creating a `HttpClient`"

maliming:
Reverts abpframework/abp#2937

Since there is no `AddHttpClient `method to register `IHttpClientFactory`. It will cause dependency injection problems.

image

### DIFF
--- a/framework/src/Volo.Abp.IdentityModel/Volo.Abp.IdentityModel.csproj
+++ b/framework/src/Volo.Abp.IdentityModel/Volo.Abp.IdentityModel.csproj
@@ -16,7 +16,6 @@
 
   <ItemGroup>
     <PackageReference Include="IdentityModel" Version="4.1.1" />
-    <PackageReference Include="Microsoft.Extensions.Http" Version="3.1.2" />
     <ProjectReference Include="..\Volo.Abp.Threading\Volo.Abp.Threading.csproj" />
   </ItemGroup>
 
--- a/framework/src/Volo.Abp.IdentityModel/Volo/Abp/IdentityModel/IdentityModelAuthenticationService.cs
+++ b/framework/src/Volo.Abp.IdentityModel/Volo/Abp/IdentityModel/IdentityModelAuthenticationService.cs
@@ -21,15 +21,12 @@ namespace Volo.Abp.IdentityModel
         public ILogger<IdentityModelAuthenticationService> Logger { get; set; }
         protected AbpIdentityClientOptions ClientOptions { get; }
         protected ICancellationTokenProvider CancellationTokenProvider { get; }
-        protected IHttpClientFactory HttpClientFactory { get; }
 
         public IdentityModelAuthenticationService(
             IOptions<AbpIdentityClientOptions> options,
-            ICancellationTokenProvider cancellationTokenProvider,
-            IHttpClientFactory httpClientFactory)
+            ICancellationTokenProvider cancellationTokenProvider)
         {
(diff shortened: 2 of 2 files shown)


NecatiMeral:
hi @maliming,

in the current dev branch there's a `AddHttpClient` method inside the `IdentityModelModule`.
Can u re-validate this one?

Would be nice if this could land in 2.2.

maliming:
hi @NecatiMeral 

`dev` branch now has build errors.  see https://github.com/abpframework/abp/pull/2962

Can you open a new PR? :) 

And should we use named Clients?  https://docs.microsoft.com/en-us/aspnet/core/fundamentals/http-requests?view=aspnetcore-3.1#named-clients

NecatiMeral:
I'll open a new PR that hopefully sorts this out.
The PR should be up in no time.

NecatiMeral:
hi @maliming,

I'm currently workin on a PR and was evaluating the benefit of using named clients inside the identitymodel-module. Since the client is only used to fire the IdentityModel requests (such as `PasswordTokenRequest` or `ClientCredentialsTokenRequest`) the client couldn't be much more pre-configured as far I see. 

We're using the `IHttpClientFactory` to implement our own certificate checking, which doesn't work here yet.

maliming:
> We're using the IHttpClientFactory to implement our own certificate checking, which doesn't work here yet.

Is this certificate error?
`The certificate cannot be verified up to a trusted certification authority`

NecatiMeral:
Look so.
With the `IHttpClientFactory` you could implement a custom certificate check, such as validating the cert footprint or other criterias. Are you using a self-signed certificate? This would be the case then.

To implement something like this you have to implement a `IHttpMessageHandlerBuilderFilter` where you can check the certificate by yourself.
A sample implementation, which allows all certificates:

**DON'T USE THIS IN PRODUCTION! ALL CERTIFICATES WILL BE CONSIDERED VALID!**

```csharp
public interface IServerCertificateValidator
{
    bool Validate(HttpRequestMessage message, X509Certificate2 certificate, X509Chain chain, SslPolicyErrors policyErrors);
}

class AllowAllServerCertificateValidator : IServerCertificateValidator
{
    ILogger<AllowAllServerCertificateValidator> Logger { get; set; }

    public AllowAllServerCertificateValidator()
    {
        Logger = NullLogger<AllowAllServerCertificateValidator>.Instance;
    }

    public bool Validate(HttpRequestMessage message, X509Certificate2 certificate, X509Chain chain, SslPolicyErrors policyErrors)
    {
        Logger.LogWarning("You really replace this validator with a more secure one!");
        return true;
    }
}

[Dependency(ServiceLifetime.Singleton, ReplaceServices = true)
    , ExposeServices(typeof(IHttpMessageHandlerBuilderFilter))]
public class WinHttpMessageHandlerBuilderFilter : IHttpMessageHandlerBuilderFilter, ISingletonDependency
{
    public IServerCertificateValidator ServerCertificateValidator { get; set; }

    public WinHttpMessageHandlerBuilderFilter()
    {
        ServerCertificateValidator = NullServerCertificateValidator.Instance;
    }

    public Action<HttpMessageHandlerBuilder> Configure(Action<HttpMessageHandlerBuilder> next)
    {
        return (builder) =>
        {
            var winHttpHandler = new WinHttpHandler()
            {
                ServerCertificateValidationCallback = ServerCertificateValidator.Validate
            };
            builder.PrimaryHandler = winHttpHandler;
        };
    }
}
```

PR #2904, opened 2020-02-24
quartz.net background job&worker integration

liangshiw:
Background jobs are fully compatible with ABP native, the background worker uses the quart.net library API.
If merged, I will write documentation to introduce them

PR #2958, opened 2020-03-03
Fix .sln file

liangshiw:


maliming:
Caused by https://github.com/abpframework/abp/pull/2904

PR #2971, opened 2020-03-04
Cascading entities should not be deleted when soft deleting entities.

maliming:
We can manually delete the associated entities.

https://docs.microsoft.com/en-us/ef/core/what-is-new/ef-core-3.0/breaking-changes#cascade-deletions-now-happen-immediately-by-default

Current behavior: **The cascaded entities of an entity are immediately deleted.**


hikalkan:
Great! The only think I think is that we are changing a default behaviour. I will work on that.

framework/src/Volo.Abp.EntityFrameworkCore/Volo/Abp/Uow/EntityFrameworkCore/UnitOfWorkDbContextProvider.cs
```diff
@@ -60,6 +61,8 @@ private TDbContext CreateDbContext(IUnitOfWork unitOfWork, string connectionStri
                     dbContext.Database.SetCommandTimeout(unitOfWork.Options.Timeout.Value.TotalSeconds.To<int>());
                 }
 
+                dbContext.ChangeTracker.CascadeDeleteTiming = CascadeTiming.OnSaveChanges;
+                dbContext.ChangeTracker.DeleteOrphansTiming = CascadeTiming.OnSaveChanges;
```
hikalkan: Can we make this inside the constructor of the `AbpDbContext`? In this way, any dbcontext can change easy for itself.
What do you think?

maliming: It is safe to use `ChangeTracker `in the constructor, I will update the PR.

maliming:
> Can we make this inside the constructor of the AbpDbContext?

This caused `AbpDateTimeValueConverter_Tests `to fail, which I was checking.

maliming:
After `CreateDbContext `creates `dbcontext`, call a method in `AbpDbContext`, such as `ConfigDbContext()`. This way developers can change the properties of dbcontext, etc.

```c#
private TDbContext CreateDbContext(IUnitOfWork unitOfWork, string connectionStringName, string connectionString)
{
	var dbContext = CreateDbContext(unitOfWork);
	
	//code
	
	dbContext.ConfigDbContext();
	return dbContext;
}
```

hikalkan:
>After CreateDbContext creates dbcontext, call a method in AbpDbContext, such as ConfigDbContext(). This way developers can change the properties of dbcontext, etc.

Agree. I will do it. Thanks ;)

hikalkan:
Done: https://github.com/abpframework/abp/issues/3176

PR #2972, opened 2020-03-04
Allow replace & delete BundleContributor.

maliming:
A js library may have many localization files, but their file names may be different from the culture name.

such as:
```
zh-Hans: zh-cn
pt-BR: ptBr
es-MX: es
```

We should allow developers to map them manually.
``` c#
public class MyBootstrapDatepickerScriptContributor : BootstrapDatepickerScriptContributor
{
    protected override string MapCultureName(string cultureName)
    {
        return cultureName == "zh-Hans" ? "zh-CN" : base.MapCultureName(cultureName);
    }
}

Configure<AbpBundlingOptions>(options =>
{
	options.ScriptBundles.Get(StandardBundles.Scripts.Global).Contributors
		.Replace<BootstrapDatepickerScriptContributor, MyBootstrapDatepickerScriptContributor>();
});
```

### DIFF
--- a/framework/src/Volo.Abp.AspNetCore.Mvc.UI.Bundling/Volo/Abp/AspNetCore/Mvc/UI/Bundling/BundleContributorCollection.cs
+++ b/framework/src/Volo.Abp.AspNetCore.Mvc.UI.Bundling/Volo/Abp/AspNetCore/Mvc/UI/Bundling/BundleContributorCollection.cs
@@ -39,6 +39,43 @@ namespace Volo.Abp.AspNetCore.Mvc.UI.Bundling
             AddWithDependencies(contributorType);
         }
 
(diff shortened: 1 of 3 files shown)


hikalkan:
You know a json localization file has the `"culture": "en"` which determines the culture name. we don't use the file name.

maliming:
`BootstrapDatepicker `is loading its localization script based on the `culture name`.

https://github.com/abpframework/abp/blob/e76c92df4e7af127289b5448e79543f5b1b6dc5b/framework/src/Volo.Abp.AspNetCore.Mvc.UI.Packages/Volo/Abp/AspNetCore/Mvc/UI/Packages/BootstrapDatepicker/BootstrapDatepickerScriptContributor.cs#L32

But its Chinese localization script is `zh-CN.js`

PR #2973, opened 2020-03-04
Create Background-Workers' documentation

cotur:


liangshiw:
@cotur Hi. Adding dependencies to module classes section is missing from the documentation

leonkosak:
@cotur, @liangshiw  I also get an exception in OnApplicationInitialization method if the class (business logic - as you have IAppStatusService appStatusService) (derived from ITransientDependency)

Autofac.Core.DependencyResolutionException: 'An exception was thrown while activating...

What am I still missing?

liangshiw:
@leonkosak Create a new issue to discuss

PR #2985, opened 2020-03-05
Create an endpoint to query all roles.

### LINKED ISSUE #2983: Create an endpoint to query all roles
* Add `GetAllListAsync()` method to the `IIdentityRoleAppService`
* Set route as `[Route("all")]` in the `IdentityRoleController`
* Use this method from MVC UI to query roles when assigning to a user

Based on https://github.com/abpframework/abp/issues/2629

maliming:
Resolve #2983

### DIFF
--- a/modules/identity/src/Volo.Abp.Identity.Application.Contracts/Volo/Abp/Identity/IIdentityRoleAppService.cs
+++ b/modules/identity/src/Volo.Abp.Identity.Application.Contracts/Volo/Abp/Identity/IIdentityRoleAppService.cs
@@ -8,7 +8,7 @@ namespace Volo.Abp.Identity
 {
     public interface IIdentityRoleAppService : IApplicationService
     {
-        Task<List<IdentityRoleDto>> GetAllListAsync();
+        Task<ListResultDto<IdentityRoleDto>> GetAllListAsync();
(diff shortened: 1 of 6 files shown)


modules/identity/src/Volo.Abp.Identity.Application.Contracts/Volo/Abp/Identity/IIdentityRoleAppService.cs
```diff
@@ -8,6 +8,8 @@ namespace Volo.Abp.Identity
 {
     public interface IIdentityRoleAppService : IApplicationService
     {
+        Task<List<IdentityRoleDto>> GetAllListAsync();
```
hikalkan: Please use ListResultDto wtihh has Items collection. We use this "wrapping" to be able to add new properties to the object without breaking existing client applications (backward compability).

maliming: `ListResultDto` has been used.

maliming:
@hikalkan Already done.

hikalkan:
Thanks!

PR #2986, opened 2020-03-05
added multi tenancy info and current tenant dto class

### LINKED ISSUE #2982: AbpApplicationConfigurationController should also return a value indicates that if multi-tenancy is enabled or not.
1. We can add `public MultiTenancyInfoDto MultiTenancy { get; set; }` that has `IsEnabled` which gets its value from the `AbpMultiTenancyOptions`
2. We can add `public CurrentTenantDto CurrentTenant { get; set; }` just like the user. (as an additional feature)

AkinSabriCam:
Resolve #2982

### DIFF
--- a/framework/src/Volo.Abp.AspNetCore.Mvc.Contracts/Volo/Abp/AspNetCore/Mvc/MultiTenancy/CurrentTenantDto.cs
+++ b/framework/src/Volo.Abp.AspNetCore.Mvc.Contracts/Volo/Abp/AspNetCore/Mvc/MultiTenancy/CurrentTenantDto.cs
@@ -10,6 +10,6 @@ namespace Volo.Abp.AspNetCore.Mvc.MultiTenancy
 
         public string Name { get; set; }
 
-        public bool isAvailable  { get; set; }
+        public bool IsAvailable  { get; set; }
     }
 }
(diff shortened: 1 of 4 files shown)


framework/src/Volo.Abp.AspNetCore.Mvc.Contracts/Volo/Abp/AspNetCore/Mvc/MultiTenancy/CurrentTenantDto.cs
```diff
@@ -0,0 +1,15 @@
...
+
+namespace Volo.Abp.AspNetCore.Mvc.MultiTenancy
+{
+    public class CurrentTenantDto
+    {
+        public Guid? Id { get; set; }
+
+        public string Name { get; set; }
+
+        public bool isAvailable  { get; set; }
```
hikalkan:
```suggestion
        public bool IsAvailable  { get; set; }
```

hikalkan: Always PascalCase on server side for class properties.

PR #3028, opened 2020-03-09
Set the default role when creating external user.

### LINKED ISSUE #2977: Should set default roles for new external users
CreateExternalUserAsync doesn't support SetDefaultRolesAsync

maliming:
Resolve #2977

### DIFF
--- a/modules/account/src/Volo.Abp.Account.Web/Pages/Account/Login.cshtml.cs
+++ b/modules/account/src/Volo.Abp.Account.Web/Pages/Account/Login.cshtml.cs
@@ -211,6 +211,7 @@ namespace Volo.Abp.Account.Web.Pages.Account
 
             CheckIdentityErrors(await UserManager.CreateAsync(user));
             CheckIdentityErrors(await UserManager.SetEmailAsync(user, emailAddress));
+            CheckIdentityErrors(await UserManager.SetRolesAsync(user, RoleManager.Roles.Where(r => r.IsDefault).Select(r => r.Name)));
             CheckIdentityErrors(await UserManager.AddLoginAsync(user, info));
 
             return user;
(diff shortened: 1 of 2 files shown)


modules/account/src/Volo.Abp.Account.Web/Pages/Account/Login.cshtml.cs
```diff
@@ -211,6 +211,7 @@ protected virtual async Task<IdentityUser> CreateExternalUserAsync(ExternalLogin
 
             CheckIdentityErrors(await UserManager.CreateAsync(user));
             CheckIdentityErrors(await UserManager.SetEmailAsync(user, emailAddress));
+            CheckIdentityErrors(await UserManager.SetRolesAsync(user, RoleManager.Roles.Where(r => r.IsDefault).Select(r => r.Name)));
```
hikalkan: Don't use RoleManager.Roles.Where... which executes a query to the database synchronously (because now way to make it async with this code)

Instead, do it just like the AccountAppService:
https://github.com/abpframework/abp/blob/ecac0358fbc9a0acbbda91111d41d5ee6bfb795a/modules/account/src/Volo.Abp.Account.Application/Volo/Abp/Account/AccountAppService.cs#L39-L44

maliming: How to get the default role in the web layer?

Should we add two endpoints to the Role application service to get all the `static/default` roles. like https://github.com/abpframework/abp/pull/2985

Or in `RoleManager`?

hikalkan: I've done it with https://github.com/abpframework/abp/commit/99f23a51bc039d4de5d36d821476be73e8536175

PR #3057, opened 2020-03-10
Russian localization

romkij:
Added Russian localization.


hikalkan:
Great! Thank you for your effort :)

Issue #3054, opened 2020-03-10
abp cli bug


ChiakiYu:
```
abp new DemoProject -t app -u angular
abp add-module Volo.Account --with-source-code
abp add-module Volo.Identity --with-source-code
```

then compile error.
```
找不到项目“D:\DemoProject\aspnet-core\modules\Volo.Identity\test\Volo.Abp.Identity.AspNetCore\Volo.Abp.Identity.AspNetCore.csproj”。请检查项目引用是否有效以及项目文件是否存在。	Volo.Abp.Account.Application.Tests	D:\DemoProject\aspnet-core\modules\Volo.Account\test\Volo.Abp.Account.Application.Tests\Volo.Abp.Account.Application.Tests.csproj	1	

```

In `D:\DemoProject\aspnet-core\modules\Volo.Account\test\Volo.Abp.Account.Application.Tests\Volo.Abp.Account.Application.Tests.csproj`
```
<ProjectReference Include="..\..\..\Volo.Identity\test\Volo.Abp.Identity.AspNetCore\Volo.Abp.Identity.AspNetCore.csproj" />
```
`test` should be `src`?

maliming:
I will check it.

maliming:
Move to https://github.com/abpframework/abp/issues/3075

PR #3060, opened 2020-03-10
Restore the sourceFile variable.

maliming:
Related: https://github.com/abpframework/abp/issues/3054

hi @yekalkan 
Volo.Identity's `masterModules `include `Volo.Account`. Should it be reversed?

image


yekalkan:
> Volo.Identity's masterModules include Volo.Account. Should it be reversed?

No. It's true. When you add source code of a module, masters of it must be added too. So, when you add Volo.Identity, Volo.Account is added as well. (This relation could be named better, i wasn't sure with the name "master" but i didn't want to waste hours with naming :) )

maliming:
> when you add Volo.Identity, Volo.Account is added as well.

But the account module depends on the identity module.

image

image

PR #3064, opened 2020-03-10
Link typo

emrekara37:


hikalkan:
I think it is working for MD format, but there is a problem when it is rendered to HTML by the docs module.

@ebicoglu what do you think? If it is not working, we can just remove the link.

Issue #3016, opened 2020-03-08
Cli  use of local templates caused an error.


Loongle:
I custom module template，I removed Web.Host and Web.Unified and MyProjectName.Web from the solution. Throws an exception: `System. ApplicationException`

Execute the command：`abp new Acme.BookStore -t module --no-ui -ts D:\localTemplate`
![image](https://user-images.githubusercontent.com/7290992/76156185-9a67b100-6131-11ea-88c9-69e6a41fa12b.png)

If I use a local template, can the cli remove some unnecessary checks? Thank you @liangshiw

Loongle:
If I just delete the file and don't remove the project from the solution, then I don't have this problem

maliming:
What is the file structure of directory `D:\localTemplate`?

Loongle:
@maliming There is only one file in the directory named `module-2.2.1.zip`

liangshiw:
@Loongle  I try to implement it

liangshiw:
> CLI should automatically remove these unnecessary projects based on the parameters.
> 
> I think user custom project templates should not change the structure of the solution.

Loongle:
@liangshiw I have a different point of view. I don't think a custom project template should examine the solution structure. why? because maybe custom project template is simple struct.  Just like the documentation case :[https://docs.abp.io/en/abp/latest/Getting-Started-AspNetCore-Application](url). Can we discuss how to solve it

liangshiw:
@Loongle The CLI cannot adapt to all scenarios, it is customized for the official template. In addition, it is open source, you can change it for your own scenarios.

PR #3078, opened 2020-03-11
CLI removes unnecessary checks when using custom template sources

liangshiw:
Resolved : https://github.com/abpframework/abp/issues/3016


maliming:
> I custom module template，I removed Web.Host and Web.Unified and MyProjectName.Web from the solution. Throws an exception: System. ApplicationException

CLI should automatically remove these unnecessary projects based on the parameters.

I think user custom project templates should not change the structure of the solution.

liangshiw:
@maliming  Agree, custom projects should keep the basic structure so that the CLI can handle them well

Issue #3088, opened 2020-03-12
Get password & email address of the admin while creating a new tenant


hikalkan:
Instead of;

````csharp
await DataSeeder.SeedAsync(tenant.Id);
````

It is possible to pass pass and email:

````csharp
await DataSeeder.SeedAsync(
    new DataSeedContext(tenant.Id)
        .WithProperty("AdminEmail", "admin@mytenant.com")
        .WithProperty("AdminPassword", "123456")
);
````

Get these from UI and pass to the seeder.

maliming:
Related: https://github.com/abpframework/abp/issues/2458

PR #3193, opened 2020-03-18
feature: add admin email and password fields to new tenant form

mehmet-erim:


olicooper:
Related: https://github.com/abpframework/abp/issues/3088

PR #3233, opened 2020-03-19
Update Background-Jobs-Hangfire.md

suravi999:
Add Hangfire Configuration

### DIFF
--- a/docs/en/Background-Jobs-Hangfire.md
+++ b/docs/en/Background-Jobs-Hangfire.md
@@ -39,5 +39,50 @@ public class YourModule : AbpModule
 ````
 
 ## Configuration
+You can install any storage for Hangfire. The most common one is SQL Server (see the [Hangfire.SqlServer](https://www.nuget.org/packages/Hangfire.SqlServer) NuGet package).
 
-TODO...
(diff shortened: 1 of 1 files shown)


docs/en/Background-Jobs-Hangfire.md
```diff
@@ -39,5 +39,50 @@ public class YourModule : AbpModule
...
+  private void ConfigureHangfire(ServiceConfigurationContext context, IConfiguration configuration)
+  {
+      context.Services.AddHangfire(config =>
+      {
+          config.UsePostgreSqlStorage(configuration.GetConnectionString("Default"));
+      });
+  }
+````
+
+2. We need to add `UseHangfireServer` call in the Configure method in `Startup` class
```
maliming: We can call `UseXXX()` in the module's `OnApplicationInitialization `method.

https://github.com/abpframework/abp/blob/dev/templates/app/aspnet-core/src/MyCompanyName.MyProjectName.Web/MyProjectNameWebModule.cs#L189

suravi999: Added changes with this commit : https://github.com/suravi999/abp/commit/0e70f9e9b6fc8fc19a8e88a4eaba8eccf9180e40

PR #3245, opened 2020-03-20
Config auto api controller for Web.Unified and HttpApi.Host projects of the module template

wakuflair:
During module development, I found it very common to enable auto api controller feature for host projects.


maliming:
@hikalkan What do you think?

hikalkan:
We don't use auto api controller feature for modules as a best practice. When you use it, it doesn't work with MVC tiered architecture.

PR #3259, opened 2020-03-20
MongoDbRepository needs to filter when deleting entity based on Id.

### LINKED ISSUE #3256: Default MongoDB repository delete by id does not take into account ISoftDelete interface
Hi,

I am using a MongoDB database and found that when I send a delete request for an entity that inherited from `FullAuditedAggregateRoot<Guid>`, it would actually be deleted (and not soft deleted). 

It appears that the `DeleteAsync(
            TKey id,
            bool autoSave = false,
            CancellationToken cancellationToken = default)` method on the default `MongoDbRepository<TMongoDbContext, TEntity, TKey>` repository actually just calls the `DeleteOneAsync` method on the MongoDB collection object without considering the `ISoftDelete` interface.

I created a fix/hack by overriding the method on a custom default repository class

```
Type type = typeof(TEntity);
if (typeof(ISoftDelete).IsAssignableFrom(type))
{
	var entity = await GetAsync(id, false, cancellationToken);

	await DeleteAsync(entity, autoSave, cancellationToken);
}
else
{
	await base.DeleteAsync(id, autoSave, cancellationToken);
}
```
This could be optimised in MongoDB to just perform a relevant update, rather than a Get and then an update, but just wanted to get it working for the time being.

Or - have I missed something?

Many thanks.

maliming:
Resolve #3256

### DIFF
--- a/framework/src/Volo.Abp.MongoDB/Volo/Abp/Domain/Repositories/MongoDB/MongoDbRepository.cs
+++ b/framework/src/Volo.Abp.MongoDB/Volo/Abp/Domain/Repositories/MongoDB/MongoDbRepository.cs
@@ -369,10 +369,7 @@ namespace Volo.Abp.Domain.Repositories.MongoDB
             bool autoSave = false,
             CancellationToken cancellationToken = default)
         {
-            return Collection.DeleteOneAsync(
-                CreateEntityFilter(id),
-                GetCancellationToken(cancellationToken)
-            );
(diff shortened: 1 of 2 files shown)


maliming:
@hikalkan Should I open PR on the `master `branch? Or wait until `v2.4`?

hikalkan:
Yes, please send to master.

maliming:
Let's move to 2.4.

PR #3263, opened 2020-03-20
Create documentation for Collapse Tag Helper

### LINKED ISSUE #3220: Create documentation for Collapse Tag Helper
Follow up from #3200

gterdem:
resolves #3220

### DIFF
--- a/docs/en/UI/AspNetCore/Tag-Helpers/Collapse.md
+++ b/docs/en/UI/AspNetCore/Tag-Helpers/Collapse.md
@@ -32,7 +32,7 @@ A value indicates if the collapse body will be initialized visible or hidden. Sh
 
 ### multi
 
-A value indicates if an `abp-button` or `a` can show and hide multiple collapse body element. Should be one of the following values:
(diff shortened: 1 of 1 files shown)


docs/en/UI/AspNetCore/Tag-Helpers/Collapse.md
```diff
@@ -0,0 +1,92 @@
...
+### show
+
+A value indicates if the collapse body will be initialized visible or hidden. Should be one of the following values:
+
+* `false` (default value)
+* `true`
+
+### multi
+
+A value indicates if an `abp-button` or `a` can show and hide multiple collapse body element. Should be one of the following values:
```
yekalkan: This is not right. We use this attribute only on `abp-collapse-body` tag. 
We don't need "multi" attribute on `abp-button` or `a` because we can determine it automatically. But we need it on `abp-collapse-body` tag. See example below.

So this should be: 
"A value indicates if an `abp-collapse-body` can be shown or hidden by an element that can show/hide multiple collapse bodies. Basically, this attribute adds "multi-collapse" class to `abp-collapse-body`. Should be one of the following values:"

yekalkan: @gterdem i approved changes by mistake.

gterdem: Corrected.

PR #3275, opened 2020-03-20
Add Angular 9 '@Injectable()' to tutorials and book store sample.

olicooper:
BookStore sample doesn't seem to be updated to Angular 9 yet but I think Injectable() should still be fine to include.


mehmet-erim:
Huge thanks @olicooper

PR #3289, opened 2020-03-21
Make properties virtual on TenantCreateDto

olicooper:


olicooper:
Messed up branching, whoops!

PR #3290, opened 2020-03-21
Make properties virtual on TenantCreateDto

olicooper:


maliming:
hi
Can you share what is the use case? :)

olicooper:
I would like to override the attributes on the password and email properties to make them not required, so I can provide a default value in the `IConfiguration`. Because the create tenant method doesn't use interfaces, I need to supply an instance of `TenantCreateDto` and this was the way I could think of that wouldn't be a breaking change.

maliming:
> Because the create tenant method doesn't use interfaces

Can you point it out? Thanks.

olicooper:
> > Because the create tenant method doesn't use interfaces
> 
> Can you point it out? Thanks.

Yeah sure, what I meant was that the `TenantAppService` CreateAsync method accepts a concrete class `TenantCreateDto` rather than an `ITenantCreateDto`interface  so I can't just override the method and supply my own class that inherits from this...
https://github.com/abpframework/abp/blob/1f2418b8721e0d991f2289bc4c57ca666de14c60/modules/tenant-management/src/Volo.Abp.TenantManagement.Application/Volo/Abp/TenantManagement/TenantAppService.cs#L45-L48

I think I prefer using virtual properties instead anyway. Or is there another way to specify that the `AdminEmailAddress` and `AdminPassword` properties not required that I might have missed?

PR #3307, opened 2020-03-23
Migrate only tenants with connection strings.

### LINKED ISSUE #3292: Only run tenant migrations for tenants with external databases?
I noticed the migrator console application runs for the host and every tenant. Data seeding takes a little while for my application so with multi-tenant migrations enabled it takes a long time.

Is it possible to migrate the host database and then generate a list of tenants with external databases and only run the migrations again for those tenants?

Related: https://github.com/abpframework/abp/issues/1371

maliming:
Resolve #3292


gterdem:
@maliming thank you.

PR #3329, opened 2020-03-24
Failing tests of Identity Server module because of shared tempkey.rsa

### LINKED ISSUE #3194: Failing tests of Identity Server module because of shared tempkey.rsa
````
System.IO.IOException : The process cannot access the file 'D:\Github\abp\modules\identityserver\test\Volo.Abp.IdentityServer.MongoDB.Tests\bin\Debug\netcoreapp3.1\tempkey.rsa' because it is being used by another process.
````

Is there any way to use an in-memory tempkey instead of a shared file?

### PR DESCRIPTION
**Pull request by gterdem**
resolves #3194

### DIFF
--- a/modules/identityserver/test/Volo.Abp.IdentityServer.TestBase/Volo/Abp/IdentityServer/AbpIdentityServerTestBaseModule.cs
+++ b/modules/identityserver/test/Volo.Abp.IdentityServer.TestBase/Volo/Abp/IdentityServer/AbpIdentityServerTestBaseModule.cs
@@ -12,6 +12,18 @@ namespace Volo.Abp.IdentityServer
     )]
     public class AbpIdentityServerTestBaseModule : AbpModule
     {
+        public override void PreConfigureServices(ServiceConfigurationContext context)
+        {
+            PreConfigure<AbpIdentityServerBuilderOptions>(options =>
+            {
+                options.AddDeveloperSigningCredential = false;
+            });
+
+            PreConfigure<IIdentityServerBuilder>(identityServerBuilder =>
+            {
+                identityServerBuilder.AddDeveloperSigningCredential(false);
+            });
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: modules/identityserver/test/Volo.Abp.IdentityServer.TestBase/Volo/Abp/IdentityServer/AbpIdentityServerTestBaseModule.cs [hunk 1]
```diff
@@ -12,6 +12,18 @@ namespace Volo.Abp.IdentityServer
...
+        public override void PreConfigureServices(ServiceConfigurationContext context)
+        {
+            PreConfigure<AbpIdentityServerBuilderOptions>(options =>
+            {
+                options.AddDeveloperSigningCredential = false;
+            });
+
+            PreConfigure<IIdentityServerBuilder>(identityServerBuilder =>
+            {
+                identityServerBuilder.AddDeveloperSigningCredential(false);
```
maliming: When the `AddDeveloperSigningCredential `method finds a `tempkey.rsa`  file in the directory, it tries to use it.
We can specify a file name to prevent the previous `tempkey.rsa` from being used.
This is why I use Guid as the file name. 

https://github.com/IdentityServer/IdentityServer4/blob/4ef0886e03d2b10acdbd2e876f521d6b636fc81d/src/IdentityServer4/src/Configuration/DependencyInjection/BuilderExtensions/Crypto.cs#L175

gterdem: Well I thought it's unnecessery io check for file existence since it will be clean/cleaned project but maybe caution is better this case.
I'll update.

PR #3331, opened 2020-03-24
Update Caching.md

380086154:


maliming:
hi

I don't think it needs to be updated, we need to keep the English documentation consistent.

Issue #3314, opened 2020-03-23
Module startup template projects may need full directory structure by default.


gdlcf88:
![image](https://user-images.githubusercontent.com/30018771/77307041-862bd280-6d33-11ea-85c8-0ea866c5166f.png)

May need to:
1. Generate the full directory structure includes `CompanyName/ModuleName/` by default for the module startup template.
1. Configure `<RootNamespace />` in .csproj files by default.

The reasons:
1. Blogging module sets an example, it may be a best practice.
1. If app developers want to override module's code, this structure makes the overridden files more reasonable to place.

gdlcf88:
I try to express my ideas in code.

PR #3337, opened 2020-03-24
Module startup template adjustment.

gdlcf88:
Relate to: #3314

* Adjusted module startup template projects folder structure.
* Adjusted root namespace of module startup template projects.
* Cleanup projects folder hierarchy for module without a company name.


maliming:
@hikalkan  I'm not sure if this is needed.

hikalkan:
Thanks. I will review this.

gdlcf88:
This code does not seem to be processed correctly by the (SolutionRenameStep)[https://github.com/abpframework/abp/blob/1a19895a55fc53a4b5802ff8cb12fd1da5b4f3c3/framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/ProjectBuilding/Building/Steps/SolutionRenameStep.cs]:
https://github.com/abpframework/abp/pull/3337/files#diff-6775e202f30745c8fd666be34690f480R15-R16
https://github.com/abpframework/abp/pull/3337/files#diff-f13fbb623ff6b035a06b42d1c043d2fcR28

If we add `MyCompanyName/MyProjectName` and `MyCompanyName\MyProjectName` to SolutionRenameStep, it may take three times longer to replace the names.

I don't know how to optimize it for now. 😃

gdlcf88:
Another problem:
If a module developer uses a complex name like `Volo.Abp.Identity`, the solution name will be `Volo.Abp` for now.

https://github.com/abpframework/abp/blob/42f37c5ff01ad853a5425d15539d4222cd0dab69/framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/ProjectBuilding/SolutionName.cs#L32-L34

But in this PR, the expected result for the path is `Volo/Abp` and `Volo\Abp`.

hikalkan:
Hi @gdlcf88 

We appreciate for your effort. However, we've intentionally done like current and don't want to change as like in this PR.

PR #3355, opened 2020-03-25
Update Tabs.md

netzheng:
link error


maliming:
Thanks @netzheng

liangshiw:
English document has same error. https://github.com/abpframework/abp/blob/dev/docs/en/UI/AspNetCore/Tag-Helpers/Cards.md
卡面demo页面 should be 标签页demo页面

Issue #3365, opened 2020-03-26
Does hangfire currently support setting IsJobExecutionEnabled as false to disable background jobs?


Huangzc88:
Does hangfire currently support setting IsJobExecutionEnabled as false to disable background jobs? 
I use hangfire, but I need to disable background jobs in the development environment, how should I do that?

maliming:
Yes, `IsJobExecutionEnabled `of `AbpBackgroundJobOptions `can control whether it is enabled.

You can enable or disable it depending on the environment.

Huangzc88:
At abp 1.1.2, I tested that setting this didn't work for hangfire, and then looked at the source code
![image](https://user-images.githubusercontent.com/31158857/77625442-879d0b00-6f7e-11ea-9e94-f12d7fda9cb9.png)
This is the source code to see today, it seems to have not been updated. 
I will update to abp 2.3.0 to test whether it works again, thank you

liangshiw:
Hangfire currently doesn't support  `IsJobExecutionEnabled ` and Quartz also.  I will try implement it.

PR #3382, opened 2020-03-27
Refactoring background jobs&worker integration

liangshiw:
Resolved: https://github.com/abpframework/abp/issues/3365.
Resolved: https://github.com/abpframework/abp/issues/3422
- [x] Implement disabled for background job & worker providers.
- [x] Make method virtual.
- [x] Allow developers decide whether to automatically register workers.
- [x] Background job quartz integration exception retry and customize.


liangshiw:
In addition, the quartz background worker provider did not implement `IsEnabled`, I will sent another PR to solve this problem

framework/src/Volo.Abp.BackgroundJobs.HangFire/Volo/Abp/BackgroundJobs/Hangfire/AbpBackgroundJobsHangfireModule.cs
```diff
@@ -1,14 +1,22 @@
...
+    )]
     public class AbpBackgroundJobsHangfireModule : AbpModule
     {
-
+        public override void ConfigureServices(ServiceConfigurationContext context)
+        {
+            var options = context.Services.ExecutePreConfiguredActions<AbpBackgroundJobOptions>();
+            if (!options.IsJobExecutionEnabled)
+            {
+                context.Services.AddSingleton<IBackgroundJobManager>(x => new NullBackgroundJobManager());
```
maliming: Use `Replace `method to replace components in DI.

framework/src/Volo.Abp.BackgroundJobs.Quartz/Volo/Abp/BackgroundJobs/Quartz/AbpBackgroundJobsQuartzModule.cs
```diff
@@ -12,7 +12,15 @@ public class AbpBackgroundJobsQuartzModule :AbpModule
...
-            context.Services.AddTransient(typeof(QuartzJobExecutionAdapter<>));
+            var options = context.Services.ExecutePreConfiguredActions<AbpBackgroundJobOptions>();
+            if (!options.IsJobExecutionEnabled)
+            {
+                context.Services.AddSingleton<IBackgroundJobManager>(x => new NullBackgroundJobManager());
+            }
+            else
+            {
+                context.Services.AddTransient(typeof(QuartzJobExecutionAdapter<>));
+            }
```
maliming: Use `Replace `method to replace components in DI.

liangshiw:
In order to be compatible with [background job documents](https://docs.abp.io/en/abp/latest/Background-Jobs#disable-job-execution), I judge the `IsJobExecutionEnabled` property when executing the job

framework/src/Volo.Abp.BackgroundJobs.HangFire/Volo/Abp/BackgroundJobs/Hangfire/HangfireJobExecutionAdapter.cs
```diff
@@ -1,27 +1,38 @@
...
             Options = options.Value;
+            Logger = NullLogger<HangfireJobExecutionAdapter<TArgs>>.Instance;
         }
 
         public void Execute(TArgs args)
         {
+            if (!Options.IsJobExecutionEnabled)
+            {
+                Logger.LogWarning("Background jobs system is disabled");
+                return;
```
hikalkan, line 33: When you return, what happens to the job data? Is it lost? I suppose so. Is so, this is dangerous. A better implementation should completely disable hangfire for this application instance, so it does not call the Execute method at all and the job data waits in the database for another application to execute it. What do you think?

liangshiw: I agree that there may be a risk of losing job data. 
But we should not completely disable hangfire, because for abp `Volo.Abp.Hangfire` **!=** `Volo.Abp.BackgourndJob.Hangfire`, hangfire may be used as a background worker integration. When we disable the background job system, we don't want to disable the background worker system. I will try to find other solutions.

hikalkan: I believe that we can disable backgroung jobs & workers together, no problem, because they are related to each other and we normally want to disable together.

liangshiw: If it is disabled together, then we don't need two enabled switches: `IsEnabled` and `IsJobExecutionEnabled`. Just one switch is enough

liangshiw:
@olicooper Could you check if it has been resolved https://github.com/abpframework/abp/issues/3422 ? Thanks.

olicooper:
@liangshiw In your tests (see https://github.com/abpframework/abp/pull/3427#issuecomment-611488819) you used `StoreDurablility(true)` which is something I didn't have configured for y jobs. I imagine this would have to be added to every `JobDetail` that is created. I wonder if we can use the other `AddJob` overload so it doesn't throw the exception "Jobs added with no trigger must be durable" like this one does. I haven't tried it out myself yet.

https://github.com/quartznet/quartznet/blob/d4241f02e8519cab8ea6af1fc359bd31357d383b/src/Quartz/Core/QuartzScheduler.cs#L685-L701

liangshiw:
@olicooper 
For jobs stored in memory, this is no problem.
I can provide an option to automatically add `StoreDurablility(true)` to `JobDetail`.

framework/src/Volo.Abp.BackgroundWorkers.Quartz/Volo/Abp/BackgroundWorkers/Quartz/QuartzBackgroundWorkerManager.cs
```diff
@@ -16,28 +16,42 @@ public QuartzBackgroundWorkerManager(IScheduler scheduler)
...
             {
                 Check.NotNull(quartzWork.Trigger, nameof(quartzWork.Trigger));
                 Check.NotNull(quartzWork.JobDetail, nameof(quartzWork.JobDetail));
-
-                AsyncHelper.RunSync(() => _scheduler.ScheduleJob(quartzWork.JobDetail, quartzWork.Trigger));
+                
+                if (await _scheduler.CheckExists(quartzWork.JobDetail.Key))
+                {
+                    await _scheduler.AddJob(quartzWork.JobDetail, true);
+                    await _scheduler.ResumeJob(quartzWork.JobDetail.Key);
```
olicooper, line 47: Why do we call this before Rescheduling? What does it do?

The description for the`ResumeJob` method is: "Resume (un-pause) the IJobDetail with the given name."
The description for the`RescheduleJob` method is: "Remove (delete) the ITrigger with the given name, and store the new given one - which must be associated with the same job."

So does this allow the misfire to be handled before rescheduling?

liangshiw: 1. The `AddJob` method updates job information, including` JobData`, etc ...
2. The `ResumeJob` method makes Quartz check the Job to execute the missed job.
3. The `RescheduleJob` method updates trigger information.

> So does this allow the misfire to be handled before rescheduling?

Yes, Missed jobs will be executed. You can test it.

olicooper:
I have noticed that Quartz takes the timezone from the host system rather than from `IClock`. It means that Quartz thinks that jobs have misfired when they haven't. You can see the timezone data in `QRTZ_CRON_TRIGGERS` table:
image

Is this something you'd be able to implement too?

framework/src/Volo.Abp.BackgroundWorkers.Quartz/Volo/Abp/BackgroundWorkers/Quartz/QuartzBackgroundWorkerManager.cs
olicooper: Yes this seems to be resuming misfired jobs well and is updating triggers and job details as expected. Good work!
I modified `ReScheduleJobAsync` slightly to use `await _scheduler.AddJob(quartzWork.JobDetail, true, true);` and this seemed to work well too with RAMJobStore and AdoJobStore (JobStoreTX). This would mean we can avoid having to use `StoreDurability`.

liangshiw: I will check it. 

> Is this something you'd be able to implement too?

What do you mean? Can you explain in detail? thanks.

olicooper:
Sorry for this.. After more testing I have found an edge case..
I have a job that generates statistics for each tenant at 4am every day for the previous day. If this job is missed I need ALL missed jobs to fire. Using `WithMisfireHandlingInstructionIgnoreMisfires` works well but calling `RescheduleJob` straight after `ResumeJob` means only 1 instance is executed before the trigger is deleted and restarted. If I comment out `RescheduleJob` in `QuartzBackgroundWorkerManager` it works as expected (but won't update the trigger!). This is proving to be a complicated thing to implement!

https://github.com/abpframework/abp/blob/ae72356b3ba653ecf372e95cafbe4d503b400074/framework/src/Volo.Abp.BackgroundWorkers.Quartz/Volo/Abp/BackgroundWorkers/Quartz/QuartzBackgroundWorkerManager.cs#L48

What would be great is a method similar to 'AddJob' called 'AddTrigger' where we can replace the trigger but keep the old fire times until all the misfires have been fired. Quartz fires events off when things happen in the system, and I wonder if we can utilise these to call 'reschedule' at the appropriate time?

Also, thank you for adding 'virtual' to the methods, this has helped a lot with testing! :)

This is my job trigger:
```csharp
Trigger = TriggerBuilder.Create()
    .WithIdentity($"{nameof(TenantStatisticsGenerationJob)}")
    // If the schedule is missed, fire ALL missed jobs up to the current time
    .WithCronSchedule("0 0 4 * * ?", opt => opt.WithMisfireHandlingInstructionIgnoreMisfires())
    .StartNow().Build();
```

liangshiw:
Hi @olicooper  
This is a complicated matter and should be decided by the developer, so I added the `AutoRegister` property to skip the automatic registration.

liangshiw:
Hi @olicooper
This is a complicated matter and should be decided by the developer, so I added the AutoRegister property to skip the automatic registration.

olicooper:
Yeah I figured this wouldn't be simple to fix. I will see if I can find anything that might help. Thank you for your work so far liangshiw.

@hikalkan The current version of this code improves the existing implementation and fixes some bugs so it is probably worth merging in for the next version and maybe we can open a ticket to resolve the remaining issues?

olicooper:
@liangshiw I didn't see your most recent commit until now, I will try to test this tomorrow.

I have been working on an idea I had about updating Quartz triggers. The only time we need to update/reschedule a trigger is when it has changed. Quartz doesn't have very good comparison mechanisms for ITrigger (maybe we can raise a ticket with them) and `ITrigger` can come in one of four 'flavours' (cron, simple, daily etc.) so I created my own trigger comparison method (which can likely be improved, it feels verbose!).

The idea is simple, compare all the known properties of the old trigger against the new trigger and if any fail to match, then reschedule the job. The modified code is below for review, what do you think?


_**QuartzBackgroundWorkerManager.cs**_
```csharp
protected override async Task ReScheduleJobAsync(IBackgroundWorker worker)
{
    if (worker is IQuartzBackgroundWorker quartzWork)
    {
        Check.NotNull(quartzWork.Trigger, nameof(quartzWork.Trigger));
        Check.NotNull(quartzWork.JobDetail, nameof(quartzWork.JobDetail));

        if (await _scheduler.CheckExists(quartzWork.JobDetail.Key))
        {
            await _scheduler.AddJob(quartzWork.JobDetail, true, true);

            var originalTrigger = await _scheduler.GetTrigger(quartzWork.Trigger.Key);
            if (IsTriggerEqual(originalTrigger, quartzWork.Trigger))
            {
                await _scheduler.ResumeJob(quartzWork.JobDetail.Key);
            }
            else
            {
                await _scheduler.RescheduleJob(quartzWork.Trigger.Key, quartzWork.Trigger);
            }
        }
        else
        {
            await _scheduler.ScheduleJob(quartzWork.JobDetail, quartzWork.Trigger);
        }
    }
}

private bool IsTriggerEqual(ITrigger oldTrigger, ITrigger newTrigger)
{
    if (!oldTrigger.Key.Equals(newTrigger.Key)) return false;
    //if (!oldTrigger.JobKey.Equals(newTrigger.JobKey)) return false;
    //if (oldTrigger.CalendarName != newTrigger.CalendarName) return false;
    if (oldTrigger.Description != newTrigger.Description) return false;
    if (oldTrigger.Priority != newTrigger.Priority) return false;
    if (oldTrigger.MisfireInstruction != newTrigger.MisfireInstruction) return false;
    if (oldTrigger.HasMillisecondPrecision != newTrigger.HasMillisecondPrecision) return false;

    // todo: required?
    //if ((oldTrigger as IOperableTrigger).FireInstanceId != (newTrigger as IOperableTrigger).FireInstanceId) return false;

    // bypass object casting if the type doesn't match anyway
    if (oldTrigger.GetType() != newTrigger.GetType()) return false;

    // CRON TRIGGER
    if (oldTrigger is ICronTrigger oldCronTrigger)
    {
        if (!(newTrigger is ICronTrigger newCronTrigger)) return false;

        if (oldCronTrigger.CronExpressionString != newCronTrigger.CronExpressionString) return false;
        if (!oldCronTrigger.TimeZone.Equals(newCronTrigger.TimeZone)) return false;
    }
    // SIMPLE TRIGGER
    else if (oldTrigger is ISimpleTrigger oldSimplTrigger)
    {
        if (!(newTrigger is ISimpleTrigger newSimplTrigger)) return false;

        if (oldSimplTrigger.RepeatCount != newSimplTrigger.RepeatCount) return false;
        if (oldSimplTrigger.RepeatInterval != newSimplTrigger.RepeatInterval) return false;
        //if (oldSimplTrigger.TimesTriggered != newSimplTrigger.TimesTriggered) return false;
    }
    // DAILY INTERVAL TRIGGER
    else if (oldTrigger is IDailyTimeIntervalTrigger oldDailyTrigger)
    {
        if (!(newTrigger is IDailyTimeIntervalTrigger newDailyTrigger)) return false;

        if (oldDailyTrigger.RepeatCount != newDailyTrigger.RepeatCount) return false;
        if (oldDailyTrigger.RepeatIntervalUnit != newDailyTrigger.RepeatIntervalUnit) return false;
        if (oldDailyTrigger.RepeatInterval != newDailyTrigger.RepeatInterval) return false;
        if (oldDailyTrigger.StartTimeOfDay != newDailyTrigger.StartTimeOfDay) return false;
        if (oldDailyTrigger.EndTimeOfDay != newDailyTrigger.EndTimeOfDay) return false;
        if (oldDailyTrigger.DaysOfWeek != newDailyTrigger.DaysOfWeek) return false;
        //if (oldDailyTrigger.TimesTriggered != newDailyTrigger.TimesTriggered) return false;
        if (!oldDailyTrigger.TimeZone.Equals(newDailyTrigger.TimeZone)) return false;
    }
    // CALENDAR INTERVAL TRIGGER
    else if (oldTrigger is ICalendarIntervalTrigger oldCalendarTrigger)
    {
        if (!(newTrigger is ICalendarIntervalTrigger newCalendarTrigger)) return false;

        if (oldCalendarTrigger.RepeatIntervalUnit != newCalendarTrigger.RepeatIntervalUnit) return false;
        if (oldCalendarTrigger.RepeatInterval != newCalendarTrigger.RepeatInterval) return false;
        //if (oldCalendarTrigger.TimesTriggered != newCalendarTrigger.TimesTriggered) return false;
        if (!oldCalendarTrigger.TimeZone.Equals(newCalendarTrigger.TimeZone)) return false;
        if (oldCalendarTrigger.PreserveHourOfDayAcrossDaylightSavings != newCalendarTrigger.PreserveHourOfDayAcrossDaylightSavings) return false;
        if (oldCalendarTrigger.SkipDayIfHourDoesNotExist != newCalendarTrigger.SkipDayIfHourDoesNotExist) return false;
    }
    else
    {
        // If there is a new type of trigger that is introduced in a new version 
        // then we should return false for safety reasons.
        return false;
    }

    // Do this last as it may be more CPU heavy for large objects.
    // todo: How good is this at comparison? Does it deep/recursive compare objects?
    if (!oldTrigger.JobDataMap.Equals(newTrigger.JobDataMap)) return false;

    return true;
}
```

liangshiw:
I think this is too complicated, the developer should be responsible for job, and the framework should provide extensions where appropriate.

olicooper:
@liangshiw I have tried to simplify and improve the code I wrote above. The problem I am trying to solve is the issue with misfire instructions not being properly handled (i.e. `WithMisfireHandlingInstructionIgnoreMisfires()`). I have found another way to make the `RescheduleJob` update the trigger, but also **preserve the original fire times** so that the scheduler resumes the jobs properly! My previous code didn't really handle it well but this fixes all the issues in one. This code should allow the Quartz scheduler to work as intended for any `Trigger` modification. The code still _looks_ a little bit complicated but it is really simple and because it fixes an existing bug, I believe it is required for the Quartz implementation to work properly with ABP.

This time I check if any of the **schedule related properties** have changed for the `ITrigger`. If the properties have not changed then I copy the previous and next fire times etc. over to the new trigger before calling `RescheduleJob`. I carefully checked the Quartz source code to see how `RescheduleJob` works and I found that this was a feasible solution.

I have tested changing the type of trigger schedule, updating various Trigger details, causing misfires by leaving application not running for a while etc. and all scenarios seem to work as expected. More testing will be required before releasing it though.

You can refactor it how you like but this is the basics:
```csharp
protected override async Task ReScheduleJobAsync(IBackgroundWorker worker)
{
    if (worker is IQuartzBackgroundWorker quartzWork)
    {
        Check.NotNull(quartzWork.Trigger, nameof(quartzWork.Trigger));
        Check.NotNull(quartzWork.JobDetail, nameof(quartzWork.JobDetail));

        if (await _scheduler.CheckExists(quartzWork.JobDetail.Key))
        {
            await _scheduler.AddJob(quartzWork.JobDetail, true, true);

            var originalTrigger = await _scheduler.GetTrigger(quartzWork.Trigger.Key);

            if (ShouldPreserveTriggerFireState(originalTrigger, quartzWork.Trigger))
            {
                quartzWork.Trigger = CopyTriggerFireState(originalTrigger, quartzWork.Trigger);
            }
                    
            await _scheduler.RescheduleJob(quartzWork.Trigger.Key, quartzWork.Trigger);
        }
        else
        {
            await _scheduler.ScheduleJob(quartzWork.JobDetail, quartzWork.Trigger);
        }
    }
}

private bool ShouldPreserveTriggerFireState(ITrigger oldTrigger, ITrigger newTrigger)
{
    // bypass object casting if the type doesn't match anyway
    if (oldTrigger.GetType() != newTrigger.GetType()) return false;
    //if (oldTrigger.HasMillisecondPrecision != newTrigger.HasMillisecondPrecision) return false;

    switch (oldTrigger)
    {
        // CRON TRIGGER
        case ICronTrigger oldCronTrigger when newTrigger is ICronTrigger newCronTrigger:
            if (oldCronTrigger.CronExpressionString != newCronTrigger.CronExpressionString) return false;
            if (!oldCronTrigger.TimeZone.Equals(newCronTrigger.TimeZone)) return false;
            break;
        // SIMPLE TRIGGER
        case ISimpleTrigger oldSimplTrigger when newTrigger is ISimpleTrigger newSimplTrigger:
            if (oldSimplTrigger.RepeatCount != newSimplTrigger.RepeatCount) return false;
            if (oldSimplTrigger.RepeatInterval != newSimplTrigger.RepeatInterval) return false;
            break;
        // DAILY INTERVAL TRIGGER
        case IDailyTimeIntervalTrigger oldDailyTrigger when newTrigger is IDailyTimeIntervalTrigger newDailyTrigger:
            if (oldDailyTrigger.RepeatCount != newDailyTrigger.RepeatCount) return false;
            if (oldDailyTrigger.RepeatIntervalUnit != newDailyTrigger.RepeatIntervalUnit) return false;
            if (oldDailyTrigger.RepeatInterval != newDailyTrigger.RepeatInterval) return false;
            if (((oldDailyTrigger.StartTimeOfDay == null) != (newDailyTrigger.StartTimeOfDay == null)) || !oldDailyTrigger.StartTimeOfDay.Equals(newDailyTrigger.StartTimeOfDay)) return false;
            if (((oldDailyTrigger.EndTimeOfDay == null) != (newDailyTrigger.EndTimeOfDay == null)) || !oldDailyTrigger.EndTimeOfDay.Equals(newDailyTrigger.EndTimeOfDay)) return false;
            if (((oldDailyTrigger.DaysOfWeek == null) != (newDailyTrigger.DaysOfWeek == null)) || oldDailyTrigger.DaysOfWeek.Count != newDailyTrigger.DaysOfWeek.Count) return false;
            if (!oldDailyTrigger.TimeZone.Equals(newDailyTrigger.TimeZone)) return false;
            break;
        // CALENDAR INTERVAL TRIGGER
        case ICalendarIntervalTrigger oldCalendarTrigger when newTrigger is ICalendarIntervalTrigger newCalendarTrigger:
            if (oldCalendarTrigger.RepeatIntervalUnit != newCalendarTrigger.RepeatIntervalUnit) return false;
            if (oldCalendarTrigger.RepeatInterval != newCalendarTrigger.RepeatInterval) return false;
            if (!oldCalendarTrigger.TimeZone.Equals(newCalendarTrigger.TimeZone)) return false;
            if (oldCalendarTrigger.PreserveHourOfDayAcrossDaylightSavings != newCalendarTrigger.PreserveHourOfDayAcrossDaylightSavings) return false;
            if (oldCalendarTrigger.SkipDayIfHourDoesNotExist != newCalendarTrigger.SkipDayIfHourDoesNotExist) return false;
            break;
        default:
            return false;
    }
            
    return true;
}

private ITrigger CopyTriggerFireState(ITrigger fromTrigger, ITrigger toTrigger)
{
    var triggertoUpdate = toTrigger;

    if (triggertoUpdate is AbstractTrigger abstractTrigger)
    {
        abstractTrigger.SetPreviousFireTimeUtc(fromTrigger.GetPreviousFireTimeUtc());
        abstractTrigger.SetNextFireTimeUtc(fromTrigger.GetNextFireTimeUtc());
        abstractTrigger.StartTimeUtc = fromTrigger.StartTimeUtc;
    }

    switch (fromTrigger)
    {
        // CRON TRIGGER
        //case ICronTrigger fromCronTrigger when triggertoUpdate is ICronTrigger toCronTrigger:
        //    break;
        // SIMPLE TRIGGER
        case ISimpleTrigger fromSimplTrigger when triggertoUpdate is ISimpleTrigger toSimplTrigger:
            toSimplTrigger.TimesTriggered = fromSimplTrigger.TimesTriggered;
            break;
        // DAILY INTERVAL TRIGGER
        case IDailyTimeIntervalTrigger fromDailyTrigger when triggertoUpdate is IDailyTimeIntervalTrigger toDailyTrigger:
            toDailyTrigger.TimesTriggered = fromDailyTrigger.TimesTriggered;
            break;
        // CALENDAR INTERVAL TRIGGER
        case ICalendarIntervalTrigger fromCalendarTrigger when triggertoUpdate is ICalendarIntervalTrigger toCalendarTrigger:
            toCalendarTrigger.TimesTriggered = fromCalendarTrigger.TimesTriggered;
            break;
    }

    return triggertoUpdate;
}
```

framework/src/Volo.Abp.BackgroundJobs.Quartz/Volo/Abp/BackgroundJobs/Quartz/QuartzBackgroundJobManager.cs
```diff
@@ -1,26 +1,45 @@
...
-            var jobDetail = JobBuilder.Create<QuartzJobExecutionAdapter<TArgs>>().SetJobData(new JobDataMap { { nameof(TArgs), args } }).Build();
+            return await ReEnqueueAsync(args, Options.RetryCount, Options.RetryIntervalMillisecond, priority, delay);
+        }
+
+        public virtual async Task<string> ReEnqueueAsync<TArgs>(TArgs args, int retryCount, int retryIntervalMillisecond,
+            BackgroundJobPriority priority = BackgroundJobPriority.Normal, TimeSpan? delay = null)
+        {
+            var jobDataMap = new JobDataMap
+            {
+                {nameof(TArgs), args},
```
olicooper, line 34: I know this is an edge case, but what if the developer called their `TArgs` the same as one of the other arguments in the `JobDataMap`? Could we call the arguments "AbpRetryCount", "AbpRetryIntervalMillisecond" and "AbpRetryIndex" or check `nameof(TArgs)` for a name clash?

liangshiw: I will update it

liangshiw:
@olicooper 
I think allowed to disable automatic registration and provide extensions, this is enough.

olicooper:
Hi @liangshiw would it be possible to create a separate PR to make the methods virtual in the [QuartzBackgroundWorkerManager](https://github.com/abpframework/abp/blob/ae72356b3ba653ecf372e95cafbe4d503b400074/framework/src/Volo.Abp.BackgroundWorkers.Quartz/Volo/Abp/BackgroundWorkers/Quartz/QuartzBackgroundWorkerManager.cs) so that it can be merged for the next (v2.7) release? @hikalkan moved this PR to v2.8 which isn't going to be released until 21st May.

liangshiw:
Hi @olicooper,
Currently you can replace the implementation class.

hikalkan:
@liangshiw I suppose this work has been finalized, so I will review and merge?

liangshiw:
@hikalkan  Yes, it's readly, please review.

hikalkan:
Thank you for the implementation. I tested with hangfire and rabbitmq, it is working as expected:

* Used demos on the https://github.com/abpframework/abp/tree/dev/modules/background-jobs/app to test it
* I disabled job execution, created some jobs and I see they are added to queue but not executed.
* I stopped the application, enabled job execution, re-started and I see previous jobs also has executed. So, no job loss.

For Quartz, jobs are lost. Probably because I haven't used a persistence system for quartz, I don't have much experience on it.

BTW, I fixed a bug: https://github.com/abpframework/abp/commit/783c1625124d18002677ffb317769c5aeec16f43

PR #3406, opened 2020-03-29
added exception for  user registered multiple organizations

AkinSabriCam:


framework/src/Volo.Abp.IdentityModel/Volo/Abp/IdentityModel/IdentityModelAuthenticationService.cs
```diff
@@ -72,6 +73,12 @@ public virtual async Task<string> GetAccessTokenAsync(IdentityClientConfiguratio
             var tokenResponse = await GetTokenResponse(discoveryResponse, configuration);
             if (tokenResponse.IsError)
             {
+                if (tokenResponse.HttpStatusCode == HttpStatusCode.Forbidden ||
```
ebicoglu: this is wrong. you cannot throw the same exception for all forbidden or internal server errors!

Issue #3422, opened 2020-03-30
Issue with Quartz job persistence


olicooper:
I've been working on job persistence with the new Quartz jobs and there seems to be an issue with it.

Example:
```csharp
public class DataArchiveJob : QuartzBackgroundWorkerBase, ISingletonDependency
{
    public DataArchiveJob()
    {
        JobDetail = JobBuilder.Create<DataArchiveJob>()
            .WithIdentity($"{nameof(DataArchiveJob)}")
            .Build();

        Trigger = TriggerBuilder.Create()
            .WithIdentity($"{nameof(DataArchiveJob)}")
            .WithCronSchedule("0 30 1/6 ? * * *")
            .StartNow().Build();
    }
}
```

The code above create a job that runs every 6 hours. The job is inserted in the database when running the application for the first time, but when it is run for a seconds time I receive an exception about a duplicate job: 
> Quartz.JobPersistenceException: 'Couldn't store job: Unable to store Job: 'DEFAULT.DataArchiveJob', because one already exists with this identification.'

I believe we need a check in the job factory to ensure jobs aren't created twice or they are deleted and re-added? I'm not sure how you'd handle changes to the job trigger, job detail or the JobDataMap? I've seen this as an example: https://stackoverflow.com/a/51724460

If any more information is needed just let me know.

liangshiw:
I will check this

olicooper:
Thanks for the quick update! I have a question...

When I do a go-live, the in-memory jobs are deleted and re-added after the application is restarted which causes the scheduled job to be missed until the next time determined by the CronExpression. I want to use the persistent job store to ensure this doesn't happen.

For example, if I have a job that runs once a day at 18:00 (`0 0 18 ? * * *`) and I do a go-live around this time, the job will normally be missed and the next one will be scheduled for tomorrow rather than the one that was missed at 18:00. 

Will your code make sure that missed jobs are still executed as soon as they can, even if the application is restarted?

PR #3427, opened 2020-03-31
Quartz Backgroundworker Implements IsEnable and judges whether the jobkey is duplicated

liangshiw:
Resolved : https://github.com/abpframework/abp/issues/3422


olicooper:
Looking at changes shown below, I wonder if deleting and re-adding the job is going to be an issue?
https://github.com/abpframework/abp/blob/c59e092476ddbf489d82474d819fc133db678b48/framework/src/Volo.Abp.BackgroundWorkers.Quartz/Volo/Abp/BackgroundWorkers/Quartz/QuartzBackgroundWorkerManager.cs#L50-L55

If we delete and re-add the job, it also deletes the trigger, which removes the next fire time, which has to be recalculated when the job is added again. I think this is going to cause certain jobs to miss one or more execution times.

Just an idea.. could we use the `IScheduler.RescheduleJob` method and then replace the job and trigger data/description/schedule if it has changed?

liangshiw:
@olicooper  That's great! This is the better way.

liangshiw:
@olicooper Now it can be handled miss execution times. You can check this.

maliming:
@olicooper Do you have any suggestions?

olicooper:
I have download and compiled liangshiw's version and it has fixed the initial problem but I haven't checked if the jobs resume as expected so I will test that out soon and get back to you with any suggestions.

olicooper:
Just want to update you on some of my findings so far. I am testing with abp v2.4.0 created from the abp cli. I also have the additional Quartz tables built from [this sql script](https://github.com/quartznet/quartznet/blob/master/database/tables/tables_sqlServer.sql) and custom BackgoundJobs.* modules built from liangshiw's repository.

1. When the Trigger or JobDetail name is changed (`TriggerBuilder.WithIdentity("myJobName")`), it causes the job to be duplicated in the `QRTZ_JOB_DETAILS` and `QRTZ_TRIGGERS` tables as seen here:
image
  I'm not sure how or if this can be handled but I wanted to share this with you.
1. When using the cron schedule `0 0/2 * ? * * *`, if I kill the application and restart it after the next scheduled time (simulating a misfire), existing jobs are executed immediately as expected. I will do more testing on this though.
1. Updating the JobDetail doesn't update the existing job. As an example, if we added `JobBuilder.UsingJobData("test", "test")` to an existing job after the job entry has been created in the database, this new data wouldn't be available when the job is run.
1. Updating the Trigger seems to work fine. I added `TriggerBuilder.WithDescription("description1")` and I saw this change reflected in the database.

liangshiw:
HI @olicooper 

1. Different Job Identity are different jobs for quartz (trigger identity is also). Developers should be responsible for job identity rather than framework.
2. That sounds good.
3. That may be a problem,  The `RescheduleJob` Method only update trigger, It should be deleted before adding the job.
4. I don't understanding what you mean.

olicooper:
This is an example of the configuration I have:
```csharp
public DataArchiveJob()
{
    JobDetail = JobBuilder.Create<DataArchiveJob>()
        .WithIdentity($"{nameof(DataArchiveJob)}")
        .WithDescription("job-description1") // Added this AFTER running the application for the first time
        .Build();

    Trigger = TriggerBuilder.Create()
        .WithIdentity($"{nameof(DataArchiveJob)}")
        .WithCronSchedule("0 0/2 * ? * * *") 
        .WithDescription("trigger-description1") // Added this AFTER running the application for the first time
        .StartNow().Build();
}
```
* (1). I agree with this. I wanted to mention just in case though.
* (3 & 4). I create a job and specify the `Trigger` and `JobDetail` in the constructor as seen in the code snippet above. I run the application which creates entries in the `QRTZ_JOB_DETAILS` and `QRTZ_TRIGGERS` tables. I then update the Job `Trigger` and `JobDetail` to add `.WithDescription()`. The trigger entry in `QRTZ_TRIGGERS` is updated with the new description, but the entry in `QRTZ_JOB_DETAILS` doesn't get updated. I think that when you call [IScheduler.RescheduleJob()](https://github.com/liangshiw/abp/blob/ce65338dda530561c3aa05f7d310588b55dcb7db/framework/src/Volo.Abp.BackgroundWorkers.Quartz/Volo/Abp/BackgroundWorkers/Quartz/QuartzBackgroundWorkerManager.cs#L53), it updates the trigger but not the job detail. We need a way to update the JobDetail without deleting it, so that the trigger is left intact (there is a foreign key constraint between `QRTZ_JOB_DETAILS` and `QRTZ_TRIGGERS`, so you can't delete the job detail without also having to delete the trigger).

Does this help explain?

If we delete the trigger then the `PREV_FIRE_TIME` and `NEXT_FIRE_TIME` will be deleted and Quartz won't know if the job has misfired so we can't delete the trigger really. I think this method is called when calling `RescheduleJob`:
https://github.com/quartznet/quartznet/blob/03fa4641c543833460d696823d010cf2c64269ab/src/Quartz/Xml/XMLSchedulingDataProcessor.cs#L933-L947

[QuartzScheduler.RescheduleJob](https://github.com/quartznet/quartznet/blob/03fa4641c543833460d696823d010cf2c64269ab/src/Quartz/Core/QuartzScheduler.cs#L947) calls [JobStoreSupport.ReplaceTrigger](https://github.com/quartznet/quartznet/blob/03fa4641c543833460d696823d010cf2c64269ab/src/Quartz/Impl/AdoJobStore/JobStoreSupport.cs#L1300) which deletes the original trigger and re-adds it with the original TriggerKey. Maybe there is something similar for Jobs which re-assigns the trigger to the new job so that the trigger is left intact? I will have a look if I can find anything.

liangshiw:
Hi @olicooper 
I have updated the logic of ReScheduleAsync method. Can you check this?

This is my test case:

```csharp
JobDetail = JobBuilder.Create<QaWork>()
         .WithIdentity($"{nameof(QaWork)}")
         .WithDescription("job-description1")
         .StoreDurably(true)
         .UsingJobData("qa", "qa5")
         .Build();
Trigger = TriggerBuilder.Create()
          .WithIdentity($"{nameof(QaWork)}")
          .WithCronSchedule("0 53 19 ? * * *")
          .WithDescription("trigger-description1")
          .StartNow().Build();
```

liangshiw:
@olicooper
I want to redesign the disable logic, I am going to close this PR and move it to another PR . Please follower https://github.com/abpframework/abp/pull/3382

PR #3432, opened 2020-03-31
added description field  in posts and updated create and edit page fo…

### LINKED ISSUE #3426: Blog module - Add Description and Title fields to a post blog
When creating a new blog post & editing an existing blog post:

- Users can be able to set a **description** and **title** fields. Put 1 textbox for the Description field and save it to the database field `Description`. 
- Title field already exists in the table, you just need to set it in` <title>goes here</title>`.  Google suggests titles to max 60 chars. Therefore, when the user exceeds 60 character limit, show a warning alert `Keep your title size under 60 characters to be SEO friendly!`
- This description field will be rendered as Meta Description
  `<meta name="Description" content="description-goes-here">`
- If the description field is empty, take the first 200 characters to be the description.

AkinSabriCam:
Resolve #3426

### DIFF
--- a/modules/blogging/src/Volo.Blogging.Web/Pages/Blogs/Posts/Edit.cshtml.cs
+++ b/modules/blogging/src/Volo.Blogging.Web/Pages/Blogs/Posts/Edit.cshtml.cs
@@ -56,7 +56,8 @@ namespace Volo.Blogging.Pages.Blog.Posts
                 Url = Post.Url,
                 CoverImage = Post.CoverImage,
                 Content = Post.Content,
-                Tags = Post.Tags
+                Tags = Post.Tags,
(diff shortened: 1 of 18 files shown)


modules/blogging/src/Volo.Blogging.Web/Pages/Blogs/Posts/Edit.cshtml.cs
```diff
@@ -56,7 +56,8 @@ public async Task<ActionResult> OnPostAsync()
                 Url = Post.Url,
                 CoverImage = Post.CoverImage,
                 Content = Post.Content,
-                Tags = Post.Tags
+                Tags = Post.Tags,
+                Description = !Post.Description.IsNullOrWhiteSpace() ? Post.Description : Post.Content.Substring(0,200)
```
liangshiw: `ArgumentOutOfRangeException` is thrown when the content field length is less than 200

PR #3454, opened 2020-04-02
Add Quartz StartupDelay option

olicooper:


framework/src/Volo.Abp.Quartz/Volo/Abp/Quartz/AbpQuartzModule.cs
```diff
@@ -20,9 +21,15 @@ public override void ConfigureServices(ServiceConfigurationContext context)
...
+            var options = context.ServiceProvider.GetRequiredService<IOptions<AbpQuartzPreOptions>>().Value;
+            
             _scheduler = context.ServiceProvider.GetService<IScheduler>();
             _scheduler.JobFactory = context.ServiceProvider.GetService<IJobFactory>();
-            _scheduler.Start();
+
+            if (options.StartDelay != null && options.StartDelay.Ticks > 0)
+                _scheduler.StartDelayed(options.StartDelay);
+            else
+                _scheduler.Start();
```
maliming, line 32: Please add curly brackets to keep the code style.

`StartDelay `is never be `null`.

maliming, line 32: I see `_scheduler.Start()` method return type is `Task`. This could be a problem.

olicooper: Yeah I noticed this too. @liangshiw any thoughts? I can only say that it hasn't caused any issues for me so far and that all Quartz and IJob exceptions have been caught and logged okay. The `Start()` method does have an exception that will be ignored if we discard the task.

This is the [QuartzScheduler.Start()](https://github.com/quartznet/quartznet/blob/master/src/Quartz/Core/QuartzScheduler.cs#L360) method. There is also a [RemoteScheduler.Start()](https://github.com/quartznet/quartznet/blob/master/src/Quartz/Impl/RemoteScheduler.cs#L209) method too.

maliming:
hi @olicooper 
Can you talk about a use case? Although I know that tasks can be delayed, why are they delayed in ABP?

liangshiw:
@olicooper Hi ,I see the `StartDelayed` method for those applications that need to be initialized, but the service registration of the ABP framework is completed before OnApplicationInitialization. I don't think it is necessary to add this option. Can you introduce other usage case?

olicooper:
@maliming my use case for this is that I need some jobs to run before any Quartz jobs and I also want to give the rest of the system time to load before jobs are loaded and configured (as they are less important than other core areas of the system).

maliming:
Quartz's delay is just `Task.Delay`. Can't you implement it yourself?

https://github.com/quartznet/quartznet/blob/b2b655db1819c8d8256981e2865ebf30433ec0fb/src/Quartz/Core/QuartzScheduler.cs#L398


And `StartupDelay `is a global configuration, it should support each task to configure its own delay time. We can add the `AbpQuartzDelayAttribute `and then `OnApplicationInitialization `method to delay startup according the attribute.

olicooper:
When Jobs are persisted in a database the previously saved start/end times tell Quartz if the job failed to run (due to down time etc). By adding a start delay to the whole Quartz scheduler, we delay all of the database queries that Quartz executes to restart those jobs which reduces the burden when the system starts.

I don't think I am able to add the `Task.Delay` to the Quartz module unless I override `AbpQuartzModule.OnApplicationInitialization` and provide my own implementation, but this seems overkill.


> @olicooper Hi ,I see the `StartDelayed` method for those applications that need to be initialized, but the service registration of the ABP framework is completed before OnApplicationInitialization. I don't think it is necessary to add this option. Can you introduce other usage case?

^^ @liangshiw Sorry, I am not sure what you mean by this. I don't think the scheduler starts to process jobs until the `Start`/`StartDelayed` method is called, so `StartDelayed` should be okay to call even if Quartz has already been registered in `OnApplicationInitialization`.

framework/src/Volo.Abp.Quartz/Volo/Abp/Quartz/AbpQuartzModule.cs
olicooper: Resolved: https://github.com/abpframework/abp/pull/3470

hikalkan:
Thanks.

PR #3457, opened 2020-04-02
enhance code

amrbadawy:
enhance code


maliming:
Thank you very much for your contribution, but this change is not necessary.

PR #3488, opened 2020-04-06
added description method and description tag

AkinSabriCam:


modules/docs/src/Volo.Docs.Web/Pages/Documents/Project/Index.cshtml.cs
```diff
@@ -591,5 +593,17 @@ public async Task SetDocumentPreferencesAsync()
                 }
             }
         }
+
+        public string GetDescription()
+        {
+            var startIndex = Document.Content.IndexOf("<p>", StringComparison.Ordinal);
```
ebicoglu, line 599: Ordinal is not enough, you should use `StringComparison.OrdinalIgnoreCase`

ebicoglu: This is a better way of getting description

 ```
public string GetDescription()
        {
            var firstParagraph = new Regex(@"<p>(.*?)</p>");
            var match = firstParagraph.Match(Document.Content);
            if (!match.Success)
            {
                return null;
            }

            var description = HttpUtility.HtmlDecode(match.Value);

            var htmlTagReplacer = new Regex(@"<[^>]*>", RegexOptions.IgnoreCase);
            description = htmlTagReplacer.Replace(description, m => string.Empty);

            return description.Truncate(MaxDescriptionMetaTagLength);
        }
```

PR #3538, opened 2020-04-09
PaginationTagHelper show info bug fix

emrekara37:
Hi,

https://bootstrap-taghelpers.abp.io/Components/Paginator
Screenshot_2


yekalkan:
Hi @emrekara37 

I'm sorry but i will not approve this pr. Because those localizations are needed in datatables and should remain same.

See https://github.com/abpframework/abp/blob/dev/framework/src/Volo.Abp.AspNetCore.Mvc.UI.Theme.Shared/wwwroot/libs/abp/aspnetcore-mvc-ui-theme-shared/datatables/datatables-extensions.js#L350

I'll fix this issue. Thanks for your effort.

Issue #3539, opened 2020-04-10
Navigation link problem for docs module multiple projects with the same name


liangshiw:
After adding multiple projects with the same name in the docs module, a list page will be displayed to select different items, which is good.
But the same navigation link results in only one project being displayed.

![image](https://user-images.githubusercontent.com/16813853/78963508-3b2eff00-7b2a-11ea-9f22-d2555d1c423a.png)

Is this a issue? If we have multiple projects, may we can consider adding GUID as a condition.

https://github.com/abpframework/abp/blob/94156f70b527d50e905b9cc903384051618117ca/modules/docs/src/Volo.Docs.Web/Pages/Documents/Project/Index.cshtml.cs#L177

maliming:
Your project name should be unique.

liangshiw:
Should you think `ShortName` is unique?  Because I can add duplicate projects in the admin page.

maliming:
> A short name of the project to be seen in URLs.

According to the current URL routing template, the `ShortName `of the project cannot be duplicate,
We should make sure it is unique in the domain.

PR #3556, opened 2020-04-11
Make docs module shortName unique

liangshiw:
Fixed :https://github.com/abpframework/abp/issues/3539


modules/docs/src/Volo.Docs.EntityFrameworkCore/Volo/Docs/Projects/EfCoreProjectRepository.cs
```diff
@@ -39,5 +38,10 @@ public async Task<Project> GetByShortNameAsync(string shortName)
 
             return project;
         }
+
+        public async Task<bool> ShortNameExistsAsync(string shortName)
+        {
+            return await DbSet.AnyAsync(x => x.ShortName == shortName);
```
yekalkan, line 44: This comparison should be case sensitive.

modules/docs/src/Volo.Docs.MongoDB/Volo/Docs/Projects/MongoProjectRepository.cs
```diff
@@ -38,5 +39,10 @@ public async Task<Project> GetByShortNameAsync(string shortName)
 
             return project;
         }
+
+        public async Task<bool> ShortNameExistsAsync(string shortName)
+        {
+            return await GetMongoQueryable().AnyAsync(x => x.ShortName == shortName);
```
yekalkan, line 45: This comparison should be case sensitive.

modules/docs/src/Volo.Docs.EntityFrameworkCore/Volo/Docs/Projects/EfCoreProjectRepository.cs
liangshiw: Ef core does not seem to support such a function. In addition, sqlserver is not case sensitive but not all databases are like this. So we should normalize ShortName.

PR #3592, opened 2020-04-14
docs: Added ExposeServices attribute.

berkansasmaz:
[ExposeServices(typeof(IdentityUserAppService))] attribute is required here since IdentityUserManager does not define an interface (like IIdentityUserAppService).


docs/en/Customizing-Application-Modules-Overriding-Services.md
```diff
@@ -60,6 +60,7 @@ In most cases, you will want to change one or a few methods of the current imple
 
 ````csharp
+[ExposeServices(typeof(IdentityUserAppService))]
```
maliming, line 63: It's should be `[ExposeServices(typeof(IIdentityUserAppService), typeof(IdentityUserAppService))]`

PR #3619, opened 2020-04-15
Add select2-bootstrap-modal-patch.js to Select2ScriptContributor.

### LINKED ISSUE #3609: Select2 is not working in opened modal by default
When I want to make a selectbox in modal to searchable with `Select2` it does not work. It looks like disabled, doesn't accept any writing.

resim

The solution is changing the modal contructor prototype by the following code: 
```$.fn.modal.Constructor.prototype._enforceFocus = function () { };```
https://github.com/select2/select2/issues/1436#issuecomment-21028474
https://select2.org/troubleshooting/common-problems

we should inspect and find possible effects by changing this prototype. If it does not occurs any problem, we can implement this in framework layer.

maliming:
Resolve #3609


hikalkan:
Should also do this immediately: https://github.com/abpframework/abp/issues/3620

PR #3651, opened 2020-04-17
Added abp generate-proxy - output switch

### LINKED ISSUE #3295: abp generate-proxy - output switch
It will bi nice if you can extend ABP CLI generate-proxy command with output directory option.

Sometimes we need for selected module to put generated code inside custom folder.

Thanks.

erolarkat:
Resolves #3295

### DIFF
--- a/framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/Commands/GenerateProxyCommand.cs
+++ b/framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/Commands/GenerateProxyCommand.cs
@@ -19,6 +19,7 @@ namespace Volo.Abp.Cli.Commands
     public class GenerateProxyCommand : IConsoleCommand, ITransientDependency
     {
         public static Dictionary<string, Dictionary<string, string>> propertyList = new Dictionary<string, Dictionary<string, string>>();
+        public static string output = "";
         public ILogger<GenerateProxyCommand> Logger { get; set; }
 
         protected TemplateProjectBuilder TemplateProjectBuilder { get; }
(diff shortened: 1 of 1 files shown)


framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/Commands/GenerateProxyCommand.cs
```diff
@@ -19,6 +19,7 @@ namespace Volo.Abp.Cli.Commands
     public class GenerateProxyCommand : IConsoleCommand, ITransientDependency
     {
         public static Dictionary<string, Dictionary<string, string>> propertyList = new Dictionary<string, Dictionary<string, string>>();
+        public static string output = "";
```
hikalkan, line 22: Why created a static variable instead of a local one at the line below:
````csharp
output = commandLineArgs.Options.GetOrNull(Options.Output.Short, Options.Output.Long);
````

Also, `output = ""` value is always overwritten.

Comment (erolarkat) [reply framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/Commands/GenerateProxyCommand.cs:22]: Output value is used in more than one method. Instead of moving the variable between methods, I preferred to define it globally. I chose to use static considering that we might need to override the output value when writing the files.

File: framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/Commands/GenerateProxyCommand.cs [hunk 2]
```diff
@@ -56,6 +57,12 @@ public async Task ExecuteAsync(CommandLineArgs commandLineArgs)
 
             var uiFramework = GetUiFramework(commandLineArgs);
 
+            output = commandLineArgs.Options.GetOrNull(Options.Output.Short, Options.Output.Long);
+            if (!string.IsNullOrWhiteSpace(output) && (!output.EndsWith("/") || !output.EndsWith("\\")))
```
carcer, line 61: Instead of this and using `output + ""`, why not use [Path.Combine](https://docs.microsoft.com/en-us/dotnet/api/system.io.path.combine?view=netcore-3.1)?

hikalkan:
I am moving this to 2.8. It has some conflicts should be resolved.

Issue #2939, opened 2020-02-28
Optimize MyProjectNameDbMigrationService


hikalkan:
* If we have 1000 tenants and they all using the same database, it runs the migration against to the same database 1000 times. It should check if tenant's connection string was migrated before (in this execution) and skip that tenant.
* Also, unify MigrateHostDatabaseAsync and MigrateTenantDatabasesAsync methods.

PR #3683, opened 2020-04-20
Optimize MyProjectNameDbMigrationService

liangshiw:
Resolved : https://github.com/abpframework/abp/issues/2939


templates/app/aspnet-core/src/MyCompanyName.MyProjectName.Domain/Data/MyProjectNameDbMigrationService.cs
```diff
@@ -37,51 +37,45 @@ public async Task MigrateAsync()
...
                 {
-                    Logger.LogInformation($"Migrating {tenant.Name} database schema... ({++i} of {tenants.Count})");
-                    await MigrateTenantDatabasesAsync(tenant);
-                    Logger.LogInformation($"Successfully completed {tenant.Name} database migrations.");
+                    using (_currentTenant.Change(tenant.Id))
+                    {
+                        await MigrateDatabaseAsync(tenant);
+                    }
+
+                    tenantConnectionStrings.ForEach(x => migratedDatabases.Add(x));
```
maliming, line 56: `migratedDatabases.AddIfNotContains(migratedDatabases);`

templates/app/aspnet-core/src/MyCompanyName.MyProjectName.Domain/Data/MyProjectNameDbMigrationService.cs
```diff
@@ -37,51 +37,45 @@ public async Task MigrateAsync()
...
+                
+                if (!migratedDatabases.Any() || !migratedDatabases.IsSupersetOf(tenantConnectionStrings))
                 {
-                    Logger.LogInformation($"Migrating {tenant.Name} database schema... ({++i} of {tenants.Count})");
-                    await MigrateTenantDatabasesAsync(tenant);
-                    Logger.LogInformation($"Successfully completed {tenant.Name} database migrations.");
+                    using (_currentTenant.Change(tenant.Id))
+                    {
+                        await MigrateDatabaseAsync(tenant);
+                    }
```
maliming, line 54: If two tenants use the same connection string, the data of the second tenant will not be migrated.

maliming: You need to combine the` tenant.Id` and connection string to determine whether the data has been migrated.

liangshiw:
I just judge whether the collection of tenant database connection strings is a superset of the migrated database connection strings．

In fact, a better way is to only judge the connection string of the main database (`Default`), But the connection string name is hardcoded in MyProjectNameDbContext.

https://github.com/abpframework/abp/blob/f0efccf877c9dafcd21bad1cde5e69ed1ba210b1/templates/app/aspnet-core/src/MyCompanyName.MyProjectName.EntityFrameworkCore/EntityFrameworkCore/MyProjectNameDbContext.cs#L20

Tenant management module uses `ConnectionStrings. DefaultConnectionStringName` constant, So the migration program cannot determine the connection string name very well.

PR #3700, opened 2020-04-22
Wrong polish culture code in translation file name

CrazyBaran:
Culture code for polish is pl-PL.


maliming:
Currently all projects are using `pl.json`

```json
{
  "culture": "pl",
  "texts": {

  }
}
```

CrazyBaran:
So it is breaking change?

maliming:
image

CrazyBaran:
You exlude test in this search, don't change it in tests or make push for them also?

maliming:
I forgot to unexclude the folder.

image

CrazyBaran:
Ok, so all files allign.

maliming:
I think this is a breaking change.

hikalkan:
Why not `pl-PL`, and not `pl`?
As you know, `pl-PL` (and actually all `pl-*`) fallbacks to `pl`, so `pl` covers more.
We prefer `pl` unless there are multiple `pl` accents.

CrazyBaran:
Well I think it is good practice to keep everything in one convention. Not mixing up pl alone with en-US for example. Anyway if you look into polish documentation on microsoft they also used pl-pl.

hikalkan:
@maliming I agree this is a breaking change, but it is very minor and easy to fix for existing applications. If they've used `pl`, they need to switch to `pl-PL`. Since `pl-PL` seems the only valid value (there is no other `pl-x`), this way seems better. I marked it as breaking change to get attention.

PR #3720, opened 2020-04-23
Blogging Module - Unwanted Exception Suppressing

cotur:
Example exceptions.

```
There is no such an entity. Entity type: Volo.Blogging.Blogs.Blog, id: favicon.ico
There is no such an entity. Entity type: Volo.Blogging.Blogs.Blog, id: robots.txt
There is no such an entity. Entity type: Volo.Blogging.Blogs.Blog, id: wp-login.php
```


modules/blogging/src/Volo.Blogging.Application/Volo/Blogging/Blogs/BlogAppService.cs
```diff
@@ -32,6 +33,13 @@ public async Task<BlogDto> GetByShortNameAsync(string shortName)
 
             if (blog == null)
             {
+                FileInfo fi = new FileInfo(shortName);
+
+                if (!string.IsNullOrEmpty(fi.Extension))
+                {
+                    return null;
```
hikalkan, line 40: Did you check if it is OK to return null from this method (check all usages)? We don't want "object reference not set..." exceptions.

hikalkan:
What about the status of this PR? Is it still draft?

cotur:
@hikalkan yes it is. I've just decided to what is best way to do for it.

modules/blogging/src/Volo.Blogging.Web/Pages/Blogs/Shared/Helpers/BlogNameControlHelper.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    public static class BlogNameControlHelper
+    {
+        public static readonly string[] ProhibitedFileExtensions = new string[] {".ico", ".txt", ".php"};
+
+        public static bool IsProhibitedFileFormatName(string blogShortName)
+        {
+            if (!string.IsNullOrWhiteSpace(blogShortName))
+            {
+                var fileInfo = new FileInfo(blogShortName);
```
hikalkan, line 14: don't use FileInfo, but only use simple case insensitive comparison

PR #3779, opened 2020-04-29
Explicitly define @Injectable() decoration requirement.

mgochmuradov:
From the documentation, it was not obvious to add `@Injectable()` decorator, So I had to do some googling to realize it.
I think this change could save some time for other developers.


mehmet-erim:
Thanks @mgochmuradov 🙂

Issue #3796, opened 2020-04-30
Virtual file system could not get the root directory


liangshiw:
![image](https://user-images.githubusercontent.com/16813853/80670204-df2c1a80-8ad8-11ea-88e0-0298dc4b520f.png)

maliming:
Related: https://github.com/abpframework/abp/issues/1609#issuecomment-520242986

PR #3801, opened 2020-04-30
make virtual file system return the root directory

liangshiw:
Resolved : https://github.com/abpframework/abp/issues/3796


framework/src/Volo.Abp.VirtualFileSystem/Volo/Abp/VirtualFileSystem/VirtualFilePathHelper.cs
```diff
@@ -10,6 +10,11 @@ internal static class VirtualFilePathHelper
 
         public static string NormalizePath(string fullPath)
         {
+            if (fullPath == "/")
```
maliming, line 13: `string.Equals(fullPath, "/", StringComparison.Ordinal)`

PR #3834, opened 2020-05-05
add Localize Emailing resource for zh-hant

ChangYinShung:


maliming:
Thanks @ChangYinShung

PR #3850, opened 2020-05-07
Added onOpen callback, to trigger when the modal opens.

Znow:
```js
// Example
editModal.onOpen(function (e) {
  console.log("modal opened");
});
```


hikalkan:
Thanks for your contribution.

PR #3918, opened 2020-05-12
Make ICurrentPrincipalAccessor changeable.

### LINKED ISSUE #3913: CurrentUser will be invalid when LongPolling is used as Transports.
We need to use Hub's `HubCallerContext `property to get clsims.

maliming:
Resolve #3913

This way we can replace `ClaimsPrincipal` with `HubCallerContext.User` in signalr.

Because many services are using `ICurrentPrincipalAccessor`. Developers may call these methods, we must ensure that `ClaimsPrincipal `is always obtained correctly.


hikalkan:
Good implementation. Thanks.

gdlcf88:
I think this is a missing breaking change in v2.8.0

hikalkan:
Agree. We missed that anyone could implement this public interface.

maliming:
We can inherit the `ThreadCurrentPrincipalAccessor` class to mock this interface.

eg https://github.com/abpframework/abp/blob/5a2b36e/templates/app/aspnet-core/test/MyCompanyName.MyProjectName.TestBase/Security/FakeCurrentPrincipalAccessor.cs#L9

Issue #3924, opened 2020-05-12
abp-paginator tag helper generates incorrect href attribute value


liangshiw:
I saw that using js in the demo fixed the problem, but it is not a universal solution. Maybe we can generate the correct href attribute in the tag helper service.

https://github.com/abpframework/abp/blob/7246b227bed0fcd93e083e4a0ce4bb5a937c7053/framework/test/Volo.Abp.AspNetCore.Mvc.UI.Theme.Basic.Demo/Pages/Components/Paginator/Index.js#L6

PR #3930, opened 2020-05-13
Make AbpPaginationTagHelper generate correct href attribute

liangshiw:
Resolves https://github.com/abpframework/abp/issues/3924

### DIFF
--- a/framework/src/Volo.Abp.AspNetCore.Mvc.UI.Bootstrap/TagHelpers/Pagination/PagerModel.cs
+++ b/framework/src/Volo.Abp.AspNetCore.Mvc.UI.Bootstrap/TagHelpers/Pagination/PagerModel.cs
@@ -38,6 +38,11 @@ namespace Volo.Abp.AspNetCore.Mvc.UI.Bootstrap.TagHelpers.Pagination
             PageSize = pageSize;
             TotalPageCount = (int)Math.Ceiling(Convert.ToDouble((decimal)TotalItemsCount / PageSize));
             Sort = sort;
+
+            if (pageUrl.IsNullOrWhiteSpace() || !pageUrl.StartsWith("/"))
+            {
+                pageUrl = "/" + pageUrl;
+            }
             PageUrl = pageUrl;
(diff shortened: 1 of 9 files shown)


framework/src/Volo.Abp.AspNetCore.Mvc.UI.Bootstrap/TagHelpers/Pagination/PagerModel.cs
```diff
@@ -38,6 +38,11 @@ public PagerModel(long totalCount, int shownItemsCount, int currentPage, int pag
             PageSize = pageSize;
             TotalPageCount = (int)Math.Ceiling(Convert.ToDouble((decimal)TotalItemsCount / PageSize));
             Sort = sort;
+
+            if (pageUrl.IsNullOrWhiteSpace() || !pageUrl.StartsWith("/"))
```
maliming, line 42: `pageUrl = "/" + pageUrl;` will error when `pageUrl` is `null`

liangshiw: See: https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/strings/#null-strings-and-empty-strings
These overloads of the binary + operator perform string concatenation. If an operand of string concatenation is null, an empty string is substituted.

yekalkan, line 42: You can use `pageUrl = pageUrl?.EnsureStartsWith('/') ?? "/"`.

See https://github.com/abpframework/abp/blob/dev/framework/src/Volo.Abp.Core/System/AbpStringExtensions.cs#L33

liangshiw: Ok

PR #3948, opened 2020-05-13
Update CLI.md

proovit:
new win install, new vs2019 install, abp install and then: 
'abp' is not recognized as an internal or external command ...
after restart this was ok


maliming:
I think this may have something to do with net core sdk instead of abp cli need to restart.

Issue #1609, opened 2019-08-09
Virtual File System Explorer


vnetonline:
@hikalkan I want to create a Virtual File System browser as a module 

Can you let me know what your approach would be? It would be good to see what is in the Virtual File System when developing to override views and pages etc etc 

Once developed happy for you to include it as a module in the repository.

hikalkan:
That is a good idea!  It can be useful in development.

Inject `IVirtualFileProvider` and use `GetDirectoryContents("/")` to get the root folder contents. It will return file and folders. Then you can reuse it to enter another folder and so on...

vnetonline:
@hikalkan `GetDirectoryContents` Is not a Async function, how do I use it in the ApplicationService

Also it seems it only returns physical files should it return files and directors from all three file providers memory, physical and dynamic ?

hikalkan:
It should return all, not only physical files.

A sync method can directly usable inside an async method. Do you have any problem of that?

vnetonline:
No problem I thought that a ApplicationService should always return Async because the UI layer will call it using HttpAPI and whenever calling a HttpApi it should be done Async way as best practice ?? Is my understanding wrong ?

hikalkan:
You are true. However, not possible (or meaningfull) to make all methods async.
Calling an async code from a sync method may cause problems. But vise verse does not.

BTW, this method is defined by Microsoft: https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.fileproviders.ifileprovider?view=aspnetcore-2.2
So, I can't make it async :)

vnetonline:
But the method I write in my application project should be Async right ?

Should I just put a task.delay ??

hikalkan:
>But the method I write in my application project should be Async right ?
No, not required. However, I suggest it. Never do task.delay :) Remove async keyword and return Task.CompletedTask; (for void methods) or Task.FromResult(...) (for method returns a value).

vnetonline:
@hikalkan 

thanks for your help that makes sense, I am not getting anything when I try and call `GetDirectoryContents("/")`

<img width="1185" alt="Screen Shot 2019-08-21 at 12 11 39 pm" src="https://user-images.githubusercontent.com/15242923/63397286-68993200-c40d-11e9-8dc9-aaf51a57713f.png">

I have changed the `launchSetting.json` as follows to simulate Production mode

```
{
  "iisSettings": {
    "windowsAuthentication": false,
    "anonymousAuthentication": true,
[...]

PR #3971, opened 2020-05-16
Virtual File System Explorer Module

liangshiw:
Resolve https://github.com/abpframework/abp/issues/1609

Provide a simple page to view all virtual files.

Use:

```csharp
[DependsOn(typeof(AbpVirtualFileExplorerWebModule))]
public class <YourProjectWebModule> : AbpModule
```

Options:

```csharp
public class AbpVirtualFileExplorerOptions
{
     /// <summary>
     /// Default: true.
     /// </summary>
     public bool IsEnabled { get; set; } = true;
}
```

Demo:

image

image


modules/virtual-file-explorer/src/Volo.Abp.VirtualFileExplorer.Web/FodyWeavers.xml
```diff
@@ -0,0 +1,3 @@
+﻿<Weavers xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="FodyWeavers.xsd">
+  <ConfigureAwait />
```
maliming, line 2: <ConfigureAwait ContinueOnCapturedContext="false" />

liangshiw: Ok

hikalkan:
That's a great module! The only thing is: You embedded primjs into the module. However, using a NPM package would be great. As a result, this module will have a NPM package (just like https://github.com/abpframework/abp/blob/dev/npm/packs/docs/package.json). Its package name can be "@abp/virtual-file-explorer" Once the final application adds this package and runs yarn & gulp, prismjs will be added and perfectly works. You just need to use the PrismJs contributors. 

@maliming can help you if you need?

maliming:
This requires some work and testing. Move to v2.9.

maliming:
@hikalkan We can release it in 2.8.

kurtwilbies:
I dont like the naming, e.g. DemoApp.

hikalkan:
Thanks a lot @liangshiw

PR #3982, opened 2020-05-17
Define CollectionFixture in each test project.

maliming:
Related issue: https://github.com/abpframework/abp/pull/2414

Because: **xUnit collection definitions must be in the same assembly as the test that uses them.**

When we use mongodb as the database provider, unit tests need to share `MongoDbRunner`.


maliming:
When using EF Core we can consider removing these unnecessary classes.

hikalkan:
Can we move this to 2.9, I want to understand if this is really needed, because it brings additional complexities.

maliming:
Because `mongo.exe` will block the github action pipeline.

https://github.com/abpframework/abp-samples/pull/6/commits/bee955155f6710a7e04826f0c287ffba1fb03919

hikalkan:
I want to move this to the next milestone. Can you fix the conflict? Thanks.

maliming:
Reproducing the problem is simple. After the unit test, three `mongod.exe` processes are not terminated.  


```
abp new Acme.BookStore -d mongodb
dotnet build
dotnet test
```

Because unit tests for `Web.Tests, Application.Tests, Domain.Tests` depend on `MongoDB.Tests`. But **xUnit collection definitions must be in the same assembly as the test that uses them**.

image

PR #3985, opened 2020-05-17
Add AbpClaimsMapMiddleware.

### LINKED ISSUE #3975: Create middleware for claims map. Make it configurable.
```c#
app.Use(async (ctx, next) =>
{
	var currentPrincipalAccessor = ctx.RequestServices.GetRequiredService<ICurrentPrincipalAccessor>();
	var map = new Dictionary<string, string>()
	{
		{ "sub", AbpClaimTypes.UserId },
		{ "role", AbpClaimTypes.Role },
		{ "email", AbpClaimTypes.Email },
		//any other map
	};
	var mapClaims = currentPrincipalAccessor.Principal.Claims.Where(p => map.Keys.Contains(p.Type)).ToList();
	currentPrincipalAccessor.Principal.AddIdentity(new ClaimsIdentity(mapClaims.Select(p => new Claim(map[p.Type], p.Value, p.ValueType, p.Issuer))));
	await next();
});
```

maliming:
Resolve #3975


hikalkan:
Great! Should we add these to the options by default?

````
		{ "sub", AbpClaimTypes.UserId },
		{ "role", AbpClaimTypes.Role },
		{ "email", AbpClaimTypes.Email },
````

What do you think?

And also, we may replace current code duplications with this middleware?

maliming:
OK, I will continue.

maliming:
@hikalkan Done.

hikalkan:
I did one more improvement: https://github.com/abpframework/abp/commit/34ec1ad0e73e1f243fa79f2fdf9e58346749673c

You know a developer may set AbpClaimTypes.UserId, since it is static (not const). Getting the value like this (Func<string>) make it getting the latest value, not copied the value on the mapping time.

PR #3989, opened 2020-05-18
"L" function in the text template rendering system should support parametric text.

### LINKED ISSUE #3977: "L" function in the text template rendering system should support parametric text
For a translation like `Hello {0}, how are you?`, we need to pass the `{0}` as a parameter to the `L` function. Currently, it is no possible.

A good implementation can be like that:

````
{{ L "HelloKey" "John" }}
````

or (using a model property)

````
{{ L "HelloKey" model.name }}
````

which produces `Hello John, how are you?`

If it can be implemented, also update [the document](https://github.com/abpframework/abp/blob/dev/docs/en/Text-Templating.md#localization).

maliming:
Resolve #3977


hikalkan:
Great!

Issue #3976, opened 2020-05-16
Console application startup template


hikalkan:
Prepare an application startup template to create a simple console application using the CLI.

PR #3993, opened 2020-05-18
Add console application startup template

liangshiw:
Resolve https://github.com/abpframework/abp/issues/3976


maliming:
hi @yekalkan 
We need to add the `template` entity of the `console` to the abp.io platform. Can you help?

hikalkan:
Thank you @liangshiw for your contribution :)

@maliming I added it.

Issue #2036, opened 2019-10-30
Swashbuckle has wrong framework dependency


MaiReo:
![image](https://user-images.githubusercontent.com/11777709/67820212-4d286280-faf3-11e9-9de8-52749e0f50ba.png)

I want to know why you decided to use Swashbuckle.AspNetCore.
It has wrong deps aslo at past, e.g. JsonFormatter 1.0.0 
thank you

maliming:
What impact does this have on your application?

Maybe you should ask [Swashbuckle.AspNetCore ](https://github.com/domaindrivendev/Swashbuckle.AspNetCore)

MaiReo:
@maliming 
There is no issue what I found yet.

hikalkan:
>I want to know why you decided to use Swashbuckle.AspNetCore.

What do you suggest?

PR #3997, opened 2020-05-18
created a test project and moved page toolbar test to abp framework solution

AkinSabriCam:


hikalkan:
Are you sure that it resolves #2036 ?

PR #4016, opened 2020-05-20
Optimize RazorRuntimeCompilation.

### LINKED ISSUE #4014: Test whether RazorRuntimeCompilation can be made configurable
In a production environment, runtime compilation does not seem to be commonly used and affects performance.

maliming:
Resolve #4014

I removed the `cshtml` files from the `virtual file system`. These `cshtml` files will be pre-compiled by dotnet. This will speed up the first time the page opens.

When we want to replace `cshtml`, we can add these `cshtml` files in the `virtual file system`, of course, the physical cshtml files during development also take effect. (`ReplaceEmbeddedByPhysical`)

`Account/Register`  page first opening speed comparison:

Before: `990.9776ms` vs after `284.0663ms`

```
2020-05-20 15:36:58.175 +08:00 [INF] Request starting HTTP/2 GET https://localhost:44303/Account/Register  
//...
2020-05-20 15:36:59.166 +08:00 [INF] Request finished in 990.9776ms 200 text/html; charset=utf-8


2020-05-20 15:38:07.742 +08:00 [INF] Request starting HTTP/2 GET https://localhost:44303/Account/Register  
//...
2020-05-20 15:38:08.026 +08:00 [INF] Request finished in 284.0663ms 200 text/html; charset=utf-8
```

I also need to test other details.


framework/src/Volo.Abp.AspNetCore.Mvc/Volo/Abp/AspNetCore/Mvc/ViewFeatures/AbpRazorCompiledItemFeatureProvider.cs
```diff
@@ -0,0 +1,62 @@
...
+using Microsoft.Extensions.DependencyInjection;
+using Volo.Abp.VirtualFileSystem;
+
+namespace Volo.Abp.AspNetCore.Mvc.ViewFeatures
+{
+    /// <summary>
+    /// The code for this class comes from
+    /// https://github.com/dotnet/aspnetcore/blob/master/src/Mvc/Mvc.Razor/src/ApplicationParts/RazorCompiledItemFeatureProvider.cs
+    /// </summary>
+    public class AbpRazorCompiledItemFeatureProvider : IApplicationFeatureProvider<ViewsFeature>
```
maliming, line 15: This class does not seem to be needed, because razor runtime compilation will use the virtual file system to compare whether the page needs to be recompiled.

maliming:
If the razor runtime compilation finds that there is a pre-compiled page, it will use it directly, unless the file provider (`MvcRazorRuntimeCompilationOptions.FileProviders`) finds that there is a page and the page has changed.

So we don't need to add `cshtml` to the `virtual file system` by default.

This will not cause breaking changes.

hikalkan:
That's great! I see a huge performance improvement on application startup.

I also made a few more improvements that truly sorts application parts considering the module dependencies, so a module can override a view of a depended module.

gdlcf88:
Hi @maliming, could you please add the "abp-template-module" label?

PR #4032, opened 2020-05-21
Add useful services to IToolbarConfigurationContext and MenuConfigurationContext.

### LINKED ISSUE #4017: Add useful services to IToolbarConfigurationContext and MenuConfigurationContext
There are some common operations, like checking permissions and localizing strings, are frequently done in these kind of contributors.
We can design shortcuts on these contexts.

Example:

````csharp
public class ChatToolbarContributor : IToolbarContributor
{
    public async Task ConfigureToolbarAsync(IToolbarConfigurationContext context)
    {
        if (context.Toolbar.Name == StandardToolbars.Main)
        {
            var authorizationService = context.ServiceProvider
                .GetRequiredService<IAuthorizationService>();

            if (await authorizationService.IsGrantedAsync(ChatPermissions.Messaging))
            {
                context.Toolbar.Items
                    .Insert(0, new ToolbarItem(typeof(MessagesToolbarItemViewComponent)));
            }
        }
    }
}
````

Can be written easier if we have a `context.IsGrantedAsync` method. We should resolve these services as lazy.

maliming:
Resolve #4017

**Todo**: Update usage codes in modules and templates.


hikalkan:
Thanks. I added a few more methods.

Issue #4040, opened 2020-05-21
SettingManagementStore not refreshing cache within the same UnitOfWork


javiercampos:
I made this test for [SettingManagementStore](https://github.com/abpframework/abp/blob/dev/modules/setting-management/src/Volo.Abp.SettingManagement.Domain/Volo/Abp/SettingManagement/SettingManagementStore.cs), which fails (on its [test file](https://github.com/abpframework/abp/blob/dev/modules/setting-management/test/Volo.Abp.SettingManagement.Tests/Volo/Abp/SettingManagement/SettingManagementStore_Tests.cs)):

```c#
/* .. declare and get service of `IUnitOfWorkManager` on constructor of the test class... */
[Fact]
public async Task SetThenGetOnUnitOfWorkAsync()
{
    using (_unitOfWorkManager.Begin())
    {
        var value = await _settingManagementStore.GetOrNullAsync("MySetting1",
            GlobalSettingValueProvider.ProviderName, null);
        value.ShouldBe("42");

        await _settingManagementStore.SetAsync("MySetting1", "43", GlobalSettingValueProvider.ProviderName,
            null);

        var valueAfterSet =
            await _settingManagementStore.GetOrNullAsync("MySetting1", GlobalSettingValueProvider.ProviderName,
                null);
        valueAfterSet.ShouldBe("43");
    }
}
```

Test result:
```
Shouldly.ShouldAssertException : valueAfterSet
    should be
"43"
    but was
"42"
```

That's because the cache is invalidated only when the `UnitOfWork` finished (it uses an `ILocalEventHandler` on [SettingCacheItemInvalidator](https://github.com/abpframework/abp/blob/dev/modules/setting-management/src/Volo.Abp.SettingManagement.Domain/Volo/Abp/SettingManagement/SettingCacheItemInvalidator.cs))

The scenario above is *pretty common*: if you have an API endpoint which updates a setting, it's  *VERY NORMAL* wanting to use the new value from other service (the one using the setting), potentially in the same request, called from the same app service (with the same UnitOfWork).

[...]

PR #4048, opened 2020-05-22
Invalidate cache on setting update, allow UnitOfWork re-reads

javiercampos:
Invalidate cache on setting `Set` (references #4040).

There are still some problems with this approach: mainly, if we do a: `get (creates cache) -> set (invalidates cache) -> get (creates cache with updated setting) -> unit of work fails saving`, the cache is kept in inconsistent state.

This is just a first approach, extra thinking should be taken to maintain consistent cache state in all scenarios (mainly UOW scenarios), which would probably require tracking the cache changes and rolling back the changes in case the UnitOfWork fails committing/saving changes.

### DIFF
--- a/modules/setting-management/src/Volo.Abp.SettingManagement.Domain/Volo/Abp/SettingManagement/SettingManagementStore.cs
+++ b/modules/setting-management/src/Volo.Abp.SettingManagement.Domain/Volo/Abp/SettingManagement/SettingManagementStore.cs
@@ -42,6 +42,10 @@ namespace Volo.Abp.SettingManagement
             {
                 setting.Value = value;
                 await SettingRepository.UpdateAsync(setting);
+                
+                /* Re-set on cache */
+                var cacheKey = CalculateCacheKey(name, providerName, providerKey);
+                await Cache.RemoveAsync(cacheKey);
             }
         }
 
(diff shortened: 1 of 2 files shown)


maliming:
`UnitOfWork `has an `Items `property,  we can store the `CacheKey `to `Items `in the `Set `method of `SettingManagementStore `and check the `Items `property of `UnitOfWork `in the `Get `method of `SettingManagementStore`.

javiercampos:
@maliming I've added more comments on #4040 about possible approaches, yes, that seems the way to go on this

maliming:
hi @javiercampos  Will you continue to commit code?

javiercampos:
@maliming I wasn't really planning to, I thought the discussion on #4040 about this was not over and @hikalkan was thinking about the best solution (as he stated on the issue)

hikalkan:
I am closing this since it has problems as we know.

@maliming can you please implement the solution (Uow.Items) you've offered. Hold a dictionary as an item in the items of the current UOW and try get from the current UOW if the value is available there.

You can also extract a class like `SettingManagementStoreCache` to separate concerns (but don't use the repository [inside it](https://github.com/abpframework/abp/blob/dev/modules/setting-management/src/Volo.Abp.SettingManagement.Domain/Volo/Abp/SettingManagement/SettingManagementStore.cs#L73), instead get a factory method which is sent by the store and implemented with using the repository)

Issue #4052, opened 2020-05-22
Customizing-Application-Module doc could remind to disable RemoteService


gdlcf88:
By default, ABP apps have **Auto API Controllers** enabled. When developers need to override module's application services, the new class should have the attribute `[RemoteService(IsEnabled = false)]`.

For example:
```csharp
    [RemoteService(IsEnabled = false)]
    [Dependency(ReplaceServices = true)]
    [ExposeServices(typeof(IResourceAppService), typeof(ResourceAppService))]
    public class MyResourceAppService : ResourceAppService
    {
        // ...
    }
```

If the `[RemoteService(IsEnabled = false)]` attribute is not added, these API will be generated:
![MEPWU){YW6F0F97X7YPS5GJ](https://user-images.githubusercontent.com/30018771/82651298-8a4e7080-9c4e-11ea-8f63-b9808534d073.jpg)

Maybe [the document](https://docs.abp.io/en/abp/latest/Customizing-Application-Modules-Overriding-Services#example-overriding-an-application-service) could show a tip to remind devlopers to add `[RemoteService(IsEnabled = false)]` attribute.

PR #4054, opened 2020-05-22
Add a RemoteService tip to the module overriding doc

gdlcf88:
See #4052

### DIFF
--- a/docs/en/Customizing-Application-Modules-Overriding-Services.md
+++ b/docs/en/Customizing-Application-Modules-Overriding-Services.md
@@ -59,6 +59,7 @@ In most cases, you will want to change one or a few methods of the current imple
 ### Example: Overriding an Application Service
 
 ````csharp
+//[RemoteService(IsEnabled = false)] // If you use dynamic controller feature you can disable remote service. Prevent creating duplicate controller for the application service.
 [Dependency(ReplaceServices = true)]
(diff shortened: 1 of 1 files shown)


hikalkan:
@maliming can you check if they really generated (even if we use `[Dependency(ReplaceServices = true)]` and `[ExposeServices(typeof(IdentityUserManager))]`).

docs/en/Customizing-Application-Modules-Overriding-Services.md
```diff
@@ -99,6 +99,8 @@ This class **overrides** the `CreateAsync` method of the `IdentityUserAppService
 
 You could completely **re-write** the entire business logic for a user creation without calling the base method.
 
+> It's better to add the attribute `[RemoteService(IsEnabled = false)]` if the [Auto API Controllers](API/Auto-API-Controllers.md) is enabled, otherwise the extra APIs for MyIdentityUser will be generated.
+
```
maliming, line 103:
```c#
//[RemoteService(IsEnabled = false)] // If you use a dynamic controller you can disable remote services. Prevent creating duplicate controllers for the application service.
[ExposeServices(typeof(IIdentityUserAppService), typeof(IdentityUserAppService))]
public class MyIdentityUserAppService : IdentityUserAppService

```

PR #4055, opened 2020-05-22
Convert c# type to typescript type.

### LINKED ISSUE #4051: use generate-proxy get more error
in c#,I defined a dto
``` csharp
public class UserListDto : EntityDto<int>
    {
        public string RealName { get; set; }
    }
```
Next use `abp generate-proxy`.I get error!
image
The resulting ts files are as follows：

1. in user-list-dto.ts
``` ts
import { EntityDto } from '@abp/ng.core';
export class UserListDto extends EntityDto<System.Int32> {
  realName: string;
  id: number;

  constructor(initialValues: Partial<UserListDto> = {}) {
    super(initialValues);
  }
}
```
2. in int32.ts
``` ts
import { EntityDto } from '@abp/ng.core';
export class Int32 extends EntityDto {
  id: number;

  constructor(initialValues: Partial<Int32> = {}) {
    super(initialValues);
  }
}

```
The above is an error with upgrade 2.8.0

maliming:
Resolve #4051

I am confused about this code. Will it cause problems if a class contains `G/guid` replaced by `string`?

eg `public class MyGuidDto`

```c#
if (baseTypeName.Contains("guid") || baseTypeName.Contains("Guid"))
{
	baseTypeName = "string";
}
```
https://github.com/abpframework/abp/pull/4055/files#diff-ae62d3da26b0d94a192555bce181d5a7R532-R535

### DIFF
--- a/framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/Commands/GenerateProxyCommand.cs
+++ b/framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/Commands/GenerateProxyCommand.cs
@@ -13,6 +13,7 @@ using Volo.Abp.DependencyInjection;
 using System.Collections.Generic;
 using System.Linq;
 using System.Net;
+using Volo.Abp.Reflection;
 
 namespace Volo.Abp.Cli.Commands
 {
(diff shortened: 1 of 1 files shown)


maliming:
I think we can split the function of `GenerateProxyCommand `class into multiple methods or services, because it currently completes all operations in one method, it is difficult for other developers to quickly understand this class. : )

erolarkat:
Yes you are right. We have a plan to refactor the code into services and methods. You are also right about guid :)

Issue #3352, opened 2020-03-25
Improve relative paths in ABP bundling


Xeevis:
Related to #3341 

In my opinion ABP bundling should use relative paths by following patterns in ASP.NET Core. Because if you are moving existing working stylesheet links into ABP bundles, it creates surface for new bugs.

While all of these patterns work
```html
<link rel="stylesheet" href="site1.css"> <!-- working relative link -->
<link rel="stylesheet" href="/site2.css"> <!-- working relative link -->
<link rel="stylesheet" href="./site3.css"> <!-- working relative link -->
<link rel="stylesheet" href="~/site4.css"> <!-- working relative link -->
```

For ABP bundles 2 out of those 4 don't

```html
<abp-style src="site5.css" /> <!-- broken -->
<abp-style src="/site6.css" /> <!-- working -->
<abp-style src="./site7.css" /> <!-- working -->
<abp-style src="~/site8.css" /> <!-- broken -->
```

Considering both are relative to wwwroot, shouldn't they work the same?

maliming:
@hikalkan  What do you think? https://github.com/abpframework/abp/issues/3341#issuecomment-603688351

hikalkan:
It seems reasonable, we can implement if possible.

alexandis:
I have published a couple of _web applications_ (**IdentityServer** and **HttpApi.Host** from **abp.io** boilerplate) under _web site_ on my VM and while I have managed to fix the issues with relative paths, I still experience the issue with loading resources in **IdentityServer**: the path to __bundle is calculated from WEB SITE root, not from ITS APPLICATION root. I've tried to set `webBuilder.UseWebRoot` / `webBuilder.UseContentRoot` / `hostingEnvironment.WebRootPath` / `hostingEnvironment.ContentRootPath`, but no luck.

hikalkan:
It is done.

maliming:
Binding still needs to deal with path issues.

alexandis:
Hi, what is the progress with this?
I'm on ABP 2.9.0 and tried to publish **Angular\dist** + **IdentityServer** + **HttpApi.Host** to _IIS_ on _Azure VM_. And despite the fact I've tried to fix all the paths (**base url** when making build, root folder in **angular.json**, links to **IdentityServer** and **HttpApi.Host** in **environment.ts**), I cannot run **Angular** successfully:

[...]

PR #4087, opened 2020-05-25
Consistent with aspnet core behavior of referencing static resources

liangshiw:
Resolve https://github.com/abpframework/abp/issues/3352

`~`  in aspnet core will be converted into subpath (if using iis sub-application deployment).I added 
 support sub-application deployment way for the bundling system.

Example :

image


maliming:
Have you tried `RequestPath` of `StaticFileOptions`?

https://docs.microsoft.com/en-us/aspnet/core/fundamentals/static-files?view=aspnetcore-3.1#serve-files-outside-of-web-root

liangshiw:
I tried, but it didn't work

PR #4097, opened 2020-05-26
Change the ErrorMessage of StringLengthAttribute according to the situation.

### LINKED ISSUE #4094: DataAnnotationsLocalization is not working as expected.
The content of data-val-length should be **The field Title must be a string with a minimum length of 5 and a maximum length of 256.**

```
[Required]
[StringLength(QuestionConsts.MaxTitleLength, MinimumLength = 5)]
public string Title { get; set; }

<input 
data-val-length="The field Title must be a string with a maximum length of 256." 
data-val-length-max="256" 
data-val-length-min="5" 
data-val-required="The Title field is required." maxlength="256">
```

maliming:
Resolve #4094

It seems that only `StringLengthAttribute `is special, so we can change its `ErrorMessage`.


hikalkan:
Great!

Issue #4110, opened 2020-05-27
Configure the virtual file system module for the template


liangshiw:
Some developers often replace the module page, we should configure the virtual file system module for the template

Issue #3987, opened 2020-05-18
overriding-a-razor-page-cshtml is not visable when publish to docker


xyfy:
I am flow the [docs](https://docs.abp.io/en/abp/latest/UI/AspNetCore/Customization-User-Interface#overriding-a-razor-page-cshtml) to override the razor page, it work well in local,but when I publish it to docker.    it looks the built-in one , not my custom page. 
Is there any steps I need to do?

my docker file is below

```
#See https://aka.ms/containerfastmode to understand how Visual Studio uses this Dockerfile to build your images for faster debugging.

FROM mcr.microsoft.com/dotnet/core/aspnet:3.1-buster-slim AS base
WORKDIR /app
EXPOSE 80
EXPOSE 443

FROM mcr.microsoft.com/dotnet/core/sdk:3.1-buster AS build

WORKDIR /src
COPY . .
WORKDIR "/src/applications/Yun.AuthServer.Host"

RUN dotnet restore "Yun.AuthServer.Host.csproj" -nowarn:msb3202,nu1503
# Prevent 'Warning: apt-key output should not be parsed (stdout is not a terminal)'
ENV APT_KEY_DONT_WARN_ON_DANGEROUS_USAGE=1

# use aliyun mirrors
#RUN cp -r /etc/apt/sources.list /etc/apt/sources.list.bak 
#RUN sed -i "s/http:\/\/deb.debian.org/https:\/\/mirrors.aliyun.com/g" /etc/apt/sources.list
#RUN sed -i "s/http:\/\/security.debian.org/https:\/\/mirrors.aliyun.com/g" /etc/apt/sources.list
#RUN sed -i "s/http:\/\/archive.debian.org/https:\/\/mirrors.aliyun.com/g" /etc/apt/sources.list
# install NodeJS 12.x
# see https://github.com/nodesource/distributions/blob/master/README.md#deb
RUN apt-get update -yq && apt-get upgrade -yq && apt-get install -yq curl
RUN curl -sL https://deb.nodesource.com/setup_12.x | bash -
RUN apt-get install -y nodejs
# use taobao mirrors
RUN npm install -g cnpm --registry=https://registry.npm.taobao.org
RUN npm install -g gulp
RUN npm install
RUN gulp

RUN dotnet build "Yun.AuthServer.Host.csproj" --no-restore -c Release -o /app

FROM build AS publish
RUN dotnet publish "Yun.AuthServer.Host.csproj" --no-restore -c Release -o /app

FROM base AS final
[...]

Issue #3766, opened 2020-04-28
.CSHTML override invalid when an app run in Linux + Docker + Production env.


gdlcf88:
![image](https://user-images.githubusercontent.com/30018771/80467693-c99be100-8970-11ea-88c2-b11207f3d91c.png)
```
...
        <abp-modal-body>
            Hello
            <abp-form-content/>
        </abp-modal-body>
...
```
I overrode a .CSHTML file, when I run the web project in Windows locally (tried: debug/release, production/development), it works:
![image](https://user-images.githubusercontent.com/30018771/80467951-28f9f100-8971-11ea-9386-1e9b79790780.png)
But when I published it to the formal environment: Linux + Docker + Production env, the inserted text "Hello" cannot be found in the create modal.

I created a repository to show my usage:
https://github.com/gdlcf88/ModuleOverrideTest

maliming:
Try to test the application after publishing it locally (windows).

gdlcf88:
> Try to test the application after publishing it locally (windows).

I tried published .dll file locally (Win10), it **can be** reproduced.

And I tried to replace the login page following the [example](https://docs.abp.io/en/abp/latest/UI/AspNetCore/Customization-User-Interface#example-1), the problem **can be** reproduced too.

maliming:
Try this demo.

https://github.com/maliming/LoginPageDemo/blob/master/src/LoginPageDemo.Web/Pages/Account/Login.cshtml

gdlcf88:
![image](https://user-images.githubusercontent.com/30018771/80477289-45505a80-897e-11ea-9674-e83cf17bdfac.png)
It works in this demo.

gdlcf88:
I think it is because I did not add this configuration:
```
            Configure<AbpVirtualFileSystemOptions>(options =>
            {
                options.FileSets.AddEmbedded<ModuleOverrideTestWebModule>("ModuleOverrideTest.Web");
            });
```

Hi @maliming, could you please tell me why the AbpVirtualFileSystemOptions is not configured by default in the startup template? thx.

maliming:
> why the AbpVirtualFileSystemOptions is not configured by default in the startup template?

The startup template does not use the [virtual file system](https://docs.abp.io/en/abp/latest/Virtual-File-System) by default.

gdlcf88:
Overriding a CSHTML file may not need VFS after this PR merged: https://github.com/abpframework/abp/pull/4016

PR #4113, opened 2020-05-27
Configure virtual files for app templates

liangshiw:
Resolve https://github.com/abpframework/abp/issues/4110


hikalkan:
Why we need to this, can you explain please?

liangshiw:
[Custom user interface documentation](https://docs.abp.io/en/abp/latest/UI/AspNetCore/Customization-User-Interface) introduce how to replace existing module pages, but some developers get some problems according to the documentation:  https://github.com/abpframework/abp/issues/3766 https://github.com/abpframework/abp/issues/3987.
We should pre-configure the template.

hikalkan:
Normally, physical files overrides embedded files. I suppose when we publish the application, physical files are not in the published path (since they are probably pre-compiled). So, this PR would be helpful for such a case. I am just curious about changing the default behaviour of AspNet Core.
I will think on that. Thanks.

hikalkan:
I support this problem will be solved by https://github.com/abpframework/abp/pull/4016

I am still working on...

hikalkan:
I am canceling this since we already canceled VFS usage for module razor files: https://github.com/abpframework/abp/pull/4016
So, no need to this, everything should work as expected.

Issue #4172, opened 2020-06-01
IdentityModelAuthenticationService should support multi-tenancy


liangshiw:
Currently unable to use the tenant for authentication, We should consider supporting multi-tenancy

https://github.com/abpframework/abp/blob/8c917e390ee0ebf61a5caef71246e290acee8d01/framework/src/Volo.Abp.IdentityModel/Volo/Abp/IdentityModel/IdentityModelAuthenticationService.cs#L122

PR #4173, opened 2020-06-01
IdentityModelAuthenticationService support multi-tenancy

liangshiw:
Resolve https://github.com/abpframework/abp/issues/4172

Maybe we should consider moving the `TenantKey` property to the `AbpMultiTenancyOptions` class.  @maliming  What do you think?

### DIFF
--- a/framework/src/Volo.Abp.IdentityModel/Volo/Abp/IdentityModel/IdentityModelAuthenticationService.cs
+++ b/framework/src/Volo.Abp.IdentityModel/Volo/Abp/IdentityModel/IdentityModelAuthenticationService.cs
@@ -126,6 +126,7 @@ namespace Volo.Abp.IdentityModel
             IdentityClientConfiguration configuration)
         {
             var httpClient = HttpClientFactory.CreateClient(HttpClientName);
+
             AddHeaders(httpClient);
 
             switch (configuration.GrantType)
--- a/framework/src/Volo.Abp.IdentityModel/Volo/Abp/IdentityModel/IdentityModelAuthenticationService.cs
+++ b/framework/src/Volo.Abp.IdentityModel/Volo/Abp/IdentityModel/IdentityModelAuthenticationService.cs
@@ -11,6 +11,7 @@ using System.Net.Http;
 using System.Net.Http.Headers;
 using System.Threading.Tasks;
 using Volo.Abp.DependencyInjection;
+using Volo.Abp.MultiTenancy;
 using Volo.Abp.Threading;
 
 namespace Volo.Abp.IdentityModel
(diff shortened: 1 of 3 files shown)


framework/src/Volo.Abp.IdentityModel/Volo/Abp/IdentityModel/IdentityModelAuthenticationService.cs
```diff
@@ -125,6 +129,8 @@ private IdentityClientConfiguration GetClientConfiguration(string identityClient
         {
             using (var httpClient = HttpClientFactory.CreateClient())
             {
+                AddHeaders(httpClient);
```
maliming, line 132: `httpClient `should be named so that we can configure it.  
After that we can reuse `httpClient` in `GetDiscoveryResponse `and `GetTokenResponse `methods

liangshiw: Yes, we can

framework/src/Volo.Abp.IdentityModel/Volo/Abp/IdentityModel/IdentityModelAuthenticationService.cs
```diff
@@ -106,40 +110,39 @@ private IdentityClientConfiguration GetClientConfiguration(string identityClient
...
+                }
+            });
         }
 
         protected virtual async Task<TokenResponse> GetTokenResponse(
             DiscoveryDocumentResponse discoveryResponse,
             IdentityClientConfiguration configuration)
         {
-            using (var httpClient = HttpClientFactory.CreateClient())
+            var httpClient = HttpClientFactory.CreateClient(HttpClientName);
```
hikalkan, line 128: Why not disposing the client (I didn't know if it is needed when you use HttpClientFactory.CreateClient)?
If you are re-using the same instance, should we care about multi-threading. In addition, should we check if tenantId added before and override it?
Can you explain a bit what this PR is trying to do.

liangshiw: There is no need to dispose of the HttpClient instances from HttpClientFactory. Disposal will not actually do anything in this case because the factory manages the handler and connection lifetimes and not the HttpClient instances. Calling dispose on HttpClient has no effect for instances provided by the factory and is simply redundant code.

Related https://support.abp.io/QA/Questions/171/How-to-call-abp-vnext-dynamic--api#answer-31b96b0e-ecd5-13ea-abca-39f56671b24d

PR #4200, opened 2020-06-03
Implement Blob Storing Azure Provider.

### LINKED ISSUE #4098: Blob Storing Azure Provider
Based on https://github.com/abpframework/abp/issues/4092

maliming:
Resolve #4098

To test AzureBlobProvider first set the correct connection string in `secrets.json`

```json
{
  "Azure": {
    "ConnectionString": "DefaultEndpointsProtocol=https;AccountName=xxxxx"
  }
}
```

After that, Remove code comments.

image

image


maliming:
Todo: Multi-tenant compatible

framework/src/Volo.Abp.BlobStoring.Azure/Volo/Abp/BlobStoring/Azure/AzureBlobProvider.cs
```diff
@@ -0,0 +1,100 @@
...
+    public class AzureBlobProvider : BlobProviderBase, ITransientDependency
+    {
+        protected IAzureBlobNameCalculator AzureBlobNameCalculator { get; }
+
+        public AzureBlobProvider(IAzureBlobNameCalculator azureBlobNameCalculator)
+        {
+            AzureBlobNameCalculator = azureBlobNameCalculator;
+        }
+
+        public override async Task SaveAsync(BlobProviderSaveArgs args)
```
hikalkan, line 18: You never use args.ContainerName. Should we use it if `AzureBlobProviderConfiguration.ContainerName` was not specified?

framework/src/Volo.Abp.BlobStoring.Azure/Volo/Abp/BlobStoring/Azure/AzureBlobProvider.cs
```diff
@@ -0,0 +1,100 @@
...
+            var configuration = args.Configuration.GetAzureConfiguration();
+
+            if (!args.OverrideExisting && await BlobExistsAsync(configuration, blobName))
+            {
+                throw new BlobAlreadyExistsException($"Saving BLOB '{args.BlobName}' does already exists in the container '{args.ContainerName}'! Set {nameof(args.OverrideExisting)} if it should be overwritten.");
+            }
+
+            if (configuration.CreateContainerIfNotExists)
+            {
+                await CreateContainerIfNotExists(args.Configuration.GetAzureConfiguration());
```
hikalkan, line 30: Can reuse the `configuration` created above.

framework/src/Volo.Abp.BlobStoring.Azure/Volo/Abp/BlobStoring/Azure/AzureBlobProvider.cs
maliming: Although their names are the same (`ContainerName`), there is still a difference. The container name of Azure should conform to a little rule.

- This name may only contain lowercase letters, numbers, and hyphens, and must begin with a letter or a number.
- Each hyphen must be preceded and followed by a non-hyphen character.
- The name must also be between 3 and 63 characters long.

I think the `ContainerName `of `AzureBlobProviderConfiguration `must be specified. If it does not exist in azure, create the container. (`CreateContainerIfNotExists`)

maliming: I have updated the code.

`Use ContainerName of the BlobProviderArgs if ContainerName of the AzureBlobProviderConfiguration is not specified.`

There are some rules for naming Azure container, which we need to explain in the documentation.

- This name may only contain lowercase letters, numbers, and hyphens, and must begin with a letter or a number.
- Each hyphen must be preceded and followed by a non-hyphen character.
- The name must also be between 3 and 63 characters long.

hikalkan: Thanks. The documentation nicely explains it. We will also take care to follow this rules for container names to not have a problem in case of developer uses azure but not specify the container name.

PR #4316, opened 2020-06-11
Make WebRequestResources store resources according to the current request Path.

### LINKED ISSUE #4309: Multiple bound script & stype is empty.
In a non-development environment.

```
@section scripts {
    <abp-script-bundle>
        <abp-script src="/Pages/Index.js" />
    </abp-script-bundle>

    <abp-script-bundle>
        <abp-script src="/Pages/Index.js" />
    </abp-script-bundle>
}
```

image

**The situation in a production environment is:**

An exception occurred in the `index page`, and then the application redirected to the `error page`.
If the `error page` uses the same binding `script or style` as an `Index page`, the above bug occurs.

maliming:
Resolve #4309


hikalkan:
Thanks for finding and fixing that.

PR #4320, opened 2020-06-11
Update spanish traslations, addedd keys for v 2.9

LuisPignataro:
I added the entries that were added to the en.json file in v2.9


maliming:
Thanks @LuisPignataro

PR #4346, opened 2020-06-14
Split `ICrudAppservice` and implementations

### LINKED ISSUE #4282: Introduce read-only app service
I am wondering if it might be nice to introduce a `IReadOnlyAppService` which is similar to `CrudAppService` but excludes:
* CreateAsync
* UpdateAsync
* DeleteAsync

The reason is that sometimes there are cases where we may only want to read via the API, but the updating happens somewhere else like another service.

In my case I have a 'Policy' which has sub-types like 'SimplePolicy' and 'ScheduledPolicy' which have their own 'update' and 'create' functionality, but share common set of 'Policy' properties and I want to be able to 'get' their high level properties.
Right now I will have to re-implement all the features from the `CrudAppService` and exclude these 3 endpoints, because the class is templated and I don't have the CRUD Dto's on the base class.

NecatiMeral:
Resolve #4282

Updated `AbstractKeyCrudAppService` to use the newly introduced `AbstractKeyReadOnlyAppService`.
Implemented `ReadOnlyAppService` to provide a easy point to implement own (readonly) services.

I have not provided default implementations for `ICreateAppService`, `IUpdateAppService` and `IDeleteAppService` because this would lead to a load of duplicated code. 
Maybe the `CrudAppService` implementation should act as a aggregating application service which delegates the methods to the actual services implementation, this would create a new layer of complexity and breaking changes. 
**This has to be discussed first.** 

The issues requirements are adressed.


hikalkan:
Thanks. I will review.

hikalkan:
Thanks @NecatiMeral for your contribution. I reviewed and it seems OK.

However, this is a critical change and I want to get a deeper review from @maliming and @liangshiw 
Guys, please review it deeply, test it and also consider namings... etc.
Thanks a lot.

framework/src/Volo.Abp.Ddd.Application.Contracts/Volo/Abp/Application/Services/ICreateAppService.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
```
liangshiw, line 1: I think these useless `using` can be removed.  Other interfaces and classes are also

image

framework/src/Volo.Abp.Ddd.Application.Contracts/Volo/Abp/Application/Services/ICreateUpdateAppService.cs
```diff
@@ -0,0 +1,26 @@
...
+    public interface ICreateUpdateAppService<TEntityDto, in TKey, in TCreateUpdateInput>
+        : ICreateUpdateAppService<TEntityDto, TKey, TCreateUpdateInput, TCreateUpdateInput>
+    {
+
+    }
+
+    public interface ICreateUpdateAppService<TGetOutputDto, in TKey, in TCreateUpdateInput, in TUpdateInput>
+        : ICreateAppService<TGetOutputDto, TCreateUpdateInput>,
+            IUpdateAppService<TGetOutputDto, TKey, TUpdateInput>,
+            IApplicationService
```
liangshiw, line 22: The `ICreateAppService` interface has implemented `IApplicationService`, which can be remove here. 

image

framework/src/Volo.Abp.Ddd.Application.Contracts/Volo/Abp/Application/Services/ICrudAppService.cs
```diff
@@ -28,16 +29,11 @@ public interface ICrudAppService<TEntityDto, in TKey, in TGetListInput, in TCrea
     }
 
     public interface ICrudAppService<TGetOutputDto, TGetListOutputDto, in TKey, in TGetListInput, in TCreateInput, in TUpdateInput>
-        : IApplicationService
+        : IReadOnlyAppService<TGetOutputDto, TGetListOutputDto, TKey, TGetListInput>,
+            ICreateUpdateAppService<TGetOutputDto, TKey, TCreateInput, TUpdateInput>,
+            IDeleteAppService<TKey>,
+            IApplicationService
```
liangshiw, line 35: `ICreateAppService` interface also can be remove `IApplicationService`.

framework/src/Volo.Abp.Ddd.Application.Contracts/Volo/Abp/Application/Services/IReadOnlyAppService.cs
```diff
@@ -0,0 +1,28 @@
...
+
+    }
+
+    public interface IReadOnlyAppService<TEntityDto, in TKey, in TGetListInput>
+        : IReadOnlyAppService<TEntityDto, TEntityDto, TKey, TGetListInput>
+    {
+
+    }
+
+    public interface IReadOnlyAppService<TGetOutputDto, TGetListOutputDto, in TKey, in TGetListInput>
```
liangshiw, line 22: Why the `IReadOnlyAppService` interface does not implement `IApplicationService`

framework/src/Volo.Abp.Ddd.Application/Volo/Abp/Application/Services/AbstractKeyReadOnlyAppService.cs
```diff
@@ -0,0 +1,185 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Linq.Dynamic.Core;
+using System.Text;
+using System.Threading.Tasks;
+
```
liangshiw, line 7: You can format code and remove useless `using`,  `AbstractKeyCrudAppService` and `ReadOnlyAppService` are also.

image

NecatiMeral: Thanks for your review. I've fixed the usings and unified the code format.

framework/src/Volo.Abp.Ddd.Application.Contracts/Volo/Abp/Application/Services/IReadOnlyAppService.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System.Threading.Tasks;
+
```
liangshiw, line 2: Just one thing. can trim this line : )

NecatiMeral: Resolved :-)

maliming:
hi @NecatiMeral 

I think we should use read-only repository in read-only application services. eg: `IReadOnlyRepository`. what do you think?

https://github.com/abpframework/abp/blob/42f37c5ff01ad853a5425d15539d4222cd0dab69/framework/src/Volo.Abp.Ddd.Domain/Volo/Abp/Domain/Repositories/IReadOnlyRepository.cs#L8

NecatiMeral:
hi @maliming,

thanks for your review. This makes totally sense. I'll check the possibilities and update this PR.

framework/src/Volo.Abp.Ddd.Application/Volo/Abp/Application/Services/AbstractKeyCrudAppService.cs
```diff
@@ -62,6 +62,8 @@ public abstract class AbstractKeyCrudAppService<TEntity, TGetOutputDto, TGetList
             ICrudAppService<TGetOutputDto, TGetListOutputDto, TKey, TGetListInput, TCreateInput, TUpdateInput>
         where TEntity : class, IEntity
     {
+        protected new IRepository<TEntity> Repository { get; }
```
NecatiMeral, line 65: How about this?
This saves us double property assignments.

```csharp
protected new IRepository<TEntity> Repository => (IRepository<TEntity>)base.Repository;
```

maliming: `protected new` doesn't seem to be a good way, I think we can use different names.

```cs
 protected IReadOnlyRepository<TEntity> ReadOnlyRepository { get; }

 protected IRepository<TEntity> Repository { get; }

```

NecatiMeral: Okay, I'll update the property assignments.

gdlcf88:
Hi @NecatiMeral, any plan to implement `CreateAppService` and `CreateUpdateAppService`?

Issue #4062, opened 2020-05-23
Try to consume both of REST and gRPC services from the console client of the gRPC demo sample


hikalkan:
I created the sample: https://github.com/abpframework/abp-samples/tree/master/GrpcDemo

It works. However, we have a problem: [Microsoft document](https://docs.microsoft.com/en-us/aspnet/core/tutorials/grpc/grpc-start) leads to make the server to support HTTP/2. However, client application can not call REST endpoints in this case. I couldn't be able to configure all HttpClients (those used to authenticate to IDS4 and call REST endpoints) to use Http/2. There are [solutions](https://stackoverflow.com/questions/32685151/how-to-make-the-net-httpclient-use-http-2-0) on the web, but I didn't spend much time to be honest.

I disabled the related lines: https://github.com/abpframework/abp-samples/blob/master/GrpcDemo/test/GrpcDemo.HttpApi.Client.ConsoleTestApp/ConsoleTestAppHostedService.cs#L28

@maliming can you look at that. Thanks.

maliming:
Set the `DefaultRequestVersion `property of `HttpClient `to `HttpVersion.Version20` to solve. 

Wait for abp v 2.9 https://github.com/abpframework/abp/pull/4138

maliming:
Wait for https://github.com/abpframework/abp-samples/pull/9

maliming:
The `DefaultRequestVersion` property doesn't apply to the `HttpClient.SendAsync` method. The `Http.HttpRequestMessage` parameter passed as the argument to the `HttpClient.SendAsync` method has its own `Version` property that controls the HTTP version used for the request.

https://github.com/dotnet/runtime/issues/31190

Components using `SendAsync` need to be changed to be compatible with `http2`.
Including methods such as IdentityModel's `RequestClientCredentialsTokenAsync`.
However, `DefaultRequestVersion` does not support `netstandard2.0`. This problem also needs to be resolved.

maliming:
Wait: https://github.com/abpframework/abp/issues/3907

PR #4353, opened 2020-06-15
Configure IdentityModel request message & configure all http clients.

maliming:
Related #4062

This allow configure all the dynamic http client and configure the request message of the IdentityModel extension method. It can make httpclient support `http 2`.

example:
```cs
Configure<AbpHttpClientOptions>(options =>
{
	options.HttpClientActions.Add(clientName =>
	{
		if (clientName == "MyClient")
		{
			return c => c.DefaultRequestVersion = HttpVersion.Version10;
		}

		return c => c.DefaultRequestVersion = HttpVersion.Version20;
	});
}

Configure<IdentityModelHttpRequestMessageOptions>(options =>
{
	options.ConfigureHttpRequestMessage = request => request.Version = HttpVersion.Version20;
});
```


hikalkan:
Thanks for the solution.

PR #4371, opened 2020-06-17
Introducing UnitOfWorkCache.

### LINKED ISSUE #4040: SettingManagementStore not refreshing cache within the same UnitOfWork

maliming:
Resolve #4040

**UnitOfWorkCache** can cache some values in the current unit of work. Keep the cached values consistent. 

`UnitOfWorkCache` is an abstract class, you need to implement it in your application and specify the name of the cache:

```cs
public class MyUnitOfWorkCache : UnitOfWorkCache<MyCacheItem>, ITransientDependency
{
	public MyUnitOfWorkCache(IUnitOfWorkManager unitOfWorkManager)
		: base(unitOfWorkManager)
	{
	}

	public override string Name => "MyUnitOfWorkCacheName";
}
```

Then you can use `MyUnitOfWorkCache `to cache the value in the current unit of work.

`IUnitOfWorkCacheWithFallback` supports finding uncached values in other caches etc.

```cs
public class MyUnitOfWorkCacheWithFallback : UnitOfWorkCacheWithFallback<MyCacheItem>, ITransientDependency
{
	public MyUnitOfWorkCacheWithFallback(IUnitOfWorkManager unitOfWorkManager)
		: base(unitOfWorkManager)
	{
	}

	public override string Name => "MyUnitOfWorkCacheWithFallback";

	public override Task<MyCacheItem> GetFallbackCacheItem(string key)
	{
		//Get the uncached value of the unit of work.
	}

	public override Task<MyCacheItem> SetFallbackCacheItem(string key, MyCacheItem item)
	{
		 //Set the generated cache value(GetOrAddAsync).
	}
}

```


maliming:
Maybe it can change its name. For example: `UnitOfWorkStore`, @hikalkan  what do you think?

PR #4374, opened 2020-06-17
Resource Localization for base and account

ebubekirdinc:
Resource localization for base and account files.


yekalkan:
@ebubekirdinc Thank you.

PR #4390, opened 2020-06-18
Remove CustomSchemaIds in swagger options.

### LINKED ISSUE #4206: NSwag and others cannot generate client due to backticks
Not sure if this is an ABP issue or something wrong with NSwag.  The PagedResultDto generic causes backticks to be added by the swagger scheme generators as you can see frmo this file.

image

We tried NSwag studio but the resulting typescript file is broken due to the backtick.  Im unsure if the other stuff like Culture, PublicKeyToken etc should all be appearing in there.

We tried using the ABP proxy generator but we need clean typescript and it seems to have multiple dependencies to angular router and others.

maliming:
Resolve #4206

By default, we remove the `CustomSchemaIds `method.

https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/752#issuecomment-467817189


hikalkan:
We had added this to not have conflicts when we have same controller names in different modules, because swagger gives error in this case. Doesn't it give error in this case anymore?

maliming:
> Doesn't it give error in this case anymore?

It seems that there is no error, if necessary, developers can customize the `CustomSchemaIds `method. 

You can try maybe there is something wrong with my test code.

hikalkan:
What if we use [GetFullNameHandlingNullableAndGenerics](https://github.com/abpframework/abp/blob/dev/framework/src/Volo.Abp.Core/Volo/Abp/Reflection/TypeHelper.cs) instead of Type.FullName? Can you try if it solves the issue.

I don't want to remove this since we are creating a modular system and it is probable two different modules use the same controller name in different namespaces (I remember it even somewhere in our own modules).

Issue #4430, opened 2020-06-19
Unit Test error


liangshiw:
![image](https://user-images.githubusercontent.com/16813853/85162322-c13a9500-b293-11ea-9b84-52307976fbdf.png)

liangshiw:
https://github.com/abpframework/abp/commit/3321bf4e4b91162cfe9b6fc633b6e99ec36cdcda

PR #4431, opened 2020-06-19
Fix unit test error

liangshiw:
Resolve https://github.com/abpframework/abp/issues/4430


liangshiw:
https://github.com/abpframework/abp/commit/3321bf4e4b91162cfe9b6fc633b6e99ec36cdcda

PR #4446, opened 2020-06-22
Remove unnecessary property.

### LINKED ISSUE #4438: UserLoginInfo's Property typo
https://github.com/abpframework/abp/blob/0ea1e59d3be1241433434c73c81bff293f5aaeb5/modules/account/src/Volo.Abp.Account.Web/Areas/Account/Controllers/Models/UserLoginInfo.cs#L21

maliming:
Resolve #4438

I checked the source code and did not use the `TenanId ` property.

https://github.com/abpframework/abp/commit/56798a74dc1dd43512bd7414ac46b5f7a44f7153#diff-af613efddf8524d26fba26eca0168827R1-R21


ebicoglu:
yes it seems like it's not used. also it must have been `TenantId` not `TenanId`

PR #4460, opened 2020-06-22
Store cache in the unit of work.

### LINKED ISSUE #4040: SettingManagementStore not refreshing cache within the same UnitOfWork

maliming:
Resolve #4040

Use ite in the setting and feature modules.


maliming:
There have been many changes to the DistributedCache recently, and this PR needs to be refactored.

maliming:
Also need to be compatible with https://github.com/abpframework/abp/pull/4500

maliming:
After merging the commit of the **dev** branch in the feature branch, Github shows all the commits in pull request. :confused:

PR #4471, opened 2020-06-23
Add NamingNormalizerProviders to BlobContainerConfiguration.

### LINKED ISSUE #4408: Normalize blob and container names in the providers
- [ ] File system provider can normalize (clear illegal chars from) file/container names before saving or getting the file.
- [ ] Azure provider can normalize the container name (you know, it only allows lower case chars... etc.) before using it.

@maliming what do you think?

maliming:
Resolve #4408

And extract the `IOSPlatformProvider `interface in `Volo.Abp.Core` to get the current OSPlatform.

`DefaultAzureBlobNamingNormalizerProvider` and `DefaultFileSystemBlobNamingNormalizerProvider` are implemented by default

1. According to the operating system, replace some illegal characters.
2. Replace the container name according to the container naming rules of azure blob.

Example
```cs
{
	options.Containers
		.Configure<TestContainer1>(container =>
		{

			container.ProviderType = typeof(MyBlobProvider1);
			container.NamingNormalizerProviders.Add<MyNamingNormalizerProvider>();
		});
});


public class MyNamingNormalizerProvider : IBlobNamingNormalizerProvider, ITransientDependency
{
	public virtual string NormalizeContainerName(string containerName)
	{
		return containerName;
	}

	public virtual string NormalizeBlobName(string blobName)
	{
		return blobName;
	}
}
```

### DIFF
--- a/framework/src/Volo.Abp.BlobStoring.Azure/Volo/Abp/BlobStoring/Azure/AzureBlobContainerConfigurationExtensions.cs
+++ b/framework/src/Volo.Abp.BlobStoring.Azure/Volo/Abp/BlobStoring/Azure/AzureBlobContainerConfigurationExtensions.cs
@@ -15,6 +15,7 @@ namespace Volo.Abp.BlobStoring.Azure
             Action<AzureBlobProviderConfiguration> azureConfigureAction)
         {
             containerConfiguration.ProviderType = typeof(AzureBlobProvider);
+            containerConfiguration.NamingNormalizerProviders.Add<DefaultAzureBlobNamingNormalizerProvider>();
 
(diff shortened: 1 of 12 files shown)


hikalkan:
@maliming I am writing for https://github.com/abpframework/abp/pull/4471/files#diff-a9808b9caa387704a1a5c4db9e5baa5dR31

What if we start with a Windows server, then later switch to a Linux server, copy all files to the new server? We, then can not access to the files anymore.

An idea: Without looking the the OS, we may remove all illegal chars (based on all restrictions in linux, win and mac)

What do you think about it?

BTW, I made some refactoring, you check them.

Issue #94, opened 2017-09-09
Object2Object mapping enhancements


hikalkan:
- [ ] If a custom map defined for T->K then we should also can map ``ICollection<T>`` -> ``ICollection<K>``.

genoher:
Would it be possible to integrate this fix in the next release 5.3 or 6.0 or 7.0, please. 
I understand that if it hasn't been integrated it will be because it breaks something, or for some good reason that I don't understand, maybe because the tests are missing?

If not possible, what is the best way to do it? implement an IObjectMapper<TSource, TDestination> Interface
where TSource is ICollection<source> and TDdestination is ICollection<destination>?

Thank you very much.

https://stackoverflow.com/questions/70107428/abp-does-not-automatically-use-custom-mapper-class

hikalkan:
@maliming can you check if we can implement this?

https://github.com/abpframework/abp/blob/b270e1aa6cee199fe7c98526496027e276d30e45/framework/src/Volo.Abp.ObjectMapping/Volo/Abp/ObjectMapping/DefaultObjectMapper.cs#L42-L49

Here, if we don't find a `specificMapper`, and if `TSource` implements `ICollection<TS>` and `TDestination` implements `ICollection<TD>` , then we can try to resolve `IObjectMapper<TS, TD>`. If it is registered to DI, we can use it to map items one by one and create the target collection manually. Can you check if that possible and efficient?
If you can implement this, please add tests and update the document as well.

PR #4498, opened 2020-06-25
#94 resolved

sercantimocin:
I have to use reflection instead of generic because I couldn't  find a way to get strongly type of custom mapper object from DI


maliming:
hi @sercantimocin  Github action failed. https://github.com/abpframework/abp/pull/4498/checks

hikalkan:
@maliming I am not sure this fail is related to this PR. Can you check and try to understand please?

maliming:
@sercantimocin 
We have to handle or ignore calls in this way.
```cs
 ObjectMapper.Map<List<Entity>, EntityDto[]>(entitys)
```

hikalkan:
I am closing this. You re-implement it when you start to work with the core ABP Framework team :)

PR #4500, opened 2020-06-26
Implement the ICacheSupportsMultipleItems for the AbpRedisCache

liangshiw:
Resolve https://github.com/abpframework/abp/issues/4489

Since the reflection Invoke method will reduce performance, I copied the `RedisCache` class source code to the project.
I used pipeline for batch operations https://stackexchange.github.io/StackExchange.Redis/PipelinesMultiplexers.


hikalkan:
Thanks @liangshiw for the PR. However;

1. We don't want to copy all the source code, because then we should maintain it ourselves. Reflection cost is not so much (I think 1000 times executing reflection code gets less time than single execution duration of the total method call). Maintainablity is more important than this very small performance lost. However, I simply eleminated the reflection by this simple change: https://github.com/abpframework/abp/commit/58cb055a994bdcf2dd29d09487eb9f04b3d8e83f (please merge this to your branch and remove the copied code).

2. What about using the MSET command (https://redis.io/commands/mset) instead of pipeline? I suppose this should be much more performant since it only calls the server one time. As similar, MGET (https://redis.io/commands/mget) can be used to get multiple values at once.

liangshiw:
@hikalkan 

OK , I agree that maintainability is important. In real application scenarios, there will be some performance-sensitive requirements (especially regarding caching), they can customize the cache implementation for higher performance.

We cannot use the `mget` command because it is only used for `string` type. Distributed cache uses a `hash table` type, and redis does not provide an API to get multiple hash tables : https://redis.io/commands#hash, I compared multiple implementations, `pipeline` performance is the best (Usually only one packet is sent in the pipeline).

liangshiw:
Performance test results:

1000 keys :

Set Execution time: 1195 ms
SetMany Execution time: 10 ms

SetAsync Execution time: 1224 ms
SetManyAsync Execution time: 14 ms

Get Execution time: 2316 ms
GetMany Execution time: 11 ms

GetAsync Execution time: 2406 ms
GetManyAsync Execution time: 12 ms

10000 keys :

Set Execution time: 11978 ms
SetMany Execution time: 86 ms

SetAsync Execution time: 11913 ms
SetManyAsync Execution time: 106 ms

Get Execution time: 23609 ms
GetMany Execution time: 103 ms

GetAsync Execution time: 23782 ms
GetManyAsync Execution time: 114 ms

hikalkan:
Performance result seems great. Thanks.

PR #4521, opened 2020-06-29
Exclude primitive type in CreateType method.

### LINKED ISSUE #4486: System. Int32 in Angular when primary key is int
When we use int as primary key  in dotnetcore then dto generated in angular is 

 **AuditedEntityDto<System.Int32>**
 
 _System.Int32_  which is not valid in angular so please replace it with number.
And after correcting it manually it works but when we develop more APIs in dotnetcore and then again run the proxy command in angular it will again do System.Int32 in the old model as well.

**Steps:**

Create Dto

```
 public class ABCDto : AuditedEntityDto<int>
 {

        public string Description { get; set; }
  }

then use this dto in service

and then run proxy in angular you will get this

export class ABCDto extends AuditedEntityDto<System.Int32>
{
      name: string;

      description: string;
}
```

maliming:
Resolve #4486


hikalkan:
Thanks.

PR #4523, opened 2020-06-29
Store cache in the unit of work.

### LINKED ISSUE #4040: SettingManagementStore not refreshing cache within the same UnitOfWork

maliming:
Resolve #4040

Resolve https://github.com/abpframework/abp/issues/4796

Use `considerUow ` in the **setting** and **feature** modules.

Due to an unexpected mess in https://github.com/abpframework/abp/pull/4460#issuecomment-651004379 I reopened the new PR.


framework/src/Volo.Abp.Caching/Volo/Abp/Caching/DistributedCache.cs
```diff
@@ -50,6 +53,8 @@ public class DistributedCache<TCacheItem> : DistributedCache<TCacheItem, string>
     public class DistributedCache<TCacheItem, TCacheKey> : IDistributedCache<TCacheItem, TCacheKey>
         where TCacheItem : class
     {
+        public const string DistributedCacheName = "AbpDistributedCache";
```
hikalkan, line 56:
```suggestion
        public const string UowCacheName = "AbpDistributedCache";
```

framework/src/Volo.Abp.Caching/Volo/Abp/Caching/DistributedCache.cs
```diff
@@ -133,14 +142,25 @@ protected virtual void SetDefaultOptions()
         /// Gets a cache item with the given key. If no cache item is found for the given key then returns null.
         /// </summary>
         /// <param name="key">The key of cached item to be retrieved from the cache.</param>
+        /// <param name="considerUow">This will store the cache in the current unit of work until the end of the current unit of work does not really affect the cache.</param>
         /// <param name="hideErrors">Indicates to throw or hide the exceptions for the distributed cache.</param>
         /// <returns>The cache item, or null.</returns>
         public virtual TCacheItem Get(
             TCacheKey key,
+            bool considerUow = false,
```
hikalkan, line 150: It would be better to add an optional parameter to the end of the parameter list. In this case, types (bool) is same and if someone set the `hideErrors` before, they will be set the `considerUow` when they upgrade the ABP Framework, which will be an implicit breaking change.

Same for other methods.

framework/src/Volo.Abp.Caching/Volo/Abp/Caching/DistributedCache.cs
```diff
@@ -172,33 +193,57 @@ protected virtual void SetDefaultOptions()
             {
                 return GetManyFallback(
                     keyArray,
+                    considerUow,
                     hideErrors
                 );
             }
 
+            var cachedValues = new List<KeyValuePair<TCacheKey, TCacheItem>>();
```
hikalkan, line 201: Move into `if (ShouldConsiderUow(considerUow))` block

framework/src/Volo.Abp.Caching/Volo/Abp/Caching/DistributedCache.cs
```diff
@@ -172,33 +193,57 @@ protected virtual void SetDefaultOptions()
...
+            var notCachedKeys = new List<TCacheKey>();
+            if (ShouldConsiderUow(considerUow))
+            {
+                var cache = GetUnitOfWorkCache();
```
hikalkan, line 205: Rename to `uowCache` to be more explicity

framework/src/Volo.Abp.Caching/Volo/Abp/Caching/DistributedCache.cs
```diff
@@ -172,33 +193,57 @@ protected virtual void SetDefaultOptions()
...
+            }
+
             hideErrors = hideErrors ?? _distributedCacheOption.HideErrors;
             byte[][] cachedBytes;
 
+            var readKeys = notCachedKeys.Any() ? notCachedKeys.ToArray() : keyArray;
             try
             {
-                cachedBytes = cacheSupportsMultipleItems.GetMany(keyArray.Select(NormalizeKey));
+                cachedBytes = cacheSupportsMultipleItems.GetMany(readKeys.Select(NormalizeKey));
```
hikalkan, line 228: I didn't understand that. It seems the method is returning only the notCachedKeys instead of all keys in this case. Shouldn't it return the values of all the keys?

framework/src/Volo.Abp.Caching/Volo/Abp/Caching/DistributedCache.cs
```diff
@@ -786,12 +1089,33 @@ protected virtual TCacheItem ToCacheItem([CanBeNull] byte[] bytes)
...
+        protected virtual string GetUnitOfWorkCacheKey()
+        {
+            return DistributedCacheName + CacheName;
+        }
+
+        protected virtual Dictionary<TCacheKey, UnitOfWorkCacheItem<TCacheItem>> GetUnitOfWorkCache()
+        {
+            if (UnitOfWorkManager.Current == null)
+            {
+                throw new AbpException($"There is no unit of work in the current context, The {GetType().Name} can only be used in a unit of work.");
```
hikalkan, line 1114: Actually, DistributedCache.cs can be used without UOW. This error message it wrong.
If this method is called out of UOW, ABP Framework has a bug and the error message should indicate that. At least, "There is no active UOW" would be better.

hikalkan:
Can you also please add a few more test scenarios.

framework/src/Volo.Abp.Caching/Volo/Abp/Caching/DistributedCache.cs
maliming: The `cachedValues `  variable will be used later.

framework/src/Volo.Abp.Caching/Volo/Abp/Caching/DistributedCache.cs
maliming: Yes, I revert the changes, it now returns `keyArray`

maliming: The logic here is:
First read the keys in the uow cache, of course, some keys may not be in the uow cache, then read the remaining keys from the cache, and finally merge the results.

```cs
return cachedValues.Concat(ToCacheItems(cachedBytes, readKeys)).ToArray();
```

PR #4549, opened 2020-06-30
Allowed auditing of selected props of an entity

iyilm4z:
Resolveds #5417

### DIFF
--- a/framework/src/Volo.Abp.EntityFrameworkCore/Volo/Abp/EntityFrameworkCore/EntityHistory/EntityHistoryHelper.cs
+++ b/framework/src/Volo.Abp.EntityFrameworkCore/Volo/Abp/EntityFrameworkCore/EntityHistory/EntityHistoryHelper.cs
@@ -246,6 +246,16 @@ namespace Volo.Abp.EntityFrameworkCore.EntityHistory
             {
                 return false;
             }
+            
+            if (!entityType.IsDefined(typeof(AuditedAttribute), true)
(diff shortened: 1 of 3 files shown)


hikalkan:
I found that if condition too complicated and may lead us potential bugs in the feature. I think we should implement it in a different way. So, I will move this to 3.1 to re-think on.
Thanks for the contribution :)

iyilm4z:
That's okay. Actually I could follow the same design of aspnetboilerplate but i found that if statement would be the minimal change i could do:)  If you have any other recommendation, please let me know.

hikalkan:
@maliming [the documentation](https://docs.abp.io/en/abp/latest/Audit-Logging#enable-disable-for-entities-properties) says:

image

So, it seems it is already possible to do that? Can you test & proove it please? If so, you can close this PR.

maliming:
**We have unit tests to prove this feature.**

https://github.com/abpframework/abp/blob/eb5ac8d2b645c70b66f79022920bc8669f0cc098/framework/test/Volo.Abp.Auditing.Tests/Volo/Abp/Auditing/Auditing_Tests.cs#L160-L175

https://github.com/abpframework/abp/blob/eb5ac8d2b645c70b66f79022920bc8669f0cc098/framework/test/Volo.Abp.Auditing.Tests/Volo/Abp/Auditing/App/Entities/AppEntityWithDisableAuditingAndPropertyHasAudited.cs#L5-L30

PR #4555, opened 2020-07-01
Switch to the tenant before calling the handleevent method.

### LINKED ISSUE #4508: how to determine current tenant in the distributedEventBus
1. IDistributedEventHandler is different from mvc, it does not go through middleware, such as `.UseMiddleware<MultiTenancyMiddleware>`. So the DbContext uses the current host ,
, Such as BlogUserSynchronizer in the Volo.Blogging module, it can only handle the host,
If you need to handle tenant,do we need to add `_currentTenant.Change(**)`,like this
```C#
  public async Task HandleEventAsync(EntityUpdatedEto<UserEto> eventData)
        {
            using (_currentTenant.Change(eventData.Entity.TenantId))
            {
                var user = await UserRepository.FindAsync(eventData.Entity.Id);
                if (user == null)
                {
                    user = await UserLookupService.FindByIdAsync(eventData.Entity.Id);
                    if (user == null)
                    {
                        return;
                    }
                }
                if (user.Update(eventData.Entity))
                {
                    await UserRepository.UpdateAsync(user);
                }
            }          
        }
```
2. Volo.Abp.Domain.Entities.Events.Distributed.EntityEto, it does not include tenantId, do we noed to add tenantId?

maliming:
Resolve #4508

An event data object (or event transfer object) can implement `IEventDataMayHaveTenantId `interface to indicate that this event may be related to a tenant.

If an event data class is always related to a tenant, then directly implement the `IsMultiTenant `interface instead of `IEventDataMayHaveTenantId`.

The `IEventDataMayHaveTenantId `interface is typically implemented by generic event handlers where the generic parameter may implement `IsMultiTenant `or not.

```cs
protected virtual Guid? GetEventDataTenantId(object eventData)
{
	return eventData switch
	{
		IMultiTenant multiTenantEventData => multiTenantEventData.TenantId,
		IEventDataMayHaveTenantId eventDataMayHaveTenantId when eventDataMayHaveTenantId.IsMultiTenant(out var tenantId) => tenantId,
		_ => CurrentTenant.Id
	};
}
```


hikalkan:
I introduced the new `IEventDataMayHaveTenantId` interface to solve the reflection problem: https://github.com/abpframework/abp/commit/beadedc0c69cc060b577797d7087b1c4b2f8cdcf

Get it from dev and update the code please.

maliming:
@hikalkan Done  https://github.com/abpframework/abp/pull/4555#issue-442425885

PR #4572, opened 2020-07-02
Throw an AbpAuthorizationException when ThrowExceptionForResponseAsync in DynamicHttpProxyInterceptor

xyfy:
Throw an AbpAuthorizationException when ThrowExceptionForResponseAsync in DynamicHttpProxyInterceptor

when  the remote  send 401 or 403

so user can do some bussiness logic such as Relogin  or Alert


xyfy:
image
when I changed it ,I got the correct httpstatuscode

hikalkan:
see my comment please: https://github.com/abpframework/abp/issues/4590#issuecomment-653932463

PR #4583, opened 2020-07-02
Provide a way to suppress the label generation for `abp-input` taghelper

wakuflair:
When using the `abp-input` taghelper, sometimes developers may not want to generate the label tag.

So this PR checks if there is a special attribute `nolabel` exists, if so, suppress the label tag generation:

``` html
<abp-input asp-for="Content" nolabel></abp-input>
```


hikalkan:
Good, but we should make it an explicit type-safe attribute.

wakuflair:
Hi @hikalkan 
I updated this PR, now the `AbpInputTagHelper` has one more property `SuppressLabel` instead of original `no-label` attribute.
I also added demo for it.


image
image
image

Issue #4582, opened 2020-07-02
AbpDbContext.OnModelCreating issue


zhk0603:
abp version 3.0.0
ef core 3.1.5

My entity：
```
public class WorkflowInstanceEntity : FullAuditedAggregateRoot<int>, IMultiTenant
{
    //other properties
    public ICollection<LogEntry> ExecutionLog { get ; set ; }
    //other properties
}

```

LogEntry :
```
    classe publique LogEntry
    {
        chaîne publique ActivityId { get ; set ; }
        public Instant Timestamp { get ; set ; } // using NodaTime , key point!!
        public bool Faulted { get ; set ; }
        public string Message { get ; set ; }
    }
```

DbContext:
```
    public class ElsaContext : AbpDbContext<ElsaContext>
    {
        public ElsaContext(DbContextOptions<ElsaContext> options) : base(options)
        {
        }

        public DbSet<WorkflowInstanceEntity> WorkflowInstances { get; set; }

        protected override void OnModelCreating(ModelBuilder builder)
        {
            base.OnModelCreating(builder); // key point!!

            builder.ConfigureElsa();
        }
    }
```

ElsaDbContextModelCreatingExtensions.cs
```
builder.Entity<WorkflowInstanceEntity>(b =>
{
    b.Property(x => x.ExecutionLog)
                    .HasConversion(x => serialized operation,
                        x => deserialization operation);
}
```

up to this point, everything works fine, "Update-database" is executed, it is generated correctly, and the ExecutionLog field is also of type string.

however, at runtime, ef core throws an exception.

```
System.InvalidOperationException: The property 'LogEntry.Timestamp' could not be mapped, because it is of type 'Instant' which is not a supported primitive type or a valid entity type. Either explicitly map this property, or ignore it using the '[NotMapped]' attribute or by using 'EntityTypeBuilder.Ignore' in 'OnModelCreating'.
   at Microsoft.EntityFrameworkCore.Infrastructure.ModelValidator.ValidatePropertyMapping(IModel model, IDiagnosticsLogger`1 logger)
   at Microsoft.EntityFrameworkCore.Infrastructure.ModelValidator.Validate(IModel model, IDiagnosticsLogger`1 logger)
   at Microsoft.EntityFrameworkCore.Infrastructure.RelationalModelValidator.Validate(IModel model, IDiagnosticsLogger`1 logger)
[...]

PR #4595, opened 2020-07-02
Ignore classes that are not entities while configuring the base properties

liangshiw:
Resolve https://github.com/abpframework/abp/issues/4582


hikalkan:
Thanks ;)

Issue #4596, opened 2020-07-02
Nuget common.ps1 files missing Volo.Abp.BlobStoring.Azure


liangshiw:

maliming:
https://github.com/abpframework/abp/commit/484ef2e583ca3e0797535dbe47e451822a7db7da

PR #4597, opened 2020-07-02
Added Volo.Abp.BlobStoring.Azure to common.ps1

liangshiw:
Resolve https://github.com/abpframework/abp/issues/4596


maliming:
https://github.com/abpframework/abp/commit/484ef2e583ca3e0797535dbe47e451822a7db7da

PR #4633, opened 2020-07-06
Rename the text "MY_PROJECT_NAME" when generating a new project

### LINKED ISSUE #4592: ABP CLI should edit the text "MY_PROJECT_NAME" (snake case & upper case) when generating a new project
CLI can't convert `MY_PROJECT_NAME` (snake & upper case) while a new project generating. 
We are using this constant in Angular UI module template

image

maliming:
Resolve #4592

`ToSnakeCase` can also be` ToKebabCase.Replace("-", "_")` 

@hikalkan What do you think?


hikalkan:
>ToSnakeCase can also be ToKebabCase.Replace("-", "_")

I don't think it is necessary. That's fine. That's not a business code that may change later and code duplication is not important - but the performance is important.

PR #4634, opened 2020-07-06
Try to set GUID Id in the InsertAsync method of the EF Core repository.

### LINKED ISSUE #4631: Try to set GUID Id in the InsertAsync method of the EF Core repository
In this way, we don't need to wait until SaveChangesAsync if we use GUID Id but not set it before InsertAsync call.
It is currently like that for the MongoDbRepository.

Note: Apply the same logic inside the AbpDbContext.CheckAndSetId

maliming:
Resolve #4631

### DIFF
--- a/framework/test/Volo.Abp.MongoDB.Tests/Volo/Abp/MongoDB/Repositories/Repository_Basic_Tests.cs
+++ b/framework/test/Volo.Abp.MongoDB.Tests/Volo/Abp/MongoDB/Repositories/Repository_Basic_Tests.cs
@@ -64,5 +64,15 @@ namespace Volo.Abp.MongoDB.Repositories
             person.Phones.Count.ShouldBe(1);
             person.Phones.Any(p => p.PersonId == person.Id && p.Number == "1234567890").ShouldBeTrue();
         }
+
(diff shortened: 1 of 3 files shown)


framework/test/Volo.Abp.MongoDB.Tests/Volo/Abp/MongoDB/Repositories/Repository_Basic_Tests.cs
```diff
@@ -64,5 +64,15 @@ public override async Task InsertAsync()
             person.Phones.Count.ShouldBe(1);
             person.Phones.Any(p => p.PersonId == person.Id && p.Number == "1234567890").ShouldBeTrue();
         }
+
+        [Fact]
+        public async Task Insert_Should_Set_Guid_Id()
```
hikalkan, line 69: Should't this be tested for EF Core (instead of MongoDB)? Or can we move it to the base `Repository_Basic_Tests<T>`?

hikalkan:
I merged to dev, but please implement my comment: https://github.com/abpframework/abp/pull/4634#discussion_r456366756

PR #4636, opened 2020-07-06
Add dataFormat option to the datatables column definitions.

### LINKED ISSUE #4629: Add dataFormat option to the datatables column definitions
If I want to render a datetime, currently I should write such a column definition:

````js
{
    title: l('CreationTime'), data: "creationTime",
    render: function (data) {
        return luxon
            .DateTime
            .fromISO(data, {
                locale: abp.localization.currentCulture.name
            }).toLocaleString(luxon.DateTime.DATETIME_SHORT);
    }
}
````

It can be that simple:

````js
{
    title: l('CreationTime'), data: "creationTime",
    dataFormat: `datetime`
}
````

It can match the `dataFormat` to the pre-defined `datatables.defaultRenderers`:

https://github.com/abpframework/abp/blob/9f2a89a5a8a96db3656140a4ae1db0ef92cb812b/framework/src/Volo.Abp.AspNetCore.Mvc.UI.Theme.Shared/wwwroot/libs/abp/aspnetcore-mvc-ui-theme-shared/datatables/datatables-extensions.js#L319-L345

We can evaluate the `dataFormat` at the `datatables.normalizeConfiguration`:

https://github.com/abpframework/abp/blob/9f2a89a5a8a96db3656140a4ae1db0ef92cb812b/framework/src/Volo.Abp.AspNetCore.Mvc.UI.Theme.Shared/wwwroot/libs/abp/aspnetcore-mvc-ui-theme-shared/datatables/datatables-extensions.js#L292-L317

### PR DESCRIPTION
**Pull request by maliming**
Resolve #4629

```cs
{
	data: "userName"
},
{
	data: "email"
},
{
	data: "creationTime",
	dataFormat: "datetime"
},
{
	data: "creationTime",
	dataFormat: "date"
}
```
image
image

### DIFF
--- a/framework/src/Volo.Abp.AspNetCore.Mvc.UI.Theme.Shared/wwwroot/libs/abp/aspnetcore-mvc-ui-theme-shared/datatables/datatables-extensions.js
+++ b/framework/src/Volo.Abp.AspNetCore.Mvc.UI.Theme.Shared/wwwroot/libs/abp/aspnetcore-mvc-ui-theme-shared/datatables/datatables-extensions.js
@@ -299,9 +299,11 @@
                     column.targets = i;
                 }
 
-                if (!column.render && column.dataFormat){
+                if (!column.render && column.dataFormat) {
                     var render = datatables.defaultRenderers[column.dataFormat];
-                    column.render = render ? render : ISOStringToDateTimeLocaleString(column.dataFormat);
+                    if (render) {
+                        column.render = render;
+                    }
                 }
 
                 if (column.rowAction) {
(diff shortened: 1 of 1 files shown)


framework/src/Volo.Abp.AspNetCore.Mvc.UI.Theme.Shared/wwwroot/libs/abp/aspnetcore-mvc-ui-theme-shared/datatables/datatables-extensions.js
```diff
@@ -299,6 +299,11 @@
                     column.targets = i;
                 }
 
+                if (!column.render && column.dataFormat){
+                    var render = datatables.defaultRenderers[column.dataFormat];
+                    column.render = render ? render : ISOStringToDateTimeLocaleString(column.dataFormat);
```
hikalkan, line 304: Do nothing if there is no renderer registered with the given name. We don't assume that the data is always date/datetime. `datatables.defaultRenderers` can have any type of renderer, just like the `datatables.defaultRenderers['boolean']` registered before. 

If we also want to accept luxon-specific format arguments, we can introduce another option like `dataFormatArgs` (optional) which can be an object and passed to the renderer function. But the `dataFormat` always should be simple string that is matched an item in the datatables.defaultRenderers.

maliming:
@hikalkan Changed.

PR #4674, opened 2020-07-08
Add permissions authorization for consuming framework tenant API endpoints

SoloHam:
These API endpoints are available freely to access, without any permission management, and it's not documented how to hide these default endpoints or override these and add authentication.


maliming:
hi

CI checks have failed.

PR #4675, opened 2020-07-08
Introduce security log.

maliming:
How to write security logs.

You can inject and use `ISecurityLogManager `to write security logs. It will create a log object by default and fill in some common values, such as CreationTime, ClientIpAddress, BrowserInfo, current user/tenant, etc. Of course, you can override them.
```cs
await SecurityLogManager.SaveAsync(securityLog =>
{
	securityLog.Identity = "IdentityServer";
	securityLog.Action = "ResourceOwnerPassword";
});
```

You can publish an event to write to the security log, (**Identity module**).
The handler in Identity will try to get the user's information, which is useful for unauthenticated users.
https://github.com/abpframework/abp/pull/4675/files#diff-8b6aaf4be3cd3e62a966b7efa7f0d3a5
```cs
await LocalEventBus.PublishAsync(new IdentitySecurityLogEvent
{
	Identity = "IdentityServer",
	Action = "ResourceOwnerPassword"
});
```

The security log store is implemented in **Identity module**.

Configure `AbpSecurityLogOptions `to provide the application name for the log or disable this feature. **enabed** by default.

```cs
Configure<AbpSecurityLogOptions>(x =>
{
	x.ApplicationName = "AbpSecurityTest";
});
```


modules/identity/src/Volo.Abp.Identity.Domain/Volo/Abp/Identity/IdentitySecurityLogHandler.cs
```diff
@@ -0,0 +1,86 @@
...
+            {
+                securityLog.Identity = eventData.Identity;
+                securityLog.Action = eventData.Action;
+
+                if (securityLog.UserName.IsNullOrWhiteSpace())
+                {
+                    securityLog.UserName = eventData.UserName;
+                }
+
+                if (securityLog.ClientId.IsNullOrWhiteSpace())
```
hikalkan, line 47: Shouldn't be vice verse? I mean a filled eventData.ClientId should override the securityLog.ClientId. What do you think?
I didn't think much, but it seems like that. If that's true, no problem.

modules/identity/src/Volo.Abp.Identity.Domain/Volo/Abp/Identity/IdentitySecurityLogStore.cs
```diff
@@ -0,0 +1,49 @@
...
+        }
+
+        public async Task SaveAsync(SecurityLogInfo securityLogInfo)
+        {
+            if (!SecurityLogOptions.IsEnabled)
+            {
+                return;
+            }
+
+            using (var uow = UnitOfWorkManager.Begin(requiresNew: true))
```
hikalkan, line 42: This can also be a "non transactional" UOW since it contains only one insert. It will reduce the dedlock probability.

hikalkan:
I accepted the PR. Added two notes, so you can check and fix theme if necessary.

But the essentional design question is that: Why we are using event bus to save a security log. We should be able to directly use the ISecurityLogManager. Using evenbus is not a good way if we can use direct method calls.
If you want to decouple something, you can simply create an interface and implement in the Identity module.

Please re-consider this design and drop using the event bus.

modules/identity/src/Volo.Abp.Identity.Domain/Volo/Abp/Identity/IdentitySecurityLogStore.cs
maliming: `Begin(requiresNew: true)` is extension method. https://github.com/abpframework/abp/blob/ae8095d1eac23cad4d7f652012578157a5ba5a86/framework/src/Volo.Abp.Uow/Volo/Abp/Uow/UnitOfWorkManagerExtensions.cs#L10-L16

maliming:
Open a new PR for the above comment. https://github.com/abpframework/abp/pull/4810

Issue #4318, opened 2020-06-11
Configure Prettier for the solutions


hikalkan:
We want to configure the [Prettier](https://prettier.io/) for, especially for JavaScripts, for all solutions (including startup templates).

It is available to [VS](https://marketplace.visualstudio.com/items?itemName=MadsKristensen.JavaScriptPrettier), [VS Code](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode) and [Rider](https://www.jetbrains.com/help/rider/Prettier.html) as I searched.

PR #4706, opened 2020-07-09
Javascript files reformatted with Prettier and Prettier settings created for each module

cotur:
Resolves: https://github.com/abpframework/abp/issues/4318

## General Prettier Settings

We use follow settings for abp modules. This `.prettierrc` file is located for all published modules.

```json
{
    "singleQuote": true,
    "useTabs": false,
    "tabWidth": 4
}
```

## How to use prettier?

You need to setup the prettier for your IDE.

### VS Code

Search "`Prettier - Code formatter`" at the extensions and download it.

> You can set `true` Format on save settings at the `File > Preferences > Settings : search "Format on Save"`

### Visual Studio

Open `Extensions > Manage Extensions` then type "`JavaScript Prettier`" and install it.

> After downloading, restart your IDE.

> You can set `true` Format on save settings at the `Tools > Options > Prettier > General : Format On Save`

### Rider

You can follow [this documentation](https://www.jetbrains.com/help/rider/Prettier.html) for more information.

Also check [this documentation](https://prettier.io/docs/en/webstorm.html#running-prettier-on-save-using-file-watcher) for `Format on save`.


maliming:
It looks very good. 👍  Can you configure it for template projects?

PR #4707, opened 2020-07-09
Update Local-Event-Bus.md

gdlcf88:


liangshiw:
Thanks a lot @gdlcf88

Issue #4711, opened 2020-07-10
Pre defined events not being emitted for entity types


feanz:
I am trying to use the DistributedEventBus and the pre defined events for entity changes are not happening.  Any custom event I make and manually emit works with no issue.  

I have this configuration for distributed entity events

```
Configure<AbpDistributedEntityEventOptions>(options =>
{  
    options.AutoEventSelectors.Add<Client>();
    options.EtoMappings.Add<Client, ClientEto>();
});
````

I have an ETO:

```
 [AutoMap(typeof(Client)]
    public class ClientEto : EntityEto
    {
        public string Name { get; set; }
    }
```

I also have a handler 

```
public class ClientUpdatedHandler: 
        IDistributedEventHandler<EntityCreatedEto<ClientEto>>,
        ITransientDependency
    {
        public async Task HandleEventAsync(EntityCreatedEto<ClientEto> eventData)
        {

        }
    }
```

This never gets fired.  However if I take a dependency on IDistributedEventBus and manually Publish a custom event those events are picked up by handlers and messages are published by via RabbitMQ.  Is there some way to debug this to find out why the messages would not be being emitted?

liangshiw:
I will check it out.

liangshiw:
This is a bug.

We should use `eventEntry.EventBus` instead of `eventPublisher`.

![image](https://user-images.githubusercontent.com/16813853/87138039-6f07f500-c2d0-11ea-8aef-5f6935692482.png)

liangshiw:
@feanz 
You can override `IEntityChangeEventHelper` to temporarily fix this bug.

```csharp
[Dependency(ReplaceServices = true)]
[ExposeServices(typeof(IEntityChangeEventHelper))]
public class MyEntityChangeEventHelper : EntityChangeEventHelper
{
    public MyEntityChangeEventHelper(IUnitOfWorkManager unitOfWorkManager, IEntityToEtoMapper entityToEtoMapper, IOptions<AbpDistributedEntityEventOptions> distributedEntityEventOptions) : base(unitOfWorkManager, entityToEtoMapper, distributedEntityEventOptions)
    {
    }

    private class EntityChangeEventEntry
    {
        public IEventBus EventBus { get; }

        public Type EventType { get; }

        public object EntityOrEto { get; }

        public object OriginalEntity { get; }

        public EntityChangeEventEntry(IEventBus eventBus, Type eventType, object entityOrEto, object originalEntity)
        {
            EventType = eventType;
[...]

PR #4715, opened 2020-07-10
Use eventEntry.EventBus instead of eventPublisher

liangshiw:
Resolve https://github.com/abpframework/abp/issues/4711


hikalkan:
Can you do this also on master branch please. We may release a patch version and it would be better to include this to the patch.

liangshiw:
OK

PR #4727, opened 2020-07-13
Avoid NullReferenceException if context is null

1d0n7kn0w:


maliming:
hi @1d0n7kn0w

1d0n7kn0w:
Hi @maliming 
I created a new solution, then I deleted all the things related to multitenancy
So in the class "src/{name}.Domain/Data/{name}DbMigrationService.cs" and specifically in the private method "SeedDataAsync()" I made this change from "await _dataSeeder.SeedAsync(tenant?.Id);" to "await _dataSeeder.SeedAsync(null);"
At first, I didn't know why the DbMigrator was throwing a NullReferenceException, after investigation with null "Volo.Abp.Data.IDataSeeder.SeedAsync(DataSeedContext context)" is used, so to avoid this I had to use the parameterless one so it uses "Volo.Abp.Data.DataSeederExtensions.SeedAsync(this IDataSeeder seeder, Guid? tenantId = null)"

maliming:
hi

You can use this extension method instead of passing **null**.

https://github.com/abpframework/abp/blob/42f37c5ff01ad853a5425d15539d4222cd0dab69/framework/src/Volo.Abp.Data/Volo/Abp/Data/DataSeederExtensions.cs#L8

1d0n7kn0w:
yes that's what I did as I mentioned in my comment, but to simplify things for other developers, passing null or nothing should have the same result I guess

maliming:
Ok, can you also handle **IdentityDataSeedContributor**?

https://github.com/abpframework/abp/blob/ace4b5526eb00d0ed6429834693fc618dc6c54b2/modules/identity/src/Volo.Abp.Identity.Domain/Volo/Abp/Identity/IdentityDataSeedContributor.cs#L16

image

1d0n7kn0w:
Done, with pleasure 😄

modules/identity/src/Volo.Abp.Identity.Domain/Volo/Abp/Identity/IdentityDataSeedContributor.cs
```diff
@@ -18,7 +18,7 @@ public virtual Task SeedAsync(DataSeedContext context)
             return IdentityDataSeeder.SeedAsync(
                 context["AdminEmail"] as string ?? "admin@abp.io",
                 context["AdminPassword"] as string ?? "1q2w3E*",
-                context.TenantId
+                context?.TenantId
```
maliming, line 21:
```cs
var adminUserName = context?["AdminEmail"] as string ?? "admin@abp.io";
var adminEmail = context?["AdminPassword"] as string ?? "1q2w3E*";
return IdentityDataSeeder.SeedAsync(adminUserName, adminEmail, context?.TenantId);
```

1d0n7kn0w: 😞 my bad I don't know how I unsaw this

PR #4730, opened 2020-07-13
Fix dynamic layout not updating

### LINKED ISSUE #4729: Replacement of layout in Angular not working
**ABP Framework version**: 3.0.3
**User Interface**:  Angular

I've followed the documentation [here](https://docs.abp.io/en/abp/latest/UI/Angular/Component-Replacement#how-to-replace-a-layout) to replace `ApplicationLayoutComponent`, but nothing changes when run. The issue happens even in a newly create ABP project. Replacement of other components like logo, etc. works fine, though.

I've did some digging into the source code of `DynamicLayoutComponent` and found something of interest.

https://github.com/abpframework/abp/blob/a70d21bc5bc73a9a52986b942d2b2d020ec4e346/npm/ng-packs/packages/core/src/lib/components/dynamic-layout.component.ts#L39-L65

https://github.com/abpframework/abp/blob/a70d21bc5bc73a9a52986b942d2b2d020ec4e346/npm/ng-packs/packages/core/src/lib/components/dynamic-layout.component.ts#L79-L81

What I understand from this code is that layout components are only fetched once when the application starts, any subsequent replacements of layout are not picked up.

haivp3010:
Fix #4729

### DIFF
--- a/npm/ng-packs/packages/core/src/lib/components/dynamic-layout.component.ts
+++ b/npm/ng-packs/packages/core/src/lib/components/dynamic-layout.component.ts
@@ -10,6 +10,7 @@ import { ReplaceableComponentsState } from '../states/replaceable-components.sta
 import { findRoute, getRoutePath } from '../utils/route-utils';
 import { takeUntilDestroy } from '../utils/rxjs-utils';
 import { TreeNode } from '../utils/tree-utils';
+import { distinctUntilChanged } from 'rxjs/operators';
 
 @Component({
   selector: 'abp-dynamic-layout',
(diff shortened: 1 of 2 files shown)


armanozak:
Hi @haivp3010,

Thank you very much for the PR. We really appreciate your contribution. However, there are two major problems with the proposed changes:

- The layout is getting replaced at runtime, but the content of the layout disappears when that happens (if you change the active layout). This is due to the nature of `DynamicLayoutComponent`. It includes a router outlet and when a the layout is swapped, the content is gone together with the old layout.
- The `AddReplaceableComponent` action did not cause an immediate layout change before, but this PR would cause it to do so. This could cause a breaking change for some people, so, if possible, we should find a non-breaking way to do so.

This is why we are not able to accept the changes proposed in this PR.

Please do not let this change your mind about contribution, because we are always looking forward to receiving your support and ideas. We just cannot merge the proposed changes for this individual case.

I wish you a very nice day.

PR #4738, opened 2020-07-13
Check Replaceable Layout Component on Every Route Change

### LINKED ISSUE #4729: Replacement of layout in Angular not working

armanozak:
Fixes #4729

### DIFF
--- a/npm/ng-packs/packages/core/src/lib/components/dynamic-layout.component.ts
+++ b/npm/ng-packs/packages/core/src/lib/components/dynamic-layout.component.ts
@@ -24,6 +24,13 @@ import { TreeNode } from '../utils/tree-utils';
 export class DynamicLayoutComponent implements OnDestroy {
   layout: Type<any>;
 
+  // TODO: Consider a shared enum (eThemeSharedComponents) for known layouts
+  readonly layouts = new Map([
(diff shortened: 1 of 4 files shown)


armanozak:
Closed due to broken GitHub actions.

PR #4742, opened 2020-07-13
Cache the AccessToken in IdentityModelAuthenticationService.

### LINKED ISSUE #4603: Is it necessary to save access_Token to cache?
Authorization call between internal services will request IDS server to get token each time. Is it necessary to cache client or user's access_token according to grant_type and ExpiresIn after the first request is successful to reduce the pressure on IDS server for each call? Of course, the token needs to be encrypted again

This is often the case when a domain event requires the invocation of a remote service, and events are generally stateless

https://github.com/abpframework/abp/blob/cf39e470a524ae60a1f56209d1f084f4c70732e2/framework/src/Volo.Abp.IdentityModel/Volo/Abp/IdentityModel/IdentityModelAuthenticationService.cs#L71-L94

maliming:
Resolve #4603


framework/src/Volo.Abp.IdentityModel/Volo/Abp/IdentityModel/IdentityModelAuthenticationService.cs
```diff
@@ -76,21 +81,36 @@ public virtual async Task<string> GetAccessTokenAsync(IdentityClientConfiguratio
                 throw new AbpException($"Could not retrieve the OpenId Connect discovery document! ErrorType: {discoveryResponse.ErrorType}. Error: {discoveryResponse.Error}");
             }
 
-            var tokenResponse = await GetTokenResponse(discoveryResponse, configuration);
-
-            if (tokenResponse.IsError)
+            var cacheKey = CalculateCacheKey(discoveryResponse, configuration);
```
hikalkan: Why not including the `GetDiscoveryResponse` call to the cached part? Should we always call it?

framework/src/Volo.Abp.IdentityModel/Volo/Abp/IdentityModel/IdentityModelTokenCacheItem.cs
```diff
@@ -0,0 +1,29 @@
...
+        }
+
+        public IdentityModelTokenCacheItem(string accessToken)
+        {
+            AccessToken = accessToken;
+        }
+
+        public static string CalculateCacheKey(DiscoveryDocumentResponse discoveryResponse, IdentityClientConfiguration configuration)
+        {
+            return discoveryResponse.TokenEndpoint + string.Join(",", configuration.Select(x => x.Key + ":" + x.Value));
```
hikalkan: Instead of such a long key (which may have problems with cache providers), can we get MD5 (or use a better algorithm) hash of the string. I am just thiking and not sure about it. What do you think?

framework/src/Volo.Abp.IdentityModel/Volo/Abp/IdentityModel/IdentityModelAuthenticationService.cs
```diff
@@ -76,21 +81,36 @@ public virtual async Task<string> GetAccessTokenAsync(IdentityClientConfiguratio
...
                 }
 
-                var rawError = tokenResponse.Raw;
-                var withoutInnerException = rawError.Split(new string[] { "<eof/>" }, StringSplitOptions.RemoveEmptyEntries);
-                throw new AbpException(withoutInnerException[0]);
+                await Cache.SetAsync(cacheKey, new IdentityModelTokenCacheItem(tokenResponse.AccessToken),
+                    new DistributedCacheEntryOptions()
+                    {
+                        //Subtract 10 seconds of network request time.
+                        AbsoluteExpirationRelativeToNow = TimeSpan.FromSeconds(tokenResponse.ExpiresIn - 10)
```
hikalkan: Should we remove the token from cache if we get an authentication/authorization related error from the server? If, somehow, server side invalidates the token, we continously get errors.

In addition, I think it is better to make the cache duration smaller, like 30 minutes for example (configurable), to force to re-authenticate the client. It would be better to refresh the token since client claims might be changed in this period.

framework/src/Volo.Abp.IdentityModel/Volo/Abp/IdentityModel/IdentityModelTokenCacheItem.cs
maliming: I have considered md5, but the name of the key will be meaningless.

framework/src/Volo.Abp.IdentityModel/Volo/Abp/IdentityModel/IdentityModelAuthenticationService.cs
maliming: Sorry I forgot it, I will cache it.
But can we use `configuration.Authority + /connect/token` directly?  Because this url is seem fixed(not sure), of course, developers can also rewrite the method to call the api.

framework/src/Volo.Abp.IdentityModel/Volo/Abp/IdentityModel/IdentityModelTokenCacheItem.cs
maliming: If the name of the key does not need to have any meaning, **md5** is no problem.

framework/src/Volo.Abp.IdentityModel/Volo/Abp/IdentityModel/IdentityModelAuthenticationService.cs
maliming: I added `CacheAbsoluteExpiration ` property to `IdentityClientConfiguration`. You can continue to review.

framework/src/Volo.Abp.IdentityModel/Volo/Abp/IdentityModel/IdentityClientConfiguration.cs
```diff
@@ -81,21 +82,32 @@ public bool RequireHttps
             get => this.GetOrDefault(nameof(RequireHttps))?.To<bool>() ?? true;
             set => this[nameof(RequireHttps)] = value.ToString().ToLowerInvariant();
         }
-        
+
+        /// <summary>
+        /// Cache absolute expiration
+        /// Default: 30 minutes.
```
hikalkan:
```suggestion
        /// Absolute expiration duration (as seconds) for the access token cache.
        /// Default: 1800 seconds (30 minutes)
```

framework/src/Volo.Abp.IdentityModel/Volo/Abp/IdentityModel/IdentityClientConfiguration.cs
```diff
@@ -81,21 +82,32 @@ public bool RequireHttps
...
+        /// </summary>
+        public double CacheAbsoluteExpiration
+        {
+            get => this.GetOrDefault(nameof(CacheAbsoluteExpiration ))?.To<double>() ?? 60 * 30;
```
hikalkan: Use integer instead of double.

framework/src/Volo.Abp.IdentityModel/Volo/Abp/IdentityModel/IdentityClientConfiguration.cs
```diff
@@ -81,21 +82,32 @@ public bool RequireHttps
...
-            string clientId, 
-            string clientSecret, 
+            string clientId,
+            string clientSecret,
             string grantType = OidcConstants.GrantTypes.ClientCredentials,
             string userName = null,
             string userPassword = null,
-            bool requireHttps = true)
+            bool requireHttps = true,
+            double cacheAbsoluteExpiration = 60 * 30)
```
hikalkan: use integer instead of double

PR #4744, opened 2020-07-13
Update Connection-Strings.md

buaziz:


maliming:
hi @buaziz 

I think there is no problem with the Connection-Strings.md.

https://github.com/abpframework/abp/blob/ee70f4bce1a3c8a6833d1eeb3518ba3753f867a7/docs/en/Connection-Strings.md#L21-L25

PR #4746, opened 2020-07-13
Remove the use of UseHangfireServer in the Hangfire background job doc

gdlcf88:
Fix #4746


liangshiw:
Thanks a lot @gdlcf88

Issue #4743, opened 2020-07-13
New Module: CMS Kit / Prepare the initial solution


hikalkan:
## Introduction

CMS Kit (CmsKit) is a new module idea that contains CMS (Content Management System) components. It is not a CMS application, but actually a CMS framework or component set.

Some of the planned components are;

* Reactions (thumbs up, smile... etc.)
* Rating
* Comments
* Tags
* Favorites

These components are entity independent and provides reusable widgets. For example, you can add a "comment" or "tag" functionality to an image gallery or a blog post.

**This is currently experimental and you are free to comment on this issue for your ideas.**

This issue is to create the initial solution structure for the development.

### Switch to Local Framework References

* We should change project references to the local framework, instead of NuGet package references.

### Change the Solution Structure

- host
  * (keep current projects)
- core
  * Volo.CmsKit.Domain.Shared
  * Volo.CmsKit.Domain
  * Volo.CmsKit.EntityFrameworkCore
  * Volo.CmsKit.MongoDB
- public
  * Volo.CmsKit.Public.Application.Contracts
  * Volo.CmsKit.Public.Application
  * Volo.CmsKit.Public.HttpApi
  * Volo.CmsKit.Public.HttpApi.Client
  * Volo.CmsKit.Public.Web
- admin
  * Volo.CmsKit.Admin.Application.Contracts
  * Volo.CmsKit.Admin.Application
  * Volo.CmsKit.Admin.HttpApi
  * Volo.CmsKit.Admin.HttpApi.Client
  * Volo.CmsKit.Admin.Web
- unified
  * Volo.CmsKit.Application.Contracts
  * Volo.CmsKit.Application
  * Volo.CmsKit.HttpApi
  * Volo.CmsKit.HttpApi.Client
  * Volo.CmsKit.Web

#### Additional Info

* Unified packages are shortcut to add admin+public. They don't have their own functionality (so, they are meta packages).

### Apply Module Standards

* Root namespaces should be empty. So, re-arrange folders & files.
* Configure Fody.
* Configure common.props
* ...

wakuflair:
I'm excited to see this.

I think the key of a CMS is to be able to handle dynamic content flexibly. Users should easily define a content (blog, questionnaire, webpage, etc.) without hardcode.

So is there anything you can share about dynamic content management?

gdlcf88:
> It is not a CMS application, but actually a CMS framework or component set.

[...]

PR #4750, opened 2020-07-14
CMSKit initial solution is prepared

cotur:
Closes: https://github.com/abpframework/abp/issues/4743


modules/cms-kit/src/Volo.CmsKit.Admin.Application.Contracts/Volo/CmsKit/Admin/Permissions/AdminPermissions.cs
```diff
@@ -0,0 +1,14 @@
+﻿using Volo.Abp.Reflection;
+
+namespace Volo.CmsKit.Admin.Permissions
+{
+    public class AdminPermissions
+    {
+        public const string GroupName = "Admin";
```
liangshiw, line 7: I think this is not a good name. `Admin` and `Public` is too common.has no special meaning.

`CmsKit.Admin` is better.

gdlcf88: If it is a CMS framework or component set like hikalkan's introduction, it may provide abstract entity class, derived repository base class, derived AppService class, UI components, and more.

If so, maybe `AdminPermissions`, `AdminMenu`, `RemoteServiceName` and their classes should not be defined, since it is not an application module and is used to create some other modules or apps.

cotur: Right, 

updated by https://github.com/abpframework/abp/commit/6d402e47186bf70028dc60854d902cf90de4a31e

cotur: This is only preparing initial solution. This will get a lot of update soon. :)

liangshiw: @cotur 
`AdminMenu`, `RemoteServiceName` should also be changed.

Issue #4721, opened 2020-07-10
Object Disposed Error in IDistributedEventHandler that take a dependency on IRepository


feanz:
I have IDistributedEventHandler  handlers that take a dependency on IRepository.  This breaks though as you get ObjectDisposed errors on the DB context.  

I have faced similar issues in other solutions that were not based on ABP using entity framework in both asp.net and in event handlers.  It could be that the DBContext has been disposed by the request context but is resolved by the handler.  I had to solve this with different DBContext registrations in the DI container.  I'm not sure how this would work in ABP?

Do you have working examples of EventHandlers using repositories?  I'm happy to update the documentation if users need to handle data access code in event handlers in a different way.

maliming:
Please shared your handler code.

feanz:
Sure

```
 public class ClientSearchIndexUpdatingEventHandler :
        IDistributedEventHandler<EntityCreatedEto<ClientEto>>,
        ITransientDependency
    {
        private readonly IRepository<Client> _repository;

        public ClientSearchIndexUpdatingEventHandler(
            IRepository<Client> repository)
        {
            _repository = repository;
        }

        public async Task HandleEventAsync(EntityCreatedEto<ClientEto> eventData)
        {
            var currentClient = await _repository.Where(x => x.Id == eventData.Entity.Id)
                .SingleOrDefaultAsync();

        }
}
}
```

The error you get when the handler code is fired is 

```
System.ObjectDisposedException: Cannot access a disposed object. A common cause of this error is disposing a context that was resolved from dependency injection and then later trying to use the same context instance elsewhere in your application. This may occur if you are calling Dispose() on the context, or wrapping the context in a using statement. If you are using dependency injection, you should let the dependency injection container take care of disposing context instances.
Object name: 'MyDbContext'.
   at Microsoft.EntityFrameworkCore.DbContext.CheckDisposed()
   at Microsoft.EntityFrameworkCore.DbContext.get_InternalServiceProvider()
   at Microsoft.EntityFrameworkCore.DbContext.get_ChangeTracker()
   at Microsoft.EntityFrameworkCore.Query.CompiledQueryCacheKeyGenerator.GenerateCacheKeyCore(Expression 
```

maliming:
See https://stackoverflow.com/questions/62777358/objectdisposedexception-when-trying-to-use-the-dbcontext-inside-an-idistributede

feanz:
[...]

PR #4768, opened 2020-07-15
Add details on the correct use of unit of work in event handlers

feanz:
If you try to use Repository methods in EventHandlers you can get `ObjectDisposedException` this adds details on how to correctly use UnitOfWorkManger to avoid this issue. 

Raise in this issue https://github.com/abpframework/abp/issues/4721

### DIFF
--- a/docs/en/Distributed-Event-Bus.md
+++ b/docs/en/Distributed-Event-Bus.md
@@ -178,7 +178,10 @@ That's all.
 
 You can inject any service and perform any required logic here. A single event handler class can **subscribe to multiple events** but implementing the `IDistributedEventHandler<TEvent>` interface for each event type.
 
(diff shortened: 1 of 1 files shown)


maliming:
hi

I think it is more appropriate to explain this in the documentation of the unit of work or repository.  What do you think?

eg:
https://aspnetboilerplate.com/Pages/Documents/Unit-Of-Work#irepository-getall-method

feanz:
Hi

I would say that if people are going to see the issue they will see it in
the handler and will go to the documentation for that first so it's more
helpful to be there.

I think it's a good idea to establish a policy. Do you flag issues where
they will happen in this case the handler or in the place they are solved
in this case the unit of work manager. I would pick one and stick with it
throughout.

hikalkan:
I slightly changed the explanation: https://github.com/abpframework/abp/commit/541324dbbe8021eecd2dd0bf6506e45f8a1ab1d4
Thanks for noticing it.

PR #4774, opened 2020-07-15
Do not create server if the Hangfire is not configured

### LINKED ISSUE #4773: Using Hangfire in Domain project makes the DbMigrator project not runnable.
The `AbpHangfireModule.OnApplicationInitialization()` method invokes `AbpHangfireOptions.CreateJobServer()` method, it requires that the `JobStorage` be in IOC, but in DbMigrator project, we don't add Hangfire, so `JobStorage` is never in the IOC, it raises the problem.

gdlcf88:
Fix #4773


liangshiw:
https://github.com/abpframework/abp/issues/4773#issuecomment-659426192

PR #4776, opened 2020-07-16
Update Account module zh-Hans.json

xyfy:
Update Account module zh-Hans.json

for the loggedout page


xyfy:
为啥中文要用英文标点符号~~ @maliming

xyfy:
Why use English punctuation instead of Chinese  @maliming

Issue #4678, opened 2020-07-08
abp generate-proxy dto baseType error


snowchenlei:
**problem:**
1. create StoreListDto.cs
``` csharp
// StoreListDto.cs
namespace A.B.Dtos
{
    public class StoreListDto : FullAuditedEntityDto<Guid>,
    {
        public string Name { get; set; }
    }
}
```
2. create StoreAppService.cs
``` csharp
// StoreAppService.cs
namespace A.B
{
    public Task<PagedResultDto<StoreListDto>> GetListAsync()
    {
      throw new NotImplementedException();
    }
}
```
3. using `abp generate-proxy` will get the following results
``` ts
// StoreListDto.ts
import { ListResultDto} from '@abp/ng.core';
export class StoreListDto extends ListResultDto<A.B.Dtos.StoreListDto> {
  totalCount: number;
  items: any[];
  constructor(initialValues: Partial<StoreListDto> = {}) {
    super(initialValues);
  }
}
```
**The temporary solution is as follows:**
change StoreAppService.cs
``` csharp
// StoreAppService.cs
namespace A.B
{
    // This method must be in front of GetListAsync.
    public Task<StoreListDto> GetAsync(Guid id)
    {
      throw new NotImplementedException();
    }
    public Task<PagedResultDto<StoreListDto>> GetListAsync()
    {
      throw new NotImplementedException();
    }
}
```
In this way, we will get the right results.
``` ts
// StoreListDto.ts
import { FullAuditedEntityDto } from '@abp/ng.core';

export class StoreListDto extends FullAuditedEntityDto<string> {
  name: string;

  constructor(initialValues: Partial<StoreListDto> = {}) {
    super(initialValues);
  }
}
```

Please fix this problem, thank you.

hikalkan:
Closing this since we are replacing the implementation: https://github.com/abpframework/abp/issues/4881

PR #4802, opened 2020-07-18
Ignore the generic dto predefined by abp in the return type

realLiangshiwei:
Resolve https://github.com/abpframework/abp/issues/4678


hikalkan:
I am closing this since we'are retiring this implementation and re-implementing it.

Issue #4610, opened 2020-07-03
Blob storing AWS S3 Integration


johnnygiter:
Please add s3 object storage support upload, delete etc.

liangshiw:
Hi, I added it to the backlog,  If you need this feature now, you can integrate aws s3 according to the [documentation](https://docs.abp.io/en/abp/latest/Blob-Storing-Custom-Provider
)

maharatha:
+1

liangshiw:
I will try to implement.

PR #4826, opened 2020-07-21
Add Blob storing AWS S3 Integration

liangshiw:
Resolve https://github.com/abpframework/abp/issues/4610


maliming:
Please add it to the `common.ps1`

https://github.com/abpframework/abp/blob/dev/nupkg/common.ps1#L62

framework/src/Volo.Abp.BlobStoring.Aws/Volo/Abp/BlobStoring/Aws/AwsTemporaryCredentialsCacheItem.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System;
+
+namespace Volo.Abp.BlobStoring.Aws
+{
+    [Serializable]
+    public class AwsTemporaryCredentialsCacheItem
+    {
+        public const string Key = "AwsBlobTemporaryCredentialsCache";
```
maliming, line 8: Use calculated key instead of fixed one.

hikalkan:
That's great. Did we document this?

realLiangshiwei:
Yes, https://github.com/abpframework/abp/pull/4842

PR #4842, opened 2020-07-22
Add blob storing AWS Sintegration document

liangshiw:
Resolve https://github.com/abpframework/abp/issues/4836


docs/en/docs-nav.json
```diff
@@ -224,6 +224,10 @@
                   "text": "Minio Provider",
                   "path": "Blob-Storing-Minio.md"
                 },
+                {
+                  "text": "Minio Provider",
```
maliming, line 228: It's should be AWS.

PR #4869, opened 2020-07-24
Map the BookStore code namespace.

maliming:


hikalkan:
@maliming can you add this code mapping to the startup template?

PR #4872, opened 2020-07-24
Delete related entities when deleting Identity Server Client.

### LINKED ISSUE #4870: Delete related entities when deleting Identity Server Client.
https://github.com/abpframework/abp/blob/dev/modules/identityserver/src/Volo.Abp.IdentityServer.EntityFrameworkCore/Volo/Abp/IdentityServer/Clients/ClientRepository.cs#L14

maliming:
Fix #4870


hikalkan:
@yekalkan if that's OK, should we merge it?

PR #4882, opened 2020-07-26
Introduce AbpSqliteUnitOfWorkDbContextProvider to disable sqlite transaction.

maliming:
It's useful for unit test.

```cs
{
	Configure<AbpEntityFrameworkCoreSqliteOptions>(options =>
	{
		options.DisableTransaction = true;
	});
}
```

If this can accept we should update all projects to disable the SQLite transaction.


hikalkan:
Can't we just use the `AbpUnitOfWorkDefaultOptions.TransactionBehavior` https://github.com/abpframework/abp/blob/dev/framework/src/Volo.Abp.Uow/Volo/Abp/Uow/AbpUnitOfWorkDefaultOptions.cs#L16

maliming:
I forgot about this.

PR #4893, opened 2020-07-27
Use EntityTypeBuilder & PropertyBuilder in the MapEfCoreProperty method.

### LINKED ISSUE #4891: Use EntityTypeBuilder and PropertyBuilder in the MapEfCoreProperty method.
It is currently impossible to set **index** for the property.

maliming:
Resolve #4891

This will bring a **breaking change.**  

@hikalkan Do you have any suggested?


hikalkan:
@maliming instead of the breaking change, please create an overload of this method that gets an action with two arguments.

hikalkan:
I merged. Can you add [Obsolete] note to the method with action with single parameter? We may remove it in the future.

PR #4906, opened 2020-07-28
Fix NRE on EntityHistoryHelper.ShouldSavePropertyHistory

skomis-mm:
ShouldSavePropertyHistory fails on shadow property while checking if it is base audit property

### DIFF
--- a/framework/test/Volo.Abp.Auditing.Tests/Volo/Abp/Auditing/App/EntityFrameworkCore/AbpAuditingTestDbContext.cs
+++ b/framework/test/Volo.Abp.Auditing.Tests/Volo/Abp/Auditing/App/EntityFrameworkCore/AbpAuditingTestDbContext.cs
@@ -27,5 +27,15 @@ namespace Volo.Abp.Auditing.App.EntityFrameworkCore
         {
 
         }
+
+        protected override void OnModelCreating(ModelBuilder modelBuilder)
+        {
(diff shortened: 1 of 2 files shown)


framework/test/Volo.Abp.Auditing.Tests/Volo/Abp/Auditing/App/EntityFrameworkCore/AbpAuditingTestDbContext.cs
```diff
@@ -27,5 +27,15 @@ public AbpAuditingTestDbContext(DbContextOptions<AbpAuditingTestDbContext> optio
...
         }
+
+        protected override void OnModelCreating(ModelBuilder modelBuilder)
+        {
+            base.OnModelCreating(modelBuilder);
+
+            modelBuilder.Entity<AppFullAuditedEntityWithAudited>(entity =>
+            {
+                entity.Property<string>("Shadow");
+            });
```
maliming, line 38: Maybe we can remove those lines, what do you think?

skomis-mm: Ok, removed

maliming:
Thanks @skomis-mm .

PR #4910, opened 2020-07-29
Implement StringValueTypeJsonConverter.

### LINKED ISSUE #4337: FeatureManagement.FeatureDto.IStringValueType  Cause Newtonsoft.Json.JsonSerializationException
` public interface IFeatureAppService : IApplicationService
    {
        Task<FeatureListDto> GetAsync([NotNull] string providerName, [NotNull] string providerKey); 
    }`
 the GetAsync method  Return  FeatureListDto that contains IStringValueType. 
In a monolithic application,It work well. But In microservices，IFeatureAppService  as IRemoteService, and the returned Json format result,The Client cannot be serialized,it cause Newtonsoft.Json.JsonSerializationException: Could not create an instance of type Volo.Abp.Validation.StringValues.IStringValueType. Type is an interface or abstract class and cannot be instantiated.

How to deal with it？

maliming:
Resolve #4337

### DIFF
--- a/modules/feature-management/src/Volo.Abp.FeatureManagement.Application.Contracts/Volo/Abp/FeatureManagement/StringValueTypeJsonConverter.cs
+++ b/modules/feature-management/src/Volo.Abp.FeatureManagement.Application.Contracts/Volo/Abp/FeatureManagement/StringValueTypeJsonConverter.cs
@@ -1,4 +1,7 @@
 ﻿using System;
+using System.Collections.Generic;
+using System.IO;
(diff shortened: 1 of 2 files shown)


maliming:
Why not move the `StringValueTypeJsonConverter `to the `framework`.

hikalkan:
I thought that, but than see that it is only needed if we use it on the application layer of the feature-management module. No need to serialize/deserialize it in any other place and I don't see any other use case.

We move to the framework if we need it later.

PR #4920, opened 2020-07-30
Convert entity property to entry property type.

### LINKED ISSUE #4917: ObjectExtension: Handle extra property's type when deserialzing json with integer property value
ABP version: 3.0.4

When you extending a object with a integer property **p1**, the json will look like this:

```
{
   "A": value,
   "extraProperties": {
    "p1": 10
  }
}
```

When you call your api with json above, after deserialization, the extra property __p1__'s type will be object(long), and because long can not be cast to integer Implicitly, efcore will throw exception of type is not matched.

According to Json.NET:

> Json.NET by default reads integer values as Int64 because there is no way to know whether the value should be Int32 or Int64, and Int64 is less likely to overflow. For a typed property the deserializer knows to convert the Int64 to a Int32 but because your property is untyped you are getting an Int64. [...] It is just the way Json.NET has to work.

So, I think current object extension implementation is missing some kind of type conversion.

maliming:
Fix #4917


framework/src/Volo.Abp.EntityFrameworkCore/Volo/Abp/EntityFrameworkCore/AbpDbContext.cs
```diff
@@ -314,7 +315,25 @@ protected virtual void HandleExtraPropertiesOnSave(EntityEntry entry)
                     continue;
                 }
 
-                entry.Property(property.Name).CurrentValue = entity.GetProperty(property.Name);
+                var entityProperty = entity.GetProperty(property.Name);
+                if (entityProperty == null)
+                {
+                    continue;
```
hikalkan, line 331: But we also want to set null if `entity.GetProperty(property.Name)` returns null. Because we already checked `entity.HasProperty(property.Name)` before. So, if `entity.GetProperty(property.Name)` returns `null` that means it is intentionally set to `null`. It doesn't mean the property is missing. So, skipping if `entityProperty == null` makes a breaking change and breaks the current behavour. This would be better:

````csharp
if (entityProperty == null)
                {
entry.Property(property.Name).CurrentValue = null;
                    continue;
}
````

maliming: I updated the code and unit test.

PR #4967, opened 2020-08-05
Implemented the getting environment from remote URL

### LINKED ISSUE #4386: Angular: Read settings from Json files
Please modify Angular modules to read settings from Json files instead of **environment.ts** to allow overwriting settings in CD process, for instance change API Url per deploying environment.
   We saw the implementation of **AppPreBootstrap** which load settings from Json in ASP.NET Zero and tried to implement the logic in AbpCommercial but could not make it work somehow.

mehmet-erim:
Resolves #4386

Remote environment configuration can be added to the environment like below:

```ts
import { Config } from '@abp/ng.core';

export const environment = {
  production: false,
  hmr: false,
  remoteEnv: {
    url: '/assets/appsettings.json',
    method: 'GET',
    headers: {}
  }
} as Config.Environment;

```

### DIFF
--- a/docs/en/UI/Angular/Localization.md
+++ b/docs/en/UI/Angular/Localization.md
@@ -199,7 +199,7 @@ Some of the culture names defined in .NET do not match Angular locales. In such 
 
 ![locale-error](./images/locale-error.png)
 
-If you see an error like this, you should pass the `cultureNameToLocaleFileNameMapping` property like below to CoreModule's forRoot static method.
+If you see an error like this, you should pass the `cultureNameLocaleFileMap` property like below to CoreModule's forRoot static method.
 
 ```js
 // app.module.ts
@@ -209,7 +209,7 @@ If you see an error like this, you should pass the `cultureNameToLocaleFileNameM
     // other imports
      CoreModule.forRoot({
       // other options
-      cultureNameToLocaleFileNameMapping: { 
+      cultureNameLocaleFileMap: { 
         "DotnetCultureName": "AngularLocaleFileName",
         "pt-BR": "pt"  // example
       }
--- a/docs/zh-Hans/UI/Angular/Localization.md
+++ b/docs/zh-Hans/UI/Angular/Localization.md
@@ -199,7 +199,7 @@ export class AppComponent {}
 
 ![locale-error](./images/locale-error.png)
 
-如果你看到这样的错误,你应该像下面这样传递 `cultureNameToLocaleFileNameMapping` 属性到CoreModule的forRoot静态方法.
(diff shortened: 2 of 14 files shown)


olicooper:
My implementation ([see here](https://github.com/abpframework/abp/issues/4386#issuecomment-668626553)) allowed for partial replacement and appending of nested properties using a package called `deepmerge`. Is this something you'd be able to add in to make it more flexible and similar to .NET Core's `appsettings.json` replacement? 

I could have the default `environment.ts` like this:
```javascript
{
  recaptcha: { enabled: false, siteKey: 'AAAA' }
}
```

And my `/assets/appsettings.json` like this:
```javascript
{
  recaptcha: { enabled: true  },
  anotherSetting: "hi"
}
```

Which results in:
```javascript
{
  recaptcha: { enabled: true, siteKey: 'AAAA' },
  anotherSetting: "hi"
}
```

You don't have to use the `deepmerge` package, but if you do then I'm not sure how you'd want to handle array merging...
https://github.com/TehShrike/deepmerge#arraymerge

mehmet-erim:
Hi @olicooper

Thanks for your suggestion. We'll consider this.

olicooper:
Could you also consider the option to suppress the `RestError` and gracefully fall back to the default environment configuration if the remoteEnv is not available? Maybe a simple console debug/warning log rather than passing it to the error handler and showing to the end user?

olicooper:
I have just tried this on a staging environment and found an issue on the Angular UI...

I'm not sure why but the request to `/.well-known/openid-configuration` is made before the remoteEnv is configured. so the request is made to the wrong API (https://api.app.example.com instead of https://api.staging.example.com).

`environment.prod.ts`:
```javascript
{
  oAuthConfig: {
    issuer: 'https://api.app.example.com',
    // etc...
  },
  apis: {
    default: {
      url: 'https://api.app.example.com',
    },
  }
}
```

`appsettings.json`:
```javascript
{
  "oAuthConfig": {
    "issuer": "https://api.staging.example.com",
    // etc...
  },
  "apis": {
    "default": {
      "url": "https://api.staging.example.com"
    }
  }
}
```

I am using `@abp/ng.core 3.1.0-preview20200813`

mehmet-erim:
> Could you also consider the option to suppress the `RestError` and gracefully fall back to the default environment configuration if the remoteEnv is not available? Maybe a simple console debug/warning log rather than passing it to the error handler and showing to the end user?

We will create an injection token to be able to get catch error function from the `AppModule`. See the related issue: https://github.com/abpframework/abp/issues/4992

I will check it. Thanks for the information.

olicooper:
Perfect, great work! Thank you 👍

olicooper:
@mehmet-erim I think I have fixed the problem.. https://github.com/abpframework/abp/pull/5058

PR #4978, opened 2020-08-06
Support IEnumerable when adding query strings.

### LINKED ISSUE #4969: How to call a "Get" RESTful service with parameters of "IEnumarable<int>" from HttpAip.Client?
### Use Case
I have a scenario like this: Users supply a bunch of Id and suppose to get relative entitiesDto by calling a HttpAip.Client method like :

```
IEnumerable<int> ids = UserSuppliedIdsSomewhere();
IEnumerable<MyTypeDto> dto = await _myAppService.GetCustomListAsync(ids);
```

and the signature of GetCustomListAsync method is:

`Task<IEnumerable<MyTypeDto>> GetDetailListAsync(IEnumerable<int> ids);`

### Issue
The operation above will cause an exception since the IEnumerable<int> argument couldn't bind to  parameter properly. However, if I try this Api from swagger, everything is fine, and the argument will be serialized to querystring of url. So I doubt the HttpAip.Client failed to form proper querystring. My temporary solution is to change my api to Post by attribute [HttpPost], which will force the api to get data from request body.

### Question
Could you team supply a guide to me about solving the problem? After all, changing a Get behavior to Post is weird, and retrieving a collection of entities by there ids is such a common use case.

Thank you so much.

maliming:
Resolve #4969

### DIFF
--- a/framework/src/Volo.Abp.Http.Client/Volo/Abp/Http/Client/DynamicProxying/UrlBuilder.cs
+++ b/framework/src/Volo.Abp.Http.Client/Volo/Abp/Http/Client/DynamicProxying/UrlBuilder.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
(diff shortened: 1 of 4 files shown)


jimmidier:
Hi @maliming 
Thanks for the fix.
But I think we should support `Dictionary` too, since asp.net core's native model binding feature supports that.
I've created another pr based on yours #4980

maliming:
In fact, there are many situations, such as using the Dto in collections. We can’t be compatible with all situations. I’m considering introducing the **ObjectToQueryString** service, which is responsible for generating querystring.

Related article: https://www.hanselman.com/blog/ASPNETWireFormatForModelBindingToArraysListsCollectionsDictionaries.aspx

jimmidier:
I think you are right.

PR #4979, opened 2020-08-06
Implemented: Authentication Extensibility system to check username & password from an external source

### LINKED ISSUE #4977: Authentication Extensibility system to check username & password from an external source
ASP.NET Boilerplate has a good feature, named "External Authentication": https://aspnetboilerplate.com/Pages/Documents/Zero/User-Management#external-authentication

It makes possible to check username & password from any source (a database, a REST service, LDAP.... etc.). This is a good system which allows to easily extend the authentication mechanism.

I will implement a similar system for the ABP Framework, in the Identity module.

hikalkan:
Resolves #4977

@maliming this adds a new field (IsExternal) to the IdentityUser, so we need to add migrations after merging. I did for the application startup templates (not module templates). Can you also do it?


modules/identity/src/Volo.Abp.Identity.AspNetCore/Volo/Abp/Identity/AspNetCore/AbpSignInManager.cs
```diff
@@ -0,0 +1,66 @@
...
+            bool lockoutOnFailure)
+        {
+            foreach (var externalLoginProviderInfo in AbpOptions.ExternalLoginProviders.Values)
+            {
+                var externalLoginProvider = (IExternalLoginProvider) Context.RequestServices
+                    .GetRequiredService(externalLoginProviderInfo.Type);
+
+                if (await externalLoginProvider.TryAuthenticateAsync(userName, password))
+                {
+                    var user = await UserManager.FindByNameAsync(userName);
```
maliming, line 49: There may be a user with the same name already in the application.

maliming: Should we only query user that `IsExternal` is `true`?

hikalkan: I don't think it is necessary. If you configured LDAP as an external login provider later and username & pass matches to a user in the LDAP, then we can convert this to external user (see https://github.com/abpframework/abp/pull/4979/files/2ad0beb94a29c5876b0003da3021a56e023d2e74#diff-b4b68ccdc1d9c7beca4d9c3e3f329ba3R133).

PR #4984, opened 2020-08-07
Make Volo.Abp.Ldap support multi-tenancy.

maliming:
I removed the uncommon functions in the Ladp package and only kept the `Authenticate `method.

Developers can inherit `LdapManager `and implement more functions. eg: query or change.


hikalkan:
Is that OK to review & merge?

PR #5029, opened 2020-08-11
correct docs wording of Principle to Principal

paulfwilliams:
A "principle" is a guiding idea. The correct usage here is "principal" as in a chief actor or leader. See also the [ClaimsPrincipal class](https://docs.microsoft.com/en-us/dotnet/api/system.security.claims.claimsprincipal).


realLiangshiwei:
Thanks a lot @paulfwilliams

PR #5057, opened 2020-08-14
Add `MyProjectNameMenus.cs` for the app template

wakuflair:
Similar to https://github.com/abpframework/abp/commit/76f9e65ad08566ce2bee61238297c7ad065e3c8f


maliming:
Should we also update the app template project?

wakuflair:
This PR is for the app template only.
The module template has already been updated.

PR #5058, opened 2020-08-14
Fixed OAuth configuration bug when using remoteEnv

olicooper:
Relates to https://github.com/abpframework/abp/pull/4967
When using `remoteEnv`, OAuth was being configured with the default `environment.ts` settings rather than the one loaded from the `remoteEnv`. This should fix the problem.
Bug originally mentioned here: https://github.com/abpframework/abp/pull/4967#issuecomment-673464718

Function call timeline:
image


mehmet-erim (approved):

mehmet-erim:
Thanks @olicooper, this fixes the problem you mentioned.

PR #5084, opened 2020-08-17
Allow root "Components" folder in the web projects to have view component resources inside

### LINKED ISSUE #5072: Allow root "Components" folder in the web projects to have view component resources inside
With their related JS/CSS files.
Views, Pages and Themes folders have this privilege, "Components" also would be fine. Currently, we put the view components under Pages/Shared/Components or a similar location.

maliming:
Resolve #5072


hikalkan:
Thanks :)

PR #5089, opened 2020-08-17
Fixed BlobStoring project reference paths

olicooper:
When building the framework using a project root folder name other than `/abp` the `blob-storing` project fails to compile.


hikalkan:
Thanks!

PR #5091, opened 2020-08-17
Created utility functions that return DatePipe format

### LINKED ISSUE #5090: Create utility functions that return DatePipe format
We will create `date-utils.ts` file that contains utility functions.

First, two functions named `getShortDateFormat` and `getShortDateShortTimeFormat` will be created. These functions can be used like below:

```ts
import { ConfigStateService, getShortDateShortTimeFormat, getShortDateFormat } from '@abp/ng.core';
import { Component } from '@angular/core';

@Component({
  template: `
    <p>{{ now | date: dateTimeFormat }}</p>
    <p>{{ now | date: dateFormat }}</p>
  `,
})
export class YourComponent {
  dateTimeFormat: string;

  dateFormat: string;

  now = new Date();

  constructor(private config: ConfigStateService) {
    this.dateTimeFormat = getShortDateShortTimeFormat(this.config);
    this.dateFormat = getShortDateFormat(this.config);
  }
}

```

Result:

image

mehmet-erim:
Resolves #5090

### DIFF
--- a/npm/ng-packs/packages/core/src/lib/tests/date-utils.spec.ts
+++ b/npm/ng-packs/packages/core/src/lib/tests/date-utils.spec.ts
@@ -1,5 +1,3 @@
-import { Spectator, createComponentFactory, SpyObject } from '@ngneat/spectator/jest';
-import { Component } from '@angular/core';
 import { ConfigStateService, getShortDateFormat, getShortDateShortTimeFormat } from '@abp/ng.core';
 
 const dateTimeFormat = {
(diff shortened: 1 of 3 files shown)


npm/ng-packs/packages/core/src/lib/tests/date-utils.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+  let config: SpyObject<ConfigStateService>;
+
+  const createComponent = createComponentFactory({
+    component: DummyComponent,
+    mocks: [ConfigStateService],
+  });
+
+  beforeEach(() => {
+    spectator = createComponent();
+    config = spectator.inject(ConfigStateService);
```
bnymncoskuner: Do you need `createComponent` just to inject `ConfigStateService`? If so, we can simply create a dummy instance of `ConfigStateService` something like `const config = new ConfigStateService(undefined)`. We don't really care about `store` since we will `spyOn` `getDeep` anyway. It seems to me that this test could be shortened if we omit the creating component part.

bnymncoskuner (commented):
LGTM

Issue #4775, opened 2020-07-15
Update to Quartz 3.1.0


olicooper:
Version 3 is not official yet but there is a beta out with some great improvements [v3.1.0-beta2](https://github.com/quartznet/quartznet/releases/tag/v3.1.0-beta2)

Features such as DI, MySQL connector and appsettings integration would be useful to utilize in abp. There are a number of bug fixes and performance improvements too.

I have been using a nightly version in my project for a while and it's been working great so there should be too many issues with upgrading it.

liangshiw:
Quartz.Net v3.1.0-beta2 looks good. It contains a lot of changes, I will try integrate.

olicooper:
The latest stable version of Quartz has just been released and it looks great! OpenTelemtry and TimeZoneConverter support now added too 👍 

[GitHub release](https://github.com/quartznet/quartznet/releases/tag/v3.1.0) | [Changelog](https://github.com/quartznet/quartznet/blob/master/changelog.md#release-310-jul-24-2020)

**Update:**
I have upgraded my own project to this release but there are issues when using MySql as the provider 😢 
`Volo.Abp.EntityFrameworkCore.MySQL` uses `Pomelo.EntityframeworkCore.MySql @ v3.1.1` which depends on `MySqlConnector < 1.0.0`, but Quartz requires `MySqlConnector >= 1.0.0`, which causes a dependency conflict between the packages.
Pomelo are going to upgrade their `MySqlConnector` dependancy to >= 1.0.0 but they won't be doing it [until v5.0.0](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/pull/1103#issuecomment-643736104) (because it is a breaking change) which I assume aligns with the release of .NET 5 - so we could be waiting until November for this to be resolved! 😢 
If we upgrade this package before .NET 5, there might be a way to fix the MySQL dependency problem by using a [custom provider](https://github.com/quartznet/quartznet/issues/902#issuecomment-663678704) but I haven't looked in to it yet.

olicooper:
I have a few suggestions if that's okay?

1. Remove the `Microsoft.Data.SqlClient` dependency and let the developer choose which provider they use (the same way that the core Quartz library does it).
2. Add a dependency on `Quartz.Plugins.TimeZoneConverter` so that it integrates with the recent changes made here https://github.com/abpframework/abp/pull/3933
[...]

PR #5094, opened 2020-08-17
fix: MySqlConnector issues between ABP and Quartz

olicooper:
After a lot of playing around, I managed to fix the MySqlConnector compatibility issues between ABP and Quartz when using MySql (as [mentioned here](https://github.com/abpframework/abp/issues/4775#issuecomment-674979973)). Original advice to create a custom provider came from the developer of Quartz.net ([see here](https://github.com/quartznet/quartznet/issues/902#issuecomment-663678704))

This overrides the `MySqlConnector` provider to revert the provider namespace back to pre 1.0.0 ([see here](https://github.com/quartznet/quartznet/blob/effbc3ee162e429ea6f198cc67927cdf01747c55/src/Quartz/Impl/AdoJobStore/Common/dbproviders.properties#L88-L100))

This change will allow you to remove this hotfix from the framework without any changes for users of ABP after Pomelo update their MySqlConnector dependency >= 1.0.0 (which will be after the .NET 5 release - [see here](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/pull/1103#issuecomment-643736104))

Alternatively we can use a custom provider name like `AbpMySqlConnector` instead of overriding the default `MySqlConnector` but my testing shows that it is working fine as it is and users of ABP won't need to update their config when it is removed either.

Feel free to amend :)

### DIFF
--- a/framework/src/Volo.Abp.Quartz/Volo/Abp/Quartz/AbpQuartzModule.cs
+++ b/framework/src/Volo.Abp.Quartz/Volo/Abp/Quartz/AbpQuartzModule.cs
@@ -1,6 +1,7 @@
 ﻿using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Options;
 using Quartz;
+using Quartz.Impl.AdoJobStore.Common;
 using Volo.Abp.Modularity;
 using Volo.Abp.Threading;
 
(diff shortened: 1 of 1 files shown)


realLiangshiwei (approved):

realLiangshiwei:
Thanks a lot @olicooper

PR #5105, opened 2020-08-18
Added forget/reset password feature to account module

### LINKED ISSUE #2608: Add forgot/reset password functionality
I don't believe there is any functionality to allow the user to reset their password at the login screen if they forget it. This existed in the [UserAppService](https://github.com/aspnetboilerplate/aspnetboilerplate/blob/dev/test/Abp.Zero.SampleApp/Users/UserAppService.cs#L61) of the old framework where an email could be sent out to the user and a `PasswordResetCode` stored in the database.

It would be really nice if this could be added to this framework, and work with the EmailConfirmed property so new users can register and change their password by clicking on a link in the email sent.

realLiangshiwei:
Resolve https://github.com/abpframework/abp/issues/2608


hikalkan:
Great, thanks ;)

PR #5107, opened 2020-08-18
Cli new project default mobile "none"

### LINKED ISSUE #5106: Set `-m none` as a default for the ABP CLI
For now, [abp cli](https://docs.abp.io/en/abp/latest/CLI) creates mobile project with default.

I think the mobile project creation should be optional.

It will be more good if we have a command that creates default empty mobile project whenever we want (adding mobile project to an exist project).

yekalkan:
resolves https://github.com/abpframework/abp/issues/5106


docs/en/Getting-Started.md
```diff
@@ -62,10 +62,10 @@ dotnet tool update -g Volo.Abp.Cli
 Use the `new` command of the ABP CLI to create a new project:
 
 ````shell
-abp new Acme.BookStore{{if UI == "NG"}} -u angular {{end}}{{if DB == "Mongo"}} -d mongodb{{end}}{{if Tiered == "Yes" && UI != "NG"}} --tiered {{else if Tiered == "Yes" && UI == "NG"}}--separate-identity-server{{end}}
+abp new Acme.BookStore{{if UI == "NG"}} -u angular {{end}}{{if DB == "Mongo"}} -d mongodb{{end}}{{if Tiered == "Yes" && UI != "NG"}} --tiered {{else if Tiered == "Yes" && UI == "NG"}}--separate-identity-server{{end}} --mobile react-native
```
hikalkan: Remove ` --mobile react-native` from here and update the documentation (and screenshots) based on that

hikalkan (changes requested):

PR #5117, opened 2020-08-19
Fixed the ngcc compiling error of the @abp/utils package

### LINKED ISSUE #5005: Circular dependency @abp/utils/dist/abp-utils.d.ts.map
- Abp 3.0.5
- Angular ("@abp/ng.core": "^3.0.5")
- Steps needed to **reproduce** the problem: `ng build`
- Warning/Error during NGCC:
```
Warning: Unable to fully load LOCAL_PATH_HERE/node_modules/@abp/utils/dist/abp-utils.d.ts for source-map flattening: Circular source file mapping dependency:
 LOCAL_PATH_HERE/node_modules/@abp/utils/dist/abp-utils.d.ts.map -> LOCAL_PATH_HERE/node_modules/@abp/utils/dist/abp-utils.d.ts.map
  Error: Tried to write LOCAL_PATH_HERE/node_modules/@abp/utils/dist/abp-utils.d.ts.__ivy_ngcc_bak with an ngcc back up file but it already exists so not writi
ng, nor backing up, LOCAL_PATH_HERE/node_modules/@abp/utils/dist/abp-utils.d.ts.
  This error may be because two or more entry-points overlap and ngcc has been asked to process some files more than once.
  You should check other entry-points in this package and set up a config to ignore any that you are not using.
```

Seems like there is a circular dependency in @abp/utils

mehmet-erim:
Resolves #5005


bnymncoskuner (commented):
LGTM 👏 👍

PR #5130, opened 2020-08-20
 enable the default generic repository

xyfy:
supplier a descirption to enable the `default generic repository`


hikalkan:
I did it, thanks.

See https://github.com/abpframework/abp/commit/7435c600686c110b216642d86164c73f6e829892

PR #5134, opened 2020-08-20
CmsKit: Should focus on the comment area after login and return back

### LINKED ISSUE #5073: Should focus on the comment area after login and return back
CMS Kit commit area shows a "Login to Reply" or "Login to Write Comment" buttons when user was not authenticated. It redirects to login and comes back to the same page after login. In this case, it should focus to the related commenting area for a better UX.

I added the related comment info to the loginUrlHash (see https://github.com/abpframework/abp/blob/dev/modules/cms-kit/src/Volo.CmsKit.Public.Web/Pages/CmsKit/Shared/Components/Commenting/CommentingViewComponent.cs#L40 and the cshtml side to know how it was set), so you can extract this in the JS side and focus it.

yekalkan:
resolves https://github.com/abpframework/abp/issues/5073

### DIFF
--- a/modules/cms-kit/src/Volo.CmsKit.Public.Web/Pages/CmsKit/Shared/Components/Commenting/default.js
+++ b/modules/cms-kit/src/Volo.CmsKit.Public.Web/Pages/CmsKit/Shared/Components/Commenting/default.js
@@ -136,11 +136,15 @@
             }
 
             function focusOnHash($container) {
-                if (location.hash === ''){
+                if (!location.hash.toLowerCase().startsWith('#cms-comment')){
                     return;
                 }
 
-                $('html, body').animate({ scrollTop: $(location.hash).offset().top + $(this).height() / 2 }, 'slow');
+                $('html, body').animate(
+                    {
+                        scrollTop: $(location.hash).offset().top + $(this).height() / 2
+                    },
+                    'slow');
             }
 
             function init() {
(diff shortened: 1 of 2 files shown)


modules/cms-kit/src/Volo.CmsKit.Public.Web/Pages/CmsKit/Shared/Components/Commenting/default.js
```diff
@@ -135,13 +135,23 @@
                 });
             }
 
+            function focusOnHash($container) {
+                if (location.hash === ''){
+                    return;
+                }
+
+                $('html, body').animate({ scrollTop: $(location.hash).offset().top + $(this).height() / 2 }, 'slow');
```
hikalkan: I think this is very generic code. We should move this to a framework package, or restrict it: apply it only for hashes start with `cms-comment`.

Also, doesn it cover the "reply" scenario (should focus on the reply)

yekalkan: > Also, doesn it cover the "reply" scenario (should focus on the reply)

It can't focus on reply or a single comment because ReturnUrlHash only has EntityType & EntityId information. So it focuses on the whole related comment area.
Of course If focusing on a reply or single comment is very important, i can do it.

PR #5137, opened 2020-08-21
Replaced Angular Client Proxy Generation code with Angular Schematics

### LINKED ISSUE #4881: Replace Angular Client Proxy Generation code with Angular Schematics
We see that we have some problems with the `abp generate-proxy` command. We decided to re-write the code generation part of it using the Angular schematics. `abp generate-proxy` command will remain same and produce the same (or very similar) output, so this is not a breaking change for existing applications.

We see this problem important and want to include this to version 3.1.0. So, we postponed the 3.1.0-RC1 release data from 29.07.2020 to 13.08.2020. We think it will be completed in that date. If not, we may postpone it a few days more. I wanted to inform the community.

armanozak:
Fixes #4881


npm/ng-packs/apps/dev-app/src/environments/environment.ts
```diff
@@ -21,5 +21,8 @@ export const environment = {
     default: {
       url: 'https://localhost:44305',
     },
+    saas: {
+      url: 'https://localhost:44305',
+    },
```
mehmet-erim:
```suggestion
```

Saas module is not available in free side

npm/ng-packs/angular.json
```diff
@@ -366,6 +366,29 @@
         }
       }
     },
+    "schematics": {
+      "projectType": "library",
+      "root": "packages/schematics",
+      "sourceRoot": "packages/schematics/src",
+      "prefix": "abp",
+      "architect": {
+        "test": {
```
mehmet-erim, line 375:
```suggestion
      "architect": {
        "build": {
          // do not remove the build and options properties, the symlink-manager process breaks without these
          "options": {
            "tsConfig": "packages/commercial-ui/tsconfig.lib.json",
            "project": "packages/commercial-ui/ng-package.json"
          }
        },
        "test": {
```

If the `build` property with the `options` property does not exist, the symlink-manager breaks

mehmet-erim (changes requested):

bnymncoskuner (commented):
LGTM

mehmet-erim (approved):

PR #5163, opened 2020-08-25
fix rxjs observable<any>.next()

badre429:


armanozak:
Hi @badre429,

Thanks for your PR. Could you please elaborate on why you think this change is necessary?

Have a nice day.

badre429:
@armanozak  the type of completed$ is  Observable<true> and confirmation$ is  Observable<any> 
There for the next version of rxjs will drop next() function with no param

armanozak:
Hi @badre429,

Thank you for the explanation. We really appreciate your contribution. However, stable RxJS v7 is not released yet and we are planning to handle RxJS migration altogether only after Angular starts using RxJS v7. We may have to do similar changes as proposed here eventually. Nevertheless, we believe it is early to implement those changes and they are likely to differ from what you have suggested.

I am sincerely sorry that we cannot accept this PR, but thank you again.

PR #5190, opened 2020-08-27
Introduce link user feature of Identity module.

### LINKED ISSUE #1082: Impersonating users
Whats the current recommended way of impersonating users?

maliming:
If this PR can be merged, I will update the migration files.


modules/identity/src/Volo.Abp.Identity.Domain/Volo/Abp/Identity/IdentityLinkUser.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System;
+using Volo.Abp.Domain.Entities;
+
+namespace Volo.Abp.Identity
+{
+    public class IdentityLinkUser : AggregateRoot<Guid>
+    {
```
hikalkan, line 7: Create a protected default constructor, for orm deserialization.

modules/identity/src/Volo.Abp.Identity.Domain/Volo/Abp/Identity/IdentityLinkUser.cs
```diff
@@ -0,0 +1,36 @@
...
+            : base(id)
+        {
+            SourceUserId = sourceUserId;
+            SourceTenantId = sourceTenantId;
+
+            TargetUserId = targetUserId;
+            TargetTenantId = targetTenantId;
+        }
+
+        public virtual Guid SourceUserId { get; protected set; }
```
hikalkan: We write properties on top of the constructor, by convention.

modules/identity/src/Volo.Abp.Identity.EntityFrameworkCore/Volo/Abp/Identity/EntityFrameworkCore/IIdentityDbContext.cs
```diff
@@ -17,5 +17,7 @@ public interface IIdentityDbContext : IEfCoreDbContext
         DbSet<OrganizationUnit> OrganizationUnits { get; set; }
 
         DbSet<IdentitySecurityLog> IdentitySecurityLogs { get; set; }
+
+        DbSet<IdentityLinkUser> IdentityLinkUsers { get; set; }
```
hikalkan: `LinkUsers` would be enough for the property name. Same for `SecurityLogs` too (take Users, Roles.. as examples)

framework/src/Volo.Abp.MultiTenancy/Volo/Abp/MultiTenancy/CurrentImpersonatorTenant.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+using System.Security.Principal;
+using Volo.Abp.DependencyInjection;
+using Volo.Abp.Security.Claims;
+
+namespace Volo.Abp.MultiTenancy
+{
+    public class CurrentImpersonatorTenant : ICurrentImpersonatorTenant, ITransientDependency
```
hikalkan: Instead of introducing `ICurrentImpersonatorTenant`, just add a `ImpersonatorId` to the `ICurrentTenant`.

framework/src/Volo.Abp.Security/Volo/Abp/Users/CurrentImpersonatorUser.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+using System.Security.Principal;
+using Volo.Abp.DependencyInjection;
+using Volo.Abp.Security.Claims;
+
+namespace Volo.Abp.Users
+{
+    public class CurrentImpersonatorUser : ICurrentImpersonatorUser, ITransientDependency
```
hikalkan: Instead, add `ImpersonatorId` to the `ICurrentUser`.

hikalkan (changes requested):

olicooper:
Couple of questions on this feature...
* Will there be a security log when someone starts and ends an impersonation session?
* Will tenants be able to allow/deny permission to impersonate? The use case for this is when a host 'support' user needs to provide support for a tenant. A tenant administrator can temporarily grant permission to access their data, then the user impersonates the tenant (not necessarily the user - so all updated records have the host UserId as the LastModifierId) to see all the information for the tenant. My current implementation involves checking a 'SupportAccessGranted' setting for each tenant on each API request.

leonkosak:
@hikalkan, @maliming 
I completely agree with second suggestion from @olicooper.
The default behavior should be that login as impersonator is not allowed. User should explicitly allow impersonation for defined period of time (for instance "to" datetime).
Of course, there could be option for "unlimited impersonator access".

I have use case that tenant stored in tenants are really sensitive and impersonator should not just login as one user (like in ANZ).
Of course, that is possible to view such accesses in AuditLogs table, but it should be prevented (rather than reacting when data are stolen).

My suggestions

1) **On tenant-level**, there should be setting "Allow automatic impersonation" which have the following options in dropdown menu:
- Allow on demand (**default** selected)
- Do not allow
- Allow always  (**mark red or dangerous**)

**Allow on demand**
Tenant user have options (on topbar or somewhere in user settings - profile) for impersonalization:

- Do not allow (**default**, it goes back to this option when impersonalization allowed time ends)

- Allow limited (when user choose this option, there has to be option for how long or "to datetime" and has to be set)

- Allow always (**mark as red or dangerous**) - **this option is visible/available only it there is "Allow users to add unlimited access" checked**


**Do not allow**
Tenant user has no option to set up personalization, because it is **forbidden** on tenant-level.

**Allow always**
Tenant user has no option to set up personalization, because it is **fully-allowed** on tenant-level.


**IMPORTANT**
When "Allow on demand" is selected, there should has to be a checkbox (**which should be unselected by default**) "Allow users to add unlimited access".




I know that this is complicated logic, but it is required when there are really sensitive date (and GDPR-related data) and one user wants help from other user (for instance helpdesk which is not even the same company).

maliming:
Add `LinkLoginExtensionGrantValidator` to regenerate identity server token for re-signin(Angular).

image

maliming:
hi @olicooper @leonkosak 
I have encountered some problems with impersonating users/tenants, so I implement `link user` first, then I will open another PR. Your comments are welcome.

olicooper:
Thanks for your hard work @maliming, I will try this in my app to see if it works once it is on the nightly train.

One feature of my temporary solution ([seen here](https://github.com/abpframework/abp/issues/1082#issuecomment-574224409)) is the ability to limit what the support user can do via the permissions system. It was an accidental feature but it turned in to something quite interesting. 
* I created a static 'support' role on the host with a single 'View Tenants' permission and added a 'impersonate tenant' button to store the __tenantId in session cache and send with every request. This stops 'support' employees messing with the host or doing anything damaging.
* I created a static 'support' role in the migration seeding code which grants most permissions to the support role for each tenant.
* When a support user impersonates a tenant, they inherit the permissions of the tenant's 'support' role that they are impersonating

This worked out great and allows each tenant to have fine-grained control over what a support person can do. It might be worth considering as an extension to this feature?

Edit:
* What if the user we are impersonating/linking has accidentally denied themselves all permissions and we are using this feature to try and fix the problem? https://github.com/abpframework/abp/issues/5395
* Are we able to impersonate the tenant without impersonating the user?

ismcagdas (approved):

tvddev:
@hikalkan Do you know if the impersonating a tenant user from the host side workaround that @olicooper has suggested above and in #1082 work in v4.3?

Issue #5213, opened 2020-08-28
When tiered templates, tenants of independent databases cannot login.


realLiangshiwei:
Related https://support.abp.io/QA/Questions/379/Login-Error

leonkosak:
I have Angular UI (and separated IdentityServer) and separated databases for all tenants.
There is login issue (error) when I forgot to start Redis.

PR #5214, opened 2020-08-28
Add formTenantResolveContributor

realLiangshiwei:
Resolve https://github.com/abpframework/abp/issues/5213


framework/src/Volo.Abp.AspNetCore.MultiTenancy/Volo/Abp/AspNetCore/MultiTenancy/FormTenantResolveContributor.cs
```diff
@@ -0,0 +1,23 @@
...
+        public override string Name => ContributorName;
+
+        protected override string GetTenantIdOrNameFromHttpContextOrNull(ITenantResolveContext context, HttpContext httpContext)
+        {
+            if (httpContext.Request == null || !httpContext.Request.Form.Any())
+            {
+                return null;
+            }
+
+            return httpContext.Request.Form[context.GetAbpAspNetCoreMultiTenancyOptions().TenantKey];
```
maliming, line 20: Use the async method to get value from the form will be better.
Maybe we should make `ITenantResolver` support `async`.
```
var form = await HttpRequest.ReadAsFormAsync();
return form["context.GetAbpAspNetCoreMultiTenancyOptions().TenantKey"];
```

realLiangshiwei: I agree. May resolve tenants from `JSON`, asynchronous is better. But this is a breaking change. @hikalkan  What do you think?

framework/src/Volo.Abp.AspNetCore.Authentication.OpenIdConnect/Microsoft/Extensions/DependencyInjection/AbpOpenIdConnectExtensions.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        public static AuthenticationBuilder AddAbpOpenIdConnect(this AuthenticationBuilder builder, string authenticationScheme, string displayName, Action<OpenIdConnectOptions> configureOptions)
+        {
+            return builder.AddOpenIdConnect(authenticationScheme, displayName, options =>
+            {
+                options.ClaimActions.MapAbpClaimTypes();
+
+                options.Events = new OpenIdConnectEvents
+                {
+                    OnAuthorizationCodeReceived = receivedContext =>
```
hikalkan, line 30: I don't know that, but see it can be risky (because we are trusting a plain text value that can be changed by clients).

* When OnAuthorizationCodeReceived is called? In every request? While obtaining the first token by the client?
* What SetParameter does? Where this parameter goes and how it is used then?
* ...

Can you please explain this in details, I need to be sure that is a true approach.

Thanks.

realLiangshiwei: 1. It will be called every time when the client receives the authorization code returned by the authorization server: https://github.com/dotnet/aspnetcore/blob/bbf7c8780c42e3d32aeec8018367037784eb9181/src/Security/Authentication/OpenIdConnect/src/OpenIdConnectHandler.cs#L639
2. In code flow, the client will use the authorization code to send a request to get an access token. https://github.com/dotnet/aspnetcore/blob/bbf7c8780c42e3d32aeec8018367037784eb9181/src/Security/Authentication/OpenIdConnect/src/OpenIdConnectHandler.cs#L807v
   This parameter will be serialized as `FormUrlEncodedContent` as a request parameter, but it missing `TenantKey`, so I use the `SetParameter` method to add the `TenantKey` parameter.  (actually, `Tenantkey` is in the cookie, but due to the same-origin policy, the browser will not send the cookie)

hikalkan:
Thanks for the development and explanations ;)

framework/src/Volo.Abp.AspNetCore.MultiTenancy/Volo/Abp/AspNetCore/MultiTenancy/FormTenantResolveContributor.cs
hikalkan: We don't want to introduce such a breaking change now.

PR #5222, opened 2020-08-28
Introduced separate commands for adding, removing, and refreshing proxies

### LINKED ISSUE #5202: Proxy Generator: Additional command for removing generated proxy required

### LINKED ISSUE #5221: Proxy Generator: Regenerate previously generated modules on every run

armanozak:
Fixes #5202
Fixes #5221 

### Command Names
- `proxy-add`
- `proxy-refresh`
- `proxy-remove`

### Parameters
All use the same parameters with the legacy `proxy` command.


hikalkan:
* `abp generate-proxy` ([see](https://github.com/abpframework/abp/blob/dev/docs/en/CLI.md#generate-proxy)) uses the `proxy-add`
`abp remove-proxy` ([see](https://github.com/abpframework/abp/blob/dev/docs/en/CLI.md#remove-proxy), a new command) uses the `proxy-remove`

No need to call proxy-refresh since both commands do refresh after the actual command. So, in practical, `abp generate-proxy` refreshed the proxies if you've generated it before.

PR #5272, opened 2020-09-03
Build dictionary query string

wakuflair:
This PR handled object type of the `parameterInfo.value` in `abp.utils.buildQueryString`, in order to pass parameters as dictionary-compatible from JS to DTO.

For example, given a DTO defined:

``` csharp
public class GetListInput 
{
    public Dictionary<string, object> Filters { get; set; }
}
```

You can pass values to `Filters` from JS like this:
``` javascript
ajax: abp.libs.datatables.createAjax(myService.getList, function () {
                    return {
                        filters: {
                            name: "John",
                            age: 36
                        }
                    }
                }),
```

will generate the query string: `myService?filters.name=John&filters.age=36...`

### DIFF
--- a/framework/src/Volo.Abp.AspNetCore.Mvc/Volo/Abp/AspNetCore/Mvc/ModelBinding/AbpExtraPropertiesDictionaryModelBinderProvider.cs
+++ b/framework/src/Volo.Abp.AspNetCore.Mvc/Volo/Abp/AspNetCore/Mvc/ModelBinding/AbpExtraPropertiesDictionaryModelBinderProvider.cs
@@ -24,7 +24,7 @@ namespace Volo.Abp.AspNetCore.Mvc.ModelBinding
                 return null;
             }
 
-            if (!context.Metadata.ContainerType.IsAssignableTo<IHasExtraProperties>())
+            if (context.Metadata.ContainerType == null || !context.Metadata.ContainerType.IsAssignableTo<IHasExtraProperties>())
             {
                 return null;
             }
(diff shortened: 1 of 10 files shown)


maliming:
There may be other problems with using Dictionary as parameter. https://github.com/abpframework/abp/pull/4978

wakuflair:
Yeap this is an implementation for simple Dictionary (no nesting).
I think it's enough for most of the basic usage, better than not having this feature at all.

maliming:
The `AddQueryStringParameters` method doesn't support `Dictionary `.

hikalkan:
@maliming you test & decide to accept this ;)

maliming:
Unless we make `UrlBuilder` support dictionary parameters.

wakuflair:
> Unless we make `UrlBuilder` support dictionary parameters.

I also added dictionary support for `UrlBuilder`, it now supports specific value types( `<string, string>`, `<string, int>`, etc), but the `<string, object>` doesn't work.
I think I need some guides here 😂

wakuflair:
Hi all

I added two more commits. Here is the summary:

### JS side
* Introduced a new JS function `abp.utils.serializeToQueryString` which can serialize any JS object to a flat query string.

### C# side
1. Add `IObjectToQueryStringConverter` interface as mentioned by @maliming in https://github.com/abpframework/abp/pull/4978#issuecomment-670269019, and an implementation `ObjectToQueryStringConverter` can convert C# object to query string similar to `abp.utils.serializeToQueryString`.
1. Modify test `GetWithComplexType` to  cover complex nesting, dictionaries, etc.

maliming:
hi @wakuflair 

Could you add more unit tests for `SerializeToQueryString` method?

wakuflair:
OK. I will find a time to add more tests.

wakuflair:
It's hard to cover all the cases.

Prefer using DTO instead of query strings

PR #5292, opened 2020-09-04
OrganizationUnit should have filter parameter

jonny-xhl:
- I think OrganizationUnit should have filter parameter.
- Add LongCount


modules/identity/src/Volo.Abp.Identity.EntityFrameworkCore/Volo/Abp/Identity/EntityFrameworkCore/EfCoreOrganizationUnitRepository.cs
```diff
@@ -162,7 +168,18 @@ public override IQueryable<OrganizationUnit> WithDetails()
             DbContext.Set<IdentityUserOrganizationUnit>().RemoveRange(ouMembersQuery);
         }
 
-        protected virtual IQueryable<IdentityUser> CreateGetMembersFilteredQuery(OrganizationUnit organizationUnit, string filter = null)
+        public virtual async Task<long> GetLongCountAsync(Guid? parentId, string filter = null,
```
gterdem, line 255: Why having parentId parameter when it's not been used?

modules/identity/src/Volo.Abp.Identity.Domain/Volo/Abp/Identity/IOrganizationUnitRepository.cs
```diff
@@ -80,5 +81,10 @@ public interface IOrganizationUnitRepository : IBasicRepository<OrganizationUnit
             OrganizationUnit organizationUnit,
             CancellationToken cancellationToken = default
         );
+
+        Task<long> GetLongCountAsync(
+            Guid? parentId,
```
gterdem, line 119: Is `parentId` parameter necessary?

modules/identity/src/Volo.Abp.Identity.MongoDB/Volo/Abp/Identity/MongoDB/MongoOrganizationUnitRepository.cs
```diff
@@ -155,7 +161,18 @@ public virtual async Task RemoveAllMembersAsync(OrganizationUnit organizationUni
             }
         }
 
-        protected virtual IMongoQueryable<IdentityUser> CreateGetMembersFilteredQuery(OrganizationUnit organizationUnit, string filter = null)
+        public virtual async Task<long> GetLongCountAsync(Guid? parentId, string filter = null,
```
gterdem, line 236: Same unused `parentId` parameter.

gterdem (changes requested):
`GetLongCountAsync` method has some unused parameters. Would you mind to check it?

modules/identity/src/Volo.Abp.Identity.MongoDB/Volo/Abp/Identity/MongoDB/MongoOrganizationUnitRepository.cs
jonny-xhl: I have already solved.

1. I think it is necessary for GetListAsync to filter through parentId when the OrgnizationUnit is more complex.

**Example:**
- The organization structure tree is lazily loaded.
- When there are many suborganizational units.

gterdem: I can't see updated code on View Change set. Can you update your dev branch?

jonny-xhl: lasted commit id 79f1369(resolved conflicts).

gterdem (approved):

maliming:
hi

When the `parentId `parameter of the `GetListAsync `and `GetLongCountAsync `methods is **null**, we are actually querying the root node. Is this expected?

Is this a breaking change?

ismcagdas:
@maliming I think this is a breaking change. Maybe we shouldn't add parentId to filter if parentId is null. But in that case, there is no way to just query root items.

gterdem:
@maliming thanks. I forgot to predict the null use case.

PR #5300, opened 2020-09-04
Fix: localization data is lost when exception is on remote

yekalkan:
Also, collect data for AbpIdentityResultException


maliming:
hi, @yekalkan Can you write some unit tests?

PR #5315, opened 2020-09-07
Add more property to user claims & ICurrentUser

maliming:
Set the default `ClaimsIdentity` type of the Identity module.
```cs
Configure<IdentityOptions>(options =>
{
	options.ClaimsIdentity.UserIdClaimType = AbpClaimTypes.UserId;
	options.ClaimsIdentity.UserNameClaimType = AbpClaimTypes.UserName;
	options.ClaimsIdentity.RoleClaimType = AbpClaimTypes.Role;
});
```


framework/src/Volo.Abp.Security/Volo/Abp/Security/Claims/AbpClaimTypes.cs
```diff
@@ -44,15 +59,14 @@ public static class AbpClaimTypes
...
-        public static string TenantId { get; set; } = "tenantid";
-
+        public static string TenantId { get; set; } = "tenant_id";
 
         /// <summary>
-        /// Default: "editionid".
+        /// Default: "edition_id".
         /// </summary>
-        public static string EditionId { get; set; } = "editionid";
+        public static string EditionId { get; set; } = "edition_id";
```
hikalkan: Do not change existing claim names. It is an unnecessary breaking change. It may even effect the existing access tokens obtained by clients, when we upgrade the ABP Framework

framework/src/Volo.Abp.Security/Volo/Abp/Security/Claims/AbpClaimTypes.cs
```diff
@@ -44,15 +59,14 @@ public static class AbpClaimTypes
         public static string PhoneNumberVerified { get; set; } = "phone_number_verified";
 
         /// <summary>
-        /// Default: "tenantid".
+        /// Default: "tenant_id".
         /// </summary>
-        public static string TenantId { get; set; } = "tenantid";
-
+        public static string TenantId { get; set; } = "tenant_id";
```
hikalkan: Do not change existing claim names

framework/src/Volo.Abp.Security/Volo/Abp/Security/Claims/AbpClaimTypes.cs
```diff
@@ -13,6 +13,21 @@ public static class AbpClaimTypes
...
+
+        /// <summary>
+        /// Default: <see cref="ClaimTypes.Surname"/>
+        /// </summary>
+        public static string SurName { get; set; } = ClaimTypes.Surname;
+
+        /// <summary>
+        /// Default: security_stamp
+        /// </summary>
+        public static string SecurityStamp { get; set; } =  "security_stamp";
```
hikalkan: Do not define such a property. We don't need it.
It is already different than the ASP.NET Core's default (which is `AspNet.Identity.SecurityStamp`). We don't like to change default wherever possible.
So, remove this property and its usage.

hikalkan (changes requested):

maliming:
Action check failed, but the problem has nothing to do with my code, it is pending investigation.

PR #5351, opened 2020-09-09
Use MongoTestCollection in OrganizationUnitRepository_Tests

maliming:


realLiangshiwei:
https://github.com/abpframework/abp/blob/dev/framework/test/Volo.Abp.EntityFrameworkCore.Tests/Volo/Abp/EntityFrameworkCore/DataFiltering/HardDelete_Tests.cs

realLiangshiwei:
https://github.com/abpframework/abp/blob/dev/framework/test/Volo.Abp.MongoDB.Tests/Volo/Abp/MongoDB/DataFiltering/HardDelete_Tests.cs

`HardDelete_Tests` also need.

PR #5398, opened 2020-09-15
Categorize the files of the Identity module.

### LINKED ISSUE #5317: Create some directories to categorize the files under Volo.Abp.Identity.Domain.
It looks messy now.

https://github.com/abpframework/abp/tree/dev/modules/identity/src/Volo.Abp.Identity.Domain/Volo/Abp/Identity

maliming:
Resolve #5317

No code changes, only rename.

image

image


maliming:
I will resolve those conflicts.

hikalkan:
Categorizing the files like that seems fine, but I think we have a small problem: We strictly follow the "folder path" - "namespace" match everywhere until now. With this PR, it is broken and Resharper (and Rider) will constantly warn us. Even if they don't warn, I don't like this. If we change namespaces, this will be a big and unnecessary breaking change for the developers.

So, I will not merge this, sorry :(

PR #5423, opened 2020-09-16
GetDbContext and GetDbSet implementation for IRepository

ekaradev:


PMExtra:
Please reference my #5697 

Most of `Class<TEntity, TKey>` are both inherit of `Class<TEntity>`, and all classes in this file followed the rule.

So I don't think there should be new overload methods, just change the original method to super class is enough.

Ps. for the TODO comment at begin of the file, I think it means "for any IRepository **implementations**", such as mongo or other implementation instead efcore. So I didn't remove it.

PR #5466, opened 2020-09-18
Remove TwoFactorEnabled related UI & Application Service.

### LINKED ISSUE #5460: Identity Module: Changing Two Factor feature value should effect user create/edit modals
MVC & Tiered & EF Core

`abp new Mvc.TierEfCore -t app -u mvc --mobile none --database-provider ef --separate-identity-server`

maliming:
Resolve #5460

Developers should implement the Two-factor feature themselves.


maliming:
**Angular needs to make similar changes after accepting the PR.**

Issue #5483, opened 2020-09-18
MongoOrganizationUnitRepository should apply filters


realLiangshiwei:

realLiangshiwei:
Close by https://github.com/abpframework/abp/pull/5484

PR #5484, opened 2020-09-18
Apply filter to MongoOrganizationUnitRepository queries

realLiangshiwei:
Resolve https://github.com/abpframework/abp/issues/5483


hikalkan:
Thanks. We improve this in the next version. https://github.com/abpframework/abp/issues/5514

PR #5494, opened 2020-09-18
feat: french localization

badre429:


maliming:
hi @badre429  

Could you create PR in **dev** branch?

badre429:
@maliming  
hi if possible we need french  localization for 3.2 since localization is not  has no impact on behavior of rc

hikalkan:
I agree about to merge this to rel-3.2. In fact, we should always complete the localizations in the preview period.
So, thank you @badre429 

@maliming it seems OK to me, can you merge, then test it?

hikalkan (approved):

maliming:
I'll test it in next RC.

PR #5510, opened 2020-09-21
Refactored the IsMultiTenant logic of BlobStoring.

### LINKED ISSUE #5492: Should remove the "host/" prefix of blob name when the blob container is set to "IsMultiTenant = false"?
https://github.com/abpframework/abp/blob/87cb32955b367a637260545ff697e4e62900f236/framework/src/Volo.Abp.BlobStoring.Azure/Volo/Abp/BlobStoring/Azure/DefaultAzureBlobNameCalculator.cs#L15-L20

https://github.com/abpframework/abp/blob/cdae51ba1aebcf2749375ea0e5d27727613b9051/framework/src/Volo.Abp.BlobStoring/Volo/Abp/BlobStoring/BlobContainer.cs#L206-L214

If a blob name is `something`, and the blob container is set to `IsMultiTenant = false`, the actual blob name will be `host/something`, but I think the blob name should be `something` itself.

maliming:
Resolve #5492

### DIFF
--- a/modules/blob-storing-database/src/Volo.Abp.BlobStoring.Database.Domain/Volo/Abp/BlobStoring/Database/DatabaseBlobNameCalculator.cs
+++ b/modules/blob-storing-database/src/Volo.Abp.BlobStoring.Database.Domain/Volo/Abp/BlobStoring/Database/DatabaseBlobNameCalculator.cs
@@ -1,13 +1,13 @@
 ﻿using Volo.Abp.DependencyInjection;
 using Volo.Abp.MultiTenancy;
 
-namespace Volo.Abp.BlobStoring.Minio
+namespace Volo.Abp.BlobStoring.Database
 {
-    public class DefaultMinioBlobNameCalculator : IMinioBlobNameCalculator, ITransientDependency
+    public class DatabaseBlobNameCalculator : IDatabaseBlobNameCalculator, ITransientDependency
     {
         protected ICurrentTenant CurrentTenant { get; }
 
-        public DefaultMinioBlobNameCalculator(ICurrentTenant currentTenant)
+        public DatabaseBlobNameCalculator(ICurrentTenant currentTenant)
         {
             CurrentTenant = currentTenant;
         }
--- a/modules/blob-storing-database/src/Volo.Abp.BlobStoring.Database.Domain/Volo/Abp/BlobStoring/Database/DatabaseBlobNameCalculator.cs
+++ b/modules/blob-storing-database/src/Volo.Abp.BlobStoring.Database.Domain/Volo/Abp/BlobStoring/Database/DatabaseBlobNameCalculator.cs
@@ -21,7 +21,7 @@ namespace Volo.Abp.BlobStoring.Database
 
             return CurrentTenant.Id == null
                 ? $"host/{args.BlobName}"
-                : $"tenants/{CurrentTenant.Id.Value.ToString("D")}/{args.BlobName}";
+                : $"tenants/{args.BlobName}";
         }
     }
 }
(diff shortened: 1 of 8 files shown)


modules/blob-storing-database/src/Volo.Abp.BlobStoring.Database.Domain/Volo/Abp/BlobStoring/Database/DatabaseBlobNameCalculator.cs
```diff
@@ -0,0 +1,27 @@
...
+            if (!args.Configuration.IsMultiTenant)
+            {
+                return args.BlobName;
+            }
+
+            return CurrentTenant.Id == null
+                ? $"host/{args.BlobName}"
+                : $"tenants/{CurrentTenant.Id.Value.ToString("D")}/{args.BlobName}";
+        }
+    }
```
maliming, line 26: @hikalkan 

We can also add a property to the `DatabaseBlob` instead of change the `Name`. eg  `Shared`.

Before:
-|TenantId|Name
-----|-----|-----
Host|NULL|host/123.jpg
Tenant|1|tenant/123.jpg
Not Multi Tenant|NULL|123.jpg

After:
-|TenantId|Shared|Name
-----|-----|-----|-----
Host|NULL|FALSE|123.jpg
Tenant|1|FALSE|123.jpg
Not Multi Tenant|NULL|TRUE|123.jpg

modules/blob-storing-database/src/Volo.Abp.BlobStoring.Database.Domain/Volo/Abp/BlobStoring/Database/DatabaseBlobNameCalculator.cs
```diff
@@ -0,0 +1,27 @@
...
+        public virtual string Calculate(BlobProviderArgs args)
+        {
+            if (!args.Configuration.IsMultiTenant)
+            {
+                return args.BlobName;
+            }
+
+            return CurrentTenant.Id == null
+                ? $"host/{args.BlobName}"
+                : $"tenants/{CurrentTenant.Id.Value.ToString("D")}/{args.BlobName}";
```
maliming: @hikalkan 

We can also add a property to the `DatabaseBlob` instead of change the `Name`. eg  `Shared`.

Before:
-|TenantId|Name
-----|-----|-----
Host|NULL|host/123.jpg
Tenant|8a5cd295-9697-4d7a-8da1-62b96a6f3c8c|tenant/123.jpg
Not Multi Tenant|NULL|123.jpg

After:
-|TenantId|Shared|Name
-----|-----|-----|-----
Host|NULL|FALSE|123.jpg
Tenant|8a5cd295-9697-4d7a-8da1-62b96a6f3c8c|FALSE|123.jpg
Not Multi Tenant|NULL|TRUE|123.jpg

hikalkan:
This will be a breaking change and it may be especially hard to fix for existing applications those are using some blob providers. I don't see a strong reason to change it. I think there is no important problem to add `host` as a prefix for non multi-tenant applications. I intentionally designed it so that. If there is a bug, we can change it. However, I don't want to say "everyone, please move your blobs since our path calculation has changed" :)

eneskoroglu:
Hi @hikalkan and @maliming ,
I think you can do it with a parameter like **AppendHostNameToBasePath** as existing parameter  **AppendContainerNameToBasePath** . So no one does not have to do any change. 

I am trying to access an existing directory and I do not have permission to add a subdirectory with host name and move the files under to that (host) directory. I can only have read permission on them, so doing this request with a parameter would be great.

PR #5571, opened 2020-09-22
Added IsNumeric

AdemAygun:
Added IsNumeric method in Check.cs

### DIFF
--- a/src/Volo.Abp.Core/Volo/Abp/Check.cs
+++ b/src/Volo.Abp.Core/Volo/Abp/Check.cs
@@ -40,5 +40,15 @@ namespace Volo.Abp
 
             return value;
         }
+        public static string IsNumeric(
+            string value,
+            [InvokerParameterName][NotNull] string parameterName)
+        {
+            if (!value.IsNullOrEmpty() && !double.TryParse(value, out _))
+            {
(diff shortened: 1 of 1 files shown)


src/Volo.Abp.Core/Volo/Abp/Check.cs
```diff
@@ -40,5 +40,15 @@ public static ICollection<T> NotNullOrEmpty<T>(ICollection<T> value, [InvokerPar
 
             return value;
         }
+        public static string IsNumeric(
+            string value,
+            [InvokerParameterName][NotNull] string parameterName)
```
hikalkan, line 45: * You request a `[NotNull]` value but not checking it. I mean this method doesn't throw an exception if parameterName is null.
* To be consistent with the other methods, `value` is a better parameter name.
* Add `[ContractAnnotation("value:null => halt")]` just like the other methods.
* `IsNumeric` naming is not clear enough to understand if it accepts floating point numbers (float, double...) or only accepts whole numbers (int/long). Maybe it can get an optional parameter to allow floating point numbers, which is false by default - so only allows integers by default.

hikalkan (changes requested):

PR #5573, opened 2020-09-22
Use blob storing system to store files

### LINKED ISSUE #4879: Use the BLOB storage module for images/files in the blogging module
- [ ] Add Volo.Abp.BlobStoring package to the blogging project and use it to store and retrieve images.
- [ ] Use Database blob provider for the solutions using the blogging module (blog.abp.io & volosoft blog).
- [ ] We should somehow import the current images to the database for the live web sites (blog.abp.io & volosoft blog). You can write a small console app to run in the server. @yekalkan can help you to obtain all the files as a single zipped folder.

realLiangshiwei:
Resolve https://github.com/abpframework/abp/issues/4879


modules/blogging/src/Volo.Blogging.Application.Contracts.Shared/Volo/Blogging/BloggingFileContainer.cs
```diff
@@ -0,0 +1,10 @@
+﻿using Volo.Abp.BlobStoring;
+
+namespace Volo.Blogging
+{
+    [BlobContainerName("blogging-files")]
+    public class BloggingFileContainer
```
hikalkan: Storing BLOBs/images is an internal detail of the application layer and should not be shared with the clients. So, we should remove any BLOB related stuff from the Application.Contracts and move it to the Application project.

hikalkan (changes requested):

PR #5598, opened 2020-09-24
MVC Open Source Startup Template Home Page

armgnunlu:


armgnunlu:
GitHub Feature Request link is missing
Localizations needed. I did not do this cause it is not final (texts and maybe some contents)
If everything is ok, you can let me know to localize.

PR #5608, opened 2020-09-25
Introduce ITenantConfigurationProvider.

maliming:
Sometimes we want to get tenant info before `MultiTenancyMiddleware`.
For example, getting the tenant's settings during the authentication process.


framework/src/Volo.Abp.MultiTenancy/Volo/Abp/MultiTenancy/TenantConfigurationProvider.cs
```diff
@@ -0,0 +1,58 @@
...
+            ITenantResolver tenantResolver,
+            ITenantStore tenantStore,
+            ITenantResolveResultAccessor tenantResolveResultAccessor)
+        {
+            TenantResolver = tenantResolver;
+            TenantStore = tenantStore;
+            TenantResolveResultAccessor = tenantResolveResultAccessor;
+        }
+
+        public virtual async Task<TenantConfiguration> ResolverAndGet()
```
hikalkan:
```suggestion
        public virtual async Task<TenantConfiguration> GetAsync()
```

framework/src/Volo.Abp.MultiTenancy/Volo/Abp/MultiTenancy/TenantConfigurationProvider.cs
```diff
@@ -0,0 +1,58 @@
...
+        {
+            var resolveResult = TenantResolver.ResolveTenantIdOrName();
+            TenantResolveResultAccessor.Result = resolveResult;
```
hikalkan: The problem of this implementation is that it implicitly sets the `TenantResolveResultAccessor.Result`. This is a side effect since the method is a GET method and overrides a global-like value every time it is called.

hikalkan (commented):
I will investigate this more in the next week.

framework/src/Volo.Abp.MultiTenancy/Volo/Abp/MultiTenancy/TenantConfigurationProvider.cs
maliming: How about adding a `saveResolveResult ` parameter?
```
Task<TenantConfiguration> GetAsync(bool saveResolveResult = false)
```

hikalkan: This would be better, so we can explicitly set it only if we need it.

PR #5614, opened 2020-09-27
fix: ObjectDisposedException

Aprite:
Minio.Examples/Cases/GetObject.cs
https://github.com/minio/minio-dotnet/blob/master/Minio.Examples/Cases/GetObject.cs
问题
@realLiangshiwei


realLiangshiwei:
Hi @Aprite 
We can fix it in 3.2, could you open a new PR in the rel-3.2 branch? thanks.

PR #5615, opened 2020-09-27
fix:ObjectDisposedException

Aprite:
Minio.Examples/Cases/GetObject.cs
https://github.com/minio/minio-dotnet/blob/master/Minio.Examples/Cases/GetObject.cs
问题
@realLiangshiwei


realLiangshiwei:
Thanks a lot @Aprite

PR #5620, opened 2020-09-28
Handle with case issues in json.

### LINKED ISSUE #5618: ABP Translate command didn't work for a regular MVC solution
Just create a new solution and check it.

maliming:
Resolve #5618


maliming:
@hikalkan 

Should we make `culture` and `texts` property always lowercase? We have some json files using uppercase.

```json
{
  "Culture": "en",
  "Texts": {
    "Home": "Home",
  }
}

{
  "culture": "en",
  "texts": {
    "Home": "Home",
  }
}
```

hikalkan:
We should make them lowercase, yes, it would be good if you can fix the Uppercase usages. But supporting both is OK to tolerate errors of other developers.

Issue #5622, opened 2020-09-28
Any IDataSeedContributor should consider the TenantId of the DataSeedContext.


realLiangshiwei:
Related https://github.com/abpframework/abp/pull/5443

We are missing some IDataSeedContributor, like https://github.com/abpframework/abp/blob/dev/templates/app/aspnet-core/src/MyCompanyName.MyProjectName.Domain/IdentityServer/IdentityServerDataSeedContributor.cs

PR #5624, opened 2020-09-28
Change tenant in all derived class of IDataSeedContributor

realLiangshiwei:
Resolve https://github.com/abpframework/abp/issues/5622

### DIFF
--- a/docs/en/Data-Seeding.md
+++ b/docs/en/Data-Seeding.md
@@ -38,31 +38,37 @@ namespace Acme.BookStore
     {
         private readonly IRepository<Book, Guid> _bookRepository;
         private readonly IGuidGenerator _guidGenerator;
+        private readonly ICurrentTenant _currentTenant;
 
         public BookStoreDataSeedContributor(
             IRepository<Book, Guid> bookRepository,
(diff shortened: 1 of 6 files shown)


maliming (approved):

hikalkan:
If we are sure that we always switch to the tenant in the context, why not do it in the DataSeeder ([here](https://github.com/abpframework/abp/blob/45a246d36994974bbe1f5f77d2f03cfa945fd6b3/framework/src/Volo.Abp.Data/Volo/Abp/Data/DataSeeder.cs#L24-L36))?

realLiangshiwei:
Maybe some people directly inject the `IDataSeedContributor` service.  Do we need to consider that?

hikalkan:
>Maybe some people directly inject the IDataSeedContributor service. Do we need to consider that?

We don't need to consider that. This is not the intended use case and they should handle it in this case.

realLiangshiwei:
We can't switch to the tenant in the `DataSeeder`, `Volo.Abp.MultiTenancy` project depends on `Volo.Abp.Data`, it will produce circular dependency.

Issue #4089, opened 2020-05-25
Support virtual path deployment


liangshiw:
Related https://support.abp.io/QA/Questions/181/Problem-with-IIS-application-path

Some people use iis sub-application deployment way, but currently abp does not support this deployment way.

I added sub-application deployment support for the bundled system https://github.com/abpframework/abp/pull/4087, in addition we should always use `asp-page` or `href="~/"`  to generate correct links.

274188A:
fine but not working using latest dev branch...

realLiangshiwei:
@274188A 
Please open a new issue to discuss

274188A:
ok - done.
https://github.com/abpframework/abp/issues/5633

PR #5631, opened 2020-09-29
removed leading "~/"

274188A:
it was causing bug in url


realLiangshiwei:
Hi,
This is not a bug. It is part of the virtual path. see https://github.com/abpframework/abp/issues/4089

image

PR #5681, opened 2020-10-02
Formating settings for ABP solution based on the default C# settings

stsrki:
Added `.editorconfig` to the ABP solution. It's used to have consistent code-style settings accross all projects.

I tested it on multiple `.cs` files and it's working without any modifications on current files except for some smaller and inconsistent spacing on the current code which it's trying to fix. I found this behavior good as it will clean the code in time.

Please test before merging!


hikalkan:
Actually, we should be able to apply this to all solutions. Is there a way to put this to the root folder of the abp repository and use from all the solutions? I don't want to duplicate...

stsrki:
As far as I know, it works only from the solution folder and for per-project, depending on where `.editorconfig` is placed. I don't think it's possible to use it on multiple solutions. Maybe you can try adding it as a file-reference in the other solution? never tried that to be fair...

PR #5683, opened 2020-10-02
Fix: "Options" on a document can be a problem for search engines

### LINKED ISSUE #4877: "Options" on a document can be a problem for search engines
They probably can not index such documents: https://docs.abp.io/en/abp/latest/Tutorials/Part-1?UI=NG&DB=EF
Because there is no link on the web site, but we switch option & refresh page by JavaScript.

You can take support from @armgnunlu

yekalkan:
resolves https://github.com/abpframework/abp/issues/4877

image


hikalkan:
Is google properly works with such hidden content? I found someting on the web, but didn't check deeply.
Can't we check the agent (or any other HTTP header) in the server side to understand if this is a search engine, then render the content in a regular way - not as an hidden element.

PR #5688, opened 2020-10-02
Added distributed Entity Cache base class

### LINKED ISSUE #5668: Distributed Entity Cache base class
ASP.NET Boilerplate has a base class to easily implement caching for entities:
https://aspnetboilerplate.com/Pages/Documents/Caching#entity-caching

We can implement a similar base class for the ABP Framework too. Something like that (can be different, it is a pseudo code):

````csharp
public class PersonCache : EntityCache<Person, PersonCacheItem, TKey>, ITransientDependency
{
    public PersonCache(IDistributedCache<PersonCacheItem> cache, IRepository<Person, TKey> repository)
        : base(cache, repository)
    {

    }
}
````

`Volo.Abp.Ddd.Domain` package seems fine for this implementation, you think on.

### PR DESCRIPTION
**Pull request by realLiangshiwei**
Resolve https://github.com/abpframework/abp/issues/5668

### CODE REVIEW

File: framework/src/Volo.Abp.Ddd.Domain/Volo/Abp/Domain/Entities/Caching/IEntityCache.cs [hunk 1]
```diff
@@ -0,0 +1,19 @@
...
+    public interface IEntityCache<TCacheItem, TKey>
+        where TCacheItem : class
+    {
+        IDistributedCache<TCacheItem> InternalCache { get; }
+
+        TCacheItem this[TKey key] { get; }
+
+        string CacheName { get; }
+
+        TCacheItem Get(TKey key);
```
hikalkan: Don't add sync version. We don't want to use `AsyncHelper.RunSync`. If the developers wants, they can use `AsyncHelper.RunSync` while using the `GetAsync`.

common.DotSettings
```diff
@@ -20,6 +20,7 @@
 	<s:String x:Key="/Default/CodeStyle/Generate/=Overrides/Options/=Async/@EntryIndexedValue">False</s:String>
 	<s:String x:Key="/Default/CodeStyle/Generate/=Overrides/Options/=Mutable/@EntryIndexedValue">False</s:String>
 	<s:String x:Key="/Default/CodeStyle/Naming/CSharpNaming/Abbreviations/=SQL/@EntryIndexedValue">SQL</s:String>
+	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EFeature_002EServices_002EDaemon_002ESettings_002EMigration_002ESwaWarningsModeSettingsMigrate/@EntryIndexedValue">True</s:Boolean>
```
hikalkan: What is this? :)

framework/src/Volo.Abp.Ddd.Domain/Volo/Abp/Domain/Entities/Caching/AbstractKeyMultiTenancyEntityCache.cs
```diff
@@ -0,0 +1,39 @@
+﻿using System;
+using Volo.Abp.Caching;
+using Volo.Abp.Domain.Repositories;
+using Volo.Abp.MultiTenancy;
+
+namespace Volo.Abp.Domain.Entities.Caching
+{
+    public abstract class AbstractKeyMultiTenancyEntityCache<TEntity, TCacheItem, TKey> :
```
hikalkan: No need to such a class. I think this is needed to add tenantid to the cache key. Instead of such a class, this can be done by an if condition (like `if TEntity is IMultiTenant`) in the `AbstractKeyEntityCache.GetCacheKey` method.

framework/src/Volo.Abp.Ddd.Domain/Volo/Abp/Domain/Entities/Caching/AbstractKeyEntityCache.cs
```diff
@@ -0,0 +1,87 @@
...
+
+namespace Volo.Abp.Domain.Entities.Caching
+{
+    public abstract class AbstractKeyEntityCache<TEntity, TCacheItem, TKey> :
+        IDistributedEventHandler<EntityChangedEventData<TEntity>>,
+        IEntityCache<TCacheItem, TKey>
+        where TEntity : class, IEntity
+        where TCacheItem : class
+    {
+        public TCacheItem this[TKey key] => Get(key);
```
hikalkan: Remove the property index accessor, since it needs to a sync method call, then remove the Get method.

framework/src/Volo.Abp.Ddd.Domain/Volo/Abp/Domain/Entities/Caching/AbstractKeyEntityCache.cs
```diff
@@ -0,0 +1,87 @@
...
+
+        public string CacheName { get; }
```
hikalkan: Where the CacheName is used? Cache name should be set with a [CacheName] attribute on the TCacheItem.

framework/src/Volo.Abp.Ddd.Domain/Volo/Abp/Domain/Entities/Caching/AbstractKeyEntityCache.cs
```diff
@@ -0,0 +1,87 @@
...
+
+        public IObjectMapper ObjectMapper { get; set; }
```
hikalkan: Property injecting the `IObjectMapper` is not good. It doesn't cover the scenario where we want to use a [contextualized mapper](https://docs.abp.io/en/abp/latest/Object-To-Object-Mapping#iobjectmapper-interface).

hikalkan (changes requested):

common.DotSettings
realLiangshiwei: This is automatically added by the rider EAP version. I will remove it.

hikalkan:
Hi @realLiangshiwei there are a lot of problems with the implementation, sorry :)
I started to refactor, but then I give up since it seems I need to change almost everyting.
So, I am closing this. I will implement it in a later version. Once I finish the implementation you compare them to see what I mean :)

realLiangshiwei:
OK

PR #5694, opened 2020-10-03
Add appsettings to enable/disable Redis as distributed cache

RachaneeSaeng:
To allow switching between In-Memory and Redis as a distributed cache.
Background: sometimes we want to stop using Redis (on production) for some reason while don't want to rebuilt and redeploy the app.

### DIFF
--- a/docs/en/Redis-Cache.md
+++ b/docs/en/Redis-Cache.md
@@ -25,9 +25,11 @@ Volo.Abp.Caching.StackExchangeRedis package automatically gets the redis [config
 
 ````js
 "Redis": { 
+ "IsEnabled": "true",
  "Configuration": "127.0.0.1"
 }
 ````
+The setting `IsEnabled` is optional and will be considered `true` if it is not set.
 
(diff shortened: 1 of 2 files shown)


hikalkan (changes requested):
Thank you for your contribution. However, this is an unnecessary **breaking change** with the current implementation. `IsEnabled` option should be optional and considered `true` by default for backward compability.

rachanee-mwp:
@hikalkan I have updated code to allow the setting `Redis:IsEnabled` to be optional and consider `true` of it is not set.
However, the step `Code scanning results / CodeQL` was failed with some irrelevant errors.
Could you please review?

hikalkan:
Thanks for the changes.

I believe we should only make the changes in the AbpCachingStackExchangeRedisModule.cs and revert all the other changes (except the documentation).

Because these changes (adding Redis:IsEnabled as false to appsettings.json) changes the behaviour. Using redis in a distributed environment is essential and this change makes it easy to forget it as disabled. Previos code was disabling in only for development environment.

If you need to disable redis, you should do it in your own application. We don't want to add this to the startup templates.

Thanks for your contribution.

rachanee-mwp:
@hikalkan I've reverted changes from other files except the `AbpCachingStackExchangeRedisModule` and the documentation.
Could you please review?

hikalkan:
Great! Thanks :)

ahmadalmunajjed:
I think in the documentation of Redis-Cache it still mentions IsEnabled param even though it's no implemented anymore. So I suggest removing it from the documentation.

PR #5696, opened 2020-10-04
Update locale of zh-hant 

ChangYinShung:


maliming:
Thanks @ChangYinShung

PR #5713, opened 2020-10-06
Update Entity-Framework-Core-SQLite.md

paragpkulkarni:
Added sample connection string for SQLite


maliming:
Thanks @paragpkulkarni

PR #5728, opened 2020-10-07
Auto Validate AntiForgery Token for HTTP API Requests

hikalkan:
While AspNet Core automatically validates anti forgery token and handles the CSRF problem, it does nothing for the API Controllers by default.

If you use [AutoValidateAntiforgeryTokenAttribute](https://github.com/dotnet/aspnetcore/blob/master/src/Mvc/Mvc.ViewFeatures/src/AutoValidateAntiforgeryTokenAttribute.cs) you solve the problem. But then you have another problem: No client can consume your APIs without sending antiforgery tokens. This is a problem when your client is a non-browser client and typically don't need to an antiforgery system at all. 

I [had discussed](https://github.com/aspnet/Antiforgery/issues/115) that with the AspNet team before. However there is no clear built-in solution. We [had implemented](https://aspnetboilerplate.com/Pages/Documents/XSRF-CSRF-Protection) a system for the ASP.NET Boilerplate before. So, I am implementing a similar system for the ABP Framework too. The purpose is to protect all APIs with zero config and also allows non-browser clients to consume your APIs without dealing this problem.

I will document it & explain the configuration options.


hikalkan:
Documented: https://github.com/abpframework/abp/blob/dev/docs/en/CSRF-Anti-Forgery.md

jack-gaojz:
Today I have the same issue after upgrade ABP 3.0.5 to ABP 3.3.2.
The ABP 3.0.5 works fine, but the ABP 3.3.2 failed. Not sure how to fix it. Anyone can help me? :(
Do I need to add some new configurations to resolve it? Thanks.

```
2021-01-09 09:46:02.436 +08:00 [INF] Request starting HTTP/1.1 POST http://localhost:5000/api/app/Visualizer/visualWidget/importDefaultWidgets?overwrite=true application/json 0
2021-01-09 09:46:02.447 +08:00 [DBG] PermissionStore.GetCacheItemAsync: pn:U,pk:39f9f4bc-b2e5-0a6a-600f-ce21b280f935,n:Gms.Visualizer.Web.VisualWidget
2021-01-09 09:46:02.447 +08:00 [DBG] Found in the cache: pn:U,pk:39f9f4bc-b2e5-0a6a-600f-ce21b280f935,n:Gms.Visualizer.Web.VisualWidget
2021-01-09 09:46:02.447 +08:00 [DBG] PermissionStore.GetCacheItemAsync: pn:R,pk:admin,n:Gms.Visualizer.Web.VisualWidget
2021-01-09 09:46:02.447 +08:00 [DBG] Found in the cache: pn:R,pk:admin,n:Gms.Visualizer.Web.VisualWidget
2021-01-09 09:46:02.447 +08:00 [INF] Authorization was successful.
2021-01-09 09:46:02.448 +08:00 [INF] No CORS policy found for the specified request.
2021-01-09 09:46:02.449 +08:00 [INF] Executing endpoint 'Dxp.Gms.Visualizer.VisualWidgetAppService.ImportDefaultWidgetsAsync (Dxp.Gms.Visualizer.Application)'
2021-01-09 09:46:02.450 +08:00 [INF] Route matched with {action = "ImportDefaultWidgets", controller = "VisualWidget", area = "", page = ""}. Executing controller action with signature System.Threading.Tasks.Task ImportDefaultWidgetsAsync(Boolean) on controller Dxp.Gms.Visualizer.VisualWidgetAppService (Dxp.Gms.Visualizer.Application).
**2021-01-09 09:46:02.451 +08:00 [ERR] The required antiforgery header value "RequestVerificationToken" is not present.**
2021-01-09 09:46:02.451 +08:00 [INF] Authorization failed for the request at filter 'Volo.Abp.AspNetCore.Mvc.AntiForgery.AbpAutoValidateAntiforgeryTokenAuthorizationFilter'.
2021-01-09 09:46:02.451 +08:00 [INF] Executing HttpStatusCodeResult, setting HTTP status code 400
2021-01-09 09:46:02.451 +08:00 [INF] Executed action Dxp.Gms.Visualizer.VisualWidgetAppService.ImportDefaultWidgetsAsync (Dxp.Gms.Visualizer.Application) in 0.6896ms
```

alexandis:
@jack-gaojz 

ABP changed something lately, but adding _AntiForgery_ token supposed to be working out of the box, if request is set from a trusted source. I was able to resolve this for published site once I've changed to relative paths for my endpoints which front-end accesses (**environment.ts** settings) - there is some bug in github showing ABP source code and explaining why it needs to be set this way now:

```
apis: {
    default: {
      url: '/httpapihost' // not full address like https://..../httpapihost
    },
    ...
}
```

But now I see it is still broken for localhost, since localhost endpoints differ by ports, not path names. It is not working when I am running Angular app via https. It does work for http. Probably there are still some other nuances, I will have to investigate this.

eemogz:
I have now taken the abp framework api live, I am sending a request to the api from local with the angular application, but I cannot log in, I get an http 400 error, I do not know if I will experience this situation when I take the angular application live ? Unfortunately, I could not understand what I should do about the subject. Can you help me?

maliming:
hi @eemogz 

Please create a new issue and share the steps.

Thanks.

HDaoud:
@maliming did @eemogz create a new issue? Is my issue related to this issue? I have /api in the environment apis url. and our auth server is /auth. (unlike @eemogz we only have the issue when we go live, not locally)

Below are the details with a video showing what is going on. The first save request causes a redirect to 400 error, subsequent requests are accepted. Is this a bug that need to be fixed by abp team, or is there something we should try to do to resolve the issue?
https://github.com/abpframework/abp/issues/20945#issuecomment-2536722368

PR #5733, opened 2020-10-07
Support to return Stream from application services & handle on the Dynamic C# proxying

### LINKED ISSUE #5727: Allow streaming from application services
reported at https://support.abp.io/QA/Questions/463/Blob-storage-support-for-large-files-with-no-overhead

alexandru-bagu:
Resolved #5727

Usage example can be found here: https://github.com/alexandru-bagu/abp-remote-stream-test


hikalkan:
Thanks a lot. I will deeply review this and comment on.

hikalkan:
Your implementation seems pretty good. I will test myself, maybe a little refactor, then I will merge.

Rwing:
@alexandru-bagu Excuse me,  how to get the file name in this feature?

maliming:
@Rwing Please create a new issue.

alexandru-bagu:
> @alexandru-bagu Excuse me, how to get the file name in this feature?

a stream does now have a name as it's supposed to be just a string of bytes. you can include the name as an additional parameter to the app service method.
you will have to add the parameter as a route parameter because the call may contain only one complex type to be sent as body (the IRemoteStreamContent)

Issue #5551, opened 2020-09-21
Blazor: Complete the IUiMessageService


hikalkan:
Implement using custom later. We can try to use sweetalert

stsrki:
Reopened because there are more things to do.

PR #5740, opened 2020-10-08
Blazor message refactor

stsrki:
#5551 

- Removed UiMessageNotifierService
- Use of Action for setting the message options
- Localization
- Add CenterMessage to options


hikalkan:
It would be better to allow to set the options in a global way, so people can set the options globally, then override per method call.
I've created an issue in the backlog: https://github.com/abpframework/abp/issues/5742 We do it later.

PR #5760, opened 2020-10-11
hu translation

lszabados:
I make translations of the framework and modules into Hungarian.


maliming:
Thanks @lszabados 

You can also add `hu` language in template projects if you like.

https://github.com/abpframework/abp/search?q=options.Languages.Add%28new+LanguageInfo%28%22tr%22%2C+%22tr%22%2C+%22T%C3%BCrk%C3%A7e%22%29%29%3B

maliming:
👍

Issue #5288, opened 2020-09-04
Consider supporting Automapper dependency injection


realLiangshiwei:
Currently, we create Automapper when the application is initialized, But this way we cannot use Automapper's dependency injection function, like: https://docs.automapper.org/en/stable/Before-and-after-map-actions.html#asp-net-core-and-automapper-extensions-microsoft-dependencyinjection

https://github.com/abpframework/abp/blob/a130e0db8cc91d049be126d10bb9b11f0bf51cf1/framework/src/Volo.Abp.AutoMapper/Volo/Abp/AutoMapper/AbpAutoMapperModule.cs#L30.

We can consider creating Automapper in advance to get dependency injection. example:

```csharp
public override void PostConfigureServices(ServiceConfigurationContext context)
{
    var mapperAccessor = new MapperAccessor();

    context.Services.AddSingleton<MapperAccessor>(_ => mapperAccessor);
    context.Services.Add(new ServiceDescriptor(typeof(IMapperAccessor), CreateMappings, ServiceLifetime.Singleton));
}

private IMapperAccessor CreateMappings(IServiceProvider serviceProvider)
{
    var options = serviceProvider.GetRequiredService<IOptions<AbpAutoMapperOptions>>().Value;

    void ConfigureAll(IAbpAutoMapperConfigurationContext ctx)
    {
        foreach (var configurator in options.Configurators)
        {
            configurator(ctx);
        }
    }

    void ValidateAll(IConfigurationProvider config)
    {
        foreach (var profileType in options.ValidatingProfiles)
        {
            config.AssertConfigurationIsValid(((Profile) Activator.CreateInstance(profileType)).ProfileName);
        }
    }

    using (var scope = serviceProvider.CreateScope())
    {
        var mapperConfiguration = new MapperConfiguration(mapperConfigurationExpression =>
        {
            ConfigureAll(new AbpAutoMapperConfigurationContext(mapperConfigurationExpression,
                scope.ServiceProvider));
        });

        ValidateAll(mapperConfiguration);

        var mapperAccessor = serviceProvider.GetService<MapperAccessor>();
        mapperAccessor.Mapper = new Mapper(mapperConfiguration, serviceProvider.GetService);

        return mapperAccessor;
    }
}
```

![image](https://user-images.githubusercontent.com/16813853/92194856-39c9ee80-ee9e-11ea-86f3-fc3bfebaa7ba.png)

realLiangshiwei:
[...]

PR #5265, opened 2020-09-02
Add multi-lingual entities

realLiangshiwei:
Resolve https://github.com/abpframework/abp/issues/1754

PR #5774, opened 2020-10-12
Tntwist/multi lingual entities

tntwist:
Hi @realLiangshiwei and @hikalkan,

I solved the breaking index for the language property of the translations and added AutoMapper-Functionalities for the MultiLingual-Entities. I packed the AutoMapper code in the MultiLingualObject Module since these extensions are not relevant for devs who don`t need the MultiLingualEntities.

It would be nice if you review these changes.

Greetings


realLiangshiwei:
Hi,

Thanks for your contribution,  But I think there are the following problems

1. Your PR contains a lot of unwanted submissions. 
2. ABP abstracts object mapping and multilingual entity project should not depend on the Automapper project.  Actually, I considered your implementation, but I didn’t decide to do it.
3. I think this way has some problems, as you see, the unit tests get failed. We are considering a better implementation: https://github.com/abpframework/abp/issues/5288

     ```csharp
            {
                var sp = context.Services.BuildServiceProvider();
                options.Configurators.Add(ct => ct.MapperConfiguration.ConstructServicesUsing(sp.GetService));
            });
     ```

So, I will close the PR, You can write your suggestion here https://github.com/abpframework/abp/pull/5265.

PR #5827, opened 2020-10-16
Add "GetAll" method to ISettingDefinitionContext

### LINKED ISSUE #5826: interface  "ISettingDefinitionContext" should expose a "GetAll" method
At present, the interface ["ISettingeDefinitionContext"](https://github.com/abpframework/abp/blob/dev/framework/src/Volo.Abp.Settings/Volo/Abp/Settings/ISettingDefinitionContext.cs) has two method named "GetOrNull" and "Add", while its implementation class [SettingDefinitionContext](https://github.com/abpframework/abp/blob/dev/framework/src/Volo.Abp.Settings/Volo/Abp/Settings/SettingDefinitionContext.cs#L20) has another method named "GetAll" apart from the two.

I think it's a common demand that we change setting definitions of a depended module one by one. If the interface exposes the "GetAll" method, we can change it in a batch mode.

taujiong:
Its implementation class already has the implementation method for "GetAll"

This pr fix #5826

### DIFF
--- a/framework/src/Volo.Abp.Settings/Volo/Abp/Settings/ISettingDefinitionContext.cs
+++ b/framework/src/Volo.Abp.Settings/Volo/Abp/Settings/ISettingDefinitionContext.cs
@@ -3,6 +3,8 @@
     public interface ISettingDefinitionContext
     {
         SettingDefinition GetOrNull(string name);
+        
+        IReadOnlyList<SettingDefinition> GetAll();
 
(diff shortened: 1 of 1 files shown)


realLiangshiwei:
Hi, the CI tests get failed.

taujiong:
sorry for that, I will fix it

maliming (approved):

PR #5828, opened 2020-10-16
Cookie service implementation

### LINKED ISSUE #5731: Create an ICookieService to get/set cookies in Blazor webassembly
We need methods like https://github.com/abpframework/abp/blob/dev/npm/packs/core/src/abp.js#L614-L685

ilkayilknur:
Closes #5731


templates/app/aspnet-core/src/MyCompanyName.MyProjectName.Blazor/wwwroot/index.html
```diff
@@ -29,6 +29,7 @@
     <script src="_content/Blazorise/blazorise.js"></script>
     <script src="_content/Blazorise.Bootstrap/blazorise.bootstrap.js"></script>
     <script src="_content/Volo.Abp.AspNetCore.Components.WebAssembly.BasicTheme/theme.js"></script>
+    <script src="libs/abp/core/abp.js"></script>
```
hikalkan: I think it would be better to move the abp.js to `_content/Volo.Abp.AspNetCore.Components.WebAssembly/abp.js`. In this way we don't need to copy it to the startup template. Just check how we've done it for `Volo.Abp.AspNetCore.Components.WebAssembly.BasicTheme/theme.js`.

hikalkan (changes requested):

templates/app/aspnet-core/src/MyCompanyName.MyProjectName.Blazor/wwwroot/index.html
ilkayilknur: Thanks for the suggestion. This way looks cleaner. I made the changes. I had to change the project type of the Volo.Abp.AspNetCore.Components.WebAssembly from Microsoft.NET.Sdk to Microsoft.NET.Sdk.Razor to enable static asset delivery.

PR #5836, opened 2020-10-16
Page Header Implementation

ilkayilknur:
I've completed the initial implementation of the PageHeader component.  https://github.com/volosoft/volo/issues/3480

Sample Usage 
``` 
<Volo.Abp.BlazoriseUI.Components.PageHeader Title="@L["Languages"]" @ref="Header">
    @if (HasCreatePermission)
    {
        <Button Color="Color.Primary" Clicked="OpenCreateModalAsync">
            <i class="mr-1 fa fa-plus"></i>@L["CreateNewLanguage"]
        </Button>
    }
</Volo.Abp.BlazoriseUI.Components.PageHeader>

protected override async Task OnInitializedAsync()
{
    await base.OnInitializedAsync();
    Header.AddBreadcrumbItem(L["LanguageManagement"]);
    Header.AddBreadcrumbItem(L["Languages"], NavigationManager.ToBaseRelativePath(NavigationManager.Uri));
} 
```

I want to get your feedback about the implementation and usage. After that, I will start changing module headers.


framework/src/Volo.Abp.BlazoriseUI/Components/PageHeader.razor
```diff
@@ -0,0 +1,35 @@
+﻿<Row Class="entry-row">
+    <Column ColumnSize="ColumnSize.IsAuto">
+        <h1 class="content-header-title">@Title</h1>
```
stsrki, line 3: You can use `Heading` component instead of `<h1>` element

framework/src/Volo.Abp.BlazoriseUI/Components/PageHeader.razor
```diff
@@ -0,0 +1,35 @@
...
+    </Column>
+    @if (BreadcrumbItems.Any())
+    {
+        <Column ColumnSize="ColumnSize.IsAuto.OnWidescreen" Class="pl-lg-0">
```
stsrki: Instead of using class "pl-lg-0", I would write `Padding="Padding.Is0.FromLeft.OnLarge"`. In fact, the best practice would be to use as many Blazorise utilities as possible

framework/src/Volo.Abp.BlazoriseUI/Components/PageHeader.razor.cs
```diff
@@ -0,0 +1,34 @@
...
+        [Parameter]
+        public bool BreadcrumbShowHome { get; set; } = true;
+
+        [Parameter]
+        public bool BreadcrumbShowCurrent { get; set; } = true;
+
+        [Parameter]
+        public RenderFragment ChildContent { get; set; }
+
+        protected List<BreadcrumbItem> BreadcrumbItems { get; set; }
```
stsrki: Maybe also have a `[Parameter]` for this so it can be set declaratively through razor syntax.

framework/src/Volo.Abp.BlazoriseUI/Components/PageHeader.razor
```diff
@@ -0,0 +1,35 @@
...
+            <Breadcrumb Mode="@(BreadcrumbShowCurrent ? BreadcrumbMode.Auto : BreadcrumbMode.None)">
+                @if (BreadcrumbShowHome)
+                {
+                    <BreadcrumbItem>
+                        <BreadcrumbLink To="/"><i class="fa fa-home"></i></BreadcrumbLink>
```
stsrki: Blazorise has it's own Icon component, so instead of `<i class="fa fa-home"></i>` write `<Icon Name="IconName.Home" />`

framework/src/Volo.Abp.BlazoriseUI/Components/PageHeader.razor
ilkayilknur: Yes. I used the `Heading `component but had to change back to `<h1>` element. Heading component adds "h1" class while rendering  the `<h1>` element and it breaks the lepton-theme design.

framework/src/Volo.Abp.BlazoriseUI/Components/PageHeader.razor
ilkayilknur: Thanks. I will move the BreadcrumbItem class to the BlazoriseUI assembly. So that developers can use Blazorise IconName enum.

framework/src/Volo.Abp.BlazoriseUI/Components/PageHeader.razor
ilkayilknur: Done.

framework/src/Volo.Abp.BlazoriseUI/Components/PageHeader.razor
stsrki: I will probably need to fix that in Blazorise

framework/src/Volo.Abp.BlazoriseUI/Components/PageHeader.razor.cs
ilkayilknur: Done.

framework/src/Volo.Abp.BlazoriseUI/BreadcrumbItem.cs
```diff
@@ -1,14 +1,16 @@
...
+﻿using Blazorise;
+
+namespace Volo.Abp.BlazoriseUI
 {
     public class BreadcrumbItem
     {
         public string Text { get; set; }
 
-        public string Icon { get; set; }
+        public IconName? Icon { get; set; }
```
stsrki: It would be better to have as an `object`  in case someone wants to set a name explicitly, like "fa-home". Blazorise supports this scenario and will apply the right name be it from enum or a string name.

ilkayilknur: Done.

ilkayilknur:
The PR is ready for review @hikalkan @stsrki

PR #5839, opened 2020-10-16
Fixed generate component command

kishoresahas:
yarn ng generate component my-layout


mehmet-erim:
Hi @kishoresahas,

This command is correct. There is a mistake in the following sentence. I will be fixed it.
Thanks for the contribution but I must close the PR.

PR #5847, opened 2020-10-18
correct exception message in PermissionAppService.CheckProviderPolicy

taujiong:
change 'policyName' to 'providerName' in the exception message

Since 'policyName' is null or empty, the exception message should always be `No policy defined to get/set permissions for the provider ''. Use PermissionManagementOptions to map the policy.`

I guess the param 'policyName' should be replaced with 'providerName'


maliming:
Thanks @taujiong

Issue #5325, opened 2020-09-07
Create auto controller routes in kebap-case


ebicoglu:
The current implementation uses camelCase.
Change this behavior to kebap-case-plural or enable users to set the endpoint manually, (because plural of a word can be hard and needs a library to find.)

![image](https://user-images.githubusercontent.com/9526587/92391655-bde1d780-f125-11ea-866f-493320219c9c.png)

PMExtra:
For now, you can do that:
In `src/{Namespace}.{Project}.HttpApi.Host/{Project}HttpApiHostModule`:
```cs
private void ConfigureConventionalControllers()
{
    Configure<AbpAspNetCoreMvcOptions>(options =>
    {
        options.ConventionalControllers.Create(typeof({Project}ApplicationModule).Assembly, setting =>
        {
            setting.UrlControllerNameNormalizer = context => context.ControllerName?.Pluralize().Kebaberize();
            setting.UrlActionNameNormalizer = context => context.ActionNameInUrl?.Kebaberize();
        });
    });
}
```

Edited:

`Pluralize()` and `Kebaberize()` are both from https://github.com/Humanizr/Humanizer

If you use underline naming style you may need also `Dasherize()`.

Thank @maliming for recommending Normalizer

maliming:
It is better to use https://github.com/Humanizr/Humanizer  to convert the name. As far as I know, many libraries such as EF Core use it.

PMExtra:
@maliming Great! I got that from https://github.com/dotnet/EntityFramework.Docs/blob/master/entity-framework/core/what-is-new/ef-core-5.0/breaking-changes.md#new-behavior-12 :
> EF Core now includes a pluralizer that uses the Humanizer library. This is the same library Visual Studio uses to recommend variable names.

hikalkan:
This is done. You can see the migration guide: https://github.com/abpframework/abp/blob/dev/docs/en/Migration-Guides/Abp-4_0.md#auto-api-controller-route-changes

I added an option for who want to continue with the version 3.x approach since this breaking change may effect their client applications.

OLD STYLE:

![image](https://user-images.githubusercontent.com/1210527/97086966-4b727b80-162f-11eb-9be3-d486f473164f.png)

NEW STYLE:

![image](https://user-images.githubusercontent.com/1210527/97086962-47465e00-162f-11eb-9c6e-5a987c873b7e.png)

[...]

PR #5866, opened 2020-10-20
Unify route format: checkPassword -> check-password.

PMExtra:
I found abp usually uses dash-separated route format. Such as:
- /api/abp/api-definition
- /api/abp/application-configuration
- /api/abp/multi-tenancy/tenants/by-name/{name}
- /api/identity/my-profile/change-password

But there is a exception that uses camelCase:
- /api/account/checkPassword

I want to unify them and keep the old route for compatible. Please check the changes.

### DIFF
--- a/modules/account/src/Volo.Abp.Account.Web/Areas/Account/Controllers/AccountController.cs
+++ b/modules/account/src/Volo.Abp.Account.Web/Areas/Account/Controllers/AccountController.cs
@@ -83,6 +83,14 @@ namespace Volo.Abp.Account.Web.Areas.Account.Controllers
 
         [HttpPost]
         [Route("checkPassword")]
+        [ApiExplorerSettings(IgnoreApi = true)]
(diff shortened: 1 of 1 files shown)


PMExtra:
You may want to mark this old route was obsolete / deprecated and remove it in next major version.

maliming:
@hikalkan What do you think?

PMExtra:
I saw you marked this `4.0-preview`.
For now, it's not a broken change, the old api is still work. I want it could be published as soon as possible.
Then you can consider to remove the old api from `4.0`.

hikalkan:
Actually, this is good. We also have such an issue: https://github.com/abpframework/abp/issues/5325
I am planning to work on this to have a consistent URL structure.

PMExtra:
@hikalkan I think it should be noticed that this is not a breaking change. I still keep the old API Endpoint and just hide it from Swagger. If you want to complete it in one step, you can remove the old API endpoint.

PR #5883, opened 2020-10-21
Blazor client-side validation

### LINKED ISSUE #5524: Interactive Validation model for the Blazor UI
Can we automate or simplify the validation stuff, with localization?

* Should support data annotation validations
* Should support IValidatableObject
* Should support FluentValidation (with low priority and we can create a separate issue for that)

Initially, we can implement the validation for the user create and edit modals of the identity module and then we discuss on it (before deciding to implement for other modules).

### LINKED ISSUE #5915: Create additional before and after methods for AbpCrudPageBase operations

stsrki:
Do not merge! This is still a work in progress.

- Added client-side validation to roles and users
- Localization of error messages
- Localization of arguments for the error message
- Publish new Blazorise and update APB project references


stsrki:
@hikalkan Please review this PR once you have time. I think it should be ready now.

We still need to do FluentValidation but that can be done later.

stsrki:
@hikalkan Please review when you're ready?

Validation is now working as expected. Also, I have merged and fixed conflicts with dev branch.

framework/src/Volo.Abp.BlazoriseUI/AbpCrudPageBase.cs
```diff
@@ -279,10 +281,31 @@ protected virtual async Task OnDataGridReadAsync(DataGridReadDataEventArgs<TList
...
 
             NewEntity = new TCreateViewModel();
+
+            // Mapper will not notify Blazor that binded values are changed
+            // so we need to notify it manually by calling StateHasChanged
+            await InvokeAsync(() => StateHasChanged());
+
             CreateModal.Show();
+
+            await OnOpeningCreateModalAsync();
```
hikalkan:
```suggestion
            await OnOpenedCreateModalAsync();
```

But I think there is no point to add these methods now. I can override the `OpenCreateModalAsync` and make anything before and after it. Just remove them please.

framework/src/Volo.Abp.BlazoriseUI/AbpCrudPageBase.cs
```diff
@@ -293,12 +316,32 @@ protected virtual Task CloseCreateModalAsync()
...
             EditingEntity = MapToEditingEntity(entityDto);
+
+            await InvokeAsync(() => StateHasChanged());
+
             EditModal.Show();
+
+            await OnOpenedEditModalAsync(id);
+        }
+
+        protected virtual Task OnOpeningEditModalAsync(TKey id)
```
hikalkan, line 328: Same.. we can remove these methods.

framework/src/Volo.Abp.Core/System/AbpObjectExtensions.cs
```diff
@@ -105,5 +106,62 @@ public static T If<T>(this T obj, bool condition, Action<T> action)
...
         }
+
+        /// <summary>
+        /// Resets all fields of a given object to their default values.
+        /// </summary>
+        /// <typeparam name="T">Type of the object.</typeparam>
+        /// <param name="obj">An object</param>
+        /// <param name="customReset">Alternative custom method to reset the values.</param>
+        /// <returns>Returns the original object.</returns>
+        public static T ResetState<T>(this T obj, Action<T> customReset = null)
```
hikalkan: If we don't use this method anymore, please delete it.

framework/src/Volo.Abp.Core/System/AbpObjectExtensions.cs
```diff
@@ -105,5 +106,62 @@ public static T If<T>(this T obj, bool condition, Action<T> action)
...
+        }
+
+        /// <summary>
+        /// Applies all fields of a given object from it's counterpart.
+        /// </summary>
+        /// <typeparam name="T">Type of the object.</typeparam>
+        /// <param name="obj">An object.</param>
+        /// <param name="other">An object from which we copy the values.</param>
+        /// <returns>Returns the original object.</returns>
+        public static T ApplyState<T>(this T obj, T other)
```
hikalkan: If we don't use this method anymore, please delete it.

modules/identity/src/Volo.Abp.Identity.Blazor/AbpIdentityBlazorMessageLocalizerExtensions.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using Blazorise;
+using Microsoft.Extensions.Localization;
+
+namespace Volo.Abp.Identity.Blazor
+{
+    public static class AbpIdentityBlazorMessageLocalizerExtensions
+    {
+        public static string Localize<T>(this IStringLocalizer<T> stringLocalizer, string format, IEnumerable<string> arguments)
```
hikalkan: This method name is too general and it is especially dengerous when it is an extension method.
I suggest to not make this an extension method and change the class name as `AbpIdentityBlazorMessageLocalizerHelper` or something like that. If it is only for displayname style localization, the method name can indicate that.

modules/identity/src/Volo.Abp.Identity.Blazor/AbpIdentityBlazorMessageLocalizerExtensions.cs
```diff
@@ -0,0 +1,34 @@
...
+        {
+            try
+            {
+                return arguments?.Count() > 0
+                    ? string.Format(stringLocalizer[format], LocalizeMessageArguments(stringLocalizer, arguments)?.ToArray())
+                    : stringLocalizer[format];
```
hikalkan: `stringLocalizer[...]` has already an overload like `stringLocalizer[format, arguments]`. Why don't you use it?

Also, `IEnumerable<string> arguments` parameter can be made optional if it allows nulls, like `[CanBeNull] IEnumerable<string> arguments = null`.

Finally, as I understand, `format` can be a simple string that has no format arguments. In this case, a `message` parameter name would be better.

modules/identity/src/Volo.Abp.Identity.Blazor/Pages/Identity/RoleManagement.razor
```diff
@@ -1,17 +1,19 @@
 ﻿@page "/identity/roles"
-@attribute [Authorize(IdentityPermissions.Roles.Default)]
+@attribute [Authorize( IdentityPermissions.Roles.Default )]
```
hikalkan, line 2: Don't use extra spaces here, please change the editor formatting config :)

modules/identity/src/Volo.Abp.Identity.Blazor/Pages/Identity/RoleManagement.razor
```diff
@@ -28,7 +30,7 @@
           ShowPager="true"
           PageSize="PageSize">
     <DataGridColumns>
-        @if (ShouldShowEntityActions)
+        @if ( ShouldShowEntityActions )
```
hikalkan, line 34: Again, we don't use extra spaces. Please change the editor settings. Otherwise, we see a lot of unnecessary changes whenever you touch a code file.

hikalkan (changes requested):

modules/identity/src/Volo.Abp.Identity.Blazor/Pages/Identity/RoleManagement.razor
stsrki: Fixed

framework/src/Volo.Abp.BlazoriseUI/AbpCrudPageBase.cs
stsrki: I removed OnOpened and leave only OnOpening methods since we're using those

framework/src/Volo.Abp.Core/System/AbpObjectExtensions.cs
stsrki: Removed

framework/src/Volo.Abp.BlazoriseUI/AbpCrudPageBase.cs
stsrki: I would leave OnOpeningEditModalAsync because we use it

framework/src/Volo.Abp.Core/System/AbpObjectExtensions.cs
stsrki: Removed

modules/identity/src/Volo.Abp.Identity.Blazor/AbpIdentityBlazorMessageLocalizerExtensions.cs
stsrki: The reason why I wanted it as an extension method id because it's easier to write `L.Localize` then something like `BlazorMessageLocalizer.Localize(stringLocalizer, message, args)`

modules/identity/src/Volo.Abp.Identity.Blazor/AbpIdentityBlazorMessageLocalizerExtensions.cs
stsrki: I have now used `stringLocalizer[format, arguments]`. And the reason for all of this code is because I need to localize arguments along with messages because they have "fallback" logic.

```
return stringLocalizer[$"DisplayName:{argument}"]
  ?? stringLocalizer[argument]
  ?? argument;
```

modules/identity/src/Volo.Abp.Identity.Blazor/AbpIdentityBlazorMessageLocalizerExtensions.cs
stsrki: Created AbpIdentityBlazorMessageLocalizerHelper instead of extension methods

PR #5890, opened 2020-10-21
Blazor UI Entity actions component

### LINKED ISSUE #5560: Blazor: Entity Actions component
We add actions to the DataGrid almost in all pages. It causes repeating code. We can create a component that we can simply add the actions column.

image

Features;

* Allows to set a policy (permission) for each action and hide the action if no permission
* If none of the action policies was granted, the column should be completely invisible
* If only one action available (including non-authorized actions) then show a button, instead of dropdown.
* Allow to set an action as "primary". In this case, the dropdown becomes a split button:
image

Note that: We should allow to reuse the same dropdown everywhere, independent from the data grid as well as inside a data grid.

ilkayilknur:
Closes #5560
This PR introduces the EntityActions component to handle entity actions both inside and outside of the DataGrid.

Sample Actions Column Usage
```
<DataGridColumn Width="150px"
                Displayable="@ShouldShowEntityActions"
                TItem="LanguageDto"
                Field="@nameof(LanguageDto.DisplayName)"
                Sortable="false"
                Caption="@L["Actions"].Value">
    <DisplayTemplate>
        <EntityActions ToggleColor="Color.Primary"
                        Entity="@context"
                        TItem="LanguageDto"
                        ToggleText="@L["Actions"]">
            <EntityAction IsVisible="@HasUpdatePermission"
                            Clicked="(entity) => OpenEditModalAsync(entity.Id)"
                            TItem="LanguageDto"
                            Text="@L["Edit"]" />

            <EntityAction IsVisible="@HasChangeDefaultPermission"
                            TItem="LanguageDto"
                            Text="@L["SetAsDefaultLanguage"]"
                            Clicked="(entity) =>SetAsDefaultAsync(entity.Id)" />

            <EntityAction IsVisible="@HasDeletePermission"
                            TItem="LanguageDto"
                            Text="@L["Delete"]"
                            Clicked="(entity) => DeleteEntityAsync(entity)" />
        </EntityActions>
    </DisplayTemplate>
</DataGridColumn>
```
You can set the `Primary` property to True on the EntityActions component to render the dropdown as a split dropdown.


hikalkan:
Can we do it like that:

````xml
<EntityActionsDataGridColumn>
    <EntityAction
        Text = "@L["Edit"]"
        RequiredPolicy="@IdentityPermissions.Roles.Update"
        Clicked="() => OpenEditModalAsync(context.As<IdentityRoleDto>().Id)" />
    <EntityAction
        Text = "@L["Permissions"]"
        RequiredPolicy="@IdentityPermissions.Roles.ManagePermissions"
        Clicked="() => PermissionManagementModal.OpenAsync(PermissionProviderName, context.As<IdentityRoleDto>().Name)" />
    <EntityAction
        Text = "@L["Delete"]"
        RequiredPolicy="@IdentityPermissions.Roles.Delete"
        ConfirmationMessage="@L["TheRole{0}WillBeDeleted", context.As<IdentityRoleDto>().Name]"
        Clicked="() => DeleteEntityAsync(context.As<IdentityRoleDto>())" />
</EntityActionsDataGridColumn>
````

**Comment by ilkayilknur** [comment]
@stsrki Is it possible to develop a custom Blazorise DataGridColumn like in the previous comment right now? I know that DataGridColumn must register itself by calling the internal hook method in DataGrid. https://github.com/stsrki/Blazorise/blob/master/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs#L89 
What do you think?

**Comment by stsrki** [comment]
I think it should be possible. You can extend your custom column from `DataGridColumn<TItem>` base class. Once you place your custom column inside of DataGrid it should be picked automatically since DataGrid is cascaded and everything is registered through the `DataGridColumn<TItem>` `OnInitialized()` method. That way you don't need to access Hook directly.

There _could_ be a problem with edit fields but I don't think you have those so maybe it's not an issue. Try it, and if something is missing I will add it.

**Comment by stsrki** [comment]
PS. Maybe name it `DataGridEntityActionsColumn` ?

**Comment by ilkayilknur** [comment]
I think we can achieve something similar that @hikalkan suggested above. Since we're using the DataGridColumn as base class, we need to add the DisplayTemplate tag. We also need a parent component for action items. This parent component will check child action visibilities or policies and decides the rendering type of the component(dropdown, split dropdown, or button). 

```xml
<DataGridEntityActionsColumn TItem="ItemType">
    <DisplayTemplate>
        <EntityActions>
                <EntityAction
                    Text = "@L["Edit"]"
                    RequiredPolicy="@IdentityPermissions.Roles.Update"
                    Clicked="() => OpenEditModalAsync(context.As<IdentityRoleDto>().Id)" />
                <EntityAction
                    Text = "@L["Permissions"]"
                    RequiredPolicy="@IdentityPermissions.Roles.ManagePermissions"
                    Clicked="() => PermissionManagementModal.OpenAsync(PermissionProviderName, context.As<IdentityRoleDto>().Name)" />
                <EntityAction
                    Text = "@L["Delete"]"
                    RequiredPolicy="@IdentityPermissions.Roles.Delete"
                    ConfirmationMessage="@L["TheRole{0}WillBeDeleted", context.As<IdentityRoleDto>().Name]"
                    Clicked="() => DeleteEntityAsync(context.As<IdentityRoleDto>())" />
        </EntityActions>
    </DisplayTemplate>
</DataGridEntityActionsColumn>
```
I'm going to start working on the changes, I'll let you know if I have further questions. Thanks, @stsrki .

hikalkan:
>PS. Maybe name it DataGridEntityActionsColumn ?

I agree.

ilkayilknur:
I have completed most of the functionality we decided above. There is only one issue left that I would like to get your opinion  @stsrki 

There is a `Displayable `parameter in DataGridColumn that controls the visibility of the column. I want to change that property inside the `Actions `component because the `Actions `component decides whether it should display the column or not.  Is there a way to achieve that?

stsrki:
@ilkayilknur You mean to somehow override default `Displayable` behavior? Currently, it's not possible. but maybe I could add a handler that you can use.

ilkayilknur:
@stsrki  How can I access that handler from the component inside the `DisplayTemplate`?

stsrki:
> @stsrki How can I access that handler from the component inside the `Di splayTemplate`?

There is no handler at the moment. I would need to implement it and even before that see if is ts possible and if it's going to be a breaking change.

ilkayilknur:
> > @stsrki How can I access that handler from the component inside the `Di splayTemplate`?
> 
> There is no handler at the moment. I would need to implement it and even before that see if is ts possible and if it's going to be a breaking change.

Yes, I understand that :) My problem is accessing the parent data column component from the component inside the DisplayTemplate.
For example,
```xml
<DataGridEntityActionsColumn TItem="ItemType">
    <DisplayTemplate>
        <EntityActions>
                <EntityAction
                    Text = "@L["Edit"]"
                    RequiredPolicy="@IdentityPermissions.Roles.Update"
                    Clicked="() => OpenEditModalAsync(context.As<IdentityRoleDto>().Id)" />
                <EntityAction
                    Text = "@L["Permissions"]"
                    RequiredPolicy="@IdentityPermissions.Roles.ManagePermissions"
                    Clicked="() => PermissionManagementModal.OpenAsync(PermissionProviderName, context.As<IdentityRoleDto>().Name)" />
                <EntityAction
                    Text = "@L["Delete"]"
                    RequiredPolicy="@IdentityPermissions.Roles.Delete"
                    ConfirmationMessage="@L["TheRole{0}WillBeDeleted", context.As<IdentityRoleDto>().Name]"
                    Clicked="() => DeleteEntityAsync(context.As<IdentityRoleDto>())" />
        </EntityActions>
    </DisplayTemplate>
</DataGridEntityActionsColumn>
```
I need to access `DataGridEntityActionsColumn` inside the `EntityActions` component so that I can manage the visibility of the column. We want to do it implicitly because we don't want the developer to check policies and decide the visibility of the column. It's already done in the `EntityAction `component. If I can do that using the handler, it would be good :)

stsrki:
Not sure if it'll help, but try adding a `@ref` to DataGridEntityActionsColumn.

```cshtml
DataGridEntityActionsColumn columnRef;

<DataGridEntityActionsColumn @ref="@columnRef">
    ...
</DataGridEntityActionsColumn>
```

Maybe that would be enough.

ilkayilknur:
I've completed the component implementation. I've created a PR that shows the sample usage in a module. https://github.com/volosoft/volo/pull/3800 
Let me know if you want additional changes @hikalkan, @stsrki.

framework/src/Volo.Abp.BlazoriseUI/Components/EntityActions.razor.cs
```diff
@@ -0,0 +1,55 @@
...
+        }
+
+        protected override void OnInitialized()
+        {
+            base.OnInitialized();
+            ToggleText = UiLocalizer["Actions"];
+        }
+    }
+
+    public enum ActionType
```
stsrki: Move this to a separate file.

stsrki:
@ilkayilknur Looking good to me.

framework/src/Volo.Abp.BlazoriseUI/Components/EntityActions.razor.cs
ilkayilknur: Done.

ilkayilknur:
PR is ready for merge @hikalkan

PR #5918, opened 2020-10-23
Introduce Volo.Abp.AspNetCore.Components package

### LINKED ISSUE #5904: Introduce Volo.Abp.AspNetCore.Components package
* `Volo.Abp.AspNetCore.Components` is a part of the Volo.Abp (the framework) solution, depends on the `Microsoft.AspNetCore.Components` NuGet package.
* Move the `ServiceProviderComponentActivator` and the related code done in [that commit](https://github.com/abpframework/abp/commit/a6a5f5aca99db7d511fbd8fcf5756008fdf83ee2) to this new package.
* Add reference to this new package from Volo.Abp.AspNetCore.Components.WebAssembly package

Our goal to keep this services in a low level package, so we can share it by blazor web assembly and blazor server side.

ilkayilknur:
Closes #5904


hikalkan:
@ilkayilknur I merged this, but actually there is one missing. I've implemented with https://github.com/abpframework/abp/commit/084d65de90af70d4930573943960065326130299

* Every ABP package should have a class derived from the `AbpModule`. So, I added `AbpAspNetCoreComponentsModule`.
* I moved the related code in the `PreConfigureServices` to this package, to be consistent. Otherwise `Volo.Abp.AspNetCore.Components` package can't work properly without referencing to `Volo.Abp.AspNetCore.Components.WebAssembly` and no meaning to split this package.
* I added dependency from `AbpAspNetCoreComponentsWebAssemblyModule` to `AbpAspNetCoreComponentsModule`.
* I also moved DI related classes to `Volo.Abp.AspNetCore.Components.DependencyInjection`, but that's not related to this problem, just a better namespacing.

Issue #5933, opened 2020-10-25
IFeatureChecker.GetOrNullAsync throws an AbpException if not found


dmytroshchotkin:
Please check Volo.Abp.Features.IFeatureChecker.GetOrNullAsync(string name) method.
It throws an exception if the name is not found. And it's expected to get null instead.

maliming:
hi

**FeatureChecker** does not seem to throw an exception.

https://github.com/abpframework/abp/blob/1a622bc50fbd9b554f61078d5eeafa9def3216e3/framework/src/Volo.Abp.Features/Volo/Abp/Features/FeatureChecker.cs#L39-L66

* Your **ABP Framework version**.
* Exception message and **stack trace** if available (check the logs).
* Steps needed to **reproduce** the problem.

PR #5938, opened 2020-10-26
IFeatureChecker.GetOrNullAsync  return null if not found

zfmy:
#5933


framework/src/Volo.Abp.Features/Volo/Abp/Features/FeatureChecker.cs
```diff
@@ -37,7 +37,11 @@ public class FeatureChecker : FeatureCheckerBase
         
         public override async Task<string> GetOrNullAsync(string name)
         {
-            var featureDefinition = FeatureDefinitionManager.Get(name);
+            var featureDefinition = FeatureDefinitionManager.GetOrNull(name);
```
realLiangshiwei, line 40: I think it should throw an exception if there is no feature definition. This may be by design.

maliming:
I agree with @realLiangshiwei . Thanks @zfmy .

PR #5943, opened 2020-10-26
Created settings document for the Angular UI

### LINKED ISSUE #5772: Settings document for the Angular UI
This document seems covering it: https://docs.abp.io/en/abp/latest/UI/Angular/Config-State#how-to-get-a-specific-setting-from-the-store

However, it would be better to have a dedicated document. I created an empty one: https://github.com/abpframework/abp/blob/rel-3.2/docs/en/UI/Angular/Settings.md

* Write the document in the rel-3.2 branch (so it immediately goes to live) then merge to dev.
* Add a "This document only explains the JavaScript API. See the settings document to understand the ABP setting system." note just like I did [here](https://github.com/abpframework/abp/blob/rel-3.2/docs/en/UI/AspNetCore/JavaScript-API/Settings.md)
* Add the document to the docs-nav.json to make it visible on the left nav tree. It can be just under the "Localization":

image

armanozak:
Fixes #5772

### DIFF
--- a/docs/en/UI/Angular/Config-State.md
+++ b/docs/en/UI/Angular/Config-State.md
@@ -93,40 +93,6 @@ const searchUrl = this.config.getApiUrl("search");
 
 This method returns the `url` of a specific API based on the key given as its only parameter. If there is no key, `'default'` is used.
 
-### How to Get All Settings From the Store
-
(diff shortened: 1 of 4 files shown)


bnymncoskuner (approved):
LGTM

PR #5952, opened 2020-10-27
Move to System.Text.Json

### LINKED ISSUE #1198: Move to System.Text.Json
Since AspNet Core will move to it with 3.0.

See https://github.com/dotnet/corefx/issues/33115

The package: https://www.nuget.org/packages/System.Text.Json

maliming:
Resolve #1198

Changes:
* Using `System.Text.Json` in `Volo.Abp.AspNetCore.Mvc`.
* Try to use `System.Text.Json` instead of `Newtonsoft` in all projects.
* Expose no-argument constructors for some classes.
https://github.com/abpframework/abp/tree/dev/framework/src/Volo.Abp.Http/Volo/Abp/Http/Modeling
https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to#specify-constructor-to-use

Special changes:

* **Deserialize inferred types to object properties**, I am using simple code provided by Microsoft(**ObjectToInferredTypesConverter**). But it is may not perfect. (https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-converters-how-to#deserialize-inferred-types-to-object-properties)

* **System.Text.Json** can't [ignore properties based on attributes](https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to#conditionally-ignore-a-property) Such as `DisableAuditingAttribute, DisableDateTimeNormalizationAttribute`. 
https://github.com/dotnet/runtime/issues/43876

  We use [AbpHybridJsonInputFormatter](https://github.com/abpframework/abp/blob/maliming/system.text.json/framework/src/Volo.Abp.AspNetCore.Mvc/Volo/Abp/AspNetCore/Mvc/Json/AbpHybridJsonInputFormatter.cs#L28)  and [AbpHybridJsonOutputFormatter](https://github.com/abpframework/abp/blob/maliming/system.text.json/framework/src/Volo.Abp.AspNetCore.Mvc/Volo/Abp/AspNetCore/Mvc/Json/AbpHybridJsonOutputFormatter.cs#L28)  to choose between `System.Text.Json` and `Newtonsoft` according to the  [configuration](https://github.com/abpframework/abp/blob/maliming/system.text.json/framework/src/Volo.Abp.Json/Volo/Abp/Json/SystemTextJson/AbpSystemTextJsonSerializerOptions.cs#L10).

   ```cs
   Configure<AbpSystemTextJsonSerializerOptions>(options =>
   {
         //If the type is used as a property of other types, you also need to add it.
         options.UnsupportedTypes.Add<MyInputModel>();
   });
   ```

**I did some tests using template projects, including tiered projects.**


maliming:
I did some tests using template projects, including tiered projects.

@hikalkan Do you have any suggestions?

maliming:
There is a way to solve the problem that `system.text.json`  can't ignore properties

We rewrite the `SystemTextJsonInputFormatter` service and use `NewtonsoftJsonInputFormatter `if the `Type` contains properties that need ignored.

This can improve performance as much as possible.

framework/src/Volo.Abp.AspNetCore.Mvc/Volo/Abp/AspNetCore/Mvc/Json/MvcCoreBuilderExtensions.cs
```diff
@@ -0,0 +1,83 @@
...
+        {
+            //SystemTextJsonInputFormatter
+            builder.Services.AddTransient(provider =>
+            {
+                var jsonOptions = provider.GetRequiredService<IOptions<JsonOptions>>();
+                var logger = provider.GetRequiredService<ILoggerFactory>().CreateLogger<SystemTextJsonInputFormatter>();
+                return new SystemTextJsonInputFormatter(jsonOptions.Value, logger);
+            });
+
+            builder.Services.AddTransient<DefaultObjectPoolProvider>();
```
hikalkan, line 26: TryAddTransient would be better.

framework/src/Volo.Abp.Ddd.Domain/Volo/Abp/Domain/Entities/EntityHelper.cs
```diff
@@ -266,5 +266,36 @@ public static Type FindPrimaryKeyType([NotNull] Type entityType)
 
             property?.SetValue(entity, idFactory());
         }
+
+        public static void TrySetId<TKey>(
+            object entity,
+            TKey id,
+            bool checkForDisableIdGenerationAttribute = false)
+        {
```
hikalkan, line 274: The method above can internally use this one to avoid code duplication. Becuase `IEntity<TKey>` is already an object.

framework/src/Volo.Abp.Auditing/Volo.Abp.Auditing.csproj
```diff
@@ -23,4 +23,8 @@
     <ProjectReference Include="..\Volo.Abp.Timing\Volo.Abp.Timing.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
```
hikalkan, line 27: Should we depend on Volo.Abp.Json instead?

modules/docs/src/Volo.Docs.Domain.Shared/Volo.Docs.Domain.Shared.csproj
```diff
@@ -13,5 +13,9 @@
   <ItemGroup>
     <ProjectReference Include="..\..\..\..\framework\src\Volo.Abp.Localization\Volo.Abp.Localization.csproj" />
   </ItemGroup>
+  
+  <ItemGroup>
+    <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
```
hikalkan, line 18: Why added this?

framework/src/Volo.Abp.AspNetCore.Mvc/Volo/Abp/AspNetCore/Mvc/Json/MvcCoreBuilderExtensions.cs
maliming: https://github.com/abpframework/abp/pull/6054

PR #5973, opened 2020-10-28
Created UiNotificationService

### LINKED ISSUE #5552: Blazor: IUiNotificationService to show toastr style notifications
We can try to integrate Toastr or develop own own.

You can get help from @stsrki ;)

stsrki:
PR for #5552

**DO NOT MERGE** until we merge #5883 since this one is based on that branch. We can merge into `dev-client-validation` if that is a preferred way.

Done:

- UiNotificationService is showing messages by using Blazorise `Snackbar` extension and it's `SnackbarStack` component

**Example:**

image

To do:

- [ ] I added `title` parameter to UiNotificationService. If it's not needed I can remove it. If it's needed then I will need to implement it in Blazorise. We can discuss this!
- [ ] OK button and text to show per notification. Currently, it's the same for all notifications.

Maybe create new notifications based on alerts that are like this??

image

@hikalkan  We can discuss it all.


framework/src/Volo.Abp.BlazoriseUI/BlazoriseUiNotificationService.cs
```diff
@@ -7,34 +10,79 @@
...
             Logger = NullLogger<BlazoriseUiNotificationService>.Instance;
         }
 
-        public Task Info(string message)
+        public Task Info(string message, string title = null, Action<UiNotificationOptions> options = null)
         {
+            var uiNotificationOptions = CreateDefaultOptions();
+            options?.Invoke(uiNotificationOptions);
+
             Logger.LogInformation(message);
```
hikalkan: No need to log the notifications. I suppose it was a test code and you didn't delete it.

framework/src/Volo.Abp.AspNetCore.Components.WebAssembly/Volo/Abp/AspNetCore/Components/WebAssembly/UiNotificationOptions.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    /// <summary>
+    /// Options to override notification appearance.
+    /// </summary>
+    public class UiNotificationOptions
+    {
+        /// <summary>
+        /// Custom text for the Ok button.
+        /// </summary>
+        public string OkButtonText { get; set; }
```
hikalkan: It should be a `ILocalizableString`, not `string`

modules/identity/.editorconfig
```diff
@@ -0,0 +1,131 @@
+# Rules in this file were initially inferred by Visual Studio IntelliCode from the D:\Projects\Volosoft\abp\framework codebase based on best match to current usage at 2.10.2020.
```
hikalkan: We follow the same rules for all the solutions in the abp & volo repositories. So, we don't like to add `.editorconfig` to every individual solution. It would not be maintainable.
If we add such a file, we should add in a single place per GitHub repository.

hikalkan (changes requested):

framework/src/Volo.Abp.BlazoriseUI/BlazoriseUiNotificationService.cs
stsrki: Logging removed

modules/identity/.editorconfig
stsrki: I don't know how to best handle this. I have too many projects and solutions that are using my own formating and only ABP projects have a default format. That's why I added these `.editorconfig` files. 

Copying and removing `.editorconfig` before and after each push/pull is very practical...

framework/src/Volo.Abp.AspNetCore.Components.WebAssembly/Volo/Abp/AspNetCore/Components/WebAssembly/UiNotificationOptions.cs
stsrki: Can you explain what is ILocalizableString and why it's better that string?

modules/identity/.editorconfig
hikalkan: I think you add this file to .gitignore because no one needs it except you :)

framework/src/Volo.Abp.AspNetCore.Components.WebAssembly/Volo/Abp/AspNetCore/Components/WebAssembly/UiNotificationOptions.cs
hikalkan: If you make it [ILocalizableString](https://github.com/abpframework/abp/blob/dev/framework/src/Volo.Abp.Localization.Abstractions/Volo/Abp/Localization/ILocalizableString.cs) (and handle while showing it), the text is automatically localized based on the current language. Otherwise, the value will be same in every language. What you need to do is;

1. Make this `public ILocalizableString OkButtonText { get; set; }`

In this way, I can set it like `new LocalizableString(typeof(MyResource), "Key")`.

2. When you need to show this, use `var currentValue = OkButtonText.Localize(StringLocalizerFactory)`. Inject and pass `IStringLocalizerFactory` here, it knows how to localize it.

modules/identity/.editorconfig
stsrki: Yep, that's a good idea :) tnx

framework/src/Volo.Abp.AspNetCore.Components.WebAssembly/Volo/Abp/AspNetCore/Components/WebAssembly/UiNotificationOptions.cs
stsrki: Ok, yeah that is better indeed. Thanks

modules/identity/.editorconfig
stsrki: Removed files and added new rule to gitignore

framework/src/Volo.Abp.AspNetCore.Components.WebAssembly/Volo/Abp/AspNetCore/Components/WebAssembly/UiNotificationOptions.cs
stsrki: Converted to ILocalizableString

PR #5988, opened 2020-10-29
Use user-defined menus in Blazor's LoginDisplay.

### LINKED ISSUE #5970: wrong account management url lead to 404 in blazor wasm
abp version: 3.3.0
we should use Authority url  instead of RemoteServices's url?

https://github.com/abpframework/abp/blob/48d73aa60753d192236ce5b4a3a167f85fb5b53b/framework/src/Volo.Abp.AspNetCore.Components.WebAssembly.BasicTheme/Themes/Basic/LoginDisplay.razor.cs#L33-L34

maliming:
Fix #5970

`InvokeVoidAsync` caused an exception, and I did find a solution. ,
```cs
await JsRuntime.InvokeVoidAsync("open", uri, "_blank");
```
```cs
blazor.webassembly.js:1 Uncaught (in promise) TypeError: Converting circular structure to JSON
    --> starting at object with constructor 'Window'
    --- property 'window' closes the circle
    at JSON.stringify (<anonymous>)
    at blazor.webassembly.js:1
```

### DIFF
--- a/framework/src/Volo.Abp.AspNetCore.Components.WebAssembly.BasicTheme/Themes/Basic/LoginDisplay.razor
+++ b/framework/src/Volo.Abp.AspNetCore.Components.WebAssembly.BasicTheme/Themes/Basic/LoginDisplay.razor
@@ -25,7 +25,7 @@
                 {
                     @foreach (var menuItem in Menu.Items)
                     {
-                        <DropdownItem Clicked="@(() => NavigateToAsync(menuItem.Url, menuItem.Target))">@menuItem.DisplayName</DropdownItem>
+                        <DropdownItem Clicked="@(() => NavigateTo(menuItem.Url))">@menuItem.DisplayName</DropdownItem>
                     }
                 }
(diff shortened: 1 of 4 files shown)


hikalkan:
We should open the profile management page in the same window. So, don't use `_blank`. But I don't know why it didn't worked. Maybe `await JsRuntime.InvokeVoidAsync("window.open", uri, "_blank");` would work.

cotur:
@hikalkan yes we should open in same page.

Related issue.

https://github.com/dotnet/aspnetcore/issues/16632#issuecomment-547404523

> This is because window.open returns a WindowProxy object (see https://developer.mozilla.org/en-US/docs/Web/API/Window/open). WindowProxy is not JSON-serializable, so can't be used as a return value to .NET code.

>To fix this, don't call window.open directly, but instead call a JS function of your own that either returns nothing or returns something that is JSON-serializable.

cotur (approved):

PR #6000, opened 2020-10-31
One method 'QueueAsync' is not virtual of EmailSenderBase class

ilonze:
we inherit the class EmailSenderBase and override all method, look the method 'public async Task QueueAsync(string from, string to, string subject, string body, bool isBodyHtml = true)' can't override


maliming:
hi

Your PR has some Irrelevant changes, I make a new PR https://github.com/abpframework/abp/pull/6001

PR #6012, opened 2020-11-02
Add send sms by BackgroundSmsSendingJob.

ilonze:
Add send sms by BackgroundSmsSendingJob. And add some extensions method and the .ctor for SmsMessage class.


ilonze:
has error

PR #6013, opened 2020-11-02
Add send sms by BackgroundSmsSendingJob.

ilonze:
Add send sms by BackgroundSmsSendingJob. And add some extensions method and the .ctor for SmsMessage class.


maliming:
hi @ilonze 
SMS is an independent module, it should not reference BackgroundJobs. If you need you can implement it in your own module or application.

ilonze:
This SMS is written with reference to the Volo.Abp.Emailing. I think it is necessary to send messages in BackgroundJob, include Email and SMS.

maliming:
Hi

You can implement it in your own module or application.

PR #6014, opened 2020-11-02
Revert "Blazorise snackbar package added."

hikalkan:
Reverts abpframework/abp#6006


hikalkan:
Reverting this since https://github.com/abpframework/abp/blob/dev/framework/src/Volo.Abp.BlazoriseUI/Volo.Abp.BlazoriseUI.csproj already has a reference to the Snackbar

PR #6035, opened 2020-11-03
Tenant modals validation

### LINKED ISSUE #6026: Blazor UI: Implement validation for the tenant-management module

stsrki:
resolve #6026

- Blazorise validation for create, edit and connection string modals
- Fixed display name localization for tenant model
- modal can be submitted with enter key
- Fixed user edit modal

### DIFF
--- a/modules/tenant-management/src/Volo.Abp.TenantManagement.Blazor/Pages/TenantManagement/TenantManagement.razor.cs
+++ b/modules/tenant-management/src/Volo.Abp.TenantManagement.Blazor/Pages/TenantManagement/TenantManagement.razor.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.ComponentModel.DataAnnotations;
 using System.Threading.Tasks;
 using Blazorise;
 using Microsoft.AspNetCore.Authorization;
@@ -20,9 +21,10 @@ namespace Volo.Abp.TenantManagement.Blazor.Pages.TenantManagement
         protected FeatureManagementModal FeatureManagementModal;
 
         protected Modal ManageConnectionStringModal;
+        protected Validations ManageConnectionStringValidations;
 
         protected TenantInfoModel TenantInfo;
-        
+
         public TenantManagement()
         {
             LocalizationResource = typeof(AbpTenantManagementResource);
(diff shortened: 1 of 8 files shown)


hikalkan (approved):

modules/tenant-management/src/Volo.Abp.TenantManagement.Blazor/Pages/TenantManagement/TenantManagement.razor.cs
```diff
@@ -67,18 +71,21 @@ protected virtual Task CloseEditConnectionStringModal()
...
         {
-            await CheckPolicyAsync(ManageConnectionStringsPolicyName);
-
-            if (TenantInfo.UseSharedDatabase || TenantInfo.DefaultConnectionString.IsNullOrWhiteSpace())
+            if (ManageConnectionStringValidations.ValidateAll())
             {
-                await AppService.DeleteDefaultConnectionStringAsync(TenantInfo.Id);
+                await CheckPolicyAsync(ManageConnectionStringsPolicyName);
+
+                if (TenantInfo.UseSharedDatabase || TenantInfo.DefaultConnectionString.IsNullOrWhiteSpace())
```
cotur, line 78: You define as `[Required]` for `TenantInfoModel.DefaultConnectionString`, but here we control is that null or empty.

modules/tenant-management/src/Volo.Abp.TenantManagement.Blazor/Pages/TenantManagement/TenantManagement.razor.cs
```diff
@@ -93,6 +100,7 @@ public class TenantInfoModel
 
         public bool UseSharedDatabase { get; set; }
 
+        [Required]
```
cotur, line 103: This may be nullable. Does not required.

cotur (changes requested):

modules/tenant-management/src/Volo.Abp.TenantManagement.Blazor/Pages/TenantManagement/TenantManagement.razor.cs
stsrki: It's only run if the checkbox is ticked so I think it's best to be Required in this case

cotur (approved):

Issue #6040, opened 2020-11-03
Use AbpSwaggerUi for all modules and templates


mtozlu:
With #5728 and #5864 `UseAbpSwaggerUi` is used by some of the templates however some modules and `HttpApi.Host` project in app template is not updated so using swagger for operations that needs CSRF token causes error in those projects.
Here is the list that is not updated;
![image](https://user-images.githubusercontent.com/2939330/97997352-c92d4880-1de0-11eb-9c96-307fa06e9e22.png)

realLiangshiwei:
Hi, this is intentional. You can directly use the client's(`Web`) swagger page.

mtozlu:
If you use app template with angular ui, the solution does not contain any `Web` project (`MyCompanyName.MyProjectName.HttpApi.HostWithIds` template). Even if it did, using swagger to do some operation that needs a CSRF token, the api gives error:
> The required antiforgery header value "RequestVerificationToken" is not present.

realLiangshiwei:
Related https://github.com/abpframework/abp/pull/6124

PR #6045, opened 2020-11-03
Use Volo.Abp.Swashbuckle in application template

mtozlu:
Related #6040


realLiangshiwei:
Thanks for your contributor, but I close this issue, because it will be done in https://github.com/abpframework/abp/pull/6124

PR #6099, opened 2020-11-06
Add AbpStringToEnumConverter & AbpStringToBooleanConverter.

### LINKED ISSUE #6092: Add JsonStringEnumConverter to JsonSerializerOptions
Otherwise, it can't convert enum strings to enum members on deserialization.

maliming:
Resolve #6092

The project may use **string** in JSON as the **enum and bool** values. We want to read them correctly, but we don’t want to   write **enum and bool** as **string** anymore.

BTW: Angular is using the **enum as int** now. 

Input:
```json
{
  "remember": "false",
  "day": "Monday",
}
```

output:
```json
{
  "remember": false,
  "day": 0
}
```

### DIFF
--- a/framework/src/Volo.Abp.Json/Volo/Abp/Json/SystemTextJson/JsonConverters/AbpStringToEnumConverter.cs
+++ b/framework/src/Volo.Abp.Json/Volo/Abp/Json/SystemTextJson/JsonConverters/AbpStringToEnumConverter.cs
@@ -1,0 +1,41 @@
+﻿using System;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace Volo.Abp.Json.SystemTextJson.JsonConverters
+{
+    public class AbpStringToEnumConverter : JsonConverter<object>
+    {
+        private readonly JsonStringEnumConverter _innerJsonStringEnumConverter;
+
+        public AbpStringToEnumConverter()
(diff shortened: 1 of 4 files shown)


framework/src/Volo.Abp.Json/Volo/Abp/Json/SystemTextJson/JsonConverters/AbpStringToEnumConverter.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        public override bool CanConvert(Type typeToConvert)
+        {
+            return typeToConvert.IsEnum;
+        }
+
+        public override object Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+        {
+            var newOptions = new JsonSerializerOptions(options);
+            newOptions.Converters.Remove(this);
```
hikalkan, line 29: I didn't understand what we are trying to do. We first add this converter to JsonSerializerOptions.Converters, then removing here. Why this workaround-like code here?

maliming: If we don't remove `this(AbpStringToEnumConverter)`, it will cause infinite loop. We hope it uses `innerJsonStringEnumConverter`.

PR #6108, opened 2020-11-07
Modify AbpNoContentActionFilter

ilonze:
Hi!
The AbpNoContentActionFilter shouldn't modify StatusCode to NoContent when action result has value, because may be result modify by custum result filter.


maliming:
Hi

Could you add a unit test for this change?

maliming (approved):

maliming (changes requested):
Your PR contains some irrelevant changes

ilonze:
Re-request

PR #6133, opened 2020-11-10
Disable modal closing by escape key or lost focus

### LINKED ISSUE #6029: Blazor: Message/Confirm dialogs should not close on click outside of the modal
Should not close the dialog when we click outside of the modal. Otherwise, user may accidently click without reading an important message.

message-box

stsrki:
resolve #6029


hikalkan:
Remember to assign the PR to a milestone, otherwise I can't see it in the dashboard ;)

PR #6139, opened 2020-11-10
Forms Module: added bootstrap-select npm and contributor

gterdem:


hikalkan:
@cotur don't merge this, it will be probably changed to select2.

gterdem:
As discussed, select2 will be used instead of bootstrap-select. Closing this PR.

PR #6168, opened 2020-11-12
Removed the Angular Account Module UI

### LINKED ISSUE #5652: Remove the Angular Account Module UI
For the Angular UI (and the new Blazor UI) we are using the MVC UI for login, register, manage profile... etc. This is because we are using code authorication flow to make the user login.

So, we can drop the Angular account module UI completely. We should discuss that.

mehmet-erim:
Resolves #5652


bnymncoskuner (approved):
LGTM

PR #6194, opened 2020-11-13
Blazor UI message service documentation added.

ilkayilknur:
Related #6214


hikalkan:
Thanks. Can you please cancel this and send the PR to the rel-4.0 branch, so it can be immediately available on the https://docs.abp.io/en/abp/4.0.

docs/en/UI/Blazor/Message.md
```diff
@@ -1,3 +1,101 @@
...
+## Quick Example
+
+Simply inject `IUiMessageService` to your page or component and call the `Success` method to show a success message.
+
+```csharp
+namespace MyProject.Blazor.Pages
+{
+    public partial class Index
+    {
+        public IUiMessageService UiMessageService { get; set; }
```
hikalkan: I suppose you forget to [Inject] this. You can use constructor injection since it is supported since the version 4.0

hikalkan: In addition, if would be good to mention about the `AbpComponentBase` class that already has a `Message` property (of type `IUiMessageService`) that is pre-injected. So, if you derived from the `AbpComponentBase` class, you can directly use `await Message.Success(...);`.

hikalkan (changes requested):

docs/en/UI/Blazor/Message.md
ilkayilknur: Actually, It works without the `[Inject]` attribute but I will change the code and use the constructor injection.

hikalkan: Oh, it works because ABP (with AutoFac) enables property injection :)

PR #6200, opened 2020-11-14
fix bug of FeatureManager.GetAllWithProviderAsync and add unittest

yinchang0626:
Add a unittest to test `FeatureManager.GetAllWithProviderAsync` method ,
It should be the same result with `featureManager.GetOrNullForEditionAsync`,but not
So this pull request will fix it


maliming:
hi @yinchang0626  Thanks,

I move it to `rel-4.0` branch. https://github.com/abpframework/abp/pull/6202

PR #6207, opened 2020-11-16
Set current language to cookie

mehmet-erim:


bnymncoskuner (approved):
LGTM

npm/ng-packs/packages/core/src/lib/services/session-state.service.ts
```diff
@@ -89,5 +89,6 @@ export class SessionStateService {
 
     this.store.patch({ language });
     document.documentElement.setAttribute('lang', language);
+    document.cookie = `.AspNetCore.Culture=${encodeURIComponent(`c=${language}|uic=${language}`)}`;
```
maliming, line 92: `AspNetCore.Culture` may be changed. It is not a fixed value.

https://github.com/dotnet/aspnetcore/blob/master/src/Middleware/Localization/src/CookieRequestCultureProvider.cs#L22

PR #6228, opened 2020-11-17
Enabling split queries globally.

maliming:
EF Core uses single query mode by default in the absence of any configuration. Since it may cause performance issues, EF Core generates a warning whenever following conditions are met:

* EF Core detects that the query loads multiple collections.
* User hasn't configured query splitting mode globally.
* User hasn't used AsSingleQuery/AsSplitQuery operator on the query.

To turn off the warning, configure query splitting mode globally or at the query level to an appropriate value.

https://docs.microsoft.com/en-us/ef/core/querying/single-split-queries#enabling-split-queries-globally


hikalkan:
We should write this to the EF Core integration document, so people will know it. Can you write please?

maliming:
https://github.com/abpframework/abp/commit/acee880831a67956f82ab826ff720effef5f174c

Issue #6222, opened 2020-11-17
Angular  CorsPolicyService did not allow origin: http://localhost:4200


maliming:
```cs
[09:55:28 DBG] CORS request made for path: /.well-known/openid-configuration/jwks from origin: http://localhost:4200
[09:55:28 WRN] Origin is not allowed: http://localhost:4200
[09:55:28 WRN] CorsPolicyService did not allow origin: http://localhost:4200
[09:55:28 INF] No CORS policy found for the specified request.
[09:55:28 DBG] Request path /.well-known/openid-configuration/jwks matched to endpoint type Discovery
```

maliming:
```cs
[10:08:25 WRN] Compiling a query which loads related collections for more than one collection navigation either via 'Include' or through projection but no 'QuerySplittingBehavior' has been configured. By default Entity Framework will use 'QuerySplittingBehavior.SingleQuery' which can potentially result i
n slow query performance. See https://go.microsoft.com/fwlink/?linkid=2134277 for more information. To identify the query that's triggering this warning call 'ConfigureWarnings(w => w.Throw(RelationalEventId.MultipleCollectionIncludeWarning))'
[10:08:25 INF] Executed DbCommand (0ms) [Parameters=[@__ef_filter__p_0='?' (DbType = Boolean)], CommandType='Text', CommandTimeout='30']
SELECT [i].[Id], [i].[ConcurrencyStamp], [i].[CreationTime], [i].[CreatorId], [i].[DeleterId], [i].[DeletionTime], [i].[Description], [i].[DisplayName], [i].[Emphasize], [i].[Enabled], [i].[ExtraProperties], [i].[IsDeleted], [i].[LastModificationTime], [i].[LastModifierId], [i].[Name], [i].[Required], [i
].[ShowInDiscoveryDocument], [i0].[IdentityResourceId], [i0].[Type], [i1].[IdentityResourceId], [i1].[Key], [i1].[Value]
FROM [IdentityServerIdentityResources] AS [i]
LEFT JOIN [IdentityServerIdentityResourceClaims] AS [i0] ON [i].[Id] = [i0].[IdentityResourceId]
LEFT JOIN [IdentityServerIdentityResourceProperties] AS [i1] ON [i].[Id] = [i1].[IdentityResourceId]
[...]

PR #6230, opened 2020-11-17
Add CorsOrigin when create identity server client.

maliming:
Resolve: https://github.com/abpframework/abp/issues/6222

Currently it is only a warning in the logs because we have configured CORS for the template application.
https://github.com/abpframework/abp/blob/rel-4.0/templates/app/aspnet-core/src/MyCompanyName.MyProjectName.HttpApi.HostWithIds/MyProjectNameHttpApiHostModule.cs#L159


hikalkan:
thanks.

PR #6237, opened 2020-11-17
Fix logger generic type

### LINKED ISSUE #6235: Passed in logger of the wrong type
https://github.com/abpframework/abp/blob/42f37c5ff01ad853a5425d15539d4222cd0dab69/modules/docs/app/VoloDocs.Web/Utils/GlobalExceptionHandlerMiddleware.cs#L16

realLiangshiwei:
Resolve https://github.com/abpframework/abp/issues/6235


maliming:
We should make it in rel-4.0.

PR #6290, opened 2020-11-20
Add ParameterName property to the ParameterApiDescriptionModel.

### LINKED ISSUE #6157: Dynamic JavaScript Proxies can't handle complex types
### Summary
I'm trying to use my [AutoFilterer](https://github.com/enisn/AutoFilterer) library to filtering entities. I've complex types in such as **Range<T>** which includes`Min` and `Max` properties  in it. 

- **ABP Framework version**: `3.3.1`

- **User Interface**: `MVC`

### Example
- My used dto instead of PagedAndSortedResultRequestDto:

```csharp
    public class BookFilterDto : PaginationFilterBase
    {
        [StringFilterOptions(StringFilterOption.Contains)]
        public string Title { get; set; }

        [OperatorComparison(OperatorType.Equal)]
        public string Language { get; set; }

        [StringFilterOptions(StringFilterOption.Contains)]
        public string Author { get; set; }
        public Range<int> TotalPage { get; set; }
        public Range<int> Year { get; set; }
        public BookType[] BookType { get; set; }
    }
```

- Swagger output:
Swagger supports and works with complex types in dto:
image

Also swagger able to send request like `domain.com/books?totalPage.min=250`.

- Javascript proxy:

I've tried a couple of way to run my query:
```js
bookstore.books.book.getList({'TotalPage.min':3})
```
```js
bookstore.books.book.getList({totalPage: { min: 3}})
```

None of them didn't work. 
image

maliming:
Resolve #6157

Test controller: https://gist.github.com/maliming/a28a4033613cfe1139f0b7bf21366784

JS result: https://gist.github.com/maliming/c204b2e9ec7644dc987bfe0a634eba1d/revisions#diff-13876b4beb64b9f156474dc78f9c923952a7ca210d4507b6b3135bbe244f8a60

### DIFF
--- a/framework/src/Volo.Abp.AspNetCore.Mvc/Volo/Abp/AspNetCore/Mvc/AspNetCoreApiDescriptionModelProvider.cs
+++ b/framework/src/Volo.Abp.AspNetCore.Mvc/Volo/Abp/AspNetCore/Mvc/AspNetCoreApiDescriptionModelProvider.cs
@@ -307,6 +307,7 @@ namespace Volo.Abp.AspNetCore.Mvc
 
                 actionModel.AddParameter(ParameterApiDescriptionModel.Create(
                         parameterDescription.Name,
+                        parameterDescription.ModelMetadata?.ParameterName,
                         matchedMethodParamName,
                         parameterDescription.Type,
                         parameterDescription.RouteInfo?.IsOptional ?? false,
--- a/framework/src/Volo.Abp.Http/Volo/Abp/Http/Modeling/ParameterApiDescriptionModel.cs
+++ b/framework/src/Volo.Abp.Http/Volo/Abp/Http/Modeling/ParameterApiDescriptionModel.cs
@@ -10,6 +10,8 @@ namespace Volo.Abp.Http.Modeling
 
         public string Name { get; set; }
 
+        public string ParameterName { get; set; }
+
         public string Type { get; set; }
 
         public string TypeSimple { get; set; }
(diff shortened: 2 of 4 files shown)


enisn (approved):

hikalkan:
@maliming why 4.1-final? Is it 4.0-final actually?

maliming:
Because it currently only has problems under the special cases. I will change it to `4.0-final`

maliming:
hi @armanozak 

Does the `abp generate-proxy` command also need change?

hikalkan:
@maliming can you explain why this is a breaking change by giving an example, so me and @armanozak can understand the exact change.

maliming:
This PR mainly supports developers to customize the name of the model binding.

I Use the [bracket notation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Property_accessors) to access properties instead of [dot notation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Property_accessors). 

`input.name => input["name"]`


For methods that use `FromForm`, I use objects to pass parameters,. This is a **breaking change**.

Before:
```js
postMethod2 = function(inputMyName, inputMy_Age, ajaxParams) {
  	//...
	data: 'input.My.Name=' + input.inputMy.Name + '&' + 'input.My_Age=' + input.inputMy_Age
  	//...
};
```

Now:
```js
postMethod2 = function(input, ajaxParams) {
  	//...
	data: 'input.My.Name=' + input['my.Name'] + '&' + 'input.My_Age=' + input['my_Age']
  	//...
};
```

Does  `generate-proxy` command can generate the correct proxy script for such a method(`GetMethod1([FromQuery(Name = "n.a.m.e")] string name)`)?

PR #6292, opened 2020-11-20
Multiline notification message

### LINKED ISSUE #6245: Blazor UI: Should not truncate the notification message
The developers may trim it if they need. We should not truncate the message. Obviously it is a problem even with such a small message:

image

The button text (`CUSTOMOK` here) can be below the text line. But actually I believe we don't need to such a text button. Instead, a simple close icon is better:

image

This screenshot has been taken from the [toastr library](https://codeseven.github.io/toastr/demo.html).

stsrki:
resolve #6245

image

@hikalkan Note that for this to fully work I will need to release a new Blazorise but I still have other outstanding features to finish.


hikalkan:
Thanks 👍

PR #6293, opened 2020-11-20
Fixed fallback logic for blazor attribute localization

### LINKED ISSUE #6220: Validation displayname should fallback to the property name
Example:

image

The code:

````xml
<Validation MessageLocalizer="@LH.Localize">
    <Field>
        <FieldLabel>@L["Name"]</FieldLabel>
        <TextEdit @bind-Text="@NewEntity.Name">
            <Feedback>
                <ValidationError/>
            </Feedback>
        </TextEdit>
    </Field>
</Validation>
````

It should try to get `DisplayName:Name` from the localization. That's fine. But if not found, it should fallback to `Name` as the key and try to localize again.

stsrki:
fix #6220


hikalkan:
Merged, but then I deeply checked and see that it is not true :)
I will fix it and write here.

hikalkan:
Changed as;

````csharp
// first try to localize with "DisplayName:{Name}"
var localization = stringLocalizer[$"DisplayName:{argument}"];

if (localization.ResourceNotFound)
{
    // then try to localize with just "{Name}"
    localization = stringLocalizer[argument];
}

yield return localization;
````

stsrki:
I wasn't aware you could use it like `localization.ResourceNotFound`. Thanks

PR #6386, opened 2020-11-29
Text corrections

ebicoglu:
I read all of them 💯   It can even be a booklet 😃


hikalkan:
Thanks a lot Alper. I have something more to add. Then we may consider to distribute this as a PDF book :)

PR #6414, opened 2020-12-01
return default value when http client proxy no content

wangbin1989:
If remote service return 204(no content), the json deserialize throw a JsonReaderException: 

> The input does not contain any JSON tokens. Expected the input to start with a valid JSON token, when isFinalBlock is true.

So I check content length, if it is 0, return defult value.


maliming:
Move to 4.0 https://github.com/abpframework/abp/pull/6415

PR #6417, opened 2020-12-01
Discard the task result in Timer_Elapsed.

### LINKED ISSUE #6385: Use discard task in the Timer_Elapsed method to execute the NotifyAsync method of IExceptionNotifier.
```cs
// Discard the result
_ = ExceptionNotifier.NotifyAsync(new ExceptionNotificationContext(ex));
```

https://github.com/abpframework/abp/blob/c9e0ae2e42aa8743775f07a08aa99306fd2f8269/framework/src/Volo.Abp.Threading/Volo/Abp/Threading/AbpTimer.cs#L101-L105

https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#timer-callbacks

maliming:
https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#timer-callbacks

Resolve #6385


framework/src/Volo.Abp.BackgroundWorkers/Volo/Abp/BackgroundWorkers/AsyncPeriodicBackgroundWorkerBase.cs
```diff
@@ -35,22 +35,24 @@ public async override Task StopAsync(CancellationToken cancellationToken = defau
         }
 
         private void Timer_Elapsed(object sender, System.EventArgs e)
+        {
+            // Discard the result
+            _ = DoWorkAsync();
```
hikalkan: All is fine, except this. Because `AbpTime` is designed to prevent overlaps (it prevents to execute `Timer_Elapsed` again if the previous work hasn't been completed). If we fire & forget, it causes overlap if execution duration is longer than the period.

The solution can be creating `AbpAsyncTimer` that internally solves this problem and use it in `AsyncPeriodicBackgroundWorkerBase.cs`.

hikalkan (changes requested):

PR #6423, opened 2020-12-01
Set TaskCreationOptions.LongRunning for Kafka consumers

realLiangshiwei:


framework/src/Volo.Abp.Kafka/Volo/Abp/Kafka/KafkaMessageConsumer.cs
```diff
@@ -121,7 +121,7 @@ protected virtual void Consume()
                         AsyncHelper.RunSync(() => ExceptionNotifier.NotifyAsync(ex, logLevel: LogLevel.Warning));
```
maliming, line 121: await

maliming (changes requested):

PR #6440, opened 2020-12-02
Add antiforgery token to the HTTP request headers for the Blazor UI

### LINKED ISSUE #5732: Add antiforgery token to the HTTP request headers for the Blazor UI
As I see that the Blazor HttpClient doesn't include cookies on HTTP requests. So, it seems we don't care about the antiforgery (because ABP auto antiforgery validation validates only if the `XSRF-TOKEN` is available in the request)

Depends on https://github.com/abpframework/abp/issues/5731 (to get cookies)

realLiangshiwei:
Resolve https://github.com/abpframework/abp/issues/5732

### DIFF
--- a/framework/src/Volo.Abp.AspNetCore.Components.WebAssembly/Volo/Abp/AspNetCore/Components/WebAssembly/AbpBlazorClientHttpMessageHandler.cs
+++ b/framework/src/Volo.Abp.AspNetCore.Components.WebAssembly/Volo/Abp/AspNetCore/Components/WebAssembly/AbpBlazorClientHttpMessageHandler.cs
@@ -3,6 +3,7 @@ using System.Net.Http;
 using System.Net.Http.Headers;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.Extensions.Configuration;
 using Microsoft.JSInterop;
 using Volo.Abp.DependencyInjection;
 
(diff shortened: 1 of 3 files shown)


framework/src/Volo.Abp.AspNetCore.Components.WebAssembly/Volo/Abp/AspNetCore/Components/WebAssembly/AbpBlazorClientHttpMessageHandler.cs
```diff
@@ -38,5 +54,22 @@ private async Task SetLanguageAsync(HttpRequestMessage request, CancellationToke
                 request.Headers.AcceptLanguage.Add(new StringWithQualityHeaderValue(selectedLanguage));
             }
         }
+
+        private async Task SetAntiForgeryTokenAsync(HttpRequestMessage request)
+        {
+            var selfUri = new Uri(_navigationManager.Uri);
+
+            if (request.Method == HttpMethod.Get || request.Method == HttpMethod.Head || request.RequestUri.Host != selfUri.Host || request.RequestUri.Port != selfUri.Port)
```
hikalkan: * Align with the backend: https://github.com/abpframework/abp/blob/dev/framework/src/Volo.Abp.AspNetCore.Mvc/Volo/Abp/AspNetCore/Mvc/AntiForgery/AbpAntiForgeryOptions.cs#L66
* Refactor suggestion: Separate `selfUri` control to another if condition to make it clearer.

hikalkan (changes requested):

PR #6480, opened 2020-12-04
Use IPermissionStore in PermissionManagementProvider

### LINKED ISSUE #6373: Localize Make the PermissionAppService.GetAsync method slow
https://github.com/abpframework/abp/blob/810949f995c01d7df0c5173d184f93f0c6430cce/modules/permission-management/src/Volo.Abp.PermissionManagement.Application/Volo/Abp/PermissionManagement/PermissionAppService.cs#L73

1500 permission items：
remove Localize  before--10s+
remove Localize  after 300ms+

maliming:
Resolve #6373


hikalkan:
I merged, but then I think we may need to revert this. Because, using cache in manager may cause problems. When you make a change for a permission, then query it in the same UOW, you should get the latest change. However, cache has always delay and may be updated later.

I don't know how many seconds it takes to work with 1500 permissions, but 1-2 seconds is tolerable. How often you change permission of a user/role?

Can you think on that @maliming

PR #6501, opened 2020-12-05
Add Huaweiyun support for blob storing

charleszh-dev:
Huawei Cloud OBS
https://www.huaweicloud.com/en-us/product/obs.html


realLiangshiwei:
Hi,

I see you are customizing `ObsClient`, it will have additional maintenance costs.  we should use the official SDK.
But currently, the official SDK target is not `netstandard`. So I want to close this PR, after the official SDK supports `netstandard`, welcome you to contribute again.

charleszh-dev:
Ok, Thanks!

PR #6518, opened 2020-12-07
Remove BasicRepositoryBase from DI.

maliming:
Because `BasicRepositoryBase ` no longer implement the the `ITransientDependency `interface, I use `AbpRepositoryConventionalRegistrar `to register a custom `repository `interface.

```cs
public class ClientRepository : EfCoreRepository<IIdentityServerDbContext, Client, Guid>, IClientRepository
```


hikalkan:
Implemented in an alternative way: https://github.com/abpframework/abp/pull/6678

PR #6555, opened 2020-12-09
Update Part-1.md

haoyk:
abp4.0 AutoMapper


maliming:
we do not enable AutoMapper validation by default.

PR #6565, opened 2020-12-10
Use location.assign refresh page to avoid form re-submit.

maliming:
`reload `will cause the login form to re-submit.


hikalkan:
I didn't know the `location.assign` 😄 Doesn't matter how much you've experience, you always have a lot of basics to learn.

PR #6620, opened 2020-12-13
Update Part-3.md

vnetonline:
Removing @ from @() => GetDeleteConfirmationMessage(context) as it is wrong and doesn't need an @ at the front of ()


maliming:
Thanks @vnetonline

PR #6642, opened 2020-12-15
Spelling

274188A:
removed cookie scheme null
fixed spelling


gterdem:
Hello @274188A,

No, cookie scheme **must be explicitly set** to **null** to get it work since it adds its own cookie scheme otherwise. You can check it in here https://github.com/AzureAD/microsoft-identity-web/issues/133#issuecomment-739041457.

Also I mentioned about it at the end of the section under _Don't Forget To_ title with related link. If they change the behaviour in the future, we can also change it.

Thanks for spelling correction. 
Can you update your PR to only include the spelling correction?

gterdem (changes requested):
Hello @274188A,

No, cookie scheme **must be explicitly set to null** to get it work since it adds its own cookie scheme otherwise. You can check it from Tratcher's comment here https://github.com/AzureAD/microsoft-identity-web/issues/133#issuecomment-739041457.

Also I mentioned about it at the end of the section under _Don't Forget To_ title with related link. If they change the behaviour in the future, we can also change it.

Thanks for spelling correction.
Can you update your PR to only include the spelling correction?

274188A:
@gterdem - reverted as requested

PR #6663, opened 2020-12-16
Set to # only for null URL

realLiangshiwei:


enisn:
What is the main goal of this change? 
Can you describe this PR?

realLiangshiwei:
> What is the main goal of this change?
> Can you describe this PR?

Sure, An empty string is a valid link for Blazor. so set URL value to `#` only for null URL.

> About `MenuItem.Url.TrimStart('/')`

When deployed as a sub-application, like `https://localhost.com/subapplication`. 

Suppose we have a menu URL is `/home`, It will generate the link `https://localhost.com/home`. so we need to  remove `/` and It will generate the correct link `https://localhost.com/subapplication/home`

enisn:
Thank you @realLiangshiwei 
It looks OK now

enisn (approved):

PR #6676, opened 2020-12-16
Bulk Operations

enisn:
Resolved #6654 

## Summary
MongoDb has its own bulk operation options over MongoDb Driver. So it has a native implementation.
EfCore doesn't have a visible solution as exact bulk operation. But ChangeTracker helps to execute waiting operations at the same time. This PR has implementation which aims to improve MongoDb bulk operations behavior & performance.

~~There is a library for Bulk Operations for Ef Core
https://github.com/borisdj/EFCore.BulkExtensions
We can add an extension for this library, but I'm not sure, It doesn't have Ef Core 5.0 support.
This implementation will be included in another PR.
Default implementation solves main problem for now.~~

## Usage
You don't need to do anything more. Just use new methods over your repository:

```csharp
await repository.InsertManyAsync(entities); // collection of entities
```

```csharp
await repository.UpdateManyAsync(entities); // collection of entities
```

```csharp
await repository.DeleteManyAsync(entities); // collection of entities
```

PR #6679, opened 2020-12-16
Repository Bulk Operations | EfCore.BulkExtensions

enisn:
https://github.com/borisdj/EFCore.BulkExtensions Implementation for #6676


maliming:
Current **EFCore.BulkExtensions** is using **EF Core 3.1** and at the moment supports **Microsoft SQL Server(2008+) and SQLite.**

We have upgraded to NET 5 and need to support mysql, pgsql, etc.

enisn:
There is no acceptable & free alternative for bulk operations,
This problem can not be solved with single EF package, each provider requires their own implementations, So I'm closing this PR, 
There no single library soluton, 

_By the way Ef Core manages itself bulk operations since 5.0, if there are 3 or more entities with Added state, it executes single transaction with multiple insert._

PR #6701, opened 2020-12-18
Set job name for Hangfire background job integration

### LINKED ISSUE #6700: Set job name for Hangfire background job integration
https://support.abp.io/QA/Questions/689/BackgroundJobNameemails-ignored-when-using-hangfire

realLiangshiwei:
Resolve https://github.com/abpframework/abp/issues/6700

```csharp
public class SmsSenderJob : BackgroundJob<SmsSenderJobArgs>, ITransientDependency
{
    public override void Execute(SmsSenderJobArgs args)
    {

    }
}

public class SmsSenderJobArgs
{

}

public class EmailSenderJob : BackgroundJob<EmailSenderJobArgs>, ITransientDependency
{
    public override void Execute(EmailSenderJobArgs args)
    {
    }
}

[BackgroundJobName("EmailSenderJob")]
public class EmailSenderJobArgs
{
}
```

image

### DIFF
--- a/framework/src/Volo.Abp.BackgroundJobs.HangFire/Volo/Abp/BackgroundJobs/Hangfire/AbpBackgroundJobsHangfireModule.cs
+++ b/framework/src/Volo.Abp.BackgroundJobs.HangFire/Volo/Abp/BackgroundJobs/Hangfire/AbpBackgroundJobsHangfireModule.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Linq;
 using Hangfire;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Options;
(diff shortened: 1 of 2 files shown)


maliming (approved):

eneskoroglu:
Hi @maliming and @realLiangshiwei 

This commit seems does not solve problem. We still cannot set Job/JobName via `BackgroundJobName` attribute.  `DisplayNameFunc` seems null and does not set JobName. 

Regarding to this [link](https://support.abp.io/QA/Questions/689/BackgroundJobNameemails-ignored-when-using-hangfire) setting `DisplayNameFunc` like `BassaSolutions` user said solves problem. 

We should not have to write below custom code:

```
  var backgroundJobOptions =
                context.ServiceProvider.GetRequiredService<IOptions<AbpBackgroundJobOptions>>().Value;
   app.UseHangfireDashboard("/jobs", new DashboardOptions
    {
        DashboardTitle = "MyProject Background Jobs",
        DisplayNameFunc = (jobContext, job) =>
        {
            var jobType = job.Args.First().GetType();
            var abpJobType = backgroundJobOptions.GetJob(jobType);
            return abpJobType.JobName;
        }
    });
```

realLiangshiwei:
https://github.com/HangfireIO/Hangfire/blob/c23b01061860b4f5b5692e23e8f8cbdbe4681a8b/src/Hangfire.AspNetCore/HangfireApplicationBuilderExtensions.cs#L50

Hangfire gets `DashboardOptions` from DI by default.

I think there is no problem, however, you can create an issue if you think there is a problem.

PR #6705, opened 2020-12-18
Fix AbpStringExtensions

PMExtra:
For now, `"".RemovePreFix("x")` will return null.
It's not a expected result. I think we shouldn't make empty string be null when trim prefix or postfix.

### DIFF
--- a/framework/test/Volo.Abp.Core.Tests/System/StringExtensions_Tests.cs
+++ b/framework/test/Volo.Abp.Core.Tests/System/StringExtensions_Tests.cs
@@ -211,7 +211,7 @@ namespace System
             //empty case
             string.Empty.RemovePreFix("Test").ShouldBe(string.Empty);
 
-            "Home.Index".RemovePreFix("NotMatchedPostfix").ShouldBe("Home.Index");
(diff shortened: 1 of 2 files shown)


maliming (approved):

framework/test/Volo.Abp.Core.Tests/System/StringExtensions_Tests.cs
```diff
@@ -202,6 +205,12 @@ public void RemovePostFix_Tests()
         [Fact]
         public void RemovePreFix_Tests()
         {
+            //null case
+            (null as string).RemovePreFix("Test").ShouldBeNull();
+
+            //empty case
+            string.Empty.RemovePreFix("Test").ShouldBe(string.Empty);
+
             "Home.Index".RemovePreFix("NotMatchedPostfix").ShouldBe("Home.Index");
```
PMExtra: It's a confusing case that call Remove**Pre**Fix with "NotMatched**Post**Fix". Shall we correct the typo?

hikalkan:
Thanks @PMExtra.

PR #6725, opened 2020-12-21
RabbitMQ.ConnectionPool and  RabbitMQ.ChannelPool has Connection Close Bug Causing memory leaks

rocleegithub:
https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2?view=net-5.0

All these operations are atomic and are thread-safe with regards to all other operations on the ConcurrentDictionary<TKey,TValue> class. The only exceptions are the methods that accept a delegate, that is, AddOrUpdate and GetOrAdd. 

So Even if all pools have been disposed ,But not all Connections can be closed when they are created concurrently! Causing memory leaks


fmpfmp:
I also had  similar problem

PR #6727, opened 2020-12-21
Fix the issue for the ProxyScript can not refresh, And some comments are wrong.

ilonze:
Fix the issue for the ProxyScript can not refresh.
When after I use the url with “useCache=false” parameter request proxy script WebApi, the cache is not refreshed when I use the url without “useCache” parameter  request proxy script WebApi.
And revise some wrong for some comments.


maliming (changes requested):
hi

Can you provide steps that reproduce the problem.

ilonze:
> hi
> 
> Can you provide steps that reproduce the problem.

steps:

1. request url "/abp/serviceproxyscript?type=jquery" and got the script first
2. request url "/abp/serviceproxyscript?type=jquery&useCache=false" and got the new script
3. request url "/abp/serviceproxyscript?type=jquery" again, but the script not same with step 2.

I find the issue in gateway, gateway aggregate api definition for some service host, when add a new service host, the script result is can not refreshed. I find this code in `ProxyScriptManager`:
```
private string CreateCacheKey(ProxyScriptingModel model)
{
    return _jsonSerializer.Serialize(model).ToMd5();
}
```
Md5 encrypt by all of the class ProxyScriptingModel's properties for key.
I think UseCache property not to Md5.

maliming:
Thanks, I will check this, but I'll not use `[Newtonsoft.Json.JsonIgnore]` and `[System.Text.Json.Serialization.JsonIgnore]`

ilonze:
> Thanks, I will check this, but I'll not use `[Newtonsoft.Json.JsonIgnore]` and `[System.Text.Json.Serialization.JsonIgnore]`

IJsonSerializer is not use Newtonsoft.Json or System.Text.Json?

maliming:
You can do that:

```cs
private string CreateCacheKey(ProxyScriptingModel model)
{
	return _jsonSerializer.Serialize(new
	{
		model.GeneratorType,
		model.Modules,
		model.Controllers,
		model.Actions,
		model.Properties
	}).ToMd5();
}

```

maliming:
Your commits contain irrelevant content.

image

ilonze:
> Your commits contain irrelevant content.
> 
> image

These commits are all generated by merging them from the ABP code repository.
And revise some wrong for some comments.

maliming:
no problem, I will open another pr.

ilonze:
> no problem, I will open another pr.

Thanks!

PR #6749, opened 2020-12-22
Update Data-Filtering.md

xyfy:
I think it not need add **IsActiveFilterEnabled** to the filter expression.

### DIFF
--- a/docs/en/Data-Filtering.md
+++ b/docs/en/Data-Filtering.md
@@ -221,10 +221,10 @@ protected override Expression<Func<TEntity, bool>> CreateFilterExpression<TEntit
 {
     var expression = base.CreateFilterExpression<TEntity>();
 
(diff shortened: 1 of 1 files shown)


maliming:
https://github.com/abpframework/abp/blob/32402a99b07ea994efe1880e0655eff3c73e2166/framework/src/Volo.Abp.EntityFrameworkCore/Volo/Abp/EntityFrameworkCore/AbpDbContext.cs#L624-L641

xyfy:
@maliming  why we need add the **IsMultiTenantFilterEnabled** to the condition  ?

  Expression<Func<TEntity, bool>> multiTenantFilter = e => !IsMultiTenantFilterEnabled || EF.Property<Guid>(e, "TenantId") == CurrentTenantId;

if(IsMultiTenantFilterEnabled){
 multiTenantFilter = e => EF.Property(e, "TenantId") == CurrentTenantId;
}
what's the different of the above code ?

PR #6753, opened 2020-12-22
Avoid using array index in taghelper expressions

### LINKED ISSUE #6748: Avoid using array index in taghelper expressions.
https://github.com/abpframework/abp/pull/6747

image

realLiangshiwei:
Resolve https://github.com/abpframework/abp/issues/6748


maliming (changes requested):
https://github.com/abpframework/abp/pull/6755

maliming (approved):

PR #6779, opened 2020-12-23
DbMigrator creates the initial migration if no migration exists

yekalkan:


framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/Commands/CreateMigrationAndRunMigrator.cs
```diff
@@ -0,0 +1,61 @@
...
+            }
+
+            var dbMigratorProjectPath = GetDbMigratorProjectPath(commandLineArgs.Target);
+
+            if (dbMigratorProjectPath == null)
+            {
+                throw new Exception("DbMigrator is not found!");
+            }
+
+            var output = CmdHelper.RunCmdAndGetOutput("cd \"" + commandLineArgs.Target + "\" && dotnet ef migrations add Initial -s " + dbMigratorProjectPath);
```
ebicoglu, line 29: Backslash will not work in macOS. use https://docs.microsoft.com/tr-tr/dotnet/api/system.io.path.directoryseparatorchar?view=net-5.0

ebicoglu: check that if the `commandLineArgs.Target` is in a different drive is it still working? normally it doesn't work and you need to run `cd /d` command. check that /d command works in macOS.

framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/Commands/CreateMigrationAndRunMigrator.cs
```diff
@@ -0,0 +1,61 @@
...
+            }
+        }
+
+        private string GetDbMigratorProjectPath(string dbMigrationsFolderPath)
+        {
+            var srcFolder = Directory.GetParent(dbMigrationsFolderPath);
+
+            var dbMigratorFolderPath = Directory.GetDirectories(srcFolder.FullName).FirstOrDefault(d => d.EndsWith(".DbMigrator"));
+
+            if (dbMigratorFolderPath == null)
```
ebicoglu, line 43: throw exception here or make method name GetDbMigratorProjectPathOrNull

templates/app/aspnet-core/src/MyCompanyName.MyProjectName.Domain/Data/MyProjectNameDbMigrationService.cs
```diff
@@ -69,7 +86,8 @@ public async Task MigrateAsync()
                 Logger.LogInformation($"Successfully completed {tenant.Name} tenant database migrations.");
             }
 
-            Logger.LogInformation("Successfully completed database migrations.");
+            Logger.LogInformation("Successfully completed all database migrations.");
+            Logger.LogInformation("You can stop this process safely, if it doesn't stop itself.");
```
ebicoglu: **You can safely end this process...**

templates/app/aspnet-core/src/MyCompanyName.MyProjectName.Domain/Data/MyProjectNameDbMigrationService.cs
```diff
@@ -89,5 +107,81 @@ private async Task SeedDataAsync(Tenant tenant = null)
 
             await _dataSeeder.SeedAsync(tenant?.Id);
         }
+
+        private async Task<bool> CheckMigrationsAsync()
```
ebicoglu: Rename to `MigrationsFolderExists()`

templates/app/aspnet-core/src/MyCompanyName.MyProjectName.Domain/Data/MyProjectNameDbMigrationService.cs
```diff
@@ -89,5 +107,81 @@ private async Task SeedDataAsync(Tenant tenant = null)
...
+        {
+            var dbMigrationsProjectFolder = GetDbMigrationsProjectFolderPath();
+
+            if (!Directory.Exists(Path.Combine(dbMigrationsProjectFolder, "migrations")))
```
ebicoglu:
```
return Directory.Exists(Path.Combine(dbMigrationsProjectFolder, "migrations"))
```

templates/app/aspnet-core/src/MyCompanyName.MyProjectName.Domain/Data/MyProjectNameDbMigrationService.cs
```diff
@@ -89,5 +107,81 @@ private async Task SeedDataAsync(Tenant tenant = null)
...
+                fileName = "cmd.exe";
+            }
+
+            var procStartInfo = new ProcessStartInfo( fileName,
+                $"{argumentPrefix} \"abp create-migration-and-run-migrator {GetDbMigrationsProjectFolderPath()}\""
+            );
+
+            try
+            {
+                Process.Start(procStartInfo);
```
ebicoglu, line 142: you can use this [https://github.com/abpframework/abp/blob/dev/framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/Utils/CmdHelper.cs#L16](https://github.com/abpframework/abp/blob/dev/framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/Utils/CmdHelper.cs#L16) .. Same things are being done in this class

yekalkan: Template domain project doesn't reference Volo.Abp.Cli

framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/Commands/CreateMigrationAndRunMigrator.cs
yekalkan: > Backslash will not work in macOS. use https://docs.microsoft.com/tr-tr/dotnet/api/system.io.path.directoryseparatorchar?view=net-5.0

Backslash is not a part of the path. It is used as an escape char `\"`.

> check that if the commandLineArgs.Target is in a different drive is it still working? normally it doesn't work and you need to run cd /d command. check that /d command works in macOS.

As i know (and tested), It is guaranteed to be work in same drive.

ebicoglu (approved):

PR #6809, opened 2020-12-24
Transaction management made async

hikalkan:
We were using sync API to begin transaction. I converted it to async and it brings a significant scalability increase.
But we have a cost: We can not use LINQ extension methods directly on the `IRepository`. Instead, we should use `await IRepository.GetQueryableAsync()`. Because ABP internally may begin transaction with this method call.

This doesn't make breaking change, however I made a lot of methods Obsolete and we may remove them in the future which will be a major breaking change.

I updated all the usages including modules.

I will update all the documentation and tutorials. Also, I will write a blog post about that since it effects almost all applications.


maliming:
I wanted to do this, but I was stuck with IQueryable. 👍

hikalkan:
I want to continue to use LINQ extension methods directly on `IRepository`. However, we should find a way to defer DbContext creation until we execute the query. When the query executed, we can intercept it, create dbcontext & begin transaction, then delegate query execution to underlying provider (EF Core / MongoDB). However, I couldn't find how can I do it easily. We may need to create a custom LINQ provider for the repository but this is not easy to implement. Still thinking on. If you find a way, please write to me.

hikalkan:
I made a demo to demonstrate how we can do it with creating custom wrapper methods to all linq extension methods (Where, OrderBy, ToList, ToListAsync...).
https://github.com/hikalkan/samples/tree/master/EfCoreLinqWrapper/EfCoreLinqWrapper

It is a huge work, but separates creating dbcontext and starting transaction. We can defer starting transaction in this way.
This requires a lot of work and makes most of my work in this PR goes away. However, we will have two important gains;

* No need to use `await IRepository.GetQueryableAsync()` which is what I don't like.
* Fully support LINQ async methods without depending on EFCore/MongoDB. We currently partially support it.

I will continue to work on this.

hikalkan:
Separating dbcontext creation (and obtaining the queryable) and starting transaction didn't work for mongodb. It requires to relate the transaction while obtaining the queryable. It works for EF Core, but the implementation is complicated.

Anyway, currently, the best thing we can do is to suggest to use `await repository.GetQueryablyAsync` to obtain the queryable instead of directly write query on the repository. Since this would be a breaking change almost for all applications, I left the current usage possible. I made Obsolete wherever possible. We probably will remove `IRepository : IQueryable` inheritence in the next major version.

maliming (approved):

maliming (approved):

PR #6815, opened 2020-12-25
Lazy Resolve Dependencies To Gain Performance

maliming:
**Resolve too many services in the HTTP request will affect performance**.

image

asp net core: 
total: **4750**/sec **950**/sec avg.

local references abp with this PR:
total: **4500**/sec **900**/sec avg.

local references abp without this PR: 
total: **4250**/sec **850**/sec avg.

nuget package references abp without this PR: 
total: **4000**/sec **800**/sec avg.

local references abp with this PR and Injecting many `IRepository` in the `BookController's` constructor: 
total: **3750**/sec **750**/sec avg.

```cs
public BookController(
	IRepository<Book, Guid> bookRepository,
	IRepository<Book, Guid> bookRepository2,
	IRepository<Book, Guid> bookRepository3,
	IRepository<Book, Guid> bookRepository4,
	IRepository<Book, Guid> bookRepository5,
	IRepository<Book, Guid> bookRepository6,
	IRepository<Book, Guid> bookRepository7,
	IRepository<Book, Guid> bookRepository8,
	IRepository<Book, Guid> bookRepository9,
	IRepository<Book, Guid> bookRepository10)
{
	_bookRepository = bookRepository;
}
```


maliming:
The new results:

local references abp with this PR:
total: **4600**/sec **920**/sec avg.

hikalkan:
Please fix the conflicts ;)

framework/src/Volo.Abp.Features/Volo/Abp/Features/FeatureInterceptor.cs
```diff
@@ -29,7 +30,7 @@ public async override Task InterceptAsync(IAbpMethodInvocation invocation)
 
         protected virtual async Task CheckFeaturesAsync(IAbpMethodInvocation invocation)
         {
-            await _methodInvocationFeatureCheckerService.CheckAsync(
+            await _serviceProvider.GetRequiredService<IMethodInvocationFeatureCheckerService>().CheckAsync(
```
hikalkan: What if this interceptor is applied to a singleton service. In this case, `FeatureInterceptor` instance (related to the singleton service) will also be a single instance and `IServiceProvider` will be the root service provider that is never disposed. If we resolve `IMethodInvocationFeatureCheckerService` in every method call, then it may cause to a memory leak since `IMethodInvocationFeatureCheckerService` instaces will never be disposed.
Can you think about that. You can test for example to see if I am right.

We should extremely careful when resolving from `IServiceProvider`. It can be a problem if that's not a scoped provider.

For this example, if we cache the `IMethodInvocationFeatureCheckerService` instance (like we do with the new LazyServiceProvider), then we don't have this problem. Alternatively, we may keep using the constructor injection.

Can you check this PR from this point?

hikalkan (changes requested):

framework/src/Volo.Abp.Features/Volo/Abp/Features/FeatureInterceptor.cs
maliming: Now `IServiceScopeFactory` is used in the interceptor. Other places with HTTP context continue to use `IServiceProvider`

hikalkan (approved):

framework/src/Volo.Abp.Features/Volo/Abp/Features/FeatureInterceptor.cs
hikalkan: I rolled back MethodInvocationAuthorizationService for the same reason: https://github.com/abpframework/abp/commit/7d049361d716c3e720ea1e801e3cce1113fb0b8d

PR #6821, opened 2020-12-25
CMS-Kit Global Feature Definitions For Content And Tag

ilkayilknur:


modules/cms-kit/src/Volo.CmsKit.Domain.Shared/Volo/CmsKit/GlobalFeatures/ContentsFeature.cs
```diff
@@ -0,0 +1,17 @@
+﻿using JetBrains.Annotations;
+using Volo.Abp.GlobalFeatures;
+
+namespace Volo.CmsKit.GlobalFeatures
+{
+    [GlobalFeatureName(Name)]
+    public class ContentsFeature : GlobalFeature
```
enisn, line 7: Content is a core feature, I think content shouldn't be able to toggle on-off.
If content isn't active, Page or other features that uses content won't work properly.

ilkayilknur: I thought people may want to use the content component for their own pages instead of using the page feature so that they can use custom layouts for pages for now. We can talk about that in the review meeting and ask Halil.

enisn (approved):

cotur (approved):

PR #6853, opened 2020-12-28
Page progress

### LINKED ISSUE #5554: Blazor UI: Progress bar / loading indicator
Especially for server side API calls, we should show a progress bar or some kind of loading indicator. This should be done automatically wherever possible (we can do it with an HttpClient handler).

stsrki:
resolve #5554

A preview of new page progress for Create/Update/Delete operations. We can comment on it before merging.


framework/src/Volo.Abp.BlazoriseUI/AbpCrudPageBase.cs
```diff
@@ -356,20 +357,28 @@ protected virtual Task CloseEditModalAsync()
             return Task.CompletedTask;
         }
 
+        protected virtual async Task RunProcess(Func<Task> process)
```
hikalkan: Why this method still exists? Shouldn't we remove this?

hikalkan (changes requested):

framework/src/Volo.Abp.BlazoriseUI/AbpCrudPageBase.cs
stsrki: I was under impression that it should stay last time we spoke. I will remove it.

stsrki: Removed

docs/en/UI/Blazor/Page-Progress.md
```diff
@@ -1,3 +1,62 @@
 # Blazor UI: Page Progress
 
-TODO
\ No newline at end of file
+Page progress is the newest component introduced with Blazorise and ABP Framework. It is used to show scoped page progress on top of the page and to indicate to the user that currently a long running process is in the work. By default you don't need to do anything to show the progress indicator, as all the work is done automatically by the ABP Framework internals. This means that all calls to the ABP backend(through Rest API) will activate page progress and show the loading indicator.
```
hikalkan, line 3: >Page progress is the newest component introduced with Blazorise and ABP Framework

If we write this here, then we will need to change this document when we introduce a new component with Blazorise or ABP Framework.

hikalkan (approved):

docs/en/UI/Blazor/Page-Progress.md
hikalkan: I just made a few changes in the doc: https://github.com/abpframework/abp/commit/79575ff040747def938f1670f4926afca5b0b988 ;)

PR #6912, opened 2020-12-31
Remove MultipleActiveResultSets from the templates project connection string.

### LINKED ISSUE #6819: DbMigrator running occurred warning in 4.1.0-rc.2
Abp version 4.1.0-rc.2
User Interface: MVC
Database provider: EF Core

How to reproduce:
1. Run cmd "abp new Acme.BookStore --preview"
2. Run Acme.BookStore.DbMigrator and warning occurred

I changed the MultipleActiveResultSets in Default connection from true to false, then the warning disappeared.
 "Default": "Server=(LocalDb)\\MSSQLLocalDB;Database=BookStore;Trusted_Connection=True;MultipleActiveResultSets=true"
I also tried the  v3.3.2 and v4.0.2 start templates, there's no this warning even MultipleActiveResultSets value is true.

Below is the detail info, I would like to know why this happened, Thank you.

[18:03:50 WRN] Savepoints are disabled because Multiple Active Result Sets (MARS) is enabled. If 'SaveChanges' fails, then the transaction cannot be automatically rolled back to a known clean state. Instead, the transaction should be rolled back by the application before retrying 'SaveChanges'. See https://go.microsoft.com/fwlink/?linkid=2149338 for more information. To identify the code which triggers this warning, call 'ConfigureWarnings(w => w.Throw(RelationalEventId.SavepointsDisabledBecauseOfMARS))'.

image

real-zony:
Resolve https://github.com/abpframework/abp/issues/6819
For SQL Server, the connection string does not enable the `MultipleActiveResultSets=true` feature by default.

### DIFF
--- a/docs/en/CLI.md
+++ b/docs/en/CLI.md
@@ -105,7 +105,7 @@ abp new Acme.BookStore
 * `--preview`: Use latest preview version.
 * `--template-source` or `-ts`: Specifies a custom template source to use to build the project. Local and network sources can be used(Like `D:\local-template` or `https://.../my-template-file.zip`).
 * `--create-solution-folder` or `-csf`: Specifies if the project will be in a new folder in the output folder or directly the output folder.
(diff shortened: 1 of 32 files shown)


realLiangshiwei:
@GameBelial Also pro startup template

maliming (approved):

hikalkan (approved):

PR #6957, opened 2021-01-02
Fix shutdown logic for Volo.Abp.BackgroundWorkers.Quartz integration package

skomis-mm:
This will shutdown Quartz's `IScheduler` before upstream [AbpBackgroundWorkersModule.OnApplicationShutdown](https://github.com/abpframework/abp/blob/dev/framework/src/Volo.Abp.BackgroundWorkers/Volo/Abp/BackgroundWorkers/AbpBackgroundWorkersModule.cs#L31) will take place, so the check on [QuartzBackgroundWorkerManager.StopAsync()](https://github.com/abpframework/abp/blob/dev/framework/src/Volo.Abp.BackgroundWorkers.Quartz/Volo/Abp/BackgroundWorkers/Quartz/QuartzBackgroundWorkerManager.cs#L27) will work and won't pause the triggers.

The problem occurs with Quartz clustered configuration: when one of the instance is being shutdown, it pauses all the triggers which results on misfires across the cluster unless some application instance will (re)start again.


realLiangshiwei:
We have shut down the quartz server when on application shut down: https://github.com/abpframework/abp/blob/b4281ec1de5ff8e790ce5e6d41231c779421e9b4/framework/src/Volo.Abp.Quartz/Volo/Abp/Quartz/AbpQuartzModule.cs#L81

realLiangshiwei:
Hi,

I will look for a better way. thanks for your contribution. please see: https://github.com/abpframework/abp/issues/6960

PR #6978, opened 2021-01-03
fix identity module build problem

ismcagdas:


maliming:
The build problem has been resolved  by https://github.com/abpframework/abp/pull/6984

Maybe we should not reference`Volo.Abp.Identity.Domain` in  `Volo.Abp.PermissionManagement.Domain.Identity.`

https://github.com/abpframework/abp/commit/847f526041145b62376b760776829d5ce257da1c

PR #6997, opened 2021-01-04
Fix enum file template with string enum.

PMExtra:
I'm using `JsonStringEnumConverter` in backend json options.
Then `abp generate-proxy` generated the broken code like that:
```ts
export enum OrderStatus {
  WaitForPayment = WaitForPayment,
  Finished = Finished,
  Cancelled = Cancelled,
}
```

I think the value should be quoted as string value, so I make this fix.
After the fix, the generated code will be like:
```ts
export enum OrderStatus {
  WaitForPayment = 'WaitForPayment',
  Finished = 'Finished',
  Cancelled = 'Cancelled',
}
```

### DIFF
--- a/npm/ng-packs/packages/schematics/src/commands/api/files-enum/proxy/__namespace@dir__/__name@kebab__.enum.ts.template
+++ b/npm/ng-packs/packages/schematics/src/commands/api/files-enum/proxy/__namespace@dir__/__name@kebab__.enum.ts.template
@@ -2,7 +2,7 @@ import { mapEnumToOptions } from '@abp/ng.core';
 
 export enum <%= name %> {<%
   for (let member of members) { %>
-  <%= member.key %> = <%= member.value %>,<% } %>
+  <%= member.key %> = <%= quote(member.value) %>,<% } %>
 }
 
 export const <%= camel(name) %>Options = mapEnumToOptions(<%= name %>);
--- a/npm/ng-packs/packages/schematics/src/utils/text.ts
+++ b/npm/ng-packs/packages/schematics/src/utils/text.ts
@@ -11,7 +11,7 @@ export const dir = (text: string) =>
   strings.dasherize(text.replace(/\./g, '/').replace(/\/\//g, '/'));
 
 export const quote = (value: number | string) =>
-  typeof value === 'string' ? `'${value.replace(/'/g, `\\'`)}'` : value;
+  typeof value === 'string' ? `'${value.replace(/'/g, '\\\'')}'` : value;
 
 function _(text: string): string {
   return text.replace(/\./g, '_');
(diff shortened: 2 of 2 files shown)


npm/ng-packs/packages/schematics/src/utils/text.ts
```diff
@@ -10,6 +10,9 @@ export const macro = (text: string) => upper(snake(text));
 export const dir = (text: string) =>
   strings.dasherize(text.replace(/\./g, '/').replace(/\/\//g, '/'));
 
+// TODO: Is there a standard function to do this? There is a leak if the value contains single quote.
+export const quote = (value: number | string) => typeof value === 'string' ? `'${value}'` : value;
```
PMExtra: Is there a standard function to do this?

There is a leak if the value contains single quote, although enum won't contains any symbol.

npm/ng-packs/packages/schematics/src/commands/api/files-enum/proxy/__namespace@dir__/__name@kebab__.enum.ts.template
```diff
@@ -2,7 +2,7 @@ import { mapEnumToOptions } from '@abp/ng.core';
 
 export enum <%= name %> {<%
   for (let member of members) { %>
-  <%= member.key %> = <%= member.value %>,<% } %>
+  <%= member.key %> = <%= quote(member.value) %>,<% } %>
```
armanozak, line 5:
```suggestion
  <%= member.key %> = <%= stringify(member.value) %>,<% } %>
```

npm/ng-packs/packages/schematics/src/utils/text.ts
armanozak:
```suggestion
export const quote = (value: number | string) =>
  typeof value === 'string' ? `'${value.replace(/'/g, `\\'`)}'` : value;
```

armanozak (changes requested):
Hi @PMExtra,

Thanks a lot for discovering this and for your contribution. It is much appreciated.

I agree that we should use a standard method for stringifying enum values. I am suggesting JSON.stringify because of two reasons:

1. It escapes quotes successfully.
2. Since this is code generation, stringifying a number does not matter in templates.

You can commit suggestions if they look fine to you too.

npm/ng-packs/packages/schematics/src/utils/text.ts
PMExtra: It's awesome. But JSON used double quote symbol `"` rather than single quote symbol `'` which the Angular project prefer.

armanozak: Yeah, you're right.

How about wrapping strings with backticks? I have updated my suggestions accordingly.

PMExtra: It doesn't solve the essential. What about if the value contains backtick symbol?
I think single quote is better than backtick. They both have their own leak. But in most of cases, the `value` won't trigger the leak.

PMExtra: I think my origin solution is enough if there is no other standard function to surround a value as code content.

armanozak: Backtick is not as common as a single quote, but OK.

Take a look at my final suggestion. It should solve all cases.

armanozak (approved):

PR #7024, opened 2021-01-05
CMS Kit - Tag Refactoring

ilkayilknur:
- Some missing attributes have been added. 
- I've made some refactoring and changes on tag methods.


modules/cms-kit/src/Volo.CmsKit.Public.Application.Contracts/Volo/CmsKit/Public/Tags/ITagAppService.cs
```diff
@@ -7,6 +7,6 @@ namespace Volo.CmsKit.Public.Tags
 {
     public interface ITagAppService : IApplicationService
     {
-        Task<List<TagDto>> GetAllRelatedTagsAsync(GetRelatedTagsInput input);
+        Task<List<TagDto>> GetAllRelatedTagsAsync(string entityType, string entityId);
```
enisn, line 10: I'm not sure about this changes, as @cotur 's say each input and output parameters should be object in an AppService. That's why we created 'GetRelatedTagsInput' type.

enisn (approved):
Rest of them LGTM

modules/cms-kit/src/Volo.CmsKit.Public.Application.Contracts/Volo/CmsKit/Public/Tags/ITagAppService.cs
ilkayilknur: All other application services and controllers use the same convention. That's why I've made the changes. If those implementations are against the convention we can change the other usages.

enisn: Ok, it's better now. 👍

cotur (approved):

PR #7085, opened 2021-01-07
CmsKit - Remove huge content value from list items

enisn:
## Summary
This PR removes `Value` from `ContentDto` to better performance. It can't be transfered while listing. It might be hundreds of line rich text.


ilkayilknur (commented):
LGTM. We should also update the tests, right? @enisn

enisn:
Canyou re-review now @ilkayilknur ?

cotur (approved):

ilkayilknur (approved):

PR #7103, opened 2021-01-08
CmsKit - Tag Definitions

### LINKED ISSUE #7102: CMS Kit - Tag System Configuration Changes
Changes were discussed in a meeting. We will review the design after implementation.

enisn:
## Summary
Initial version of Cms Kit Tag definitions. Implementation is similar with Reaction definitions.

---
- Closes #7102


modules/cms-kit/src/Volo.CmsKit.Domain/Volo/CmsKit/Tags/DefaultTagDefinitionStore.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        public Task<TagDefiniton> GetTagDefinitionOrNullAsync([NotNull] string entityType)
+        {
+            Check.NotNullOrWhiteSpace(entityType, nameof(entityType));
+            return Task.FromResult(options.Tags.GetOrDefault(entityType));
+        }
+
+        public Task<List<TagDefiniton>> GetTagDefinitionsAsync()
+        {
+            return Task.FromResult(options.Tags.Values.ToList());
```
ilkayilknur: Is there a reason why we implemented these methods are `async`? They are not `virtual`. We don't need to create an additional `Task` object here.

modules/cms-kit/src/Volo.CmsKit.Domain/Volo/CmsKit/Tags/DefaultTagDefinitionStore.cs
```diff
@@ -0,0 +1,38 @@
...
+        }
+
+        public Task<bool> IsDefinedAsync([NotNull] string entityType)
+        {
+            return Task.FromResult(options.Tags.ContainsKey(entityType));
```
ilkayilknur: We should add `Check.NotNullOrWhiteSpace(entityType, nameof(entityType));`  check here like we did inside `GetTagDefinitionOrNullAsync` method.

modules/cms-kit/src/Volo.CmsKit.Domain/Volo/CmsKit/Tags/ITagManager.cs
```diff
@@ -26,5 +27,7 @@ public interface ITagManager : IDomainService
             [NotNull] string name,
             Guid? tenantId = null,
             CancellationToken cancellationToken = default);
+
+        Task<List<TagDefiniton>> GetTagDefinitionsAsync(CancellationToken cancellationToken = default);
```
ilkayilknur: No need to define this method as `async`.

modules/cms-kit/src/Volo.CmsKit.Domain/Volo/CmsKit/PolicySpecifiedDefinition.cs
```diff
@@ -0,0 +1,37 @@
...
+            [CanBeNull] string deletePolicy = null)
+        {
+            DisplayName = displayName;
+            CreatePolicy = createPolicy;
+            DeletePolicy = deletePolicy;
+            UpdatePolicy = updatePolicy;
+        }
+
+        [CanBeNull]
+        public virtual ILocalizableString DisplayName { get; }
```
ilkayilknur: Why do we need `DisplayName` in this type? I guess people can use this to show special display names for entity types. Maybe we should introduce another base type or interface for this. Can you work on this? I think we can improve the design here.

modules/cms-kit/src/Volo.CmsKit.Domain/Volo/CmsKit/Tags/TagDefinitionDictionary.cs
```diff
@@ -0,0 +1,20 @@
+﻿using JetBrains.Annotations;
+using System.Collections.Generic;
+using Volo.Abp.Localization;
+
+namespace Volo.CmsKit.Tags
+{
+    public class TagDefinitionDictionary : Dictionary<string, TagDefiniton>
+    {
+        public void AddOrReplace(
```
ilkayilknur: There is a method called `AddIfNotExists` inside `AbpCollectionExtensions`. We can use that for adding definitions. I don't see any use case for replacing definitions. Am I missing anything?

ilkayilknur (changes requested):

modules/cms-kit/src/Volo.CmsKit.Domain/Volo/CmsKit/Tags/TagDefinitionDictionary.cs
enisn: The pattern comes from here: https://github.com/abpframework/abp/blob/a92a5c98ab95aab4c456d9068c141e6fddba8109/modules/cms-kit/src/Volo.CmsKit.Domain/Volo/CmsKit/Reactions/ReactionDefinitionDictionary.cs#L9

modules/cms-kit/src/Volo.CmsKit.Domain/Volo/CmsKit/Tags/DefaultTagDefinitionStore.cs
enisn: You're right about missing `virtual` modifier. I'm adding it. But other way, `ITagDefinitionStore` declares async methods, because implementation can be changed with database, or any rest request implementation.

Also ReactionDefinitionStore uses synchronous task implementation.
https://github.com/abpframework/abp/blob/a92a5c98ab95aab4c456d9068c141e6fddba8109/modules/cms-kit/src/Volo.CmsKit.Domain/Volo/CmsKit/Reactions/DefaultReactionDefinitionStore.cs

ilkayilknur: It looks like `virtual` is still missing.

cotur (approved):

modules/cms-kit/src/Volo.CmsKit.Domain/Volo/CmsKit/CmsKitDomainModule.cs
```diff
@@ -28,7 +35,25 @@ public override void ConfigureServices(ServiceConfigurationContext context)
...
                 options.Reactions.AddOrReplace(StandardReactions.Pray);
+
             });
+
+
+            if (GlobalFeatureManager.Instance.IsEnabled<TagsFeature>())
+            {
+                Configure<CmsKitTagOptions>(options =>
+                {
+                    if (GlobalFeatureManager.Instance.IsEnabled<PagesFeature>())
```
ilkayilknur: We don't need tags inside pages. Can we remove this?

modules/cms-kit/src/Volo.CmsKit.Domain/Volo/CmsKit/Tags/EntityNotTaggableException.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+using Volo.Abp;
+
+namespace Volo.CmsKit.Tags
+{
+    [Serializable]
+    public class EntityNotTaggableException : BusinessException
+    {
+        public EntityNotTaggableException(string code = null, string message = null, string details = null, Exception innerException = null, Microsoft.Extensions.Logging.LogLevel logLevel = Microsoft.Extensions.Logging.LogLevel.Warning) : base(code, message, details, innerException, logLevel)
```
ilkayilknur: Can we split this code into two lines? It's too long for a single line. We can also add namespaces to the top.

ilkayilknur (changes requested):

modules/cms-kit/src/Volo.CmsKit.Domain/Volo/CmsKit/Tags/ITagManager.cs
enisn: https://github.com/abpframework/abp/pull/7103#discussion_r555686131

ilkayilknur (approved):

PR #7114, opened 2021-01-10
fix totalCount in IdentityRole GetList in case of using filter

kamalseisy:
totalCount always return the total count without filtering


cotur (approved):

maliming:
Thanks @kamalseisy  👍

PR #7115, opened 2021-01-10
Make AccountEmailer.NormalizeReturnUrl method protected

olicooper:
This will allow the `SendPasswordResetLinkAsync` method to be overridden but still use the base `NormalizeReturnUrl` method.
Also consider making this `virtual` if you would like.


realLiangshiwei (approved):

realLiangshiwei:
Thanks @olicooper

PR #7128, opened 2021-01-11
Add the CountAsync method of the Identityserver module Repostiory

dazsco:


realLiangshiwei:
Thanks a lot, @dazsco, I think this is a bug and we should fix it in 4.1. I will open a new PR in the rel-4.1 branch.

PR #7148, opened 2021-01-11
angular: Fix AuthGuard returning 'true' when user not authorized

olicooper:
I believe this is a bug? By returning false the page doesn't continue to load, and the user is redirected to the '/Account/Login' page as expected without causing permission errors to appear on my app.


mehmet-erim:
Thanks @olicooper

mehmet-erim (approved):

PR #7155, opened 2021-01-12
Fix bug in PagerInfo

CharlieGomez:
Correction of the bug in "pagerinfo" did not show the correct start number within the datatable.


maliming (approved):

maliming:
Thanks @CharlieGomez 👍

PR #7203, opened 2021-01-14
Enhance GetAsync to eliminate EF warnings

### LINKED ISSUE #7188: Remove EF warning logs 'The query uses a row limiting operator'
Since .NET 5, Microsoft has un-obsoleted the [CoreEventId.FirstWithoutOrderByAndFilterWarning](https://docs.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.diagnostics.coreeventid.firstwithoutorderbyandfilterwarning?f1url=%3FappId%3DDev16IDEF1%26l%3DEN-US%26k%3Dk(Microsoft.EntityFrameworkCore.Diagnostics.CoreEventId.FirstWithoutOrderByAndFilterWarning);k(DevLang-csharp)%26rd%3Dtrue&view=efcore-5.0) and [CoreEventId.RowLimitingOperationWithoutOrderByWarning](https://docs.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.diagnostics.coreeventid.rowlimitingoperationwithoutorderbywarning?view=efcore-5.0)

I get warning logs like this when using methods like the default `IReadOnlyAppService.GetAsync` method and queries with 
> [WRN][54][Microsoft.EntityFrameworkCore.Query] The query uses a row limiting operator ('Skip'/'Take') without an 'OrderBy' operator. This may lead to unpredictable results.

_Note: The log above appears for FirstOrDefault() calls as well as ToList() even though the `CoreEventId.FirstWithoutOrderByAndFilterWarning` warning should be written. I'm not sure why this happens but its probably a bug with EF._

Is there a way to apply something like the  `AbstractKeyReadOnlyAppService.ApplyDefaultSorting` method to the EfCoreRepository for FindAsync calls (and default back to sorting by Entity.Id if CreationTime doesn't exist) or change it to `SingleOrDefaultAsync`?
https://github.com/abpframework/abp/blob/158158e0aba708768657a97490a610aaf46d87d8/framework/src/Volo.Abp.EntityFrameworkCore/Volo/Abp/Domain/Repositories/EntityFrameworkCore/EfCoreRepository.cs#L412

realLiangshiwei:
Resolve https://github.com/abpframework/abp/issues/7188

https://docs.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.diagnostics.coreeventid.firstwithoutorderbyandfilterwarning?view=efcore-5.0

> A query uses First/FirstOrDefault operation without OrderBy and filter which may lead to unpredictable results.

### DIFF
--- a/framework/src/Volo.Abp.EntityFrameworkCore/Volo/Abp/Domain/Repositories/EntityFrameworkCore/EfCoreRepository.cs
+++ b/framework/src/Volo.Abp.EntityFrameworkCore/Volo/Abp/Domain/Repositories/EntityFrameworkCore/EfCoreRepository.cs
@@ -409,7 +409,7 @@ namespace Volo.Abp.Domain.Repositories.EntityFrameworkCore
         public virtual async Task<TEntity> FindAsync(TKey id, bool includeDetails = true, CancellationToken cancellationToken = default)
         {
             return includeDetails
-                ? await (await WithDetailsAsync()).FirstOrDefaultAsync(e => e.Id.Equals(id), GetCancellationToken(cancellationToken))
+                ? await (await WithDetailsAsync()).OrderBy(x => x.Id).FirstOrDefaultAsync(e => e.Id.Equals(id), GetCancellationToken(cancellationToken))
                 : await (await GetDbSetAsync()).FindAsync(new object[] {id}, GetCancellationToken(cancellationToken));
         }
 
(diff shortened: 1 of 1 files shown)


maliming (approved):

maliming:
When we include entity navigation properties, we need to **order by**, otherwise, there will be a warning.

Issue #7242, opened 2021-01-15
DbMigrator: For mongodb, we don't need to create a migration file


realLiangshiwei:
For MongoDB, we can remove the `MigrationsFolderExists` method

![image](https://user-images.githubusercontent.com/16813853/104677802-0db05b00-5725-11eb-8970-f61320d0c710.png)

PR #7244, opened 2021-01-15
Fix DbMigrator for mongodb

yekalkan:
resolves https://github.com/abpframework/abp/issues/7242

### DIFF
--- a/templates/app/aspnet-core/src/MyCompanyName.MyProjectName.Domain/Data/MyProjectNameDbMigrationService.cs
+++ b/templates/app/aspnet-core/src/MyCompanyName.MyProjectName.Domain/Data/MyProjectNameDbMigrationService.cs
@@ -41,7 +41,7 @@ namespace MyCompanyName.MyProjectName.Data
         {
             try
             {
-                if (await DbMigrationsProjectExistsAsync() && !await MigrationsFolderExistsAsync())
+                if (DbMigrationsProjectExists() && !MigrationsFolderExists())
                 {
                     await AddInitialMigrationAsync();
                     return;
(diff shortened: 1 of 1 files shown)


templates/app/aspnet-core/src/MyCompanyName.MyProjectName.Domain/Data/MyProjectNameDbMigrationService.cs
```diff
@@ -108,14 +108,21 @@ private async Task SeedDataAsync(Tenant tenant = null)
             await _dataSeeder.SeedAsync(tenant?.Id);
         }
 
-        private async Task<bool> MigrationsFolderExists()
+        private async Task<bool> DbMigrationsProjectExistsAsync()
```
realLiangshiwei: It is not an asynchronous method, I think the method should be:  `private bool DbMigrationsProjectExists()`

templates/app/aspnet-core/src/MyCompanyName.MyProjectName.Domain/Data/MyProjectNameDbMigrationService.cs
```diff
@@ -108,14 +108,21 @@ private async Task SeedDataAsync(Tenant tenant = null)
...
+
+        private async Task<bool> MigrationsFolderExistsAsync()
         {
             var dbMigrationsProjectFolder = GetDbMigrationsProjectFolderPath();
 
             return Directory.Exists(Path.Combine(dbMigrationsProjectFolder, "migrations"));
         }
 
-        private async Task AddInitialMigration()
+        private async Task AddInitialMigrationAsync()
```
realLiangshiwei: Is there a reason to use asynchrony? It is not an asynchronous method.

realLiangshiwei (approved):

PR #7250, opened 2021-01-15
Remove the redis connection string when the public website option is not specified

realLiangshiwei:


yekalkan:
it's better to add `*.json` to https://github.com/abpframework/abp/blob/dev/framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/ProjectBuilding/Building/Steps/TemplateCodeDeleteStep.cs#L11

yekalkan (approved):

PR #7293, opened 2021-01-18
Change AbpLazyServiceProvider lifetime to scoped.

### LINKED ISSUE #7227: Change AbpLazyServiceProvider DI lifetime to IScopedDependency?
Abp `AbpLazyServiceProvider` has a transient dependency, which if I understand correctly means that each service injected in a request gets a new instance? Would it be better to make it scoped so that the `CachedServices` can be used by all services for each request? Forgive me if I have misunderstood the mechanisms of DI.

https://github.com/abpframework/abp/blob/34d66ee70f6d2bf719c251331bbf1eda086960c3/framework/src/Volo.Abp.Core/Volo/Abp/DependencyInjection/AbpLazyServiceProvider.cs#L7-L8

maliming:
Resolve #7227

### DIFF
--- a/framework/src/Volo.Abp.Core/Volo/Abp/DependencyInjection/AbpLazyServiceProvider.cs
+++ b/framework/src/Volo.Abp.Core/Volo/Abp/DependencyInjection/AbpLazyServiceProvider.cs
@@ -4,7 +4,7 @@ using Microsoft.Extensions.DependencyInjection;
 
 namespace Volo.Abp.DependencyInjection
 {
-    public class AbpLazyServiceProvider : IAbpLazyServiceProvider, ITransientDependency
+    public class AbpLazyServiceProvider : IAbpLazyServiceProvider, IScopedDependency
     {
         protected IDictionary<Type, object> CachedServices { get; set; }
 


jerviscui:
I have a question.
Just like the following code.

If `IAbpLazyServiceProvider` is `IScopedDependency`, the `Baby` life will be changed to scope  and is no longer transient.
Am I right?

```csharp
public class Baby : ITransientDependency
{
}

public class ServiceA : ITransientDependency
{
    public IAbpLazyServiceProvider AbpLazyServiceProvider { get; set; }

    public Baby Baby => AbpLazyServiceProvider.LazyGetRequiredService<Baby>();
}

public class ServiceB : ITransientDependency
{
    public IAbpLazyServiceProvider AbpLazyServiceProvider { get; set; }

    public Baby Baby => AbpLazyServiceProvider.LazyGetRequiredService<Baby>();
}

public class MainA
{
    void Test()
    {
        var aBaby = new ServiceA().Baby;
        var bBaby = new ServiceB().Baby;
    }
}
```

hikalkan:
I am unsure about that and need to think detailed if we have problems (like memory leak), especially if the current scope is not HttpContext.

hikalkan:
>If IAbpLazyServiceProvider is IScopedDependency, the Baby life will be changed to scope and is no longer transient.
Am I right?

Yes, this PR makes that (while your sample is not complete, I assume you are resolving all instance in the same scope).

maliming:
Move to backlog.

PR #7304, opened 2021-01-18
Move CreateModal.Hide() to created/updated methods

### LINKED ISSUE #7291: AbpCrudPageBase.cs
` protected virtual async Task CreateEntityAsync()`
... 
move 
>> CreateModal.Hide();

```
       protected virtual Task OnCreatedEntityAsync()
        {
             CreateModal.Hide();
            return Task.CompletedTask;
        }
```

use case: Modal "Create dialog" with Button "Save & New" to create records in one go

stsrki:
resolves #7291


hikalkan:
We currently do 2 things after creating an entity:

1. `GetEntitiesAsync()`
2. `CreateModal.Hide()`

Now, we are moving the second one into `OnCreatedEntityAsync()`. I see a few problems with this approach:

* Why not moving `GetEntitiesAsync()` also into the `OnCreatedEntityAsync()` (to be consistent)
* `OnCreatedEntityAsync()` was designed as a hook point for end-developers. But currently, it performs some essential work. So, end-developer now much call `base.OnCreatedEntityAsync()` to preserve the existing behaviour while previously it was not necessary. This is a breaking change.

What do you think?

Edit: Alternatively, end-developer can completely override `CreateEntityAsync()` to customize it. It is not a long method.

stsrki:
@hikalkan I agree. Moving `GetEntitiesAsync()` into `OnCreatedEntityAsync` makes sense. The problem could potentially be in how users are using the `OnCreatedEntityAsync`. For example when whey override they have two options

1.
```cs
protected override async Task OnCreatedEntityAsync()
{
   await base.OnCreatedEntityAsync();

  // do something
}
```

2.

```cs
protected override async Task OnCreatedEntityAsync()
{
  // do something

   await base.OnCreatedEntityAsync();
}
```

If they expect `GetEntitiesAsync()` to already be called it could be a breaking change. So no.2 would not work in this case.

I can easily make changes. It's your call :)

stsrki:
@hikalkan Can you check PR now? Thanks

hikalkan (approved):
That's good now. Thanks!

PR #7310, opened 2021-01-18
Improve flexibility of email validation in the ValidationHelper

### LINKED ISSUE #7265: Login fails for 'admin@localhost' due to incorrect email validation
The `ValidationHelper.IsValidEmailAddress` doesn't validate against `admin@localhost` but does work for e.g. `admin@localhost.com` (contains a dot followed by a character). Because the validation helper returns false, the email address is treated like a username when using the login form, so the user gets 'invalid username or password' instead of being successfully logged in. 
For the login I think we should be a little less strict on the rules and use something like this (HTML5 email regex):
```
^[a-zA-Z0-9.!#$%&’*+/=?^_`{|}~-]+@[a-zA-Z0-9-]+(?:\.[a-zA-Z0-9-]+)*$
```
This site https://emailregex.com/ has a few options to consider.

olicooper:
Resolves https://github.com/abpframework/abp/issues/7265

* ValidationHelper.EmailRegEx can now be modified to suit the users needs
* Regex updated to be more inclusive. Taken from the WHATWG specification https://html.spec.whatwg.org/#valid-e-mail-address

WHATWG state that the rules defined by RFC5322 are not good enough. They offer an alternative regex which they say violates the standard, but it works better. See https://html.spec.whatwg.org/#valid-e-mail-address

Instead of using two properties `EmailRegex` and `StrictEmailRegex`, I propose we simply update the current regex and use a more inclusive regex. If you think we should include both then I will update the PR to include both and use the W3C regex for the less strict one. I have updated the tests to include a test for `admin@localhost`, which now passes. The old `1024x768@60hz` test would have passed for the same reason so I remove it.

### DIFF
--- a/framework/src/Volo.Abp.Validation/Volo/Abp/Validation/ValidationHelper.cs
+++ b/framework/src/Volo.Abp.Validation/Volo/Abp/Validation/ValidationHelper.cs
@@ -4,7 +4,9 @@ namespace Volo.Abp.Validation
 {
     public class ValidationHelper
     {
-        private const string EmailRegEx = @"[a-z0-9!#$%&'*+/=?^_`{|}~-]+(?:\.[a-z0-9!#$%&'*+/=?^_`{|}~-]+)*@(?:[a-z0-9](?:[a-z0-9-]*[a-z0-9])?\.)+[a-z0-9](?:[a-z0-9-]*[a-z0-9])?";
+        // Taken from W3C as an alternative to the RFC5322 specification: https://html.spec.whatwg.org/#valid-e-mail-address
+        // The RFC5322 regex can be found here: https://emailregex.com/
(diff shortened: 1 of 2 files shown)


olicooper:
W3C state that the rules defined by RFC5322 are not good enough. They offer an alternative regex which they say violates the standard, but it works better. See https://html.spec.whatwg.org/#valid-e-mail-address

Instead of creating a `EmailRegex` and `StrictEmailRegex`, I propose we simply update the current regex and use a more inclusive regex. If you think we should include both then I will update the PR to include both and use the W3C regex for the less strict one. I have updated the tests to include a test for `admin@localhost`, which now passes. The old `1024x768@60hz` test would have passed for the same reason so I remove it.

maliming:
@hikalkan  What do you think?

hikalkan:
I think that seems fine.

hikalkan (approved):

PR #7422, opened 2021-01-22
CmsKit - Make protected TagEntityTypeDefiniton parameterless constructor

enisn:


ilkayilknur:
I'm going to merge this but we need to find a better solution to specify the `EntityTypes`. It's not a good practice to use a dictionary and force developers to use only a specific method. @enisn

ilkayilknur (approved):

PR #7441, opened 2021-01-25
Use IDistributedCache in AbpCorsPolicyService,ClientStore,ResourceStore.

### LINKED ISSUE #6969: Optimize IdentityServer data access on token request
When we request access token from IDS endpoint, it performs 6-7 database queries. We can try to optimize these database operations. For example, we may exclude details while querying where it is not needed or try to reduce queries if possible.

Note: Adding `app.UseUnitOfWork()` middleware before `app.UseIdentityServer()` (#6968) helps to reduce DbContext creation. However, I think we can optimize it more.

maliming:
Resolve #6969

https://identityserver4.readthedocs.io/en/latest/reference/options.html#caching


maliming:
We have to consider entity changes to refresh the cache.

maliming:
Move to https://github.com/abpframework/abp/pull/7856

PR #7468, opened 2021-01-26
Introduce ICacheNameProvider.

maliming:
This way we can dynamically set the name for the cache item. For example, a generic cache item.

```cs
[Serializable]
public class MyCacheItem<T> : ICacheNameProvider
{
    public T Value { get; set; }

    public string GetCacheName()
    {
        return typeof(T).Name;
    }
}
```


hikalkan:
I think we don't need to such a dinamic way of determining the cache name. I didn't like the way we determine the name:

````csharp
return ((ICacheNameProvider)Activator.CreateInstance(cacheItemType)).GetCacheName();
````

We are creating a new unnecessary instance just to get the name. Also, what if developer wants to determine the name based on the values of the properties in the cache item class? When we create it like that, all properties will have default values and such a logic can't work.

I think we don't need that, closing. Thanks for the work ;)

PR #7483, opened 2021-01-27
Cms Kit - Change TagEntityTypeDefinition Configuration

### LINKED ISSUE #7443: Designing a better solution to add entity types
We need to find a better solution to add and configure `EntityTypes` inside options. 
Design a better solution, present it to the team, and implement.

enisn:
Closes #7443

### DIFF
--- a/modules/cms-kit/src/Volo.CmsKit.Domain/Volo/CmsKit/Tags/TagEntityTypeDefinitionDictionary.cs
+++ b/modules/cms-kit/src/Volo.CmsKit.Domain/Volo/CmsKit/Tags/TagEntityTypeDefinitionDictionary.cs
@@ -4,17 +4,7 @@ using Volo.Abp.Localization;
 
 namespace Volo.CmsKit.Tags
 {
-    public class TagEntityTypeDefinitionDictionary : Dictionary<string, TagEntityTypeDefiniton>
+    public class TagEntityTypeDefinitionDictionary : HashSet<TagEntityTypeDefiniton>
     {
-        public void AddOrReplace(
-            [NotNull] string entityType,
-            [CanBeNull] ILocalizableString displayName = null,
-            [CanBeNull] string createPolicy = null,
-            [CanBeNull] string updatePolicy = null,
-            [CanBeNull] string deletePolicy = null
-            )
-        {
-            this[entityType] = new TagEntityTypeDefiniton(entityType, displayName, createPolicy, updatePolicy, deletePolicy);
-        }
     }
 }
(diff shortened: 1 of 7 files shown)


modules/cms-kit/src/Volo.CmsKit.Domain/Volo/CmsKit/Tags/TagEntityTypeDefinitionDictionary.cs
```diff
@@ -4,17 +4,7 @@
 
 namespace Volo.CmsKit.Tags
 {
-    public class TagEntityTypeDefinitionDictionary : Dictionary<string, TagEntityTypeDefiniton>
+    public class TagEntityTypeDefinitionDictionary : HashSet<TagEntityTypeDefiniton>
```
ilkayilknur: Why did we use `HashSet<T>` instead of `List<T>`?

ilkayilknur (changes requested):
Rest LGTM

modules/cms-kit/src/Volo.CmsKit.Domain/Volo/CmsKit/Tags/TagEntityTypeDefinitionDictionary.cs
enisn: HastSet is used here to prevent adding same EntityType more than once.

enisn: You may check this test shows what I mean
https://github.com/abpframework/abp/pull/7483/files#diff-e7adfd293ef282fbfef31b79f5e3fa9a4e8441222192123b1e837fd88f33d238R54

ilkayilknur: `AddIfNotContains` is an extension method inside ABP Framework. If you implement `IEquatable<T>` , you can use this method for all collections implementing `ICollection<T>`.

enisn: Yes, It implements `IEquatable<T>`

But `Add` method is still available for collections. If user configures same `EntityType` more than once, it'll create a bug, we'll use always first one. I just wanted  to be sure that collection will never contains same EntityType more than once.

ilkayilknur: I'm not sure implementing `IEquatable<T>` prevents duplicate adding. I tested in a sample project and it did not work. Can you test again?

enisn: You're right, test passes because tries to add same instance. **HashSet** uses `IEqualityComparer<T>.Default` and it compares if they are same reference or not. It doesn't use `IEquatable<T>`. I've changed it as **List** and now works perfectly 👌 
@ilkayilknur

ilkayilknur (approved):

PR #7496, opened 2021-01-28
add AutoMapper.Collection

Nokecy:


enisn:
Hi @Nokecy,
Can you introduce main point of adding that dependency?

PR #7509, opened 2021-01-28
Multi tenancy & default localization patch

### LINKED ISSUE #5215: When tiered templates, tenants of independent databases cannot login (AngularUI).
Related https://github.com/abpframework/abp/issues/5213

Should add tenant key to the request.

image

### LINKED ISSUE #7508: Use the default language setting if user hasn't selected a language.
Use the default language setting (tenant aware) if user hasn't selected a language and the browser's language is not one of the supporterd languages.
I put this to the next patch because there was such a setting but not properly used until now.

hikalkan:
- Resolves #7508
- Resolves #5215

Some new enhancements to solve the problems;

* `[IgnoreMultiTenancy]` now can be added to a DbContext class (and to interface of the DbContext) to indicate that this DbContext should always use the host connection string because it's entities can't be stored in the tenant databases. This is important when we use separate db for tenants. Declare this attribute for a dbcontext that has no entity implements `IMultiTenant`. Even if only one entity implements `IMultiTenant`, do not use this attribute and see the next item. This was actually done with https://github.com/abpframework/abp/commit/16fdc527cc85f6fd3c2565df30753b5b9a2f80f2 and I had implemented it for the IdentityServer
* If an entity doesn't implement `IMultiTenant` interface and you use the repository of that entity, it guarantees that it always use the host database. This is also useful if a dbcontext has some entities multi-tenant while some others are not and you couldn't add `IgnoreMultiTenancy` attribute to it.

I updated related modules based on that. This PR will solve the separate tenant database problems.

### DIFF
--- a/framework/src/Volo.Abp.AspNetCore.MultiTenancy/Volo/Abp/AspNetCore/MultiTenancy/MultiTenancyMiddleware.cs
+++ b/framework/src/Volo.Abp.AspNetCore.MultiTenancy/Volo/Abp/AspNetCore/MultiTenancy/MultiTenancyMiddleware.cs
@@ -1,7 +1,15 @@
-﻿using System.Threading.Tasks;
+﻿using System;
+using System.Globalization;
+using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Localization;
+using Microsoft.AspNetCore.RequestLocalization;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
 using Volo.Abp.DependencyInjection;
+using Volo.Abp.Localization;
 using Volo.Abp.MultiTenancy;
+using Volo.Abp.Settings;
 
 namespace Volo.Abp.AspNetCore.MultiTenancy
 {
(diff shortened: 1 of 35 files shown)


maliming (approved):

maliming:
I merged, But I will check the `TryGetRequestCultureAsync` method  in `MultiTenancyMiddleware `

olicooper:
So can we add `[IgnoreMultiTenancy]` to a single entity in a DbContext that is multi-tenancy aware? I can't see any code that suggests we can.

I have some 'archive' tables which I'd like to keep in the host database, but allow tenants to have their own database for all other entities. I imagine I'd have to create a new DbContext for the archive entities and add `[IgnoreMultiTenancy]` to it right?

hikalkan:
>So can we add [IgnoreMultiTenancy] to a single entity in a DbContext that is multi-tenancy aware? I can't see any code that suggests we can.

`[IgnoreMultiTenancy]` is for a dbcontext. If all entities are not multi-tenant, you can add this to indicate it. However, for most cases, you don't need it. Actually you only need it if you use Dapper where you directly go over DbContext rather than entity repositories. If you use repositories, ABP already knows the entity and check IMultiTenant interface.

For entities, no needed to add an attribute. If entity implements `IMultiTenant`, it is already multi-tenant, otherwise it always ignores multi-tenancy.

obb1201:
Is it possible to apply this enhancement to release 3.2.1?

maliming:
@obb1201 It's recommended to use the latest version of ABP.

obb1201:
Thanks for your comment. 
If the latest one was based of .Net Core 3.1, it would be pretty much easier to migrate. 
Since this problem started from ABP v3.1, Isn't it supposed to support the version 3.x?
I don't have any plan to migrate to .NET 5.0 with ABP v4.x within 2 years. or is there any way to resolve this issue on v3.2.1?

evenlee:
@hikalkan , is it possible to add [MultiTenancy] attribute for a dbcontext?  I am asking because we have a case that each tenant has a legacy database and  impossible to implement IMultiTenancy interface, we need the legacy dbContext access tenant database but not host by default. and I've discussed it in [10757](https://github.com/abpframework/abp/issues/10757), but seems no good solution yet.

PR #7542, opened 2021-01-31
Fix MongoDbRepository DeleteManyAsync method

olicooper:
I found that `DeleteManyAsync` was not setting any deletion audit properties.

I have moved calls to `ApplyAbpConceptsForDeletedEntityAsync` so it is called after setting `IsDeleted` on the entities which means the deletion events are triggered using fully modified entities.

I have also noticed that the `CreateEntitiesFilter` is not called for the hard-deleted entities with a ConcurrencyStamp, and that `RepositoryFilterer.CreateEntitiesFilter` was incorrectly used. I don't know if there is any point in modifying the entity if is is going to be hard deleted, which is why I assume the ConcurrencyStamp is not passed to `CreateEntitiesFilter`?

### DIFF
--- a/framework/src/Volo.Abp.MongoDB/Volo/Abp/Domain/Repositories/MongoDB/MongoDbRepository.cs
+++ b/framework/src/Volo.Abp.MongoDB/Volo/Abp/Domain/Repositories/MongoDB/MongoDbRepository.cs
@@ -277,15 +277,16 @@ namespace Volo.Abp.Domain.Repositories.MongoDB
             bool autoSave = false,
             CancellationToken cancellationToken = default)
         {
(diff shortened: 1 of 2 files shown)


olicooper:
Hey @maliming - I have another PR that I want to create but it relies on this fix for the tests to pass.. can I base the other PR on this PR's branch? Here is the other PR: https://github.com/abpframework/abp/pull/7543

maliming:
hi

You can consider combining those PRs into one.

olicooper:
I can close this PR then and mark https://github.com/abpframework/abp/pull/7543 as ready for review because it hass all changes included. I will just copy the description over.

Does that seem okay? I made 2 PRs because I wasn't sure if the other one would be merged.

PR #7588, opened 2021-02-03
AutoMapper's Mapper creation fix: allow custom services construction

skomis-mm:
4.2.0 forces use DI / IServiceProvider for automapper services construction, which breaks custom configurations that worked with 4.1 and lower for services creation, e.g.
```csharp
{
      options.Configurators.Add(ctx =>
      {
          // ..         
          ctx.MapperConfiguration.ConstructServicesUsing(type => serviceProvider.GetService(type) ?? Activator.CreateInstance(type));
      });
});
```

pr fixes this by setting DI / IServiceProvider as default, but allow custom configuration.


realLiangshiwei:
I think we can fix it in 4.2, I will send a new PR. thanks a lot @skomis-mm  👍

PR #7623, opened 2021-02-06
Fix code sample in Environment doc page

okyrylchuk:
Hi,

I believe there is a mistake in the Environment doc page in the code sample. `EnvironemntService` doesn't have `getAll()` and `getAll$()` methods. This PR fixes it. There is also a small grama fix.


mehmet-erim (approved):

mehmet-erim:
Thanks for the contribution.

PR #7659, opened 2021-02-08
Remove setting connection string on the tenant management UI.

### LINKED ISSUE #7655: Remove setting connection string on the tenant management UI
This function is not working as expected. Let me explain;

The main problem is that when you create a new tenant, you can't set the connection string, which results the tenant's initial data (admin user, role and permissions) are seeding into the host database. See the "new tenant UI":

image

Then you open the connection strings modal (actions->connection strings) and change the connection string:

image

You expect that a new database is created on the fly and the existing tenant data is moved to the new database. However, nothing happens. It just updates/sets the connection string in the database. Then, when you try to login with this tenant, you can't login because there is no database with the configured connection string. What do you do? Stop the application, run `DbMigrator` and let it to create the tenant database and seed the initial data (not move existing, again).

This may be acceptable in development (I think now), but obviously not usable in production environment. I don't know why community did't complain much about it until now. maybe separate db approach is not much used or they handle it somehow 😄 

So, while the **ABP Framework perfectly supports "database per tenant" approach**, to be honest, the tenant management module and the startup template has never been designed to properly support this scenario.

Supporting this scenario requires some work. We actually have done in the ABP Commercial's SaaS module and the startup template (and will be released in the next verson). However, I think we may leave this to the developer in the open source (free) templates. While implementing it is not so complicated, we can consider this as an enterprise feature (if you need to separate tenant databases, you should have a big system and you can invest it by purchasing a commercial license). This approach fits our "core framework and features are open source and free, pre-built application UI are commercial" approach.

For who want to implement it, they can customize the tenant management module (download its source code and replace with nuget packages in your solution). You can change the UI (get connection string on tenant creation) and tenant application service's create method (if connection string is provided, create and migrate the new database). We made the hard work in the framework, you just need to complete the application UI part yourself.

Note: We will leave the dbmigrator as is. It will continue to support the separate tenant database approach. We just make the change on UI (removing an already non-working option). We will also not change the HTTP API part (the app service).

maliming:
Resolve #7655


modules/tenant-management/src/Volo.Abp.TenantManagement.Blazor/Pages/TenantManagement/TenantManagement.razor.cs
```diff
@@ -33,8 +30,6 @@ public TenantManagement()
             CreatePolicyName = TenantManagementPermissions.Tenants.Create;
             UpdatePolicyName = TenantManagementPermissions.Tenants.Update;
             DeletePolicyName = TenantManagementPermissions.Tenants.Delete;
-            ManageConnectionStringsPolicyName = TenantManagementPermissions.Tenants.ManageConnectionStrings;
-            ManageFeaturesPolicyName = TenantManagementPermissions.Tenants.ManageFeatures;
```
hikalkan, line 37: Why removed `ManageFeaturesPolicyName`?

hikalkan:
Can you open a following issue for the angular UI?

modules/tenant-management/src/Volo.Abp.TenantManagement.Blazor/Pages/TenantManagement/TenantManagement.razor.cs
hikalkan: Sorry, `ManageFeaturesPolicyName`?

maliming: Reverted.

hikalkan:
I will merge, but if you didn't do, please test in on UI.

hikalkan (approved):

maliming:
Tested.

PR #7666, opened 2021-02-09
Map ExtraProperties of an entity to regular properties of a type-safe dto

### LINKED ISSUE #3128: How to map ExtraProperties of an entity to a type-safe DTO?
[ExtraProperties](https://docs.abp.io/en/abp/latest/Entities#extra-properties) of an entity is a dictionary that contains some additional properties related to the entity.

If you map such an entity to a DTO (or vice verse), the values in the extra properties will not be mapped as you can expect. You can manually define AutoMapper mapping configuration.

Since this is pretty standard, we can define some extension methods & conventions to make this easier.

iyilm4z:
Resolves #3128

### DIFF
--- a/framework/src/Volo.Abp.ObjectExtending/Volo/Abp/Data/HasExtraPropertiesExtensions.cs
+++ b/framework/src/Volo.Abp.ObjectExtending/Volo/Abp/Data/HasExtraPropertiesExtensions.cs
@@ -2,8 +2,7 @@
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Globalization;
-using Volo.Abp.DynamicProxy;
-using Volo.Abp.Localization;
+using System.Linq;
 using Volo.Abp.ObjectExtending;
 using Volo.Abp.Reflection;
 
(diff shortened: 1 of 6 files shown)


framework/src/Volo.Abp.ObjectExtending/Volo/Abp/Data/HasExtraPropertiesExtensions.cs
```diff
@@ -106,7 +105,20 @@ public static void SetDefaultsForExtraProperties(object source, Type objectType)
...
+        public static void SetExtraPropertiesToRegularProperties(this IHasExtraProperties source)
+        {
+            var properties = source.GetType().GetProperties()
+                .Where(x => source.ExtraProperties.ContainsKey(x.Name)
+                            && x.GetSetMethod(true) != null)
+                .ToList();
+
+            foreach (var property in properties)
+            {
+                property.SetValue(source, source.ExtraProperties[property.Name]);
```
hikalkan, line 120: I think we should remove this property from source.ExtraProperties. No need to duplication. If we define a regular property, we want to use it instead of extra property.
Duplication is dangerous here. Because if we change one of them, the other one won't change.

hikalkan (changes requested):

hikalkan:
Actually, we may do vice verse too. Map regular properties to extra properties. But, this time we should only copy the extra properties defined for the target class.

I mean we define automapper config for ProductDto->Product mapping with extra properties, we can copy ProductDto.Price to Product.ExtraProperties["Price"] if we've defined the "Price" as an extra property for the Product (using the ObjectExtensionManager).

We can reuse the same mapToRegularProperties, no need to define an additional option.

Note: We can consider this as another issue/PR. Please make my change request and I merge this first ;) Thanks.

hikalkan (approved):

hikalkan:
Thanks!

michaelsudnik:
Will this support any form of conversion? 

For example we are using MongoDB and have extended the OrganizationUnit entity to include a collection which is loaded as a collection of bson values.

PR #7672, opened 2021-02-09
validate password method has been added.

ilkayilknur:
I've added `ValidatePasswordAsync` method to validate the admin password inside the SaaS module. Part of the fix for https://github.com/volosoft/volo/issues/5278

We can use this method to validate the admin password on the `dev` branch as well.

### DIFF
--- a/modules/identity/src/Volo.Abp.Identity.Domain/Volo/Abp/Identity/IdentityUserManager.cs
+++ b/modules/identity/src/Volo.Abp.Identity.Domain/Volo/Abp/Identity/IdentityUserManager.cs
@@ -14,6 +14,7 @@ using Volo.Abp.Threading;
 using Volo.Abp.Uow;
 using Volo.Abp.Settings;
 using Volo.Abp.Identity.Settings;
+using Microsoft.AspNetCore.Identity;
 
 namespace Volo.Abp.Identity
 {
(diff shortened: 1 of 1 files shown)


modules/identity/src/Volo.Abp.Identity.Domain/Volo/Abp/Identity/IdentityUserManager.cs
```diff
@@ -242,5 +243,12 @@ public virtual async Task<IdentityResult> AddDefaultRolesAsync([NotNull] Identit
 
             return await UpdateUserAsync(user);
         }
+
+        public virtual async Task ValidatePasswordAsync(Guid id, [NotNull] string username, [NotNull] string email,
+            string password, Guid? tenantId = null)
+        {
+            var user = new IdentityUser(id, username, email, tenantId);
+            (await base.ValidatePasswordAsync(user, password)).CheckErrors();
+        }
```
maliming, line 252: If there is a password validator. It requires that the password cannot be the same as the user's first or last name or other properties.

A non-existent user can cause problems. Maybe we can consider ignoring the password validator for **admin**.

maliming: https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.identity.usermanager-1.createasync?view=aspnetcore-5.0#Microsoft_AspNetCore_Identity_UserManager_1_CreateAsync__0_

hikalkan:
Cancelled volosoft/volo#5278.

Issue #7690, opened 2021-02-10
[discussion] Improvements to data filtering


olicooper:
I have spent a week working on a proof-of-concept demo to see if we could intercept database queries to apply ABP concepts to them before they are executed. I'd like to open this up to a discussion about how useful this is for people and how this could be used in ABP to make it more flexible. This need for flexibility has been [mentioned before](https://github.com/abpframework/abp/issues/1181). It could be provided as part of the core ABP implementation or possibly included as a plugin (I have not investigated this yet).

My demo intercepts queries to apply `IDataFilters` without relying on Entity Framework's [Global Query Filters](https://docs.microsoft.com/en-us/ef/core/querying/filters) which offers limited/restricted functionality. There are bugs and the code isn't clean because its a prototype so you'll probably be able to break it - I don't mind :)

My motivation for this is due to a few issues I have seen recently:
* https://github.com/abpframework/abp/issues/6680 - Queries can't be indexed by database providers due to the the dynamic parameters `IsSoftDeleteFilterEnabled` and `IsMultiTenantFilterEnabled` in AbpDbContext.
* https://github.com/abpframework/abp/issues/7482 / https://github.com/abpframework/abp/issues/1181 - Filters can't be disabled for navigation items
* https://github.com/abpframework/abp/issues/5650 / [filtering limitations](https://docs.microsoft.com/en-us/ef/core/querying/filters#limitations) - Filtering can't be customised

The wider community also wants the filters to be more flexible too:
* https://github.com/dotnet/efcore/issues/21093
* https://github.com/dotnet/efcore/issues/11691
* https://github.com/dotnet/efcore/issues/8576

I have written an [extensive readme](https://github.com/olicooper/abp-queryfilter-test/blob/main/README.md) to explain the project if you need more info.

Here is the demo: https://github.com/olicooper/abp-queryfilter-test

PR #7706, opened 2021-02-12
Extend DataFilter implementation

olicooper:
I would like to propose some updates to the DataFilter implementation. The changes should be non-breaking as the base functionality remains the same, with an additional `IsActive` attribute along with some additional useful methods.
I have made a number of changes but my main motivation is to know which filters are currently 'active' (being used within a disposable scope). Having access to information about the filters is also very useful.

Because DataFilter is `ISingletonDependency` the cached filters are kept in memory after their first use. The `IsEnabled` property isn't a reliable source because the default `IsEnabled` value can be changed per filter with `AbpDataFilterOptions`. I tried making the `DataFilter` Scoped; but the scope can change within a request so the filters reset.

I have added various tests to ensure these work within all common scenarios and I've added documentation to guide developers for the correct use of each method.

Initially, I intend to use this in conjunction with the [abp-queryfilter-test](https://github.com/olicooper/abp-queryfilter-test) project I am working on ([mentioned here](https://github.com/abpframework/abp/issues/7690)) - where you can the proposed changes in use.

Changes include:
* Add new methods (GetOrAddFilter, IsActive) and add ReadOnlyFilters property to DataFilter
* Make DataFilter extensible (virtual methods etc.)
* Add ability to set the DefaultFilterState in AbpDataFilterOptions
* Add test suite
* Improve method XML descriptions

### DIFF
--- a/framework/src/Volo.Abp.Core/System/AbpTypeExtensions.cs
+++ b/framework/src/Volo.Abp.Core/System/AbpTypeExtensions.cs
@@ -1,4 +1,5 @@
 ﻿using System.Collections.Generic;
+using System.Text;
 using JetBrains.Annotations;
 using Volo.Abp;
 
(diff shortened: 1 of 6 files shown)


hikalkan:
Thanks @olicooper for this PR.

>Having access to information about the filters is also very useful.

Can you please give more info for the use cases? You can show some usage examples and explain the reason. I didn't understand how it is supposed to be used and what are the benefit of such an active concept.

olicooper:
Yeah sure I will try 😃. The reason I created this PR was partly to facilitate the a proposed change in how data filters are applied in ABP (I created a ticket on the issue [here](https://github.com/abpframework/abp/issues/7690)). But essentially, without `IsActive` you can't see which cached filters are applied for a given query due to the Singleton nature of DataFilters and you can't make DataFilters scoped because nested scopes are created within each request.

Official filters are `ISoftDelete` and `IMultiTenant`, but the filtering system allows for anything to be specified, for example I have also created `IPassivable` (filter for active entities) and `ITemplatable` (filter for entities marked as templates). The main reason for needing `IsActive` is to perform conditional logic based on the filter value from within a nested method call or query interceptor. One example could be when filtering for template records, I also require that I search for non-tenant records. But I think it is best to see it in action so here is a real example using the filters to determine if I should modify the EF query before sending it to the database. I didn't know which filters were active within the `DbContext` because the query is being intercepted rather than being directly called by the AppService, so without `IsActive` I wouldn't be able to modify the query accordingly.

See [AbpFilterAppendingExpressionVisitor.cs](https://github.com/olicooper/abp-queryfilter-test/blob/0714d21a764061b9cf821ce4ce78700a1b1f1a2c/src/AbpQueryFilterDemo.EntityFrameworkCore/Extensions/AbpFilterAppendingExpressionVisitor.cs#L207-L241) - you can also check out the whole demo if you'd like 😀

PR #7718, opened 2021-02-14
Ignore password validation in IdentityDataSeeder.

### LINKED ISSUE #7677: Ignore password validation in IdentityDataSeeder.
`IdentityOptions.Value.Password.ClearRequirements()` only the default password verification rules will be ignored, and the custom `IPasswordValidator` can't be ignored.

```cs
public class MyPasswordValidator : IPasswordValidator<IdentityUser>
{
    public Task<IdentityResult> ValidateAsync(UserManager<IdentityUser> manager, IdentityUser user, string password)
    {
        if (password == user.UserName || password == user.Email)
        {
            return Task.FromResult(IdentityResult.Failed(new IdentityError
            {
                Description = "The password cannot be the same as the username or email."
            }));
        }

        return Task.FromResult(IdentityResult.Success);
    }
}
```

Create user without validate password. **validatePassword: false**
```cs
public virtual async Task<IdentityResult> CreateWithoutValidatePasswordAsync(IdentityUser user, string password)
{
    var result = await CreateAsync(user);
    if (!result.Succeeded)
    {
        return result;
    }

    return await UpdatePasswordHash(user, password, validatePassword: false);
}
```

maliming:
Resolve #7677


hikalkan:
Thanks for fixing this.

`IdentityManager.CreateAsync(TUser user, string password)` method first uses `UpdatePasswordHash`, then calls `CreateAsync(user)`:

https://github.com/dotnet/aspnetcore/blob/1b67848306760619552b11a9bfeaf7d9655344bc/src/Identity/Extensions.Core/src/UserManager.cs#L587-L604

But we've done Create first, then update hash. I suppose we should follow the same order. What do you think?

Also, you can delete `Password.ClearRequirements()` extension method if we don't use it anymore.

maliming:
@hikalkan Changed.

hikalkan (approved):

PR #7731, opened 2021-02-15
Cms Kit - Pages Prefix

### LINKED ISSUE #7708: CmsKit - Use 'pages' prefix for page urls
Use `/pages` prefix to catch page urls

enisn:
## Summary
Page routes must be handled with a static prefix such as `/pages/` until finding better than **match all** pattern.

---

Closes #7708


modules/cms-kit/src/Volo.CmsKit.Public.Web/CmsKitPublicWebModule.cs
```diff
@@ -61,6 +61,10 @@ public override void PostConfigureServices(ServiceConfigurationContext context)
         {
             Configure<RazorPagesOptions>(options =>
             {
+                if (GlobalFeatureManager.Instance.IsEnabled<PagesFeature>())
```
ilkayilknur: Can we move the  if check to the top of the `Configure` method? So that we won't call the configure method unnecessarily.

modules/cms-kit/src/Volo.CmsKit.Public.Web/CmsKitPublicWebModule.cs
```diff
@@ -61,6 +61,10 @@ public override void PostConfigureServices(ServiceConfigurationContext context)
...
+                {
+                    options.Conventions.AddPageRoute("/CmsKit/Pages/Index", @"/pages/{pageUrl:minlength(1)}");
```
ilkayilknur: Maybe we should change the route parameter to the `slug.` It will align with the domain.

ilkayilknur (approved):

ilkayilknur (changes requested):
I accidentally approved the PR.

cotur (approved):

ilkayilknur (approved):

PR #7742, opened 2021-02-16
Implemented: Should return the same instance in the same UOW for replaced DbContexts

### LINKED ISSUE #7737: Should return the same instance in the same UOW for replaced DbContexts
This will better work with dbcontext replacement combined joining different repository queryables. It is also more performant since it will not create unnecessary dbcontext instances.

See [Replace Other DbContextes](https://docs.abp.io/en/abp/latest/Entity-Framework-Core#replace-other-dbcontextes) to understand this feature.

hikalkan:
Resolves #7737

### DIFF
--- a/framework/src/Volo.Abp.EntityFrameworkCore/Volo/Abp/EntityFrameworkCore/AbpDbContextOptions.cs
+++ b/framework/src/Volo.Abp.EntityFrameworkCore/Volo/Abp/EntityFrameworkCore/AbpDbContextOptions.cs
@@ -7,19 +7,22 @@ namespace Volo.Abp.EntityFrameworkCore
 {
     public class AbpDbContextOptions
     {
-        internal List<Action<AbpDbContextConfigurationContext>> DefaultPreConfigureActions { get; set; }
+        internal List<Action<AbpDbContextConfigurationContext>> DefaultPreConfigureActions { get; }
 
         internal Action<AbpDbContextConfigurationContext> DefaultConfigureAction { get; set; }
 
-        internal Dictionary<Type, List<object>> PreConfigureActions { get; set; }
+        internal Dictionary<Type, List<object>> PreConfigureActions { get; }
 
-        internal Dictionary<Type, object> ConfigureActions { get; set; }
+        internal Dictionary<Type, object> ConfigureActions { get; }
+        
+        internal Dictionary<Type, Type> DbContextReplacements { get; }
 
         public AbpDbContextOptions()
         {
             DefaultPreConfigureActions = new List<Action<AbpDbContextConfigurationContext>>();
             PreConfigureActions = new Dictionary<Type, List<object>>();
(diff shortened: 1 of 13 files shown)


framework/src/Volo.Abp.EntityFrameworkCore/Volo/Abp/EntityFrameworkCore/AbpDbContextOptions.cs
```diff
@@ -7,19 +7,22 @@ namespace Volo.Abp.EntityFrameworkCore
...
 
         internal Action<AbpDbContextConfigurationContext> DefaultConfigureAction { get; set; }
 
-        internal Dictionary<Type, List<object>> PreConfigureActions { get; set; }
+        internal Dictionary<Type, List<object>> PreConfigureActions { get; }
 
-        internal Dictionary<Type, object> ConfigureActions { get; set; }
+        internal Dictionary<Type, object> ConfigureActions { get; }
+        
+        internal Dictionary<Type, Type> DbContextReplacements { get; }
```
maliming, line 18: How about `Dictionary<Type, List<Type>> DbContextReplacements { get; }`

**key** is **DbContext type** that replace the types of **value**. 

```cs
services.Configure<AbpDbContextOptions>(opts =>
{
    opts.DbContextReplacements.GetOrAdd(typeof(TDbContext), () => new List<Type>())
        .AddIfNotContains(dbContextType);
});
```

```cs
internal Type GetReplacedTypeOrSelf(Type dbContextType)
{
    return DbContextReplacements.Where(x => x.Value.Contains(dbContextType)).Select(x => x.Key).FirstOrDefault()
           ?? dbContextType;
}
```

hikalkan: `Dictionary<Type, Type>` is faster than sequential search. What is the benefit of `Dictionary<Type, List<Type>>` approach?

maliming: For me, it is easier to understand. And no need for `while (true)`

image

hikalkan: While-true is also needed for your case. It is actually an edge case. If A is replaced by B, then B is replaced then C, it finds C for A. It could be simplified (if we handle multiple replace on the replacement phase), however using a `Dictionary<Type, Type>` is faster and pretty simple.

hikalkan: If nothing else, can you please merge, I am working another feature based on it :)

maliming (approved):

PR #7766, opened 2021-02-17
Support extending a controller

### LINKED ISSUE #6288: Replacing controller is not possible [BUG ?}
@hikalkan Hi.

I'm using ABP Commercial 3.3.1 with angular.
I have integrated DevExpress reporting into ABP. 
After upgrading to 3.3.1 replacing controllers is not working any more.

After some research i noticed that you commit some changes with RemoveDuplicateController method in AbpServiceConvention.

This RemoveDuplicateController  method removes my controller (message in log):

> Removing the controller Sebit.Web.Reports.EdvinController, Sebit.Web.Reports, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null from the application model since it replaces the controller(s): DevExpress.AspNetCore.Reporting.WebDocumentViewer.WebDocumentViewerController, DevExpress.AspNetCore.Reporting.v20.2, Version=20.2.3.0, Culture=neutral, PublicKeyToken=b88d1754d700e49a

My controller is derived like this:

```
public class EdvinController : WebDocumentViewerController

public class WebDocumentViewerController : ControllerBase

public abstract class ControllerBase : Controller

```
As you can see GetBaseClasses extended method will allways return more than one bases and my controller will be allways removed.

This is braking change for me, because WebDocumentViewerController is Devexepress class and i can't make it abstract.

Please advice.

### LINKED ISSUE #7698: AppService missing in Swagger after upgrading from 2.7 to 3.3.2
I just upgraded to versio 3.3.2 because this is the last version before 5.0. I want to do the 5.0 upgrade later. However after upgrading from 2.7 the Swagger seemed to be missing some AppServices. All the AppServices that were missing do inherit from a base class.

For example:
```c#
    [ExposeServices(typeof(IIdentityUserAppService), typeof(IdentityUserAppService), typeof(MydentityUserAppService))]
    public class MyIdentityUserAppService : IdentityUserAppService
```

In 2.7 there were 2 end-points. One for users and one for myIdentityUsers. This wasn't ideal either because I would like to hide the initial users. However when upgrading to 3.3.2 I am missing the myIdentityUsers endpoint in Swagger at all. This prevents openapi from generating that service as well. Did I miss something here?

Also how would I hide the original users service in Swagger, or better keep the users name but use my own implementation that has an extra endpoint for creating a user without a password.

### LINKED ISSUE #7721: Extending an ABP controller
I know you can override an ABP controller: https://docs.abp.io/en/abp/latest/Customizing-Application-Modules-Overriding-Services#example-overriding-a-controller

But, how would you add an endpoint to the overridden controller? I have tried adding an endpoint and I can't hit it with Postman and it doesn't show up in swagger. For example:

    [ExposeServices(typeof(IdentityUserController))]
    public class MyIdentityUserController : IdentityUserController
    {
        public MyIdentityUserController(IIdentityUserAppService userAppService) : base(userAppService)
        {
        }

        public override Task<PagedResultDto<IdentityUserDto>> GetListAsync(GetIdentityUsersInput input)
        {
            // this works fine and it gets hit
            return base.GetListAsync(input);
        }

        [HttpGet, Route("mymethod")]
        public Task<string> MyMethod()
        {
            return Task.FromResult("Doesn't work");
        }
    }

How can I accomplish this?

acjh:
Resolves #6288
Resolves #7698
Resolves #7721

Changes:
- Remove existing controllers instead of the derived controller if `ExposeService` with `IncludeSelf = true`.
- Only remove the derived controller if it has base controller models, since some third-party base controllers are non-abstract,
  e.g. `DevExpress.AspNetCore.Reporting.WebDocumentViewer.WebDocumentViewerController` (#6288).
- Fix typo of `AccountController` to `MyAccountController` in *Customizing-Application-Modules-Overriding-Services.md*.
- Mention the behaviour for extending a controller in *Customizing-Application-Modules-Overriding-Services.md*.

### DIFF
--- a/docs/en/Customizing-Application-Modules-Overriding-Services.md
+++ b/docs/en/Customizing-Application-Modules-Overriding-Services.md
@@ -196,7 +196,11 @@ This example replaces the `AccountController` (An API Controller defined in the 
 
 **`[ExposeServices(typeof(AccountController))]` is essential** here since it registers this controller for the `AccountController` in the dependency injection system. `[Dependency(ReplaceServices = true)]` is also recommended to clear the old registration (even the ASP.NET Core DI system selects the last registered one).
 
-In addition, the `MyAccountController` will be removed from [`ApplicationModel`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.applicationmodels.applicationmodel.controllers) because it defines `ExposeServicesAttribute`. If you don't want to remove it, you can configure `AbpAspNetCoreMvcOptions`:
+In addition, the `MyAccountController` will be removed from [`ApplicationModel`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.applicationmodels.applicationmodel.controllers) because it defines `ExposeServicesAttribute`.
+
+If `IncludeSelf = true` is specified, i.e. `[ExposeServices(typeof(AccountController), IncludeSelf = true)]`, then `AccountController` will be removed instead. This is useful for **extending** a controller.
+
+If you don't want to remove either controller, you can configure `AbpAspNetCoreMvcOptions`:
 
 ```csharp
 Configure<AbpAspNetCoreMvcOptions>(options =>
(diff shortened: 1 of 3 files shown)


maliming:
hi @acjh 

Could you add some test controllers to the unit test to test this feature? 
If you don't have much time, I can add that.

acjh:
@maliming Sure, done.

maliming (approved):

hikalkan:
Thanks @acjh for your contribution. It is nice to see you here :)

PR #7798, opened 2021-02-22
Add MenuContributorBase to check permissions in advance.

### LINKED ISSUE #7788: bad performance Isgranted with redis
Hello,

My application has 500+ Permission Policy and  when I use redis, the average page load speed increasing by 50 seconds. (I don't use local redis).

I have attached the log files

I think you can change the caching mechanism on permissions (like ambient context
[logs-20210221.txt](https://github.com/abpframework/abp/files/6016469/logs-20210221.txt)

Thank you

maliming:
Resolve #7788

Usage:
Use **GrantedPermissions** to check the permissions.
```cs
public class TestMenuContributor1 : MenuContributorBase
{
    public TestMenuContributor1()
    {
        PreCheckPermissions.Add("MenuPermission1");
        PreCheckPermissions.Add("MenuPermission2");
    }

    public override Task ConfigureMenuAsync(MenuConfigurationContext context)
    {
        if (GrantedPermissions.Contains("MenuPermission1"))
        {
            //context.Menu.AddItem()
        }

        return Task.CompletedTask;
    }
}
```

### DIFF
--- a/framework/src/Volo.Abp.UI.Navigation/Volo/Abp/Ui/Navigation/MenuContributorBase.cs
+++ b/framework/src/Volo.Abp.UI.Navigation/Volo/Abp/Ui/Navigation/MenuContributorBase.cs
@@ -1,0 +1,26 @@
+﻿using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Threading.Tasks;
+
+namespace Volo.Abp.UI.Navigation
+{
+    public abstract class MenuContributorBase : IMenuContributor
+    {
+        public List<string> PreCheckPermissions { get; }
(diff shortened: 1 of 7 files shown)


maliming:
@hikalkan 
If you think this ok. I will do same changes to `IToolbarContributor` and apply to all modules and templates.

hikalkan:
This solution requires manual work and open to bugs (we may forget to add to the list). Also, it is a solution per menu contributor.

What if we add RequiredPermission to the `ApplicationMenuItem`. All contributors add menu items without permission checking. Then MenuManager removes items from the menu if user has no permission.
Same logic can be applied to toolbarcontributor.

PR #7806, opened 2021-02-22
Cms Kit - Refactorings and code fixes

ilkayilknur:


modules/cms-kit/src/Volo.CmsKit.Domain/Volo/CmsKit/Tags/Tag.cs
```diff
@@ -33,7 +33,7 @@ protected Tag()
 
         public virtual void SetName(string name)
```
enisn:
```suggestion
        public virtual void SetName([NotNull] string name)
```

enisn (approved):

modules/cms-kit/src/Volo.CmsKit.EntityFrameworkCore/Volo/CmsKit/Blogs/EfCoreBlogPostRepository.cs
```diff
@@ -20,47 +22,54 @@ public EfCoreBlogPostRepository(IDbContextProvider<CmsKitDbContext> dbContextPro
...
 
-        public async Task<BlogPost> GetBySlugAsync(Guid blogId, string slug, CancellationToken cancellationToken = default)
+        public async Task<BlogPost> GetBySlugAsync(Guid blogId, [NotNull] string slug,
+            CancellationToken cancellationToken = default)
         {
+            Check.NotNullOrEmpty(slug, nameof(slug));
+
             var dbSet = await GetDbSetAsync();
 
             return await dbSet
```
cotur: We can override `WithDetails`

cotur (changes requested):

cotur (approved):

PR #7816, opened 2021-02-23
Use ApplicationApiDescriptionModel.Create  insetad of new.

### LINKED ISSUE #7815: some problems with ServiceProxyScript
* Your **ABP Framework version**.4.2.1

###  1. When qualifying modules or controllers :ServiceProxyScript?modules=abp, reference object is empty of abnormal will happen, because when CreateSubModel, direct use of the new ApplicationApiDescriptionModel ()

image
image

###  2. When the service method is FormForm and the DTO type contains IFormFile, the generated JS code directly excludes the contents of IFormFile and generates the following code
image

### However, the generated JS code is missing the IFormFile attribute
image

`  
   ###  My JS code looks something like this：

    var formBody= document.getElementById("formID");
    var bookData = new FormData(formBody);
    bookData.name = "xxx";
    bookData.auth = "xxx";
    testProject.book.createBook(bookData, {
        contentType: false, //must
        processData: false, //must
        success:function(){}
    });
`

maliming:
Resolve #7815


hanfeige:
image
### Note: If the namespace is different and the controller name is the same, an exception will also occur. The code environment is still #7815

realLiangshiwei (approved):

PR #7882, opened 2021-03-01
Update Navigation-Menu and Toolbars.

maliming:


docs/en/UI/AspNetCore/Navigation-Menu.md
```diff
@@ -104,14 +104,34 @@ There are more options of a menu item (the constructor of the `ApplicationMenuIt
...
 
 ### Authorization
 
 As seen above, a menu contributor contributes to the menu dynamically. So, you can perform any custom logic or get menu items from any source.
 
-One use case is the [authorization](../../Authorization.md). You typically want to add menu items by checking a permission.
+One use case is the [authorization](../../Authorization.md).
 
-**Example: Check if the current user has a permission**
+You can set the `RequiredPermissionName` property of the menu item.
```
hikalkan: The document flow is not good like that. First show the custom check (with if), then say "For the authorization, you can use RequiredPermissionName as a shortcut. It is also more performant, ABP optimizes the permission check for all the items" then show it.

hikalkan (approved):

PR #7895, opened 2021-03-01
Add constructors to base entity/aggregateroot classes.

### LINKED ISSUE #780: Add constructors to base entity/aggregateroot classes
Current `CreationAuditedAggregateRoot<TKey>` class is like that:

````csharp
[Serializable]
public abstract class CreationAuditedAggregateRoot<TKey> : AggregateRoot<TKey>, ICreationAuditedObject
{
    /// <inheritdoc />
    public virtual DateTime CreationTime { get; set; }

    /// <inheritdoc />
    public virtual Guid? CreatorId { get; set; }
````

We can add constructors like:

````csharp
[Serializable]
public abstract class CreationAuditedAggregateRoot<TKey> : AggregateRoot<TKey>, ICreationAuditedObject
{
    /// <inheritdoc />
    public virtual DateTime CreationTime { get; set; }

    /// <inheritdoc />
    public virtual Guid? CreatorId { get; set; }

    protected CreationAuditedAggregateRoot()
    {
        
    }

    protected CreationAuditedAggregateRoot(
        TKey id, 
        DateTime creationTime = default, 
        Guid? creatorId = null)
        : base(id)
    {
        CreationTime = creationTime;
        CreatorId = creatorId;
    }
}
````

real-zony:
Resolve #780

### DIFF
--- a/framework/src/Volo.Abp.Ddd.Domain/Volo/Abp/Domain/Entities/Auditing/AuditedAggregateRoot.cs
+++ b/framework/src/Volo.Abp.Ddd.Domain/Volo/Abp/Domain/Entities/Auditing/AuditedAggregateRoot.cs
@@ -34,10 +34,13 @@ namespace Volo.Abp.Domain.Entities.Auditing
 
         }
 
-        protected AuditedAggregateRoot(TKey id)
+        protected AuditedAggregateRoot(TKey id,
+            DateTime? lastModificationTime = null,
+            Guid? lastModifierId = null)
             : base(id)
         {
-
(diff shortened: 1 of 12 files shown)


maliming (approved):

hikalkan:
Thanks @real-zony for your contribution. However, I think we should cancel this issue. Getting id is enough. It is not true to get all properties (for example, fullauditedaggregateroot has 7 properties and we get 3 in this PR. if we get all, the ctor will be unnecessarily long). Also, most of these properties are not meaningful on object creation (deletion and modification properties).

PR #7923, opened 2021-03-03
CmsKit - Add EntityType configuration to Comments

### LINKED ISSUE #7886: CMS Kit - Add CMSKitCommentOptions and Make EntityTypes Configurable
Ensure that all websites using the comment feature get updated.

enisn:
## Summary
This PR includes EntityType configuration for Comments but authorization check, becuase it has its own authorization logic on management side and doesn't require permission on public side.

---

- Closes #7886


modules/cms-kit/src/Volo.CmsKit.Domain/Volo/CmsKit/Comments/CommentManager.cs
```diff
@@ -0,0 +1,42 @@
...
+        }
+
+        public virtual async Task<Comment> CreateAsync([NotNull] CmsUser creator,
+                                                       [NotNull] string entityType,
+                                                       [NotNull] string entityId,
+                                                       [NotNull] string text,
+                                                       [CanBeNull] Guid? repliedCommentId = null)
+        {
+            Check.NotNull(creator, nameof(creator));
+
```
ilkayilknur, line 30: We should add null checks for entityType, entityId, and text as well.

ilkayilknur (changes requested):
I have small feedback about null checks. Rest of the code LGTM 👍

modules/cms-kit/src/Volo.CmsKit.Domain/Volo/CmsKit/Comments/EntityNotCommentableException.cs
```diff
@@ -0,0 +1,26 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.Serialization;
+using System.Text;
+using System.Threading.Tasks;
+using Volo.Abp;
+
+namespace Volo.CmsKit.Comments
+{
+    public class EntityNotCommentableException : BusinessException
```
cotur, line 12: Requires `[Serializable]` attribute

modules/cms-kit/src/Volo.CmsKit.Domain/Volo/CmsKit/Comments/CommentManager.cs
enisn: Those are checked in constructor of Comment entity. That's why I ignored from here. Should we ad here too? @ilkayilknur

ilkayilknur: Yes, we should add it. The implementation of the might change in the future. We also check the entityType definition below. What happens if the entityType is null? We should add checks.

enisn: Ok, it's done with 96e84af

ilkayilknur (approved):

enisn:
Awaiting to find possible breaks

PR #7996, opened 2021-03-09
Use application service provider in AbpDbContextConfigurationContext

maliming:
https://docs.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.dbcontextoptionsbuilder.useapplicationserviceprovider?view=efcore-5.0

> This is done automatically when using 'AddDbContext' or 'AddDbContextPool', so it is rare that this method needs to be called.

https://github.com/dotnet/efcore/blob/efc2924e6e36db17611c7da480d5bc97a4514cb8/src/EFCore/Extensions/EntityFrameworkServiceCollectionExtensions.cs#L952

In this way, We can use `IServiceProvider `in `OnModelCreating `and `OnConfiguring `methods of `DbContext`.


hikalkan:
Great!

hikalkan (approved):

PR #8002, opened 2021-03-09
Set RedirectUri property of AuthenticationProperties.

maliming:
This will make **oidc** redirect correctly.


framework/src/Volo.Abp.AspNetCore.Mvc/Volo/Abp/AspNetCore/Mvc/Authentication/ChallengeAccountController.cs
```diff
@@ -23,41 +23,29 @@ public ActionResult Login(string returnUrl = "", string returnUrlHash = "")
...
-                        {
-                            {"returnUrl", returnUrl},
-                            {"returnUrlHash", returnUrlHash}
-                        }
-                    },
-                    ChallengeAuthenticationSchemas
-                );
-            }
+
+            return Challenge(new AuthenticationProperties {RedirectUri = returnUrl}, ChallengeAuthenticationSchemas);
```
gterdem: Why do we discard returnUrlHash? Client side apps may be using hashed return urls, no?

maliming: Changed.

gterdem (approved):

PR #8035, opened 2021-03-11
add localizations  of  zh-Hant

ChangYinShung:
use abp translate command to modify zh-Hant


realLiangshiwei:
Thanks for your contributor, Could you translate all of these?

image

ChangYinShung:
@realLiangshiwei ok
 i will create another pr to translate  it

realLiangshiwei:
Hi

If only these keys here, then the localized content of the page will be blank or even error.
If you could translate them in this PR, it will be better.

ChangYinShung:
sorry,i will delete this and  create another pull-request to add cms-kit 's zh-Hant

Issue #7800, opened 2021-02-22
CLI should check EF Core global tool installed


hikalkan:
CLI creates the initial migration. But if user has no installed dotnet ef tooling, then it will fail.

It is installed as shown below:

````bash
dotnet tool install --global dotnet-ef
````

Also, there should be commands to check if it was installed. ABP CLI should install it if not installed.

PR #8048, opened 2021-03-13
Add Entity Framework Core tools CLI

Tbz8:
So that other users don't run into the same problem as I did, I think it's worth mentioning that the installation of Entity Framework Core tools is mandatory in order to run the application.


maliming:
hi

It will install automatically.

https://github.com/abpframework/abp/issues/7800

PR #8057, opened 2021-03-15
Introduce MyProjectNameWebThemeModule.

### LINKED ISSUE #8055: Include source code of the basic theme in the startup template (MVC UI)
We see that many people want to completely customize the basic theme for their applications. This is what we expect since the basic theme is designed to be "basic", so you build your own theme based on it.

There are some (see for [mvc](https://community.abp.io/articles/creating-a-new-ui-theme-by-copying-the-basic-theme-for-mvc-ui-yt9b18io) & [blazor](https://community.abp.io/articles/creating-a-new-ui-theme-by-copying-the-basic-theme-for-blazor-ui-qaf5ho1b)) community articles to demonstrate how to copy its source into your application code and freely change the basic theme. However, it still requires some additional work after project creation.

We think it will be good if we include the theme code inside the downloaded application startup template, so anyone can directly change the theme. We don't make changes frequently in the basic theme, so most of times you won't need to upgrade it.

We will add the theme code in a separate project, so it will be separated from the application code. In this way, you may reuse the theme in different applications or switch to another theme in the future easily.

If anyone still wants to use the basic theme as a package, we will continue to deploy it as a NuGet package. They can remove the project in the solution and add the basic theme package.

We may consider to make this optional if community wants it in the future.

maliming:
Resolve #8055

`Web`, `Web.Host`, `IdentityServer`, `HostWithIds` both use this theme.

@yekalkan I think the theme project will not affect the CLI. right?

### DIFF
--- a/templates/app/aspnet-core/MyCompanyName.MyProjectName.sln
+++ b/templates/app/aspnet-core/MyCompanyName.MyProjectName.sln
@@ -53,6 +53,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "MyCompanyName.MyProjectName
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "MyCompanyName.MyProjectName.Blazor", "src\MyCompanyName.MyProjectName.Blazor\MyCompanyName.MyProjectName.Blazor.csproj", "{27B2DDC7-8B75-4322-A312-25419C15D9D8}"
 EndProject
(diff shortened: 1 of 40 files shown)


yekalkan:
>@yekalkan I think the theme project will not affect the CLI. right?

It won't effect.

hikalkan:
I am canceling this. We should not rename the project. Instead, we should directly put the same basic theme project inside the solution. I will talk with @yekalkan for CLI support for that.

PR #8117, opened 2021-03-18
Create interceptor for global features

### LINKED ISSUE #8108: Create interceptor for global features
It is similar to the `FeatureInterceptor`, but for global features.

We have `GlobalFeatureActionFilter` and `GlobalFeaturePageFilter`, but they can work only for aspnet core controllers. They are not working for blazor server, for example. Becuase blazor server doesn't pass through these filters, but directly injects and consumes the application services.

We can create a `IGlobalFeatureCheckingEnabled` interface to not intercept every class, but only intercept classes implement this interface. Check `IValidationEnabled` and `ValidationInterceptorRegistrar` for a similar example. `ApplicationService` base class can implement this `IGlobalFeatureCheckingEnabled`, so only app services are intercepted. Other classes should implement `IGlobalFeatureCheckingEnabled` directly.

You can also use `AbpCrossCuttingConcerns.IsApplied` logic to not double check the features.

### LINKED ISSUE #8188: Cli add-package: allow to use on solution file

maliming:
Resolve #8108

### DIFF
--- a/framework/src/Volo.Abp.GlobalFeatures/Volo/Abp/GlobalFeatures/AbpGlobalFeatureNotEnableException.cs
+++ b/framework/src/Volo.Abp.GlobalFeatures/Volo/Abp/GlobalFeatures/AbpGlobalFeatureNotEnableException.cs
@@ -1,0 +1,22 @@
+﻿using System;
+using Volo.Abp.ExceptionHandling;
+
+namespace Volo.Abp.GlobalFeatures
+{
+    public class AbpGlobalFeatureNotEnableException : AbpException, IHasErrorCode
+    {
+        public string Code { get; }
+
+        public AbpGlobalFeatureNotEnableException(string message = null, string code = null, Exception innerException = null)
+            : base(message, innerException)
+        {
+            Code = code;
+        }
+
+        public AbpGlobalFeatureNotEnableException WithData(string name, object value)
+        {
+            Data[name] = value;
+            return this;
+        }
+    }
+}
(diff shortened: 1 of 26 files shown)


framework/src/Volo.Abp.GlobalFeatures/Volo/Abp/GlobalFeatures/AbpGlobalFeatureNotEnableException.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+using Volo.Abp.ExceptionHandling;
+
+namespace Volo.Abp.GlobalFeatures
+{
+    public class AbpGlobalFeatureNotEnableException : AbpException, IHasErrorCode
```
hikalkan, line 6:
```suggestion
    public class AbpGlobalFeatureNotEnabledException : AbpException, IHasErrorCode
```

hikalkan: Also, please make the exception serializable (using constructor like the AbpException class)

hikalkan: Also, please make the exception serializable (using constructor like the AbpException class)

framework/src/Volo.Abp.GlobalFeatures/Volo/Abp/GlobalFeatures/GlobalFeatureHelper.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+using Volo.Abp.Reflection;
+
+namespace Volo.Abp.GlobalFeatures
+{
+    public static class GlobalFeatureHelper
+    {
+        public static bool IsGlobalFeatureEnabled(Type controllerType, out RequiresGlobalFeatureAttribute attribute)
```
hikalkan, line 8:
```suggestion
        public static bool IsGlobalFeatureEnabled(Type type, out RequiresGlobalFeatureAttribute attribute)
```

controller is not a good naming in this layer.

hikalkan:
I've merged this. Please make the improvement suggestions as another PR.

Issue #8125, opened 2021-03-18
Upgrade to Bootstrap 4.6


hikalkan:
We should consider to upgrade to bootstrap 4.6 in ABP v4.3.

- [x] MVC UI / @maliming 
- [x] Angular UI / @mehmet-erim 
- [x] Blazor UI / @maliming, cc: @stsrki

stsrki:
Blazorise is already 4.6 compatible so I would guess for ABP we only need to update bundle files?

PR #8126, opened 2021-03-18
 Upgrade bootstrap to 4.6 & bootstrap-v4-rtl to 4.6.0-1.

maliming:
https://github.com/abpframework/abp/issues/8125


hikalkan (approved):

hikalkan:
I merged. But can you try if the files in the 4.6 packages are same with 4.6. You know MVC UI copies files. If file name was changed, then it gets error.

maliming:
Tested.

PR #8144, opened 2021-03-19
Add missing localizations for admin side

berkansasmaz:


sturlath:
The last time I tried out the framework it was super hard/confusing to localize the login page, is that still the case or can that be squeezed in also @berkansasmaz?

ebicoglu:
actually, it's very easy.

*   this is where the localization key is being used on the Login page: [https://github.com/abpframework/abp/blob/dev/modules/account/src/Volo.Abp.Account.Web/Pages/Account/Login.cshtml#L15](https://github.com/abpframework/abp/blob/dev/modules/account/src/Volo.Abp.Account.Web/Pages/Account/Login.cshtml#L15)
*   and this is the localization file https://github.com/abpframework/abp/blob/dev/modules/account/src/Volo.Abp.Account.Application.Contracts/Volo/Abp/Account/Localization/Resources/en.json#L18

ebicoglu (approved):

DonVliet:
I think you've made a spelling mistake?

**"Enum:EntityChangeType:0": "Crated",**

should be

**"Enum:EntityChangeType:0": "Created",**

berkansasmaz:
Oops! I'll fix this, thank you for the warning :)

PR #8152, opened 2021-03-22
Add GetListAsync with predicate to IReadOnlyBasicRepository

### LINKED ISSUE #8151: Add GetListAsync with predicate to IReadOnlyRepository
Here, a typical code to get a list of entities with a predicate:

````csharp
var organizationMembershipsQueryable = await _organizationMembershipsRepository.GetQueryableAsync();
var query = organizationMembershipsQueryable.Where(x => x.OrganizationId == @event.OrganizationId);
var organizationMembers = await _asyncExecuter.ToListAsync(query);
````

It would be good if we have a GetListAsync method like;

````csharp
var organizationMembers = await _organizationMembershipsRepository.GetListAsync(x => x.OrganizationId == @event.OrganizationId);
````

### PR DESCRIPTION
**Pull request by realLiangshiwei**
Resolve https://github.com/abpframework/abp/issues/8151

### CODE REVIEW

File: framework/src/Volo.Abp.Ddd.Domain/Volo/Abp/Domain/Repositories/IReadOnlyBasicRepository.cs [hunk 1]
```diff
@@ -17,6 +19,18 @@ public interface IReadOnlyBasicRepository<TEntity> : IRepository
...
+        /// <summary>
+        /// Gets a list entities by the given <paramref name="predicate"/>.
+        /// </summary>
+        /// <param name="predicate">A condition to find the entity</param>
+        /// <param name="includeDetails">Set true to include all children of this entity</param>
+        /// <param name="cancellationToken">A <see cref="T:System.Threading.CancellationToken" /> to observe while waiting for the task to complete.</param>
+        /// <returns>Entity</returns>
+        Task<List<TEntity>> GetListAsync(
+            [NotNull] Expression<Func<TEntity, bool>> predicate,
+            bool includeDetails = true,
```
hikalkan: `includeDetails` should be false for list returning methods.

hikalkan (changes requested):

hikalkan (approved):

PR #8160, opened 2021-03-23
Provide an option to DataTables to cancel the previous requests.

maliming:
It will cancel the previous request when a new request created. eg: page index or page count change.

```cs
  _dataTable = _$wrapper.find('table').DataTable(
      abp.libs.datatables.normalizeConfiguration({
          ajax: abp.libs.datatables.createAjax(_tenantAppService.getList, null, null, true)
      })
  );
```

image

### DIFF
--- a/framework/src/Volo.Abp.AspNetCore.Mvc.UI.Theme.Shared/wwwroot/libs/abp/aspnetcore-mvc-ui-theme-shared/datatables/datatables-extensions.js
+++ b/framework/src/Volo.Abp.AspNetCore.Mvc.UI.Theme.Shared/wwwroot/libs/abp/aspnetcore-mvc-ui-theme-shared/datatables/datatables-extensions.js
@@ -373,8 +373,9 @@ var abp = abp || {};
                         promise.jqXHR.abort();
                     }
                     promise = serverMethod(input);
-                    promise.then(function (result) {
+                    promise.always(function () {
                         promise = null;
+                    }).then(function (result) {
                         callback(responseCallback(result));
                     });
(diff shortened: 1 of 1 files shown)


michaelsudnik:
Will this work as expected if we have multiple datatable instances on a single page? E.g. within tabs?

maliming:
@michaelsudnik 

No, It only controls one table's request.

michaelsudnik:
I was just worried that, if table1 is refreshing and then I make a change to table2 page count, it could cancel the request for table1?

maliming:
Only control its own behavior and will not affect each other

framework/src/Volo.Abp.AspNetCore.Mvc.UI.Theme.Shared/wwwroot/libs/abp/aspnetcore-mvc-ui-theme-shared/datatables/datatables-extensions.js
```diff
@@ -368,7 +369,12 @@ var abp = abp || {};
                 }
 
                 if (callback) {
-                    serverMethod(input).then(function (result) {
+                    if(cancelPreviousRequest && promise && promise.jqXHR) {
+                        promise.jqXHR.abort();
+                    }
+                    promise = serverMethod(input);
+                    promise.then(function (result) {
+                        promise = null;
```
hikalkan, line 377: Should we set promise to null, also on failure?

hikalkan (approved):

PR #8170, opened 2021-03-23
Create DbContextOptions in a fully async.

maliming:
It is already async now!

https://github.com/abpframework/abp/blob/1d650f8bf79570e890e648fcf845fcfc5ffa89de/framework/src/Volo.Abp.EntityFrameworkCore/Volo/Abp/Uow/EntityFrameworkCore/UnitOfWorkDbContextProvider.cs#L132


maliming:
It is already async now!

https://github.com/abpframework/abp/blob/1d650f8bf79570e890e648fcf845fcfc5ffa89de/framework/src/Volo.Abp.EntityFrameworkCore/Volo/Abp/Uow/EntityFrameworkCore/UnitOfWorkDbContextProvider.cs#L132

PR #8198, opened 2021-03-25
Use lazy in connection pooling

### LINKED ISSUE #8192: Using concurrent dictionary as connection pool has concurrency problem
Using concurrent dictionary as connection pool has concurrency problem；

such as  ConnectionPool in AbpRabbitMqModule， ModelCache  in AbpMongoDbModule  and so on

ABP Framework version:  latest

realLiangshiwei:
Resolve https://github.com/abpframework/abp/issues/8192

The number of connections is limited, we should make sure to create it only once.

https://www.rabbitmq.com/connections.html

> A connection leak is a condition under which an application repeatedly opens connections without closing them, or at least closing only some of them.

> Connection leaks eventually exhaust the node (or multiple target nodes) of file handles, which means any new inbound client, peer or CLI tool connection will be rejected. A build-up in the number of concurrent connections also increases node's memory consumption.

https://github.com/dotnet/aspnetcore/blob/main/src/Mvc/Mvc.Core/src/Filters/MiddlewareFilterBuilder.cs

> 'GetOrAdd' call on the dictionary is not thread safe and we might end up creating the pipeline more than
once. To prevent this Lazy<> is used. In the worst case multiple Lazy<> objects are created for multiple
 threads but only one of the objects succeeds in creating a pipeline.

### DIFF
--- a/framework/src/Volo.Abp.Kafka/Volo/Abp/Kafka/ConsumerPool.cs
+++ b/framework/src/Volo.Abp.Kafka/Volo/Abp/Kafka/ConsumerPool.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using System.Collections.Concurrent;
-using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using Confluent.Kafka;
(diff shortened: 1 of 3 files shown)


maliming (approved):

fmpfmp:
Can't use lazy for Rabbitmq  connectionpool;
becase rabbitmq client use lock when call "ConnectionOpen";

line 242:   https://github.com/rabbitmq/rabbitmq-dotnet-client/blob/5c4c6ddfc3f18d08dcb7dc962864b1e181b161ca/projects/RabbitMQ.Client/client/impl/ModelBase.cs

if valueFactory  has lock internally, deadlocks can occur. 
https://docs.microsoft.com/zh-cn/dotnet/framework/performance/lazy-initialization
https://docs.microsoft.com/zh-cn/dotnet/api/system.threading.lazythreadsafetymode?view=net-5.0

PR #8225, opened 2021-03-26
Use Visible instead of RequiredPolicy for Blazor

berkansasmaz:
The width and height of the replaced image are almost the same:

image
<br>
Also, maybe we can remove the code block in this [section](https://github.com/abpframework/abp/blob/dev/docs/en/Tutorials/Part-5.md#showhide-the-actions) because none of these variables are used.

image

**After this PR is approved, I will update the Commercial documents.**


hikalkan:
policy names are used in the AbpCrudPageBase class: https://github.com/abpframework/abp/blob/883b2c72554f70283ce87962af9d99296c5cabff/framework/src/Volo.Abp.BlazoriseUI/AbpCrudPageBase.cs#L235-L251

hikalkan (approved):

PR #8366, opened 2021-04-03
fix:  AuthPasswordFlowStrategy init() method logs out on refresh

### LINKED ISSUE #8347: Password Flow init() method logs out on refresh
@mehmet-erim 
There is a bug that logs out the user on page refresh using password flow authentication.

Steps to reproduce:
1- `abp new pw-flow-init-test -csf -t app -ui angular -d mongodb --preview`
2- Change env.ts to password flow: https://docs.abp.io/en/abp/4.3/UI/Angular/Authorization#resource-owner-password-flow
3- Run the projects
4- Go to http://localhost:4200/account/login (this is the key to bug: navigate to this url directly by clicking this link, do not go to login page by clicking "login" button on home page)
5- Login using admin user and default password.
6- Now that you are logged in and redirected to home page: refresh the page (using F5 or CTRL+R)
7- Observe that you are logged out after page refresh is finished.

I found that this is because rememberMe cookie is not set on direct navigation to /account/login so this condition returns true and password flow strategy's `init` method logs you out. The code is here:
https://github.com/abpframework/abp/blob/3c35bce82e5b646cc3f59fb74882a31803e4c37c/npm/ng-packs/packages/core/src/lib/strategies/auth-flow.strategy.ts#L143

Alternatively you can go to login page and make a refresh so that the first entrypoint to the app is the /account/login page. Now after you login, either you refresh the page or click a link to open in new tab (meaning a new `init` call will be made) and you will see that it logs you out.

I would like to create a PR however i couldn't see the reason of logging the user out when initializing the strategy. I thought there may be a reason which i'm not aware of. I think even if user didn't choose 'remember me', we should not logout the user on page initialization.

mtozlu:


mehmet-erim:
Hi @mtozlu, thanks for the contribution.

This will be caused a bug. Init method works on app launch. This will override the existing cookie value and expression date.

mtozlu:
You are right. 
But related issue persists.
Thanks.

PR #8370, opened 2021-04-03
Add wwwroot/libs to .gitignore for aspnet-core template

### LINKED ISSUE #8369: Add wwwroot/libs folder to .gitignore for templates
`wwwroot/libs` folder can be filled with gulp by running `yarn install && gulp` command so there is no need to include them in git.
This reduces initial tracked git file count from 455 to 113 in new apps that are generated from templates.

mtozlu:
Closes #8369


maliming:
This is a fundamental design decision and it has taken by considering tradeoffs.

mtozlu:
Thank you for the information.
Mind you, this PR doesn't remove files from templates. They stay as-is. This is for only not tracking library files in git.
Would you mind sparing me motivations behind the decision?

maliming:
See https://github.com/abpframework/abp/issues/8528

PR #8399, opened 2021-04-05
Use IRemoteStreamContent for media uploading

cotur:


enisn (approved):
I've tested for tiered and existing problem looks like solved. 

We can merge after https://github.com/volosoft/volo/issues/6118 solution

ilkayilknur (approved):

Issue #8289, opened 2021-03-31
Auto login (Windows Login) with LDAP


hikalkan:
Windows Authentication for automatic login with ABP is being asked many times.
Create an LDAP environment using docker or any other tech. Try to implement this and write an article

https://github.com/abpframework/abp/issues/5169
https://github.com/abpframework/abp/issues/3638
https://github.com/abpframework/abp/issues/1189

maliming:
https://github.com/abpframework/abp/pull/8419
https://docs.microsoft.com/en-us/aspnet/core/security/authentication/windowsauth?view=aspnetcore-5.0&tabs=visual-studio

jprodriguez-nbs:
Version 6.0.1 integrates openiddict. Could you please provide an example or documentation on how to modify a Volo Abp solution using OpenIdDict to use Auto Login (Windows Login) with LDAP?

talal3000:
bump. seconding @jprodriguez-nbs question. can this work with OpenIdDict?
If a user is logged in already with Windows AD then use the current user otherwise log using a form?

PR #8419, opened 2021-04-06
Refactor ProcessWindowsLoginAsync.

maliming:
About https://github.com/abpframework/abp/issues/8289

```cs
context.Services.Configure<IISServerOptions>(iis =>
{
    iis.AuthenticationDisplayName = "Windows";
    iis.AutomaticAuthentication = false;
});
```

image
image
image


hikalkan (approved):

zafariNet:
Hello,
I want to have windows authentication and abp.io users at the same time. I mean that when user open abp.io application I want first detect Current windows username and depend on his Windows group assign a predefined abp.io role to it and sign user in and generate token and refresh token for this user and send it back to user. From now, we consider this user as an abp.io user. And if a Windows user does not have a group with the same name in abp.io roles, I would like to redirect the user to a login page and the user must enter a valid username and password.
I am using Angular project for UI.
can you please help me ?
Thank you

PR #8514, opened 2021-04-08
Add exception to audit log if it not record.

maliming:
If an exception occurs during the activation of the **controller** or **page**.

image


hikalkan (approved):

hikalkan:
Thanks. Good solution 👍

PR #8543, opened 2021-04-08
Add state check to permissions.

### LINKED ISSUE #8518: Add restriction to permissions
Add `FeatureName` and `GlobalFeatureName` properties to `PermissionDefinition`.

* Check feature name from `IFeatureChecker`. If it returns false, `PermissionChecker` should return false when we check IsGranted for this permission. Also, `PermissionAppService.GetAsync` should not return this permission in the returning list.
* Check the global feature name from `GlobalFeatureManager`. If it returns false, `PermissionChecker` should return false when we check IsGranted for this permission. Also, `PermissionAppService.GetAsync` should not return this permission in the returning list.

You should also care this on PermissionManager.

Hint: Find usages of `PermissionDefinition.IsEnabled`. You typically mostly changes is same places for checking.

maliming:
Resolve #8518

We can add an `IPermissionStateProvider` to permission or globally.

```cs

group.AddPermission("MyPermission1").AddStateProvider(new RequireEditionPermissionStateProvider());

services.Configure<AbpPermissionOptions>(options =>
{
    return options.GlobalStateProviders.Add<GlobalRequireRolePermissionStateProvider>();
});
```

If we want add build-in `RequireFeaturePermissionStateProvider` and `RequireGlobalFeaturePermissionStateProvider`.
`Volo.Abp.GlobalFeatures` and `Volo.Abp.Features` need to depend on the `Volo.Abp.Authorization` module.

I'm not sure about that.


cotur:
I think `Volo.Abp.GlobalFeatures` and `Volo.Abp.Features` can depends on to `Volo.Abp.Authorization` package, because after that, permissions are Feature Depended

hikalkan:
@maliming I think you can create a `Volo.Abp.Authorization.Abstractions` package, move the IPermissionStateProvider to that package. Then you can depend to that package from `Volo.Abp.GlobalFeatures` and `Volo.Abp.Features` packages. Actually I don't see a problem by directly depending on the authorization package, but the abstraction package usage reduces the dependency tree.

maliming:
ok, I will update.

maliming:
@hikalkan Done.

hikalkan (approved):

Issue #8549, opened 2021-04-09
[api-definition] please consider that replace GetCustomAttributes with TypeDescriptor.GetAttributes


yinchang0626:
* ABP Framework version
4.2.2
* Your **User Interface** type (Angular/MVC/React... etc.) 
don't care 
* Your database provider(EF Core/MongoDB)
don't care
* Exception message and **stack trace** if available (check the logs).
none

@maliming Hi:
When I use abp cli to `generate-proxy`,the result is depend on `/api/abp/api-definition`
Sometimes, we should modify `AreaAttribute` for a `Controller` in exist module,like below, I want to modify module name "app" to "eshop" or others:

![image](https://user-images.githubusercontent.com/16490084/114126190-8c82f180-992a-11eb-8fdc-2e211aa5d262.png)

And we know  that 'System.ComponentModel.TypeDescriptor' can help me to modify `Attributes` at runtime.
but it should modify some code in `AspNetCoreApiDescriptionModelProvider`.

I would be grateful if you could modify it for  meet such a requirement.thanks.
also i created a [branch](https://github.com/yinchang0626/abp/commit/0ebe49fbe85d6a032ba921057f32ddbfb13fd2db) to show about this .

PR #8550, opened 2021-04-09
modify GetAttributes by  System.ComponentModel.TypeDescriptor

yinchang0626:
about #8549

### DIFF
--- a/framework/src/Volo.Abp.AspNetCore.Mvc/Volo/Abp/AspNetCore/Mvc/AspNetCoreApiDescriptionModelProvider.cs
+++ b/framework/src/Volo.Abp.AspNetCore.Mvc/Volo/Abp/AspNetCore/Mvc/AspNetCoreApiDescriptionModelProvider.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.ComponentModel;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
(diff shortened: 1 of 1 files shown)


maliming:
hi

The unit tests failed.

https://github.com/abpframework/abp/blob/dev/framework/test/Volo.Abp.AspNetCore.Mvc.Versioning.Tests/Volo/Abp/AspNetCore/Mvc/Versioning/App/HelloController.cs#L6-L7

https://docs.microsoft.com/en-us/dotnet/api/system.componentmodel.typedescriptor.getattributes?view=net-5.0#System_ComponentModel_TypeDescriptor_GetAttributes_System_Type_

yinchang0626:
@maliming Hi:
Thank you for pointing out the mistake,
but because of `ApiVersionAttribute` is  a `AllowMultiple` Attribute, and according  to msdn:
```text
The default implementation of Attribute.TypeId returns the type identity regardless of the value of the 
AttributeUsageAttribute.AllowMultiple property. In order to return multiple instances of an 
AttributeUsageAttribute.AllowMultiple attribute from the AttributeCollection, your attribute must override the Attribute.TypeId 
 property.
```
So,In this P.R,
I just modify `GetAttributes` method to get `AreaAttribute` and `RemoteServiceAttribute` but not `ApiVersionAttribute`

maliming:
hi

I think we can add more `Generator` to `AspNetCoreApiDescriptionModelProviderOptions`.

https://github.com/abpframework/abp/blob/48c52625f4c4df007f04d5ac6368b07411aa7521/framework/src/Volo.Abp.AspNetCore.Mvc/Volo/Abp/AspNetCore/Mvc/AspNetCoreApiDescriptionModelProviderOptions.cs#L15-L19

PR #8554, opened 2021-04-09
Implemented Basic & ApiKey authentication for elasticsearch

### LINKED ISSUE #8553: Docs - ElasticSearch should be usable with authentication
Hi everyone,

we're using a elasticsearch cluster with authentication enabled.

It would be nice if the docs-module would show how to use the full-text-search feature with a authentication-enabled elasticsearch cluster, so more users will look into authentication and securing their services.

I know, I can just implement the `IElasticClientProvider` but it would be better to support this scenario ootb.
I'll submit a PR which adresses this issue.

NecatiMeral:
Fixes #8553

Added configuration classes to make basic and apikey based authentication possible.


maliming:
hi @NecatiMeral 

I did some refactor.

```cs
Configure<DocsElasticSearchOptions>(options =>
{
	options.Enable = true;
	options.UseBasicAuthentication("root", "123");
});
```

maliming (approved):

PR #8555, opened 2021-04-09
Store multiple connection string change info in TenantConnectionStringsUpdatedEto.

maliming:


hikalkan:
We don't want to make a breaking change here. Please cancel this PR. If multiple connection string is changed for a tenant, then publish multiple events, each for a single database.

PR #8561, opened 2021-04-09
Allow to directly enable/disable features on the GlobalFeatureManager

hikalkan:
- Allowed to create individual features (module-indepdendent) with this commit: https://github.com/abpframework/abp/commit/424595bf07b6e882be5b1890f2b8d8428495c896 (allows to not derive from the base class)
- Allow to directly enable/disable features on the GlobalFeatureManager.Instance.
- Revised the document.
- Added a TODO at the end, for you to complete.

### DIFF
--- a/docs/en/Global-Features.md
+++ b/docs/en/Global-Features.md
@@ -91,4 +91,56 @@ public class PaymentController : AbpController
 
 It is common to group global features of a module to allow the final application developer easily discover and configure the features.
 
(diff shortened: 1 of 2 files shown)


enisn:
After those changed `AddFeature` method from `GlobalModuleFeatures` class looks unuseful. Should we deprecate that method. Because features can be Enabled/Disabled without adding to **AllFeatures** list first.

@hikalkan

enisn (approved):
It's ready to merge after you reviewed the documentation @hikalkan

PR #8562, opened 2021-04-09
AbpLdapOptions never be configured

Rorschach331:
`configuration["Ldap"]` always return null, even set the "Ldap" section in appsettings.json.
so `Configure<AbpLdapOptions>(configuration.GetSection("Ldap"));`   does not execute,i think its a bug..


maliming:
hi

I move it to 4.3. Thanks.
https://github.com/abpframework/abp/pull/8563

Issue #6089, opened 2020-11-06
Define an option to add extension properties for the user to the claims (auth ticket)


hikalkan:

PMExtra:
I want add a new interface:
```csharp
interface IUserClaimProvider<TUser>
{
    Task<IEnumerable<Claim>> GetClaimsAsync(TUser user);
}
```

Then inject `IEnumerable<IUserClaimProvider<IdentityUser>> _userClaimProviders` in [`AbpUserClaimsPrincipalFactory`](https://github.com/abpframework/abp/blob/dev/modules/identity/src/Volo.Abp.Identity.Domain/Volo/Abp/Identity/AbpUserClaimsPrincipalFactory.cs)

It will call the each of providers in [`AbpUserClaimsPrincipalFactory.CreateAsync`](https://github.com/abpframework/abp/blob/48c52625f4c4df007f04d5ac6368b07411aa7521/modules/identity/src/Volo.Abp.Identity.Domain/Volo/Abp/Identity/AbpUserClaimsPrincipalFactory.cs#L35-L40):
```csharp
foreach (var provider in _userClaimProviders)
foreach (var claim in await provider.GetClaimsAsync(user))
{
    identity.AddIfNotContains(claim);
}
```

Would it solve this issue? I can make the pull request if you accept this solution.

maliming:
This feature already implemented. I will add the document to explain.

https://github.com/abpframework/abp/blob/dev/framework/src/Volo.Abp.Security/Volo/Abp/Security/Claims/AbpClaimsPrincipalFactory.cs

https://github.com/abpframework/abp/blob/dev/framework/test/Volo.Abp.Security.Tests/Volo/Abp/Security/Claims/AbpClaimsPrincipalFactory_Test.cs

https://github.com/abpframework/abp/blob/dev/modules/identity/src/Volo.Abp.Identity.Domain/Volo/Abp/Identity/AbpUserClaimsPrincipalFactory.cs#L73
https://github.com/abpframework/abp/blob/dev/modules/identity/test/Volo.Abp.Identity.Domain.Tests/Volo/Abp/Identity/AbpUserClaimsPrincipalFactory_Tests.cs

PR #8569, opened 2021-04-11
Add document of IAbpClaimsPrincipalFactory.

maliming:
https://github.com/abpframework/abp/issues/6089

### DIFF
--- a/docs/en/Authorization.md
+++ b/docs/en/Authorization.md
@@ -424,7 +424,7 @@ public class EditionClaimsPrincipalContributor : IAbpClaimsPrincipalContributor,
 
 The `EditionClaimsPrincipalContributor` will participate in it when the `CreateAsync` method of `IAbpClaimsPrincipalFactory` is called. 
 
-> The Identity module has integrated it.
+> The [Identity module](https://docs.abp.io/en/abp/latest/Modules/Identity) has integrated it.
 
 ## See Also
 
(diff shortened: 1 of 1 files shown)


hikalkan:
This is for custom properties but we are showing an example that uses EditionId (thats a standard claim defined by ABP framework), which may confuse people. I think we should provide another example that feels that it is really custom, like `SocialSecurityNumber` for example.

maliming:
@hikalkan Changed

docs/en/Authorization.md
```diff
@@ -393,6 +393,39 @@ public override void ConfigureServices(ServiceConfigurationContext context)
...
+    {
+        var identity = context.ClaimsPrincipal.Identities.FirstOrDefault();
+        if (identity != null)
+        {
+            var currentUser = context.ServiceProvider.GetRequiredService<ICurrentUser>();
+            if (currentUser.Id.HasValue)
+            {
+                var userManager = context.ServiceProvider.GetRequiredService<IdentityUserManager>();
+                var user = await userManager.GetByIdAsync(currentUser.Id.Value);
+                if (user?.SocialSecurityNumber != null)
```
PMExtra, line 415: It's not possible. `user` is an instance of `IdentityUser` that does NOT have `SocialSecurityNumber` property.
It should be something like `user.GetProperty("SocialSecurityNumber", default(string))`.

maliming: It’s just a example.

hikalkan (approved):

PR #8612, opened 2021-04-14
Use Enum.Parse to convert the enum.

maliming:
Related: https://github.com/abpframework/abp/pull/8602


ilkayilknur (approved):

ilkayilknur:
That's much safer. Thanks.

PR #8667, opened 2021-04-17
refactor: change Initialize in VirtualFileLocalization to virutal

badre429:
i need to inherit from (VirtualFileLocalizationResourceContributorBase) and apply some database logic to extract field localization 
i can copy the full code but it better if initialize function is virutal that way i can  replace and get the service that i need


maliming (approved):

maliming:
@badre429 👍

PR #8673, opened 2021-04-18
Supports the use of multiple template rendering engines in one application.

### LINKED ISSUE #8651: Supports the use of multiple template rendering engines in one application.
Related: https://github.com/abpframework/abp/pull/8392

We should be able to set the rendering engine as `TemplateDefinition.RenderEngine`. When we render a template, it should understand the engine and use the right rendering engine.  This property should be string, so it can be extensible, we can define new engines and register to the text template system.

maliming:
Resolve #8651

The `AbpTextTemplatingModule` is depend on the `AbpTextTemplatingScribanModule` and `AbpTextTemplatingRazorModule`  now.

```cs
context.Add(
  new TemplateDefinition(
      TestTemplates.HybridTemplateRazor,
      localizationResource: typeof(TestLocalizationSource),
      layout: null
  )
  .WithVirtualFilePath("/SampleTemplates/TestRazorTemplate.cshtml", true)
  .WithRenderEngine(RazorTemplateRendererProvider.ProviderName)
```


docs/en/Text-Templating-Razor.md
```diff
@@ -84,6 +84,7 @@ public class DemoTemplateDefinitionProvider : TemplateDefinitionProvider
     {
         context.Add(
             new TemplateDefinition("Hello") //template name: "Hello"
+                .WithRenderEngine(RazorTemplateRendererProvider.ProviderName)
```
hikalkan: What about to define some sugar extension methods like `.WithRazorTemplate()` and `.WithScribanTemplate()` as shortcut of `.WithRenderEngine` (keep it for extensibility).

docs/en/Text-Templating.md
```diff
@@ -24,6 +24,8 @@ ABP Framework provides two types of engines;
 * **[Razor](Text-Templating-Razor.md)**
 * **[Scriban](Text-Templating-Scriban.md)**
 
+You can use different template engines at the same time, or even create a new template engine.
```
hikalkan:
```suggestion
You can use different template engines in the same application, or even create a new custom template engine.
```

framework/src/Volo.Abp.TextTemplating.Abstractions/Volo/Abp/TextTemplating/AbpTemplateRendererProviderOptions.cs
```diff
@@ -0,0 +1,33 @@
...
+            Providers = new Dictionary<string, Type>();
+        }
+
+        public AbpTemplateRendererProviderOptions AddProvider<TProvider>(string templateRenderEngineName)
+            where TProvider : ITemplateRendererProvider
+        {
+            if (Providers.ContainsKey(templateRenderEngineName))
+            {
+                Providers.RemoveAll(x => x.Key == templateRenderEngineName);
+            }
```
hikalkan: Seems unnecessary since the following line will override it.

framework/src/Volo.Abp.TextTemplating.Abstractions/Volo/Abp/TextTemplating/AbpTemplateRendererProviderOptions.cs
```diff
@@ -0,0 +1,33 @@
...
+    public class AbpTemplateRendererProviderOptions
+    {
+        protected IDictionary<string, Type> Providers { get; }
+
+        public AbpTemplateRendererProviderOptions()
+        {
+            Providers = new Dictionary<string, Type>();
+        }
+
+        public AbpTemplateRendererProviderOptions AddProvider<TProvider>(string templateRenderEngineName)
```
hikalkan: We use two terms for the same thing: "render engine" and "renderer provider". I think we should go with one of them. "Rendering engine" seems reasonable to me. So, we should also rename `AbpTemplateRendererProviderOptions` to `AbpTemplateRenderingEngineOptions`, or I even suggest to simply define `AbpTextTemplatingOptions` and use a RenderEngines dictonary inside it.

framework/src/Volo.Abp.TextTemplating.Abstractions/Volo/Abp/TextTemplating/ITemplateRendererProvider.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System.Collections.Generic;
+using System.Globalization;
+using System.Threading.Tasks;
+using JetBrains.Annotations;
+
+namespace Volo.Abp.TextTemplating
+{
+    public interface ITemplateRendererProvider
```
hikalkan: Consider to rename as `ITemplateRenderingEngine` following my previous suggestion.

framework/src/Volo.Abp.TextTemplating.Abstractions/Volo/Abp/TextTemplating/TemplateDefinition.cs
```diff
@@ -37,6 +37,9 @@ public ILocalizableString DisplayName
         [CanBeNull]
         public string DefaultCultureName { get; }
 
+        [CanBeNull]
+        public string RenderEngine { get; protected set; }
```
hikalkan: Making the setter public would be fine if someone want to override the template definition with a different rendering engine.

framework/src/Volo.Abp.TextTemplating/Volo.Abp.TextTemplating.csproj
```diff
@@ -10,6 +10,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\Volo.Abp.TextTemplating.Scriban\Volo.Abp.TextTemplating.Scriban.csproj" />
+    <ProjectReference Include="..\Volo.Abp.TextTemplating.Razor\Volo.Abp.TextTemplating.Razor.csproj" />
```
hikalkan: I think we should not add razor to this package. It will bring additional dependencies to current applications.
We keep this package for backward compatibility. We should mark this module as obsolete by indicating to use other packages.

framework/src/Volo.Abp.TextTemplating/Volo/Abp/TextTemplating/TextTemplating/AbpTemplateRenderer.cs
```diff
@@ -0,0 +1,53 @@
...
+            if (renderEngine.IsNullOrWhiteSpace())
+            {
+                renderEngine = ScribanTemplateRendererProvider.ProviderName;
+            }
+
+            var providerType = Options.GetProviderTypeOrNull(renderEngine);
+
+            if (providerType != null)
+            {
+                var templateRendererProvider = (ITemplateRendererProvider)ServiceProvider.GetRequiredService(providerType);
```
hikalkan: We are getting a service from the service provider, but not releasing it in anywhere. Instead, we should inject IServiceScopeFactory and create an inner scope here that is disposed after the rendering process.

framework/src/Volo.Abp.TextTemplating/Volo/Abp/TextTemplating/TextTemplating/AbpTemplateRenderer.cs
```diff
@@ -0,0 +1,53 @@
...
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
+using Volo.Abp.DependencyInjection;
+using Volo.Abp.TextTemplating.Scriban;
+
+namespace Volo.Abp.TextTemplating
+{
+    public class AbpTemplateRenderer : ITemplateRenderer, ITransientDependency
```
hikalkan: I think we have a design issue here. Core template rendering system should be render engine independent. Developer should add render engine packages to add render engines to the application.

So, we are actually dropping the Volo.Abp.TextTemplating package. But, for backward compability, we keep it and add a reference to Volo.Abp.TextTemplating.Scriban to keep the current behavior. However, this module will be completely empty, and probably will be deleted in the future
. 
All the code should be moved to Volo.Abp.TextTemplating.Core package

Volo.Abp.TextTemplating.Abstractions -> rename to Volo.Abp.TextTemplating.Core
Volo.Abp.TextTemplating.Razor -> depends on Volo.Abp.TextTemplating.Core
Volo.Abp.TextTemplating.Scriban -> depends on Volo.Abp.TextTemplating.Core


So, if I want to use Razor, I add Volo.Abp.TextTemplating.Razor to my application. If I want to use both engines, I add both of the razor and scriban packages.
If I only add Volo.Abp.TextTemplating.Core package, I can't use any engine, so I get exception when I try to render.

framework/src/Volo.Abp.TextTemplating/Volo/Abp/TextTemplating/TextTemplating/AbpTemplateRenderer.cs
```diff
@@ -0,0 +1,53 @@
...
+            string cultureName = null,
+            Dictionary<string, object> globalContext = null)
+        {
+            var templateDefinition = TemplateDefinitionManager.Get(templateName);
+
+            var renderEngine = templateDefinition.RenderEngine;
+
+            if (renderEngine.IsNullOrWhiteSpace())
+            {
+                renderEngine = ScribanTemplateRendererProvider.ProviderName;
```
hikalkan: Core rendering engine can't be scriban dependent. To set a default engine, define a property (`string DefaultRenderingEngine`) in the text template options class and set it to scriban in the scriban package (force to set without checking the current value). Also, in the razor package set it to razor, but only if it is null. This will make the system backward compatible, while we allow razor to be default only if it is the only engine in the application.

hikalkan (changes requested):

hikalkan (approved):

hikalkan:
Thanks for the implementation. Great feature!

hikalkan:
I've done some renamings: https://github.com/abpframework/abp/commit/fdcfb502ed05fcb88936eb9daffe6059c6292486

PR #8694, opened 2021-04-19
Refactor state check feature.

maliming:


maliming:
Move to 4.4 https://github.com/abpframework/abp/pull/8742

PR #8732, opened 2021-04-21
Cli: Process multiple EfCore ModelBuilder Configure Method when adding a module

yekalkan:
to allow 

````
            builder.ConfigureCmsKit();
            builder.ConfigureCmsKitPro();
````

and similar usages.


hikalkan:
Ok, we merge this, but the true solution for cmskit is that: ConfigureCmsKitPro should call ConfigureCmsKit internally.

hikalkan (approved):

PR #8782, opened 2021-04-25
Standardize passing and setting of content in Tag Helpers

acjh:
Purely refactoring.

Changes:
1. Standardize passing of child content.
   - Pass child content as `TagHelperContent` instead of `string` from unnecessary call to `.GetContent()`.
   - Rename `innerContent` to `childContent`.
2. Simplify usage of `SetHtmlContent`  to `AppendHtml` instead of `.GetContent()` and string concatenation.


maliming (approved):

maliming:
@acjh  👍

PR #8814, opened 2021-04-27
Add IsActive property to TenantConfiguration.

maliming:
We didn't implement the **IsActive** feature of **Tenant**, it is **true** by default. It will not break anything.


framework/src/Volo.Abp.AspNetCore.Mvc.UI.MultiTenancy/Pages/Abp/MultiTenancy/TenantSwitchModal.cshtml.cs
```diff
@@ -48,7 +48,7 @@ public async Task OnPostAsync()
             else
             {
                 var tenant = await TenantStore.FindAsync(Input.Name);
-                if (tenant == null)
+                if (tenant == null || !tenant.IsActive)
```
hikalkan, line 51: I think a different exception message for the `!tenant.IsActive` would be better.

hikalkan (approved):

PR #8829, opened 2021-04-28
Introduce event bus error handler

### LINKED ISSUE #7657: RabbitMQ rapidly re-delivers the messages
When we have an exception while processing a distributed event (integrated to RabbitMQ), we send a NACK message to the server, so it is re-queued. However, RabbitMQ immediately re-sends the message, and probably we get the error again. It would be better to stop sending messages to this application for a while.
I don't know if RabbitMQ has such a configuration. If anyone knows, please write here, otherwise we will look for it.
Otherwise, how can we solve that problem?

realLiangshiwei:
Resolve https://github.com/abpframework/abp/issues/7657

I abstracted `IEventErrorHandler` to handle event exceptions and added `EventErrorHandlerBase` abstract class, which provides abstract methods for **retrying** and **moving to the dead letter queue**

* Local: throw original exceptions if retry fails.
* Rabbitmq: move to the dead letter if retry fails. https://www.rabbitmq.com/dlx.html
* Kafka: move to the dead letter if retry fails. 
* Rebus: https://github.com/rebus-org/Rebus/wiki/Automatic-retries-and-error-handling

### usage

```csharp
PreConfigure<AbpEventBusOptions>(options =>
{
      options.UseRetryStrategy();
});
```


hikalkan:
@realLiangshiwei thanks for the implementation. What happens to events drops into dead letter queue? Is there any recommended way to explore or recover them, or we should just forget about them?

realLiangshiwei:
Dead letter queue actually is a normal queue, message will be stored in the queue persistently. 

image

We can move the message to the work queue using Web UI or create a consumer to receive the dead letter queue. However, developers should care for them.

image

slarkerino:
Is this feature avaliable now? XD

slarkerino:
@hikalkan Can we merge this feature please :)

hikalkan (approved):

maliming (approved):

volethanh:
> Resolve #7657
> 
> I abstracted `IEventErrorHandler` to handle event exceptions and added `EventErrorHandlerBase` abstract class, which provides abstract methods for **retrying** and **moving to the dead letter queue**
> 
> * Local: throw original exceptions if retry fails.
> * Rabbitmq: move to the dead letter if retry fails. https://www.rabbitmq.com/dlx.html
> * Kafka: move to the dead letter if retry fails.
> * Rebus: https://github.com/rebus-org/Rebus/wiki/Automatic-retries-and-error-handling
> 
> ### usage
> ```cs
> PreConfigure<AbpEventBusOptions>(options =>
> {
>       options.UseRetryStrategy();
> });
> ```

Hi @realLiangshiwei ,

I'm using abp.eventbus.Rebus v4.4.0 but I cannot handle error message by implement `IEventErrorHandler`. I don't know if I do any wrong ?

Screen Shot 2021-09-11 at 11 32 28 AM

I also tried implement the way in Rebus wiki guide `IHandleMessages<IFailed<DoStuff>>` and tried all the ways but nothing enter to the method ErrorHandler

Screen Shot 2021-09-11 at 11 34 17 AM

My rebus configure in module as below

Screen Shot 2021-09-11 at 11 26 50 AM

realLiangshiwei:
@volethanh 

Please open an issue, thanks.

PR #8897, opened 2021-05-05
Introduce ReplaceDbContextAttribute

### LINKED ISSUE #8889: Introduce ReplaceDbContextAttribute
Consider the following code:

````csharp
context.Services.AddAbpDbContext<MainDbContext>(options =>
{
    options.AddDefaultRepositories();
    options.ReplaceDbContext<IIdentityDbContext>();
});
````

If we have a `ReplaceDbContextAttribute`, then we can add it on the `MainDbContext` and remove the `options.ReplaceDbContext<IIdentityDbContext>();` usage.

````csharp
[ReplaceDbContext(typeof(IIdentityDbContext))]
public partial class MainDbContext : AbpDbContext<MainDbContext>, IIdentityDbContext
{
    ...
}
````

Please also update the documentation ([ef](https://docs.abp.io/en/abp/latest/Entity-Framework-Core#replace-other-dbcontextes) & [mongo](https://docs.abp.io/en/abp/latest/MongoDB#replace-other-dbcontextes)).

### PR DESCRIPTION
**Pull request by realLiangshiwei**
Resolve https://github.com/abpframework/abp/issues/8889

### CODE REVIEW

File: framework/src/Volo.Abp.EntityFrameworkCore/Volo/Abp/EntityFrameworkCore/DependencyInjection/AbpDbContextConventionalRegistrar.cs [hunk 1]
```diff
@@ -0,0 +1,37 @@
+﻿using System;
+using System.Reflection;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection.Extensions;
+using Volo.Abp.DependencyInjection;
+
+namespace Volo.Abp.EntityFrameworkCore.DependencyInjection
+{
+    public class AbpDbContextConventionalRegistrar : DefaultConventionalRegistrar
```
hikalkan: We should not create a registrar. We can do it much simpler.

See this:

https://github.com/abpframework/abp/blob/8770fd7811de54b40762de3df11b84d8c7f8c443/framework/src/Volo.Abp.EntityFrameworkCore/Microsoft/Extensions/DependencyInjection/AbpEfCoreServiceCollectionExtensions.cs#L10-L29

Here, before `optionsBuilder?.Invoke(options);`, we can get the ReplaceDbContextAttribute attribute (or attributes, because there is no restriction - developer may use it multiple times instead of passing multiple types into a single attribute), then call `ReplaceDbContext` for each replaced context. Same for mongdb too.

hikalkan (changes requested):

framework/src/Volo.Abp.MongoDB/Microsoft/Extensions/DependencyInjection/AbpMongoDbServiceCollectionExtensions.cs
```diff
@@ -11,13 +13,20 @@ public static class AbpMongoDbServiceCollectionExtensions
             where TMongoDbContext : AbpMongoDbContext
         {
             var options = new AbpMongoDbContextRegistrationOptions(typeof(TMongoDbContext), services);
+
+            var replaceDbContextAttribute = typeof(TMongoDbContext).GetCustomAttribute<ReplaceDbContextAttribute>(true);
```
hikalkan: Get all ReplaceDbContextAttribute attributes, not single one. We may use like that:

````csharp
[ReplaceDbContextAttribute(typeof(ISomeDbContext))]
[ReplaceDbContextAttribute(typeof(IOtherDbContext))]
public class MyDbContext...
{
}
````

It should replace all.

hikalkan (approved):

hikalkan (changes requested):
Please support multiple usages of the attribute.

realLiangshiwei:
Supported

hikalkan (approved):

PR #8898, opened 2021-05-05
Allow to customize EF Core mapping of entity and dbContext of an existing module

### LINKED ISSUE #8862: Allow to customize EF Core mapping of an entity of an existing module
image

Then the configuration is auto applied in [EfCoreObjectExtensionManagerExtensions.ConfigureEfCoreEntity](https://github.com/abpframework/abp/blob/dev/framework/src/Volo.Abp.EntityFrameworkCore/Volo/Abp/ObjectExtending/EfCoreObjectExtensionManagerExtensions.cs#L116). In this way, no change required in existing modules.

### LINKED ISSUE #8863: Allow to customize configuration of a dbcontext via object extension manager
As an improvement similar to #8862

image

Here, I am trying to customize IdentityDbContext mappings in a static context (customized IdentityUser and also mapped a new entity). To make this work, IdentityDbContextModelBuilderExtensions.ConfigureIdentity method should make such a call:

image

Please make such calls in all modules (including volo)

realLiangshiwei:
Resolve https://github.com/abpframework/abp/issues/8863, https://github.com/abpframework/abp/issues/8862


realLiangshiwei:
See https://github.com/abpframework/abp/issues/8862#issuecomment-838194688

framework/src/Volo.Abp.ObjectExtending/Volo/Abp/ObjectExtending/ObjectExtensionInfo.cs
```diff
@@ -24,7 +24,7 @@ public class ObjectExtensionInfo
 
         public ObjectExtensionInfo([NotNull] Type type)
         {
-            Type = Check.AssignableTo<IHasExtraProperties>(type, nameof(type));
+            Type = type;
```
hikalkan: Checking for null would be good.

framework/src/Volo.Abp.EntityFrameworkCore/Volo/Abp/ObjectExtending/EfCoreObjectExtensionManagerExtensions.cs
```diff
@@ -113,7 +150,7 @@ public static class EfCoreObjectExtensionManagerExtensions
             );
         }
 
-        public static void ConfigureEfCoreEntity(
+        public static void ConfigureEfCoreEntityProperties(
```
hikalkan: Renaming this method (and replacing with another one which does a different thing) is a hidden breaking change which is very dangerous.
I suggest to merge these two methods (`ConfigureEfCoreEntityProperties` and `ConfigureEfCoreEntity`) that performs both entity and property mappings in a single method. This won't be a breaking change and will be more meaningful.

hikalkan (changes requested):

hikalkan (approved):

PR #8983, opened 2021-05-11
Fixing IsExternal flag for external users

ejohnson-dotnet:
External users created with the method CreateExternalUserAsync was not setting the IsExternal flag for the user.


maliming:
hi

I think the social login does not belong to external users.

https://github.com/abpframework/abp/issues/4977

PR #8987, opened 2021-05-12
Document index backgroundjob

ebicoglu:
Elastic search indexing operation is a long-lasting request and sometimes timeout occurs.

*   Move the service to a background job.
*   Notify the UI when it completes.


hikalkan:
I am moving this to 5.0 milestone to review and merge it after 4.4 release this week.

modules/docs/app/VoloDocs.Web/abp.resourcemapping.js
```diff
@@ -7,6 +7,7 @@
         "@libs"
     ],
     mappings: {
-
+        "@node_modules/@microsoft/signalr/dist/browser/signalr.js": "@libs/signalr/",
+        "@node_modules/@microsoft/signalr/dist/browser/signalr.js.map": "@libs/signalr/"
```
hikalkan: We can use the `@abp/signalr` package instead of the manual installation. See https://docs.abp.io/en/abp/latest/SignalR-Integration#asp-net-core-mvc-razor-pages-ui

modules/docs/src/Volo.Docs.Admin.Application/Volo/Docs/Admin/DocsAdminApplicationModule.cs
```diff
@@ -11,7 +12,7 @@ namespace Volo.Docs.Admin
         typeof(DocsAdminApplicationContractsModule),
         typeof(AbpCachingModule),
         typeof(AbpAutoMapperModule),
-        typeof(AbpDddApplicationModule)
+        typeof(AbpDddApplicationModule)
```
hikalkan: I suppose you wanted to add `AbpBackgroundJobsAbstractionsModule` here?

modules/docs/src/Volo.Docs.Admin.Application/Volo/Docs/Admin/Documents/DocumentAdminAppService.cs
```diff
@@ -27,23 +27,23 @@ public class DocumentAdminAppService : ApplicationService, IDocumentAdminAppServ
         private readonly IDistributedCache<DocumentUpdateInfo> _documentUpdateCache;
         private readonly IDistributedCache<List<VersionInfo>> _versionCache;
         private readonly IDistributedCache<LanguageConfig> _languageCache;
-        private readonly IDocumentFullSearch _documentFullSearch;
+        private readonly IDocumentFullSearch _elasticSearchService;
```
hikalkan: I think `_documentFullSearch` is a better field name since `_elasticSearchService` says something about the implementation and breaks encapsulation.

modules/docs/src/Volo.Docs.Admin.Web/DocsAdminWebModule.cs
```diff
@@ -27,6 +32,8 @@ public override void PreConfigureServices(ServiceConfigurationContext context)
             {
                 mvcBuilder.AddApplicationPartIfNotExists(typeof(DocsAdminWebModule).Assembly);
             });
+
+            context.Services.AddSignalR();
```
hikalkan: This is already done in [AbpAspNetCoreSignalRModule](https://github.com/abpframework/abp/blob/dev/framework/src/Volo.Abp.AspNetCore.SignalR/Volo/Abp/AspNetCore/SignalR/AbpAspNetCoreSignalRModule.cs#L35).

modules/docs/src/Volo.Docs.Admin.Web/DocsAdminWebModule.cs
```diff
@@ -46,6 +53,26 @@ public override void ConfigureServices(ServiceConfigurationContext context)
...
                 options.AddProfile<DocsAdminWebAutoMapperProfile>(validate: true);
             });
+
+            Configure<AbpBundlingOptions>(options =>
+            {
+                options
+                    .ScriptBundles
+                    .Get(StandardBundles.Scripts.Global)
+                    .AddFiles("/libs/signalr/signalr.js")
+                    .AddFiles("/Pages/Docs/Admin/docs-notification-hub.js");
```
hikalkan: I think adding signalr to the global scripts is not a good idea. Add it to only the pages we are using it.
Also, `/Pages/Docs/Admin/docs-notification-hub.js` should only be inside a specific page, not in global scripts.

modules/docs/src/Volo.Docs.Admin.Web/Pages/Docs/Admin/docs-notification-hub.js
```diff
@@ -0,0 +1,42 @@
...
+
+    docsNotificationManager.messageListeners.notificationProcessMessages.forEach(function (fn) {
+        fn({
+            message: message
+        });
+    });
+});
+
+var showNotification = function(notification) {
+    abp.message.success(notification.message);
```
hikalkan: I personally think that using SignalR and sending real-time notifications to the UI is lux for this module and it is unnecessary. It also causes a strange user experience. Think that I am on the website (in the document page, or another page with current implementation) and I suddenly see notification messages just because another user re-indexed a document. Notice that with this implementation all users are getting the notification, not only the one has started it.

I think we should drop signalr and real-time notification at all. That's a background job and we don't need to see the result.

modules/docs/src/Volo.Docs.Admin.Web/Volo.Docs.Admin.Web.csproj
```diff
@@ -23,6 +23,8 @@
     <ProjectReference Include="..\..\..\..\framework\src\Volo.Abp.AspNetCore.Mvc.UI.Bundling\Volo.Abp.AspNetCore.Mvc.UI.Bundling.csproj" />
     <ProjectReference Include="..\..\..\..\framework\src\Volo.Abp.AspNetCore.Mvc.UI.Bootstrap\Volo.Abp.AspNetCore.Mvc.UI.Bootstrap.csproj" />
     <ProjectReference Include="..\..\..\..\framework\src\Volo.Abp.AspNetCore.Mvc.UI.Packages\Volo.Abp.AspNetCore.Mvc.UI.Packages.csproj" />
+    <ProjectReference Include="..\..\..\..\framework\src\Volo.Abp.AspNetCore.Mvc.UI.Theme.Shared\Volo.Abp.AspNetCore.Mvc.UI.Theme.Shared.csproj" />
```
hikalkan: Are you sure that we need to this reference? I think it is already indirectly added. Also, if we add we should add dependson for this.

modules/docs/src/Volo.Docs.Domain/Volo/Docs/DocsDomainModule.cs
```diff
@@ -21,7 +22,8 @@ namespace Volo.Docs
     [DependsOn(
         typeof(DocsDomainSharedModule),
         typeof(AbpDddDomainModule),
-        typeof(AbpAutoMapperModule)
+        typeof(AbpAutoMapperModule),
+        typeof(AbpBackgroundJobsModule)
```
hikalkan: We are adding `AbpBackgroundJobsModule` here, but adding AbpBackgroundJobsAbstractionModule to the application layer. `AbpBackgroundJobsModule` already depends on the abstraction package and the application depends on the domain. Please revisit the dependencies.

npm/packs/docs/abp.resourcemapping.js
```diff
@@ -0,0 +1,13 @@
...
+    aliases: {
+        "@node_modules": "./node_modules",
+        "@libs": "./wwwroot/libs"
+    },
+    clean: [
+        "@libs"
+    ],
+    mappings: {
+        "@node_modules/@microsoft/signalr/dist/browser/signalr.js": "@libs/signalr/",
+        "@node_modules/@microsoft/signalr/dist/browser/signalr.js.map": "@libs/signalr/"
```
hikalkan: Instead, the docs package should depend on the @abp/signalr package.

npm/packs/docs/package.json
```diff
@@ -9,7 +9,8 @@
     "@abp/clipboard": "~4.3.0",
     "@abp/malihu-custom-scrollbar-plugin": "~4.3.0",
     "@abp/popper.js": "~4.3.0",
-    "@abp/prismjs": "~4.3.0"
+    "@abp/prismjs": "~4.3.0",
+    "@microsoft/signalr": "^5.0.6"
```
hikalkan: use @abp/signalr package instead

hikalkan (changes requested):

hikalkan:
Also, resolve the conflicts please. I also suggest to keep this work simple and minimal as I've commented.

ebicoglu:
I made the necessary changes. 
@maliming please test this.
you can use abp.io admin page to test it.
if you see any problem, please fix it.
image

maliming (approved):

maliming (approved):

PR #9002, opened 2021-05-12
postgresql connection string updated

can-guven:
error : Keyword not supported: minimumpoolsize (Parameter 'keyword')


maliming:
Can you check https://github.com/abpframework/abp/pull/7562

can-guven:
I think you can use display name or property name as keyword.

These code are same for pool sizing:
Property Name Usage
` return "Host=localhost;Port=5432;Database=MyProjectName;User ID=root;Password=myPassword;Pooling=true;MinPoolSize=0;MaxPoolSize=100;Connection Lifetime=0;";`

Display Name Usage 
` return "Host=localhost;Port=5432;Database=MyProjectName;User ID=root;Password=myPassword;Pooling=true;Minimum Pool Size=0;Maximum Pool Size=100;Connection Lifetime=0;";`

maliming:
hi

We can keep it simple. developers can add `pool` config if needs.

maliming (approved):

maliming:
@can-guven Thanks a lot.

PR #9106, opened 2021-05-20
Update Background-Workers.md

yadurajshakti:
Adding the required NuGet package detail and installation step to start with Background Worker process


cotur (changes requested):
Hi, thanks for your contribution.

There are few missing steps,

1. When you install an ABP package, you have to add `Module Dependency` also. See: https://docs.abp.io/en/abp/latest/Module-Development-Basics#module-dependencies
2. The package can be installed via [ABP Cli `add-package` command](https://docs.abp.io/en/abp/latest/CLI#add-package) also.

Here is the example **Package Installation Guide**:

https://docs.abp.io/en/abp/latest/SignalR-Integration#installation

PR #9133, opened 2021-05-21
External authentication error when confirm email/phone is selected

### LINKED ISSUE #9132: External authentication error when confirm email/phone is selected
`OnGetExternalLoginCallbackAsync` throws **email already taken exception** when confirmed email/phone setting is selected.

It should redirect to a _Not Allowed_ page and inform user about this setting instead of trying to create same user.

gterdem:
Fixes #9132


maliming:
I think an exception is enough for the free module. 

`throw new UserFriendlyException("Cannot proceed because user is not allowed!");`

gterdem:
Updated as requested

maliming (approved):

PR #9161, opened 2021-05-26
Fix ModelMetadata is required by default for validator localization

JadynWong:
> DataAnnotationsClientModelValidatorProvider will add a default '[Required]' validator for generating HTML if necessary.

https://github.com/dotnet/aspnetcore/blob/e7b5aa6f713e9f040ba0730b915ae407d35971c1/src/Mvc/Mvc.DataAnnotations/src/DataAnnotationsClientModelValidatorProvider.cs#L99-L109

### DIFF
--- a/framework/src/Volo.Abp.AspNetCore.Mvc/Volo/Abp/AspNetCore/Mvc/DataAnnotations/AbpValidationAttributeAdapterProvider.cs
+++ b/framework/src/Volo.Abp.AspNetCore.Mvc/Volo/Abp/AspNetCore/Mvc/DataAnnotations/AbpValidationAttributeAdapterProvider.cs
@@ -1,6 +1,7 @@
 ﻿using System.ComponentModel.DataAnnotations;
 using Microsoft.AspNetCore.Mvc.DataAnnotations;
 using Microsoft.Extensions.Localization;
+using Volo.Abp.AspNetCore.Mvc.Validation;
 using Volo.Abp.Validation;
 
 namespace Volo.Abp.AspNetCore.Mvc.DataAnnotations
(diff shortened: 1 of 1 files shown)


maliming:
Hi 

Can you share some code to reproduce the problem?

JadynWong:
Hi @maliming ,

The non-null value types properties without `RequiredAttribute` will be appended to the `RequiredAttribute` for client validator in the method of the `DataAnnotationsClientModelValidatorProvider`.

Abp uses `AbpModelMetadataProvider` to handle the localization of `DataAnnotations`, but the above automatically added attribute does not go through this method.

Use abp-sample project bookstore.
- Add a property `Age` on `/Author/CreateModal/CreateAuthorViewModel`.
- Add a `abp-input` for `Age` on `/Author/CreateModal.cshtml`.  

image
image

`AbpModelMetadataProvider`
https://github.com/abpframework/abp/blob/60dcef4c6eeecb2602cd43501da0a878e816ec8b/framework/src/Volo.Abp.AspNetCore.Mvc/Volo/Abp/AspNetCore/Mvc/ModelBinding/Metadata/AbpModelMetadataProvider.cs#L13-L54

maliming:
hi

I think the problem will be resolved if you add `[Required]` to `Age` manually. right?

**AbpValidationAttributeAdapterProvider** should only handle related attributes of ABP.

JadynWong:
> hi
> 
> I think the problem will be resolved if you add `[Required]` to `Age` manually. right?
> 
> **AbpValidationAttributeAdapterProvider** should only handle related attributes of ABP.

Yes.

maliming:
OK 
You can add a custom `IValidationAttributeAdapterProvider` or `[Required]`.

PR #9164, opened 2021-05-26
Fix: interception fails on synchronous calls

### LINKED ISSUE #9165: Synchronous interceptor fails if an invocation is not proceeded
If an Abp interceptor intercepts a synchronous call and does not call `invocation.ProceedAsync()` (e.g. if used for caching and returning a value from a cache) the call fails with a `System.InvalidOperationException: Interceptors failed to set a return value, or swallowed the exception thrown by the target`.

pecanw:
Fixes: #9165

### DIFF
--- a/framework/src/Volo.Abp.Castle.Core/Volo.Abp.Castle.Core.csproj
+++ b/framework/src/Volo.Abp.Castle.Core/Volo.Abp.Castle.Core.csproj
@@ -16,7 +16,7 @@
 
   <ItemGroup>
     <PackageReference Include="Castle.Core" Version="4.4.1" />
-    <PackageReference Include="Castle.Core.AsyncInterceptor" Version="2.0.21-alpha" />
+    <PackageReference Include="Castle.Core.AsyncInterceptor" Version="2.0.0" />
   </ItemGroup>
 
   <ItemGroup>
--- a/framework/src/Volo.Abp.Castle.Core/Volo/Abp/Castle/DynamicProxy/CastleAbpMethodInvocationAdapterBase.cs
+++ b/framework/src/Volo.Abp.Castle.Core/Volo/Abp/Castle/DynamicProxy/CastleAbpMethodInvocationAdapterBase.cs
@@ -20,7 +20,11 @@ namespace Volo.Abp.Castle.DynamicProxy
 
         public MethodInfo Method => Invocation.MethodInvocationTarget ?? Invocation.Method;
 
-        public object ReturnValue { get; set; }
+        public object ReturnValue
+        {
+            get => Invocation.ReturnValue;
+            set => Invocation.ReturnValue = value;
+        }
 
         protected IInvocation Invocation { get; }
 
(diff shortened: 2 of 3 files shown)


pecanw:
Maybe, we can also omit the assignment `ReturnValue = await Proceed(Invocation, ProceedInfo);` in  `CastleAbpMethodInvocationAdapterWithReturnValue` and just call `await Proceed(Invocation, ProceedInfo);`.

Or even have just one `CastleAbpMethodInvocationAdapter` istead of the `CastleAbpMethodInvocationAdapterBase`, `CastleAbpMethodInvocationAdapter` and `CastleAbpMethodInvocationAdapterWithReturnValue`...

maliming:
hi

See https://github.com/abpframework/abp/issues/2026#issuecomment-568681191

pecanw:
Hi @maliming ,

I'm sorry, but I don't understand how this is related (apart of having _some_ problems with synchronous interceptors). If you can please add some more comments.

The issue which I'm trying to address happens only if a synchronous interceptor does not call the invocation and tries to return a value.

pecanw:
I'm working on the broken test...

pecanw:
I have changed the `CastleAbpMethodInvocationAdapterBase.ReturnValue` not to read / write directly from / to the 'CastleInterceptor', but use a backing field instead (as it was done originally), but the setter writes to the `CastleInterceptor` as well as to the backing field. All tests seem to be passing now.

maliming:
Hi

Like the url I shared above, the sync method is not supported in the interceptor.

pecanw:
Are you telling me that the interception should not be used in ABP for synchronous methods at all? The interception works fine with my fix. Sou why should it be not supported? I understand the intention to have all ABP internal services and repositories with async methods only, but users (like me) can have custom methods with synchronous signature and it is IMHO very awkward to make the method async only to enable the interception...

Even if the sync interceptor is "discouraged", is there any problem with accepting the pull request which makes it possible?

maliming:
image

pecanw:
Thanks for pointing out the main idea. I do understand the intention not to use the sync methods at all. and also the reason why 

Anyway I think that if a user writes (for any reason) a synchronous method it is better to have the interceptor working than receiving a quite confusing error message "_Interceptors failed to set a return value, or swallowed the exception thrown by the target_" even though the code __does__ set the return value.

Another option would be to throw a "synchronous interception not supported exception" any time an interceptor is applied on a sync method. However I don't like this option as it would probably break a lot of existing implementations.

maliming:
hi 

You can implement this feature in your project, but we can't make these changes to the framework.

PMExtra:
I think it is a defect of [`JSkimming/Castle.Core.AsyncInterceptor`](https://github.com/JSkimming/Castle.Core.AsyncInterceptor) which is referenced by ABP framework.

It should be fixed in the upstream library rather than ABP.

Take a look at this code part:
https://github.com/JSkimming/Castle.Core.AsyncInterceptor/blob/v2.1.0/src/Castle.Core.AsyncInterceptor/AsyncInterceptorBase.cs#L116-L130

I think it should append `invocation.ReturnValue = task.Result;` at the end of the method.

According to that, there are `invocation.ReturnValue = InterceptAsync(...)` for async methods.
https://github.com/JSkimming/Castle.Core.AsyncInterceptor/blob/v2.1.0/src/Castle.Core.AsyncInterceptor/AsyncInterceptorBase.cs#L49-L67

PR #9170, opened 2021-05-27
Refactor RequestLocalizationOptionsProvider.

### LINKED ISSUE #9158: RequestLocalizationOptions needs to be updated after the default language changed.
https://github.com/abpframework/abp/blob/dev/framework/src/Volo.Abp.AspNetCore/Microsoft/AspNetCore/RequestLocalization/DefaultAbpRequestLocalizationOptionsProvider.cs

https://support.abp.io/QA/Questions/1315/How-can-I-set-default-language-to-FI#answer-12d2e6a8-1379-4f37-72d2-39fcb8910f0b

maliming:
Resolve #9158

Because the method parameters are changed. Although there may not be too many developers using it,
**it is a minor breaking change.**

The method uses a cache internally.
```
await languageProvider.GetLanguagesAsync();
await settingProvider.GetOrNullAsync(LocalizationSettingNames.DefaultLanguage);
```

### DIFF
--- a/framework/src/Volo.Abp.AspNetCore/Microsoft/AspNetCore/Builder/AbpApplicationBuilderExtensions.cs
+++ b/framework/src/Volo.Abp.AspNetCore/Microsoft/AspNetCore/Builder/AbpApplicationBuilderExtensions.cs
@@ -58,13 +58,8 @@ namespace Microsoft.AspNetCore.Builder
                 .UseMiddleware<AbpCorrelationIdMiddleware>();
         }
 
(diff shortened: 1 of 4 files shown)


hikalkan:
Creating the `RequestLocalizationOptions` in every request may slow down the requests, even if languageProvider and settingProvider internally uses caches (they are using distributed caches which is slower than a local, singleton field)

I suppose you are doing it to get the correct default language when it changes. I suggest to somehow invalidate the cached object, instead of re-calculating it in each request (I am not sure if that's possible).

maliming:
https://github.com/abpframework/abp/pull/20886

PR #9180, opened 2021-05-28
use IRemoteContentStream without form content headers

alexandru-bagu:
1. allow the usage of IRemoteContentStream or RemoteContentStream when the incoming stream is not a form content format
2. remove all rewinds related to streams; the user is the one who must ensure his streams are at the correct position when calling the api; stream rewind would exclude the ability of sending partial streams (at least skippable streams, because the tail cannot be limitted in .net core)
3. RemoteStreamContent now must get the content type associated with the stream in its constructor; it may also receive the length of the stream in a read-only form (this is because some stream classes cannot provide the length and it is provided via http headers)
4. IRemoteStreamContent should implement IDisposable to allow the auto clean up of streams (example: FileStream)


maliming:
hi 

Action failed.

maliming:
> allow the usage of IRemoteContentStream or RemoteContentStream when the incoming stream is not a form content format.

The `FormData` is the most common way in JavaScript.  
If we want to support it, we also need to change the **GenerateBody** method.
https://github.com/abpframework/abp/blob/821c08985dfe6dc143035498e098fb643a044eb6/framework/src/Volo.Abp.Http.Client/Volo/Abp/Http/Client/DynamicProxying/RequestPayloadBuilder.cs#L71

alexandru-bagu:
> > allow the usage of IRemoteContentStream or RemoteContentStream when the incoming stream is not a form content format.
> 
> The `FormData` is the most common way in JavaScript.
> If we want to support it, we also need to change the **GenerateBody** method.
> https://github.com/abpframework/abp/blob/821c08985dfe6dc143035498e098fb643a044eb6/framework/src/Volo.Abp.Http.Client/Volo/Abp/Http/Client/DynamicProxying/RequestPayloadBuilder.cs#L71

I am not sure why **GenerateBody** would have to be changed. At the moment if you receive a request with a content-type different than some type of form then it would be processed as a IRemoteContentStream or RemoteContentStream. If in turn this would have to be then proxied to the application layer, the **GenerateBody** you are referring to would then encode it within a FormFile which would be fine as the application layer would then read the proper content-type from inside the form file content headers.

From a JS point of view I guess there's no way to easily send a file or a stream. But this is mostly a change for backend api calls not necessarily for frontend. Is this what you are talking about or something else?

maliming (approved):

maliming (approved):

maliming (approved):

PR #9194, opened 2021-05-31
Remove EntityFrameworkCore.DbMigrations project from the solution & Compatible with CLI.

### LINKED ISSUE #8776: Consider to remove EntityFrameworkCore.DbMigrations project from the solution
Instead of a separate migration dbcontext, we can use application's dbcontext. This simplifies database mappings and migrations, especially when you want to setup relations to entities coming from modules. I see some developers found the current structure complicated.

One disadvantage is that we should remove the AppUser in the solution and use IdentityUser entity (coming from the Identity module) when we need users in our code. So, we can't add new type-safe properties to this entity. However, module extensibility system already provides an easy way to add extra properties to existing entities.

Making that change is easy actually, but we should update all tutorials, documents, tooling, etc.

maliming:
Resolve #8776


realLiangshiwei (approved):

framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/ProjectBuilding/TemplateProjectBuilder.cs
```diff
@@ -111,6 +112,11 @@ public async Task<ProjectBuildResult> BuildAsync(ProjectBuildArgs args)
                 args
             );
 
+            if (context.Template is AppTemplateBase appTemplateBase)
+            {
+                appTemplateBase.HasDbMigrations = SemanticVersion.Parse(templateFile.Version) < new SemanticVersion(4, 4, 0);
```
yekalkan: @maliming This will be False for 4.4.0 preview versions. ( v4.4.0 > v4.4.0-preview.1)

maliming: I think this will be fine.

`appTemplateBase.HasDbMigrations = SemanticVersion.Parse(templateFile.Version) < new SemanticVersion(4, 3, 99);`

maliming: https://github.com/abpframework/abp/pull/9194/commits/e8d287183eef35523e4def20f064ec13d267a4a2

yekalkan (approved):

hikalkan (approved):

PR #9195, opened 2021-05-31
Add appsettings.secrets.json to template

realLiangshiwei:


ebicoglu:
I see the following files have still AbpLicenseCode. Do you need to fix them?


- https://github.com/volosoft/volo/blob/dev/abp/templates/ms/shared/MyCompanyName.MyProjectName.DbMigrator/appsettings.json#L40
- https://github.com/volosoft/volo/blob/dev/abp/templates/ms/services/template/src/MyCompanyName.MyProjectName.MicroserviceName.HttpApi.Host/appsettings.json#L42

ebicoglu (approved):

PR #9231, opened 2021-06-03
Skip authenticate when action allow anonymous.

### LINKED ISSUE #9205: Dynamic client proxy at Blazor webassembly always call authorize but the target api doesn't require
* Your **ABP Framework version**: 4.3.2
* Your **User Interface** type: Blazor webassembly
* Your database provider: MongoDB

I have create app service is ILeadChannelAppService  and inject it into Blazor webassembly. If I don't login and call to the api through HttpApi.Client module - the client proxy always calls to the identity server for authorizing but It will be failed and take time before call my target api:

image

Here is the code at razor file:
@inject ILeadChannelAppService _leadChannelAppService
leadChannel = await _leadChannelAppService.GetAsync(new Guid("01dccee1-2dbb-92dc-f997-39fcbe188e9e"));

But if user login success, the client proxy don't call authorize any more and call the target api immediately.

I don't know this is a bug or not. 

Thanks.

maliming:
Resolve #9205

### DIFF
--- a/framework/src/Volo.Abp.AspNetCore.Mvc/Volo/Abp/AspNetCore/Mvc/AspNetCoreApiDescriptionModelProvider.cs
+++ b/framework/src/Volo.Abp.AspNetCore.Mvc/Volo/Abp/AspNetCore/Mvc/AspNetCoreApiDescriptionModelProvider.cs
@@ -4,6 +4,7 @@ using System.Linq;
 using System.Reflection;
 using System.Text;
 using JetBrains.Annotations;
+using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.Abstractions;
 using Microsoft.AspNetCore.Mvc.ApiExplorer;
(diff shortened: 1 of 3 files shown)


hikalkan:
Hi @maliming 

That's not a true implementation. There are two main problems with this change;

1. If none of `Authorize` or `AllowAnonymous` was used, the method is not authorized, so no need to token in that case too.
2. But the actual problem is bigger: What if developer uses `IAuthorizationService` inside code? If current client can be authenticated (has token), we should always authenticate it.

I didn't investigated the #9205 deeply, but it seems we should bypass the authentication inside `ClientAuthenticator` (maybe we simply check if user was authenticated / logged in).

maliming:
This is for the HttpApi controllers, They don't use **Authorize** or **AllowAnonymous**. 

**If the developer can confirm** that the application services don't require authorization, they can use `AllowAnonymous`, and we will skip the authentication.

If not specified, we keep the default behavior. `if (action.AllowAnonymous != true)` do` Authenticate ` 

https://github.com/abpframework/abp/blob/dev/modules/identity/src/Volo.Abp.Identity.HttpApi/Volo/Abp/Identity/IdentityUserController.cs#L9-L90

hikalkan (approved):

TarikElHajjam:
Hey! I'm kind of facing the same problem, I'm unable to use an authorized application service from within a method that lives on a controller in HttpApi which has [AllowAnonymous] Attribute, any soulution?

TarikElHajjam:
Hey! I'm kind of facing the same problem, I'm unable to use an authorized application service method from within a method that lives on a controller in HttpApi which has [AllowAnonymous] Attribute, any solution?

PR #9245, opened 2021-06-04
feat(audit-logging): add userId filter

PMExtra:
Filter audit logs by users is a very typical scene.
But the `IAuditLogRepository` is only support username now.
In the EFCore storage, there is an index with `{ TenantId, UserId, ExecutionTime }` but no index with `UserName`.
So I make this PR.


maliming (approved):

maliming:
Thanks @PMExtra

PR #9350, opened 2021-06-16
Extracted RemoteServiceConfigurationProvider.

hikalkan:
This will allow the developer to dynamically determine the remove service configurations (like URL of a remote service).


maliming (approved):

maliming:
It is best if we can make `IRemoteServiceConfigurationProvider` async. But `IServerUrlProvider` is sync now. We can consider refactor `IServerUrlProvider`.

https://github.com/abpframework/abp/blob/dev/framework/src/Volo.Abp.AspNetCore.Components.Web/Volo/Abp/AspNetCore/Components/Web/IServerUrlProvider.cs
https://github.com/abpframework/abp/blob/dev/framework/src/Volo.Abp.AspNetCore.Components.WebAssembly/Volo/Abp/AspNetCore/Components/WebAssembly/WebAssemblyServerUrlProvider.cs#L20

enisn (approved):

hikalkan:
Made it async with https://github.com/abpframework/abp/pull/9406

slarkerino:
@hikalkan Hi, is it possible to use this with consul to dynamically discovery service url

PR #9354, opened 2021-06-17
CmsKit - Add caching to MainMenu

enisn:


modules/cms-kit/src/Volo.CmsKit.Common.Application/Volo/CmsKit/Menus/MainMenuCacheKey.cs
```diff
@@ -0,0 +1,10 @@
+namespace Volo.CmsKit.Public.Menus
+{
+    public class MainMenuCacheKey
```
hikalkan: No need to a composite cache key class, just use string as the key.

modules/cms-kit/src/Volo.CmsKit.Common.Application/Volo/CmsKit/Menus/MenuUpdatedHandler.cs
```diff
@@ -0,0 +1,26 @@
...
+    {
+        public IDistributedEventBus EventBus { get; }
+
+        public MenuUpdatedHandler(IDistributedEventBus eventBus)
+        {
+            EventBus = eventBus;
+        }
+
+        public async Task HandleEventAsync(EntityUpdatedEventData<Menu> eventData)
+        {
```
hikalkan: I didn't understand why we are publishing a distributed event, but not directly using IDistributedCache to remove the cache item, like `await DistributedCache.RemoveAsync(...)`

BTW, if you one day need to publish a distributed event when an entity changes, use [predefined events](https://docs.abp.io/en/abp/latest/Distributed-Event-Bus#pre-defined-events). ABP will automatically publish the event for you.

modules/cms-kit/src/Volo.CmsKit.Public.Application/Volo/CmsKit/Public/Menus/MenuPublicAppService.cs
```diff
@@ -1,30 +1,50 @@
...
         public async Task<MenuWithDetailsDto> GetMainMenuAsync()
         {
-            var menu = await MenuRepository.FindMainMenuAsync(includeDetails: true);
+            var cachedMenu = await DistributedCache.GetOrAddAsync(
+                new MainMenuCacheKey(),
+                async () =>
+                {
+                    var menu = await MenuRepository.FindMainMenuAsync(includeDetails: true);
+                    
+                    if (menu == null)
```
hikalkan, line 34: If there is no menu in the database, it will always call this factory method again and again, because `DistributedCache.GetOrAddAsync` [calls the factory](https://github.com/abpframework/abp/blob/cfc648bc656da21b73d99578d71d0ac71b56b9c4/framework/src/Volo.Abp.Caching/Volo/Abp/Caching/DistributedCache.cs#L503-L509) if the item is null in the cache.
However, if it doesn't exists in the database, no point to call it again and again (it should cache that info too)

modules/cms-kit/src/Volo.CmsKit.Public.Application/Volo/CmsKit/Public/Menus/MenuUpdatedHandler.cs
```diff
@@ -0,0 +1,23 @@
+using System.Threading.Tasks;
+using Volo.Abp.Caching;
+using Volo.Abp.DependencyInjection;
+using Volo.Abp.EventBus.Distributed;
+using Volo.CmsKit.Menus;
+
+namespace Volo.CmsKit.Public.Menus
+{
+    public class MenuUpdatedHandler : IDistributedEventHandler<MenuUpdatedEto>, ITransientDependency
```
hikalkan: Distributed events are completely out of topic in this issue.

hikalkan (changes requested):

modules/cms-kit/src/Volo.CmsKit.Common.Application/Volo/CmsKit/Menus/MenuUpdatedHandler.cs
```diff
@@ -0,0 +1,23 @@
+using System.Threading.Tasks;
+using Volo.Abp.Caching;
+using Volo.Abp.DependencyInjection;
+using Volo.Abp.Domain.Entities.Events;
+using Volo.Abp.EventBus;
+
+namespace Volo.CmsKit.Menus
+{
+    public class MenuUpdatedHandler :  ILocalEventHandler<EntityUpdatedEventData<Menu>>, ITransientDependency
```
hikalkan: We should also handle `EntityDeletedEventData<Menu>` to remove the deleted object from the cache.

hikalkan (changes requested):

hikalkan (approved):

PR #9358, opened 2021-06-17
Use Type instead of TypeAsString to compare.

### LINKED ISSUE #9357: Generic Type in Api Dto will cause coupling between microservices.
* Your **ABP Framework version**.
From v3.1.2 (used in my team) ~ latest (checked latest version code)
Not related.
Not related.
Not related.
Not a bug.

## Problem

I have a microservice A to call api on microservice B through `Volo.Abp.Http.Client`. Sometimes will receive exception as "Could not found remote action for method: ...", but the definition for that api has never changed for almost 60 days.

## Cause
As we found codes in this line : https://github.com/abpframework/abp/blob/dev/framework/src/Volo.Abp.Http.Client/Volo/Abp/Http/Client/DynamicProxying/ApiDescriptionFinder.cs#L143

```
protected virtual bool TypeMatches(MethodParameterApiDescriptionModel actionParameter, ParameterInfo methodParameter)        {            
    return NormalizeTypeName(actionParameter.TypeAsString) ==NormalizeTypeName(methodParameter.ParameterType.GetFullNameWithAssemblyName());        
}
protected virtual string NormalizeTypeName(string typeName) 
{
const string placeholder = "%COREFX%"; 
const string netCoreLib = "System.Private.CoreLib";
const string netFxLib = "mscorlib";
            return typeName.Replace(netCoreLib, placeholder).Replace(netFxLib, placeholder);        
}
```
the  `actionParameter.TypeAsString` in `/api/abp/api-definition` ,when the type is generic type, it is :
```
...
 }, {
    "name": "granted",
    "typeAsString": "System.Collections.Generic.List`1[[DangQu.PT.PermissionManage.RoleBasedAppContentAcls.Dto.AppContentGrantedWithRolesDto, DangQu.PT.PermissionManage.Application.Contracts, Version=1.5.92.0, Culture=neutral, PublicKeyToken=null]], System.Private.CoreLib",
    "type": "System.Collections.Generic.List<DangQu.PT.PermissionManage.RoleBasedAppContentAcls.Dto.AppContentGrantedWithRolesDto>",
    "typeSimple": "[DangQu.PT.PermissionManage.RoleBasedAppContentAcls.Dto.AppContentGrantedWithRolesDto]",
    "isOptional": false,
    "defaultValue": null
}],
"parameters": [{                        
...
```
I think `Version=1.5.92.0`  make the  microservice A and B coupled.

When A call one api on B (here is `DangQu.PT.PermissionManage.Application.Contracts`), B upgrade for something not related to A, A MUST upgrade the version of `DangQu.PT.PermissionManage.Application.Contracts` nupkg for the type not matched as version not same.

## Solution

So, here I suggest :

change the default implemention for the `protected virtual bool TypeMatches(...` as below:
```
        protected override bool TypeMatches(MethodParameterApiDescriptionModel actionParameter, ParameterInfo methodParameter)
        {
            return actionParameter.Type == TypeHelper.GetFullNameHandlingNullableAndGenerics(methodParameter.ParameterType);
        }
```

This not so restrict but can be lower-coupled.

personball:
Close #9357 

Compatible for different versions and assemblies.

Type: 
**System.Collections.Generic.IEnumerable<System.Guid>**

TypeAsString: 
**System.Collections.Generic.IEnumerable`1[[System.Guid, System.Private.CoreLib, Version=5.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]], System.Private.CoreLib**

### DIFF
--- a/framework/src/Volo.Abp.Http.Client/Volo/Abp/Http/Client/DynamicProxying/ApiDescriptionFinder.cs
+++ b/framework/src/Volo.Abp.Http.Client/Volo/Abp/Http/Client/DynamicProxying/ApiDescriptionFinder.cs
@@ -10,6 +10,7 @@ using Microsoft.Extensions.Options;
 using Volo.Abp.DependencyInjection;
 using Volo.Abp.Http.Modeling;
 using Volo.Abp.MultiTenancy;
+using Volo.Abp.Reflection;
 using Volo.Abp.Threading;
 using Volo.Abp.Tracing;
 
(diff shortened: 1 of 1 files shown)


maliming (approved):

hikalkan (approved):

hikalkan:
Thanks!

PR #9382, opened 2021-06-21
Project does not build initially

Znow:
Could the following changes be a possible fix? It makes the project build without errors.


cotur:
Hi, thanks for your contribution.

I don't get any build error. How can I reproduce the problem?

enisn:
> Could the following changes be a possible fix? It makes the project build without errors.

I couldn't get a build error initially, Can you share steps to reproduce @Znow ?

cotur:
I've closed this PR because of the inactivity. Thanks.

PR #9393, opened 2021-06-22
Update Customizing-Application-Modules-Guide.md

sturlath:
I really think that this text needs a update. I read to much into the line "ABP CLI automates this process for you." and did not notice the the other packages that were in the projects and also just looked into the project that it added packages and attributes. It may be silly but I didn´t think about the other projects and that I had to remove every nuget package...

See my support ticket https://support.abp.io/QA/Questions/1502/Error-running-up-Separate-IdentityServer-after-adding-Account-module-source-code


realLiangshiwei:
Thanks a lot @sturlath

realLiangshiwei (approved):

PR #9426, opened 2021-06-24
Add FileName for RemoteStreamContent

JadynWong:
Add FileName for RemoteStreamContent
- Download file with file name.
- Upload file with file name.


JadynWong:
`RemoteStreamContent` as a stream content.  I think the `ContentType` property of `RemoteStreamContent` should use `application/octet-stream` as the default value.

[applicationoctet-stream](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types#applicationoctet-stream)

What do you think?

maliming (approved):

maliming:
Thanks @JadynWong  👍

PR #9433, opened 2021-06-25
Allowing the developer to handle ajax error.

### LINKED ISSUE #9432: abp.ajax fails silently when jqXHR.status === 0
If I understand correctly, [`abp.ajax` fails silently and leaves a hanging promise when `jqXHR.status === 0`](https://github.com/abpframework/abp/blob/1efc962aa1e15da6938ff2a3b6110ce869a0ed90/npm/packs/jquery/src/abp.jquery.js#L109). According to the comment in the abovementioned code, it is expected that the request is aborted in such case. However,
- `jqXHR.status === 0` can occur in other situations, such as the host is down, or CORS preflight has been rejected. The caller is entitled to know this;
- It's generally not a good idea to make a promise hanging.

Is it the code was supposed to be like this (which is still not good enough IMO as it still hangs the promise in certain scenarios)?
```javascript
if(jqXHR.status === 0 && jqXHR.statusText === 'abort')
//                    ↑ && instead of ||
```

realLiangshiwei:
resolve https://github.com/abpframework/abp/issues/9432

### DIFF
--- a/npm/packs/jquery/src/abp.jquery.js
+++ b/npm/packs/jquery/src/abp.jquery.js
@@ -106,7 +106,7 @@ var abp = abp || {};
                     $dfd.resolve(data);
                     userOptions.success && userOptions.success(data);
                 }).fail(function (jqXHR) {
-                    if(jqXHR.status === 0 || jqXHR.statusText === 'abort') {
+                    if(jqXHR.statusText === 'abort') {
                         //ajax request is abort, ignore error handle.
                         return;
                     }


hillin:
I'd argue that it's still better to leave the choice to the caller, even if the request is aborted. Simply returning for an aborted request will have the promise hanging, never to be fulfilled (nor rejected). An aborted request can be triggered not only by page disposal events like refreshing the page (in which it does not matter if the promise dangles), but also by manually calling the `abort` method on the `jqXHR` object, which actually works like cancelling a `CancellationToken`. Allowing the caller to handle it, like handling an `OperationCancelledException`, will just make their life easier.

npm/packs/jquery/src/abp.jquery.js
```diff
@@ -106,7 +106,7 @@ var abp = abp || {};
                     $dfd.resolve(data);
                     userOptions.success && userOptions.success(data);
                 }).fail(function (jqXHR) {
-                    if(jqXHR.status === 0 || jqXHR.statusText === 'abort') {
+                    if(jqXHR.statusText === 'abort') {
```
maliming, line 109: We can consider adding a `error callback` to allow custom the error process

maliming, line 109: We can consider adding an `error callback` to allow developers to handle it themselves.

eg:
`userOptions.success && userOptions.success(data);`

maliming (approved):

PR #9460, opened 2021-06-30
fix: Failed to compare different versions of the same type

netcore-jroger:
Failed to compare different versions of the same type.
Like the following situation:
```csharp
System.Collections.Generic.IEnumerable`1[[System.Guid, %COREFX%, Version=5.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]], %COREFX%
System.Collections.Generic.IEnumerable`1[[System.Guid, %COREFX%, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]], %COREFX%
```


maliming:
See https://github.com/abpframework/abp/pull/9358

PR #9485, opened 2021-07-02
Update Authorization.md

sturlath:
After back and fourth between me and @realLiangshiwei for few days I got the help to solve my problem https://github.com/abpframework/abp/discussions/9428 so here is giving some back.

I can add more code if you want or place this somewhere.. just let me know...

### DIFF
--- a/docs/en/Authorization.md
+++ b/docs/en/Authorization.md
@@ -124,6 +124,16 @@ The localized UI will be as seen below:
 
 ![authorization-new-permission-ui-localized](images/authorization-new-permission-ui-localized.png)
 
+#### Adding permissions to a user on registration when using a separate Identity Server
+On registration of a user, the user can be assigned permissions by utilizing the [local event bus](Local-Event-Bus.md) using ```SetForUserAsync```.
+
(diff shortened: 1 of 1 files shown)


realLiangshiwei:
Thanks, but I don't think this is a necessary change.

Because you using separate architecture, try using distributed events may be a better way.

sturlath:
Basically just swap `ILocalEventHandler` for `IDistributedEventHandler` in the `EntityCreatedEventDataHandler`
Just like this? No need to use [Eto´s](https://docs.abp.io/en/abp/4.2/Distributed-Event-Bus) (I didn't find any that matched Identity like there is one for `TenantCreatedEto`)?

This code below worked at least locally.

```csharp
public class NewUserCreatedHandler : IDistributedEventHandler<EntityCreatedEventData<IdentityUser>>, ITransientDependency
{
...
}
```

So @realLiangshiwei, you don´t want to have this document update anywhere else?

realLiangshiwei:
You can use Rabbitmq or other provider.

sturlath:
So you are telling me that its not possible with `IDistributedEventHandler` (even though I got it to work). 

Why?

Will it only work locally or?

PR #9496, opened 2021-07-05
Fix DynamicOptions of `LdapManager`

JadynWong:
`AbpLdapOptions` use `AbpDynamicOptionsManager` to read Setting settings.

https://github.com/abpframework/abp/blob/240f72524f945641fb6de318f893fcf0bb52af54/framework/src/Volo.Abp.Ldap/Volo/Abp/Ldap/AbpAbpLdapOptionsManager.cs#L9-L17

### DIFF
--- a/framework/src/Volo.Abp.Ldap/Volo/Abp/Ldap/LdapManager.cs
+++ b/framework/src/Volo.Abp.Ldap/Volo/Abp/Ldap/LdapManager.cs
@@ -12,11 +12,11 @@ namespace Volo.Abp.Ldap
     public class LdapManager : ILdapManager, ITransientDependency
     {
         public ILogger<LdapManager> Logger { get; set; }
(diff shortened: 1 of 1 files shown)


maliming:
hi

I remember that LDAP-related components have all supported async, and `AbpAbpLdapOptionsManager` is no longer needed.

https://github.com/abpframework/abp/pull/6368

JadynWong:
Hi @maliming, 

As I understand it, the `Options` that inherit from `AbpDynamicOptionsManager` must call `SetAsync` method. 
Now it won't take the settings from `SettingProvider`.
If `AbpAbpLdapOptionsManager` is no longer needed, would you consider removing this class?

Thanks.

maliming:
I will handle it.

maliming (approved):

PR #9510, opened 2021-07-06
Switch to the tenant if __tenant querystring parameter is sent.

### LINKED ISSUE #9383: Switch to the tenant if __tenant querystring parameter is sent
In the app.UseMultiTenancy() middleware, set tenant cookie if __tenant parameter is sent in the querystring.

maliming:
Resolve #9383

### DIFF
--- a/modules/account/src/Volo.Abp.Account.Application/Volo/Abp/Account/AccountAppService.cs
+++ b/modules/account/src/Volo.Abp.Account.Application/Volo/Abp/Account/AccountAppService.cs
@@ -62,19 +62,16 @@ namespace Volo.Abp.Account
 
         public virtual async Task ResetPasswordAsync(ResetPasswordDto input)
         {
-            using (CurrentTenant.Change(input.TenantId))
+            await IdentityOptions.SetAsync();
+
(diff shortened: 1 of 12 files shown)


modules/account/src/Volo.Abp.Account.Application/Volo/Abp/Account/AccountAppService.cs
```diff
@@ -62,19 +62,16 @@ public virtual async Task SendPasswordResetCodeAsync(SendPasswordResetCodeDto in
 
         public virtual async Task ResetPasswordAsync(ResetPasswordDto input)
         {
-            using (CurrentTenant.Change(input.TenantId))
```
hikalkan, line 65: I think we still continue to use TenantId here, otherwise there is no way to switch to the host side.

maliming: https://github.com/abpframework/abp/pull/9510/commits/81f1f7cb260a84576c9a439832eb95d17bcd0d35

hikalkan (approved):

hikalkan (approved):

PR #9523, opened 2021-07-07
Identity: Add role and organization filters to GetListAsync

EngincanV:


modules/identity/src/Volo.Abp.Identity.EntityFrameworkCore/Volo/Abp/Identity/EntityFrameworkCore/EfCoreIdentityUserRepository.cs
```diff
@@ -136,8 +136,15 @@ select ouRole.RoleId
...
             bool includeDetails = false,
+            Guid? roleId = null,
+            Guid? organizationUnitId = null,
             CancellationToken cancellationToken = default)
         {
+            if(roleId.HasValue || organizationUnitId.HasValue)
+            {
+                includeDetails = true;
+            }
+
```
maliming: Maybe this is not necessary.

EngincanV: It's necessary if roleId or organizationUnitId is specified. Because otherwise, the sub-collections would not be included (if includeDetails is false).

maliming: The `includeDetails` control` sub-collections` load or not, I don't know why it relates to `roleId or organizationUnitId`?

EngincanV: Because if `roleId` specified, I need to reach the [Roles collection](https://github.com/abpframework/abp/blob/dev/modules/identity/src/Volo.Abp.Identity.Domain/Volo/Abp/Identity/IdentityUser.cs#L115) for filtering. (Same is valid for `organizationUnitId`)

maliming (approved):

PR #9545, opened 2021-07-09
Replace the `culture` and `ui-culture` in the `url` while switching language.

maliming:
The problem: **if the querystring contains a specific culture, we can't switch to another language, because the querystring has the highest priority.**

image

image


hikalkan (approved):

hikalkan:
Good 👍 thanks.

PR #9552, opened 2021-07-09
Add the class name as property localization prefix.

### LINKED ISSUE #9551: DisplayName in Localization
**How do I distinguish between the same properties in a localized configuration？**

I have 2 model as follow:

```csharp
class Foo{
}

class Bar{
   public string Name{get; set;}
}
```

And i only set one message in localization file as bellow:

```json
{
    "DisplayName:Name": "name"
}
```

But they are not the same. I want to set Foo.Name as 'FooName' and 'Bar.Name' as 'BarName' .

How should i do?

maliming:
Resolve #9551

The priority is:
 * `DisplayName:ClassName:PropertyName`
 * `DisplayName:PropertyName`
 * `ClassName:PropertyName`
 * `PropertyName`

### DIFF
--- a/framework/src/Volo.Abp.AspNetCore.Mvc/Volo/Abp/AspNetCore/Mvc/AbpDataAnnotationAutoLocalizationMetadataDetailsProvider.cs
+++ b/framework/src/Volo.Abp.AspNetCore.Mvc/Volo/Abp/AspNetCore/Mvc/AbpDataAnnotationAutoLocalizationMetadataDetailsProvider.cs
@@ -54,7 +54,29 @@ namespace Volo.Abp.AspNetCore.Mvc
 
             displayMetadata.DisplayName = () =>
             {
-                var localizedString = localizer[PropertyLocalizationKeyPrefix + context.Key.Name];
+                /*
+                 * DisplayName:ClassName:PropertyName
+                 * DisplayName:PropertyName
+                 * ClassName:PropertyName
+                 * PropertyName
+                 */
+
+                LocalizedString localizedString = null;
+
+                if (context.Key.ContainerType != null)
+                {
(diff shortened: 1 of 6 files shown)


framework/src/Volo.Abp.AspNetCore.Mvc/Volo/Abp/AspNetCore/Mvc/AbpDataAnnotationAutoLocalizationMetadataDetailsProvider.cs
```diff
@@ -54,7 +54,29 @@ public void CreateDisplayMetadata(DisplayMetadataProviderContext context)
...
+                 * DisplayName:PropertyName
+                 * ClassName:PropertyName
+                 * PropertyName
+                 */
+
+                LocalizedString localizedString = null;
+
+                if (context.Key.ContainerType != null)
+                {
+                    localizedString = localizer[PropertyLocalizationKeyPrefix + context.Key.ContainerType.Name + ":" + context.Key.Name];
```
hikalkan, line 68: I am merging this assuming `context.Key` or `context.Key.ContainerType` can not be null, right?

hikalkan (approved):

framework/src/Volo.Abp.AspNetCore.Mvc/Volo/Abp/AspNetCore/Mvc/AbpDataAnnotationAutoLocalizationMetadataDetailsProvider.cs
maliming: They are never `null` by default.

Issue #8259, opened 2021-03-30
Article request: Using Syncfusion Razor components


EdmundHerbert:
I use Syncfusion controls extensively and they offer a community edition. 
Could someone add example of how to implement example Syncfusion Grid to ABP Entity Framework Razor Page.

hikalkan:
Done with https://community.abp.io/articles/using-syncfusion-components-with-the-abp-framework-5ccvi8kc

fileman:
@hikalkan to use the Syncfusion SfDataManager control, a little change is required, but I don't know if this break something in ABP:
`context.Services.AddHttpClient("abp", sp =>
            {
                sp.BaseAddress = new Uri("/");
            });`
Instead of:
`context.Services.AddTransient(sp => new HttpClient
            {
                BaseAddress = new Uri("/")
            });`
I can submit a PR if it can be changed without other issues

PR #9553, opened 2021-07-09
HttpClient services using AddHttpClient

fileman:
When I tried to use the Syncfusion SfDataManager control, I have found that a little change is required to solve URI exception, but I don't know if this break something else in ABP Framework.

If is ok, I will update even the tiered template.

https://github.com/abpframework/abp/issues/8259


maliming:
hi @fileman 

Can you complete the PR?

templates/app/aspnet-core/src/MyCompanyName.MyProjectName.Blazor.Server/MyProjectNameBlazorModule.cs
```diff
@@ -206,9 +206,9 @@ private void ConfigureSwaggerServices(IServiceCollection services)
 
         private static void ConfigureHttpClient(ServiceConfigurationContext context)
         {
-            context.Services.AddTransient(sp => new HttpClient
+            context.Services.AddHttpClient("abp", sp =>
```
maliming, line 210: You can consider using constant string.

fileman: Sure, you mean something defined in MyProjectNameConsts like DbTablePrefix or inside Blazor/Blazor.Server, since is required only in this UI version?

maliming: Can you share the document that explains the `AddHttpClient`?

fileman: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/http-requests?view=aspnetcore-5.0#httpclient-and-lifetime-management

maliming: > When I tried to use the Syncfusion SfDataManager control, I have found that a little change is required to solve URI exception, 

Hi 

Does Syncfusion have a document explain it?

Can you share the exception details?

fileman: I don't have any Syncfusion document for this problem, I found the solution myself by following Microsoft's recommendations for httpclient.

I attached sample to reproduce the following exception "System.UriFormatException: 'Invalid URI: The format of the URI could not be determined.'"
[SfDataManager.Server.zip](https://github.com/abpframework/abp/files/7228103/SfDataManager.Server.zip)

maliming:
hi @hikalkan 

Can you explain why **HttpClient** is added?

https://github.com/abpframework/abp/blob/dev/templates/app/aspnet-core/src/MyCompanyName.MyProjectName.Blazor.Server/MyProjectNameBlazorModule.cs#L210-L213

totpero:
Hi,
I try to inject "HttpClient" in my page like this:
`[Inject] protected HttpClient Http { get; set; }`
or in blazor view:
`@inject HttpClient Http`
with the old configuration `context.Services.AddTransient` and with the new `context.Services.AddHttpClient("abp", sp => `.
When I try to call WebApi Endpoint, using : `var result =  await Http.GetStringAsync("/api/....");` the BaseAddress remains null.
After a few searches I discovered that in `ServiceConfigurationContext.Services` are more then one service of type HttpClient.

My scope was different so I hard-coded the `BaseAddress` I have discovered that cookies were not set even if I was logged in the page.
Is there any method to automatically use  `HttpClient` request authorized without calling "/api/account/login" endpoint?

maliming:
hi

I will remove it from the Balzor Server. 
https://docs.microsoft.com/en-us/aspnet/core/blazor/call-web-api?view=aspnetcore-5.0&pivots=server#add-the-httpclient-service-1
https://docs.microsoft.com/en-us/aspnet/core/blazor/call-web-api?view=aspnetcore-5.0&pivots=webassembly#add-the-httpclient-service-1

maliming:
https://github.com/abpframework/abp/pull/10405

fileman:
Yes it's not required for "server" so it can be removed.
This could be usefull for someone else as a community post, I'll try to write and submit.

totpero:
Hi,
HttpClient is already injected into services from other modules. Even if you remove configurehttpclient method, the HttpClient still exists into services with BaseAdress set to null.

PR #9560, opened 2021-07-11
Document: sample project - Add-Migration

minhhungit:
When we run `Add-Migration "Created_Book_Entity"`, it will throw an error like this, update document to guide developer install necessary package

>Your startup project 'Acme.BookStore.EntityFrameworkCore' doesn't reference Microsoft.EntityFrameworkCore.Design. This package is required for the Entity Framework Core >Tools to work. Ensure your startup project is correct, install the package, and try again.

### DIFF
--- a/docs/en/Tutorials/Part-1.md
+++ b/docs/en/Tutorials/Part-1.md
@@ -187,6 +187,12 @@ Select the `Acme.BookStore.EntityFrameworkCore.DbMigrations` as the **default pr
 Add-Migration "Created_Book_Entity"
 ```
 
+If you get error something like this on the console:
+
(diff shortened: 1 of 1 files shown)


maliming:
hi 

The document already explained this. Thanks for your contribution.

> The startup template uses [EF Core Code First Migrations](https://docs.microsoft.com/en-us/ef/core/managing-schemas/migrations/) to create and maintain the database schema. Open the **Package Manager Console (PMC)** under the menu *Tools > NuGet Package Manager*.

minhhungit:
Hi @maliming 

>The startup template uses EF Core Code First Migrations to create and maintain the database schema. Open the Package Manager Console (PMC) under the menu Tools > NuGet Package Manager.

This text just say that we will open **Nuget Manage Console** tool to run migration command, not to fix the problem after running the command

Issue #9585, opened 2021-07-13
Remove abppkg.json files from the package contents


hikalkan:
Create a new solution using 4.4.0-rc.2 then open with VS 2019:

![image](https://user-images.githubusercontent.com/1210527/125478478-bfdd40c7-8254-4703-977f-6562df0a15b3.png)

We should not see these files here.

PR #9592, opened 2021-07-14
Cli: Remove abp***.json files from the package contents

yekalkan:
resolves https://github.com/abpframework/abp/issues/9585


hikalkan:
We should not delete these files from the startup template.

PR #9690, opened 2021-07-29
Use `CancellationToken` if it exists in the parameter.

maliming:
When we call the API proxy in the Console or Blazor, we may manually provide the cancellation token.


hikalkan (approved):

hikalkan:
Great.

PR #9703, opened 2021-07-30
rabbitmq delay queue

tanyongzheng:
1. rabbitmq declare delay queue
2. rabbitmq background Jobs implement delayed job


realLiangshiwei:
Sorry, after my test, but it does not work.

realLiangshiwei:
Thanks for your idea, I will create a new PR to do it.

realLiangshiwei:
https://github.com/abpframework/abp/pull/9725

PR #9725, opened 2021-08-03
Implemented delayed jobs for RabbitMQ integration

### LINKED ISSUE #9724: Implement delayed jobs for RabbitMQ integration

realLiangshiwei:
Resolve https://github.com/abpframework/abp/issues/9724

### DIFF
--- a/framework/src/Volo.Abp.BackgroundJobs.RabbitMQ/Volo/Abp/BackgroundJobs/RabbitMQ/AbpRabbitMqBackgroundJobOptions.cs
+++ b/framework/src/Volo.Abp.BackgroundJobs.RabbitMQ/Volo/Abp/BackgroundJobs/RabbitMQ/AbpRabbitMqBackgroundJobOptions.cs
@@ -15,10 +15,16 @@ namespace Volo.Abp.BackgroundJobs.RabbitMQ
         /// </summary>
         public string DefaultQueueNamePrefix { get; set; }
 
+        /// <summary>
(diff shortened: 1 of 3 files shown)


hikalkan:
As I understand, we are publishing delayed messages to another queue. Then how we consume that queue? Can you explain the implementation.

realLiangshiwei:
Yes,

I used RabbitMQ's message expiration and dead letter features.

The normal queue is bound to the delayed queue as a dead letter queue:

https://github.com/abpframework/abp/pull/9725/files#diff-58bc5b9360333e116200e7894e850259b7eb34f608e7cd67a00ee014829da188R39

When a delayed job is publishing, it is published to the delayed queue and an **expiration time** is set,  after the message expires, it will move to the normal queue according to the dead letter routing, and then the application will consume it to implement delayed jobs

hikalkan:
Why we need this feature?

realLiangshiwei:
https://docs.abp.io/en/abp/latest/Background-Jobs#queue-a-job-item
 
Because ABP background jobs support the delay parameter, should we implement it in the background job provider?

realLiangshiwei:
https://docs.abp.io/en/abp/latest/Background-Jobs#queue-a-job-item

Because the background job manager supports the delay parameter and in the default implementation it will delay the execution of the job, should we implement it in the RabbitMQ provider?

hikalkan (approved):

PR #9768, opened 2021-08-06
fix: DTR in angular not replacing placeholder from baseUrl

### LINKED ISSUE #9728: Angular tenant resolve  broken when update to version 4.4.0
After update abp packages to 4.4.0, i found tenant resolve not work on the domain server.  Then i run the abp sample project DomainTenantResolver,  it works fine with default version 4.3.2，when i update to latest version 4.4.0, the tenant resolve with error also.
qq20210803184821

bnymncoskuner:
Resolves #9728


maliming:
Tested, it works.

maliming (approved):

PR #9798, opened 2021-08-11
Remove Application.Contracts reference from DbMigrator

### LINKED ISSUE #9796: DbMigrator may not need to reference Application.Contracts
`DbMigrator` may not need to reference `Application.Contracts`.

Only referencing `EntityFrameworkCore` project gives a reference to `Domain` too, which contains `*DbMigrationService` class. Besides, `DbMigrator` is not a client of the backend.

What do you think about it?

ABP 4.4

realLiangshiwei:
Resolve https://github.com/abpframework/abp/issues/9796


realLiangshiwei:
https://github.com/abpframework/abp/issues/9796#issuecomment-896678231

PR #9813, opened 2021-08-13
CmsKit - Try to use Public Layout if presented by Theme

enisn:
CmsKit Public pages should try to use Public layout if presented by theme.
Currently, developer must add **_ViewStart.cshtml** file into the end project


hikalkan:
Please send PR to rel-4.4 branch.

enisn:
It calls `GetPublicLayout` method and default value of `fallbackToDefault` is **true**. So it'll always fallback to default layout if public layout is not provided by theme:

https://github.com/abpframework/abp/blob/48c52625f4c4df007f04d5ac6368b07411aa7521/framework/src/Volo.Abp.AspNetCore.Mvc.UI/Volo/Abp/AspNetCore/Mvc/UI/Theming/ThemeExtensions.cs#L15-L18

ebicoglu:
we are releasing 4.4.1 today. I moved it to 5.0

hikalkan (approved):

PR #9821, opened 2021-08-15
Use DisallowConcurrentExecution to keep a Job from firing concurrently

### LINKED ISSUE #9819: QuartzPeriodicBackgroundWorkerAdapter should add `[DisallowConcurrentExecution]`
abp 4.4.0 / mvc

Volo.Abp.BackgroundWorkers.Quartz

`QuartzPeriodicBackgroundWorkerAdapter` should use the `[DisallowConcurrentExecution]` attribute to prevent concurrent execution.

This is useful in multi-instance environment.

Also same for `Volo.Abp.BackgroundJobs.Quartz.QuartzJobExecutionAdapter`.

realLiangshiwei:
Resolve https://github.com/abpframework/abp/issues/9819


realLiangshiwei:
We don't need to use DisallowConcurrentExecution for background Jobs. because they are all different job keys and only executed once

maliming (approved):

PR #9865, opened 2021-08-20
Enabled PublishTrimmed for Blazor wasm project

realLiangshiwei:


sturlath:
Would you not want to enable linking in Release mode and not when developing locally (for faster start-up)?

Something like this
```xml
<PropertyGroup>
  <BlazorWebAssemblyEnableLinking Condition="'$(Configuration)'!='Release'">false</BlazorWebAssemblyEnableLinking>
</PropertyGroup>
```

But what about `PublishTrimmed`?

realLiangshiwei:
@sturlath 

https://docs.microsoft.com/en-us/aspnet/core/blazor/host-and-deploy/configure-linker?view=aspnetcore-3.1

> The linker is disabled when building in Debug configuration

maliming (approved):

PR #9901, opened 2021-08-26
Fix object reference not set to an instance of an object for http proxy

wangj90:
The method ConvertValueToString reports Object reference not set to an instance of an object if there is an array or collection with null elements in the parameter.


maliming (approved):

maliming:
Thanks @wangj90

PR #9905, opened 2021-08-26
Add C# and JavaScript Static Client Proxy Generation

### LINKED ISSUE #9864: C# and JavaScript Static Client Proxy Generation
ABP's dynamic client proxy ([for c#](https://docs.abp.io/en/abp/latest/API/Dynamic-CSharp-API-Clients) and [for javascript](https://docs.abp.io/en/abp/latest/UI/AspNetCore/Dynamic-JavaScript-Proxies)) system creates the proxies on runtime. With this issue, we are planning to build an alternative system that creates the client proxy code in development time. We are already doing it for angular with the `abp generate-proxy` command ([see doc](https://docs.abp.io/en/abp/latest/UI/Angular/Service-Proxies)). We will create a similar system for c# client proxies (that is typically used by the Blazor UI and tiered MVC applications) and javascript client proxies (for mvc UI, while calling APIs from the JavaScript code).

ABP CLI's same `abp generate-proxy` command will be used. We can get a parameter (like `-t JavaScript` or `-t csharp`) to indicate the language. To generate a client code, we should first run the server application in our development environment. It exposes the `api/abp/api-definition` endpoint. ABP CLI will use this endpoint to get the API definition (like the current Angular code generation) and create the client code.

This feature naturally solves [that problem](https://github.com/abpframework/abp/issues/9405). So, we no longer need to add HTTP API layers of microservices to API gateways to be able to create `api/abp/api-definition` endpoint with all the service definitions included. Client will already have the necessary proxy code to call the API. When it calls, Ocelot (or another gateway) will already forward request to the corresponding microservice (this is already working now).

Note that we won't remove the dynamic client system. It will remain in v5.0 just as before, to be backward compatible. With this new feature, we will have two ways to create dynamic client proxies: 1) dynamically on runtime 2) statically on development.

### C# Client Proxy Generation Notes

I will typically run this command in the directory of the `*.HttpApi.Client` project in the startup solution:

````
abp generate-proxy --type csharp --module identity --url https://localhost:8292 
````

* Shortcut for `--type` is `-t`. *Required*.
* Shortcut for `--module` is `-m`. This value is matched with the module name in [modules dictionary](https://github.com/abpframework/abp/blob/dev/framework/src/Volo.Abp.Http/Volo/Abp/Http/Modeling/ApplicationApiDescriptionModel.cs#L11) of the API definition. *Optional, assumed `app` if not provided*.
* `--url` is the URL of the server-side application. It should be running while executing this command. *Required*.

#### Implementation Notes

* Name clients like `UserClientProxy` for `UserController`.
* Derive the client proxy from a common class, like `ClientProxyBase` that is defined in `Volo.Abp.Http.Client` package.
* Implement the corresponding interface of the controller, like `IUserAppService`. Create only the methods of that interface. The controller may have more methods, but we don't want to create proxy methods for the methods not defined by that interface. Also, this interface should be available in the client side. We don't want to create interfaces or DTO types defined in the service methods, they should be available on the client side (it has a reference to the *Application.Contracts* package, so the types should already be available there).
* Create the client class partial and leave an empty partial to the developer, so they may want to add custom methods.
* Place all the client classes into `ClientProxies` namespace by default. We can get the folder name as a parameter, like `--folder MyProxies/InnerFolder`.
* Try to keep method bodies of the proxy client classes minimal, delegate the actual work to base class methods.

### JavaScript Client Proxy Generation Notes

I will run this command in the directory of the web project:

````abp generate-
abp generate-proxy --type js --module identity --url https://localhost:8292 
````

#### Implementation Notes

* Every module should be generated in a separate JavaScript file, like `identity-proxy.js` under `wwwroot/client-proxies` folder by default.
* We can get the full output path as a parameter, like `--out Pages/Identity/client-proxies.js `.

### PR DESCRIPTION
**Pull request by realLiangshiwei**
Resolve https://github.com/abpframework/abp/issues/9864

Usage:

`abp generate-proxy -t  ng`

`abp generate-proxy -t  js -m identity -u https://localhost:44305`
 
image

`abp generate-proxy -t  csharp -m identity -u https://localhost:44305`

* `*.partial.cs` is a partial class of client proxy class, it uses to add custom methods.
* `*.generate-proxy.json` is the API definition of the module, should not change it.

image

```csharp
    [ExposeServices(typeof(IIdentityRoleAppService))]
    public partial class IdentityRoleClientProxy : ClientProxyBase<IIdentityRoleAppService>, IIdentityRoleAppService
    {
        public virtual async Task<ListResultDto<IdentityRoleDto>> GetAllListAsync()
        {
            return await RequestAsync<ListResultDto<IdentityRoleDto>>(nameof(GetAllListAsync));
        }
        // .......
 
        public virtual async Task DeleteAsync(Guid id)
        {
            await RequestAsync(nameof(DeleteAsync), id);
        }
 
    }
```


maliming (approved):

hikalkan:
Hi,

I tried but failed with the following exception:

````
Could not get DynamicHttpClientProxyConfig for Volo.Abp.Identity.IIdentityRoleAppService.
Volo.Abp.AbpException: Could not get DynamicHttpClientProxyConfig for Volo.Abp.Identity.IIdentityRoleAppService.
   at Volo.Abp.Http.Client.HttpProxyExecuter.MakeRequestAsync(HttpProxyExecuterContext context) in D:\Github\abp\framework\src\Volo.Abp.Http.Client\Volo\Abp\Http\Client\HttpProxyExe
cuter.cs:line 92
   at Volo.Abp.Http.Client.HttpProxyExecuter.MakeRequestAndGetResultAsync[T](HttpProxyExecuterContext context) in D:\Github\abp\framework\src\Volo.Abp.Http.Client\Volo\Abp\Http\Clie
nt\HttpProxyExecuter.cs:line 60
   at Volo.Abp.Http.Client.ClientProxying.ClientProxyBase`1.RequestAsync[T](String methodName, Object[] arguments) in D:\Github\abp\framework\src\Volo.Abp.Http.Client\Volo\Abp\Http\
Client\ClientProxying\ClientProxyBase.cs:line 24
   at Volo.Abp.Identity.HttpApi.Client.ClientProxies.IdentityRoleClientProxy.GetListAsync(GetIdentityRolesInput input) in D:\Github\abp\modules\identity\src\Volo.Abp.Identity.HttpAp
i.Client\ClientProxies\IdentityRoleClientProxy.cs:line 25
````

I did my test in a branch: https://github.com/abpframework/abp/tree/services-proxy-trying
See me commit: https://github.com/abpframework/abp/commit/6f87944021c940d54c9bdff470e0234157193f9a

I am trying to use static c# proxies for the identity module. Actually, we will do this for all modules as a part of this PR. Can you please make it working?

I also have some findings and suggestions;

* We can set namespace from the api-definition json and add `ClientProxies` to it. Example:
For `Volo.Abp.Identity.IdentityRoleController` we can create  `IdentityRoleClientProxy` in the `Volo.Abp.Identity.ClientProxies` namespace.
* Add `// ReSharper disable once CheckNamespace` on top of the class, so we Rider developers don't get warning :)
* * identity-generate-proxy.json is not embedded resource by default. abp generate-proxy should make it embedded and also we should add it to virtual file system. How can we automate that? Adding to VFS may be manual, but embedded resource can be easily automated by manipulating the csproj file.
* I sugges to rename IdentityRoleClientProxy.partial.cs to IdentityRoleClientProxy.Generated.cs, move the generated code here and leave the IdentityRoleClientProxy.cs as empty for the developer.
* `[ExposeServices(typeof(IIdentityRoleAppService))]` should be `[ExposeServices(typeof(IIdentityRoleAppService), typeof(IdentityRoleClientProxy))]`, so we can also inject by class.

hikalkan:
JavaScript proxies perfectly works, great. I tried it for identity's user page: https://github.com/abpframework/abp/commit/5a0217b2f3f5f9c9327521fb387fde56c2cd2dad

We should make this for all modules. I mean we will use static JS proxies for the modules, from now.

hikalkan:
Currently, `IdentityUserClientProxy.cs` is like that:

````csharp
[ExposeServices(typeof(IIdentityUserAppService), typeof(IdentityUserClientProxy))]
public partial class IdentityUserClientProxy : ClientProxyBase<IIdentityUserAppService>, IIdentityUserAppService
{
}
````

and `IdentityUserClientProxy.Generated.cs` is like that:

````csharp
public partial class IdentityUserClientProxy
{
    ....
}
````

It would be better if we move attributes (`Dependency` and `ExposeServices`) and inheritance/implementation to the generated file, so use do not accidently deletes them.

hikalkan:
I am merging this. You can address my comment (https://github.com/abpframework/abp/pull/9905#issuecomment-920018556) as a separate PR.

Issue #3085, opened 2020-03-12
Extended built in module's  Localization file does not work after publish


inamvar:
I'm using abp 2.1.1 and I've extened some abp built in modules localization such as `AbpIdentity` and ... 

![image](https://user-images.githubusercontent.com/6972027/76492846-731d2680-6446-11ea-9de1-d7e89a772390.png)

And this is my Configuarion: 

``` C#
  Configure<AbpLocalizationOptions>(options =>
            {


                options.Resources
                .Get<IdentityResource>()
                .AddVirtualJson("/Localization/AbpIdentity");

                options.Resources
                    .Get<AbpPermissionManagementResource>()
                    .AddVirtualJson("/Localization/AbpPermissionManagement");


                options.Resources
                    .Get<AbpIdentityServerResource>()
                    .AddVirtualJson("/Localization/AbpIdentityServer");



                options.Resources
                    .Get<AbpFeatureManagementResource>()
                    .AddVirtualJson("/Localization/AbpFeatureManagement");




                options.Resources
                    .Get<AbpSettingManagementResource>()
                    .AddVirtualJson("/Localization/AbpSettingManagement");


                options.Resources
                    .Get<AbpTenantManagementResource>()
                    .AddVirtualJson("/Localization/AbpTenantManagement");



                options.Resources
                    .Get<AbpValidationResource>()
                    .AddVirtualJson("/Localization/AbpValidation");

                options.Resources
                    .Add<PmsCloudResource>("fa")
                     .AddBaseTypes(typeof(AbpFeatureManagementResource))
                    .AddBaseTypes(typeof(AbpValidationResource))
                    .AddVirtualJson("/Localization/PmsCloud");

            });
        }
```

everything  works properly in debug mode but after publish  only my own module localization works and other extended localisations did not work at all.

maliming:
hi
Can you share the complete code of the module class where `Configuarion `is located?

inamvar:
this my DomainShared Module: 

``` C#
using Dariksoft.PmsCloud.Localization;
using Volo.Abp.AuditLogging;
using Volo.Abp.BackgroundJobs;
using Volo.Abp.FeatureManagement;
using Volo.Abp.FeatureManagement.Localization;
using Volo.Abp.Identity;
using Volo.Abp.Identity.Localization;
using Volo.Abp.IdentityServer;
using Volo.Abp.IdentityServer.Localization;
using Volo.Abp.Localization;
[...]

PR #9907, opened 2021-08-27
Add Build Action specification for translation files

mtozlu:
New translation files (en.json etc) should be set to 'Embedded Resource', otherwise translation will not work on published projects. This needs to be documented.
Related: #3085

### DIFF
--- a/docs/en/Localization.md
+++ b/docs/en/Localization.md
@@ -63,8 +63,6 @@ public class MyModule : AbpModule
 }
 ````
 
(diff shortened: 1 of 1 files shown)


mtozlu:
I will add https://docs.abp.io/en/abp/latest/Virtual-File-System#working-with-the-embedded-files link for creating localization resources in a separate module.

mtozlu:
@maliming ping

maliming:
hi @mtozlu 

Thank you for your contribution,

I think these contents are duplicates, we have explained relevant in other documents.

PR #9909, opened 2021-08-27
Publish events while unit of work is being completed, in the same unit of work

### LINKED ISSUE #9896: Publish events while unit of work is being completed, in the same unit of work
### Current Situation

ABP auto-publishes create, update and delete events for the entities:

* Local events ([documentation](https://docs.abp.io/en/abp/latest/Local-Event-Bus#pre-built-events))
  * `EntityCreatingEventData<T>`
  * `EntityCreatedEventData<T>`
  * `EntityUpdatingEventData<T>`
  * `EntityUpdatedEventData<T>`
  * `EntityDeletingEventData<T>`
  * `EntityDeletedEventData<T>`
  * `EntityChangingEventData<T>`
  * `EntityChangedEventData<T>`
* Distributed events ([documentation](https://docs.abp.io/en/abp/latest/Distributed-Event-Bus#pre-defined-events))
  * `EntityCreatedEto<T>`
  * `EntityUpdatedEto<T>`
  * `EntityDeletedEto<T>`

These events are published on DbContext.SaveChangesAsync for EF Core, and in repository's create/update/delete methods for MongoDB. So, for EF Core, they are published when we manually call Unit Of Work's SaveChangesAsync, or the UOW is completed. However, for MongoDB, they are immediately published when we call repository's create/update/delete methods. Publishing means `PublishAsync` method of `ILocalEventBus` or `IDistributedEventBus` is called. `ILocalEventBus.PublishAsync` immediately execute the handlers and `IDistributedEventBus.PublishAsync` immediately sends event to the message broker (e.g. RabbitMQ). Remember that IDistributedEventBus is actually in-process and delivers events to the local event bus if we don't explicitly install one of the message broker integrations, like RabbitMQ.

We can also publish events using `ILocalEventBus` and `IDistributedEventBus` and these events are immediately published.

Also, past-tense events (like `EntityCreatedEventData<T>`, `EntityUpdatedEventData<T>`, `EntityCreatedEto<T>` and `EntityUpdatedEto<T>`) are published after the current unit of work. That means, if an event handler throws an exception, it can not rollback the transaction that the event was published in. Continuous tense events (like `EntityCreatingEventData<T>`) are still published in the UOW, so the handler can rollback the transaction.

### New approach

Beginning with ABP v5.0, auto-published create, update and delete events for the entities will be always published at the end of the unit of work (when `IUnitOfWork.CompleteAsync` is called) and included in the current transaction. In this way, a handler can throw exception and can rollback the transaction.

If you manually call `PublishAsync` method of an event bus, the published event is stored in the current unit of work and actually published at the end of the unit of work and included in the current transaction.

With the new approach;

* All the event handlers are executed in the current unit of work, so the whole process becomes transactional. This is very useful and simple model for non-distributed systems.
* Event publish system is being compatible between EF Core and MongoDB and it will be more consistent.
* Since we publish all events at the end of the unit of work, we can ensure the events are published in the right order. Previously, manually published events were being published before pre-built entity create/update/delete events for EF Core since entity events were being published at the end of the UOW (or when we call SaveChanges manually).

### LINKED ISSUE #9897: Deprecate EntityCreatingEventData, EntityUpdatingEventData, EntityDeletingEventData and EntityChangingEventData
Related to https://github.com/abpframework/abp/issues/9896

`EntityCreatingEventData`, `EntityUpdatingEventData`, `EntityDeletingEventData` and `EntityChangingEventData` is not necessary now, because `EntityCreatedEventData`, `EntityUpdatedEventData`, `EntityDeletedEventData` and `EntityChangedEventData` is already taken into the current unit of work.

hikalkan:
Resolved #9896
Resolved #9897


framework/src/Volo.Abp.Uow/Volo/Abp/Uow/UnitOfWork.cs
```diff
@@ -126,6 +134,30 @@ public virtual async Task CompleteAsync(CancellationToken cancellationToken = de
...
+                        var distributedEventsToBePublished = DistributedEvents.OrderBy(e => e.EventOrder).ToArray();
+                        DistributedEvents.Clear();
+                        await UnitOfWorkEventPublisher.PublishDistributedEventsAsync(
+                            distributedEventsToBePublished
+                        );
+                    }
+
+                    await SaveChangesAsync(cancellationToken);
+                }
+
```
maliming, line 160: We will delete the cache in the event of the entity, but the `considerUow` feature will set the cache in `OnCompletedAsync`.

We also need to refactor the `considerUow` feature. Do you have any suggestions?

framework/src/Volo.Abp.EventBus/Volo/Abp/EventBus/EventBusBase.cs
```diff
@@ -154,7 +154,7 @@ protected virtual async Task TriggerHandlersAsync(Type eventType, object eventDa
                     var baseEventType = eventType.GetGenericTypeDefinition().MakeGenericType(baseArg);
                     var constructorArgs = ((IEventDataWithInheritableGenericArgument)eventData).GetConstructorArgs();
                     var baseEventData = Activator.CreateInstance(baseEventType, constructorArgs);
-                    await PublishAsync(baseEventType, baseEventData);
```
hikalkan, line 136: Great! Thanks.

framework/src/Volo.Abp.Uow/Volo/Abp/Uow/UnitOfWork.cs
hikalkan: I think this development doesn't effect the `considerUow` feature of the distributed cache. There is no way to guarantee to set the cache in the same transactional because database and cache are different systems.
It is better to keep it as is. You can merge this.

maliming: 1. The `considerUow` feature sets the cache after the unit of work is completed.
2. The event of an entity changed deletes the cache.

Now their order is reversed.

You can see the unit tests failed https://github.com/abpframework/abp/runs/3458051031

maliming (approved):

PR #9940, opened 2021-09-01
Use ASP NET Core's AuthenticationScheme to handle AbpAuthorizationException.

### LINKED ISSUE #9926: Use ASP NET Core's `AuthenticationScheme` to handle `AbpAuthorizationException`.
Related #2643 #3848  #4785  #5235 #7503

> When you call the controller that requires authentication, the authentication middleware finds that the current user is not authenticated and calls ChallengeAsync (DefaultChallengeScheme is identity cookie). At this point the request has been shorted.
> If your anonymous controller calls the application service method, it will execute the ABP filter and interceptor. The framework throws AbpAuthorizationException, the filter wraps the exception into a 401, etc.

We wrap the exception in `ExceptionFilter` as `RemoteServiceErrorResponse` and specify the HTTP status code for it.

An exception will be throw after the permission check fails. It will not get an exception in ASP NET Core but is handled by the authentication scheme.

I think we should let the authentication scheme handle `AbpAuthorizationException`, This will make the above behavior consistent.

---

Update:

Cookies authentication will set 401 for AJAX requests.

https://github.com/dotnet/aspnetcore/blob/v6.0.0/src/Security/Authentication/Cookies/src/CookieAuthenticationEvents.cs#L46-L50

### Forward any requests that start with /api to the JWT scheme.

```cs
context.Services.ConfigureApplicationCookie(options =>
{
    // Forward any requests that start with /api to the JWT scheme.
    options.ForwardDefaultSelector = ctx =>
    {
        return ctx.Request.Path.StartsWithSegments("/api") ? JwtBearerDefaults.AuthenticationScheme : null;
    };
});

context.Services.AddAuthentication()
    .AddJwtBearer(options =>
    {
        options.Authority = configuration["AuthServer:Authority"];
        //...
    });
```

maliming:
Resolve #9926

The `AbpAuthorizationException` will be handled by the authentication scheme. (`Forbid/Challenge`)


hikalkan (approved):

Sunzhuokai:
only works in development not production

Sunzhuokai:
in production `UseErrorPage` will cause status code 302 not 401 
```csharp
        if (!env.IsDevelopment())
        {
            app.UseErrorPage();
        }
```

maliming:
@Sunzhuokai Please create a new issue. Thanks.

PR #9945, opened 2021-09-02
chore: disable auto generator launchSettings.json

netcore-jroger:


maliming:
If this is necessary, we can add it to `common.props` and `common.test.props`

netcore-jroger:
This is only for projects with `Sdk="Microsoft.NET.Sdk.Web"` and `<OutputType>Library</OutputType>`.

maliming:
Please make changes at a single PR. Thanks.

https://github.com/abpframework/abp/pull/9946

PR #9961, opened 2021-09-03
disable auto generator launchSettings.json

netcore-jroger:


maliming:
hi

Why don't you change this in the `common.props` ?


https://github.com/abpframework/abp/pull/9946#issuecomment-911273432

netcore-jroger:
I tried use `$(UsingMicrosoftNETSdkWeb)` global variable, but is does not work.

maliming (approved):

maliming:
Many thanks @netcore-jroger

PR #9982, opened 2021-09-08
Update Dapper.md

git102347501:
Use GetDbConnectionAsync method


realLiangshiwei:
I think there is a problem, could you send a new PR? thanks

image

PR #10013, opened 2021-09-11
make methods virtual to be overridable

muratyuceer:
I maked model methods virtual to be able to customize it in the projects used.


muratyuceer (commented):
aproved

maliming (approved):

PR #10051, opened 2021-09-16
dynamic permissions

puschie286:
+ add method to easy access permission groups at runtime
+ add method to remove permissions from permission groups ( child permissions )

would be great if we can change permissions at runtime.

this is a minimal implementation of this feature. ( currently no group adding / removing )

please feel free to suggest a more structured solution or mention problems that might happen
( i guess there might be problems with the cache when removing existing entries )


puschie286:
push

puschie286:
improved the implementation and created a clean and more "natural" interface - last commit (c6f3dab) describe all changes
This should give you all the options that are available in the PermissionDefinitionProvider.

one thing we unfortunately could not solve cleanly is the instantiation of PermissionDefinition / PermissionGroupDefinition ( which we changed to public for now ). Maybe we need to create a seperate manager class in Authorization.Abstractions.

PR #10053, opened 2021-09-16
Correct naming of ApiScopeRepository.GetByName method.

PMExtra:
https://github.com/abpframework/abp/blob/0b50e1d219874c0fec1e1537067773ea7dfd8ce1/modules/identityserver/src/Volo.Abp.IdentityServer.EntityFrameworkCore/Volo/Abp/IdentityServer/ApiScopes/ApiScopeRepository.cs#L21-L26

The method was implementation like `FindByNameAsync`, it will return null instead of throw an exception when the entity was not found.

This PR will correct its name. Notice that this is a breaking change, it may should be merged in next major version.


maliming (approved):

maliming:
Thanks @PMExtra

PR #10142, opened 2021-09-24
Improve IdentityServer module.

PMExtra:
There are 3 commits in this PR:

### [fix(identity-server): includeDetails of ApiScopeRepository.FindByNameAsync.](https://github.com/abpframework/abp/pull/10142/commits/52bc3bb57137934b3f1f4b4a64710a3e82c49e7e)
Fix the missing `includeDetails` in `ApiScopeRepository.FindByNameAsync` method.

### [chore(identity-server): some detailed improvements.](https://github.com/abpframework/abp/pull/10142/commits/3149f770eadb336fab2a6e31498300bb58249e51)
1. Use `base(id)` in constructor of entities if available.
2. Prevent some possible `NullReferenceException` in [`DeviceFlowStore`](https://github.com/abpframework/abp/pull/10142/commits/68a5bd64cd04f352b862ac4082bc32f3630c57b7#diff-e4ff39aa3f9a6a916bd0256b9eab7f798d6bb79ed69c7c6453c242694b619799).
3. Convert some LINQ from query syntax to method syntax.
4. Combine `.Where()` filter with `.FirstOrDefaultAsync()`.
5. Replace some `cancellationToken` with `GetCancellationToken(cancellationToken)`.

### [feat(identity-server): improve property and claim of Client.](https://github.com/abpframework/abp/pull/10142/commits/f6ef8e5ed221ccca2f3e6c53f44a316aeaa3fb3b)
1. Simply `ClientProperty` related methods in `Client` class.
The `Property` of other entities such as `ApiResourceProperty`, `ApiScopeProperty`, there are common domain methods such as `ApiScopeProperty FindProperty(string key)`, `void RemoveProperty(string key)`.
But in `Client` class, these methods need a extra argument `value`.
According to [this unit test of IdentityServer4](https://github.com/IdentityServer/IdentityServer4/blob/4.1.2/src/EntityFramework.Storage/test/UnitTests/Mappers/ClientMappersTests.cs#L67-L81), I got that the Client cannot have multiple properties with the same key.
So I think these methods should be unified, just drop the unnecessary `value` argument.

2. Sort the arguments of `ClientClaim`.
I change the arguments order from `value, type` to `type, value`. Because the `type` work like a `key`, it's different from `ApiResourceSecret.Type`.

3. Add some new API of `ClientClaim`.
- `void Client.RemoveClaim(string type)`
- `List<ClientClaim> Client.FindClaims(string type)`
It's different from `ClientProperty`. The `ClientClaim` can have multiple values with same type.

**Notice: This is a breaking change.**

### DIFF
--- a/modules/identityserver/src/Volo.Abp.IdentityServer.Domain/Volo/Abp/IdentityServer/ApiResources/ApiResource.cs
+++ b/modules/identityserver/src/Volo.Abp.IdentityServer.Domain/Volo/Abp/IdentityServer/ApiResources/ApiResource.cs
@@ -35,11 +35,9 @@ namespace Volo.Abp.IdentityServer.ApiResources
 
         }
 
-        public ApiResource(Guid id, [NotNull] string name, string displayName = null, string description = null)
+        public ApiResource(Guid id, [NotNull] string name, string displayName = null, string description = null) : base(id)
         {
             Check.NotNull(name, nameof(name));
-
-            Id = id;
 
             Name = name;
 
--- a/modules/identityserver/src/Volo.Abp.IdentityServer.Domain/Volo/Abp/IdentityServer/ApiScopes/ApiScope.cs
+++ b/modules/identityserver/src/Volo.Abp.IdentityServer.Domain/Volo/Abp/IdentityServer/ApiScopes/ApiScope.cs
@@ -40,11 +40,11 @@ namespace Volo.Abp.IdentityServer.ApiScopes
             bool required = false,
             bool emphasize = false,
             bool showInDiscoveryDocument = true,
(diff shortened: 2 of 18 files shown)


maliming:
hi

Some codes may not be ABP's code style, can you do the fix and enhance separately?

If your PR has many changes, you can create an issue to discuss.

PMExtra:
@maliming OK, I will rebase it to drop the commit which formatted the code.

PMExtra:
@maliming Finished, you can review it now. Thanks.

maliming:
Move to https://github.com/abpframework/abp/pull/10154

PR #10149, opened 2021-09-26
Introduce `AbpMongoDbDateTimeSerializer`.

### LINKED ISSUE #10127: Generated tests for datetime fail with MongoDB
Version 4.4.2 Angular + MongoDB

The generated datefield tests fail initially, example test:
image

Date validation fail
image

Dates are stored as UTC in mongodb, so adjusting to verify against UTC passes the test:
image

### Steps to reprocuce

Generate a new entity with a DateTime field using ABP Suite then run the generated tests.

maliming:
Resolve #10127

`MongoDb` will normalize the `datetime` based on  `AbpClockOptions` now, before `DateTime values in MongoDB are always saved as UTC`.


hikalkan:
Thanks @maliming I want two additional things:

1. Please create an option to keep the previous behavior. There are many applications already build and has a lot of data in the database and we should not force everyone to convert their datas.
2. Create a new migration guide document for 4.x too 5.0 here: https://github.com/abpframework/abp/tree/dev/docs/en/Migration-Guides and document this change and write how to keep the previous behavior.

hikalkan:
Can you please resolve the conflicts.

hikalkan (approved):

vitry:
If the kind of the input datetime value is local, and the clock DateTimeKind option is UTC, It results in the time saved in mongodb to be a future time. Because double timezone offsets. Is it a bug? @maliming 

`    public override void Serialize(BsonSerializationContext context, BsonSerializationArgs args, DateTime value)
    {
        context.Writer.WriteDateTime(DisableDateTimeNormalization
            ? ToMillisecondsSinceEpoch(value)
            : ToMillisecondsSinceEpoch(DateTime.SpecifyKind(value, DateTimeKind)));
    }`

maliming:
hi @vitry 

Please create a new issue and share some code. 

Thanks.

PR #10150, opened 2021-09-26
feat: Support multiple policies for verification in OR

netcore-jroger:
Support multiple policies for verification in OR.
sample:
```csharp
// And verify
[Authorize("mydoc")]
[Authorize("mydraft")]
public class MyAppService
{
    // ......
}

// Or verify
[Authorize("mydoc,mydraft")]
public class MyAppService
{
    // ......
}
```

### DIFF
--- a/framework/src/Volo.Abp.Authorization/Volo/Abp/Authorization/MethodInvocationAuthorizationService.cs
+++ b/framework/src/Volo.Abp.Authorization/Volo/Abp/Authorization/MethodInvocationAuthorizationService.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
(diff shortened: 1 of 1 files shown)


maliming:
hi

We should follow Microsoft's recommended approach. But we can consider adding `PermissionsRequirement` and `PermissionsRequirementHandler` to check multiple permissions at the same time and decide whether the result is **AND** or **OR**.

https://docs.microsoft.com/en-us/aspnet/core/security/authorization/policies?view=aspnetcore-5.0

https://github.com/abpframework/abp/blob/dev/framework/src/Volo.Abp.Authorization/Volo/Abp/Authorization/Permissions/RequirePermissionsSimpleBatchStateCheckerModel.cs#L10-L12

maliming:
Move to https://github.com/abpframework/abp/pull/10152

netcore-jroger (commented):
when current user is admin and Or verify, will throw exception.

PR #10171, opened 2021-09-29
Implement Hangfire Background Worker Manager

### LINKED ISSUE #10163: Hangfire Recurring Jobs provider for Background Workers
### Hangfire Recurring Jobs

There is already an implementation built to use Hangfire for Background Jobs, but if you want to use more advanced scheduling for Background Workers than `every x number of million seconds` then the current option is Quartz. 

Hangfire does provide the functionality to use advanced scheduling for Recurring Jobs by using the CRON which provides for very advanced scheduling of specific background workers.

Here is more information about the recurring jobs in Hangfire.
https://docs.hangfire.io/en/latest/background-methods/performing-recurrent-tasks.html

It would be great to have a single provider for both Background Jobs and Workers with an easy-to-manage dashboard, by providing an interface and implementation of `IHangfireBackgroundWorker` for the `IBackgroundWorker` interface.

gdeswardt:
Added an implementation to use Hangfire as the Background Worker Manager
by utilising the Recurring Jobs from Hangfire. This allows the same
Hangfire infrastructure to be used for Background Jobs and Workers.

This resolves #10163


realLiangshiwei:
Thanks for your contributor, but we plan to move [Lsw.Abp.BackgroundWorkers.Hangfire](https://github.com/realLiangshiwei/Lsw.Abp.BackgroundWorkers.Hangfire) to ABP framework.

So, close this PR. Thanks again.

PR #10185, opened 2021-09-30
 Set `IsActive` default value to `true`.

maliming:


nebula2:
Wasted 2 hours on this after migrating from 4.4.4 to 5.2.0. When to expect this to happen?

maliming:
https://docs.abp.io/en/abp/latest/Migration-Guides/Abp-5_0#user-active-passive

Issue #10311, opened 2021-10-13
Enqueue background jobs with `executionTime`.


gdlcf88:
Add an overloaded method in [IBackgroundJobManager](https://github.com/abpframework/abp/blob/48c52625f4c4df007f04d5ac6368b07411aa7521/framework/src/Volo.Abp.BackgroundJobs.Abstractions/Volo/Abp/BackgroundJobs/IBackgroundJobManager.cs):
```csharp
    /// <summary>
    /// Enqueues a job to be executed.
    /// </summary>
    /// <typeparam name="TArgs">Type of the arguments of job.</typeparam>
    /// <param name="args">Job arguments.</param>
    /// <param name="executionTime">Time to execute the job.</param>
    /// <param name="priority">Job priority.</param>
    /// <returns>Unique identifier of a background job.</returns>
    Task<string> EnqueueAsync<TArgs>(
        TArgs args,
        DateTime executionTime,
        BackgroundJobPriority priority = BackgroundJobPriority.Normal
    );
```

It could help to specify an absolute time to execute a job.

maliming:
https://github.com/abpframework/abp/pull/10313

PR #10292, opened 2021-10-12
Implement Hangfire Background Worker Manager

realLiangshiwei:
Resolve https://github.com/abpframework/abp/issues/10163

PR #10313, opened 2021-10-13
Enqueue background jobs with `executionTime`.

gdlcf88:
#10311


maliming:
@hikalkan What do you think?

hikalkan:
We can just calculate the delay (as `executionTime - Clock.Now`) and pass to the current method. Why we should define another method?

gdlcf88:
`executionTime - Clock.Now` is fine. However, the final (first) execution time may depend on the job inserted time if we use the `delay`. I'm not sure how long it will spend. It is not rigorous enough when we have an absolute execution time.

gdeswardt:
There is also a BackgroundJob Manager implementation for Hangfire (https://github.com/abpframework/abp/pull/10292) that will require to be updated for this specific change. Would also consider updating the documentation with this change

PR #10370, opened 2021-10-20
Introduce DynamicJavaScriptProxyOptions

### LINKED ISSUE #9955: Introduce option to enable/disable dynamic JavaScript proxies per module
Based on https://github.com/abpframework/abp/issues/9864

We can enable dynamic proxies for the `identity` module:

````csharp
Configure<DynamicJavaScriptProxyOptions>(options => {
    options.EnabledModules.Add("identity");
});
````

`options.EnabledModules` is a `HashSet<string>` and contains a single item, `app`, by default.

`/Abp/ServiceProxyScript` endpoint returns scripts only for the specified modules.

In this way, ABP doesn't generate dynamic JS client proxies for the modules already have static proxies created in development time.

### PR DESCRIPTION
**Pull request by realLiangshiwei**
Resolve https://github.com/abpframework/abp/issues/9955, https://github.com/abpframework/abp/issues/10419

I used disabled mode, dynamic proxy is enabled by default, we manually disable it in the module. 

For developers, they don’t need to change anything when they upgrade to 5.0, the dynamic proxy is still works.

### CODE REVIEW

File: framework/src/Volo.Abp.Http/Volo/Abp/Http/ProxyScripting/Generators/JQuery/DynamicJavaScriptProxyOptions.cs [hunk 1]
```diff
@@ -0,0 +1,16 @@
+﻿using System.Collections.Generic;
+
+namespace Volo.Abp.Http.ProxyScripting.Generators.JQuery
+{
+    public class DynamicJavaScriptProxyOptions
+    {
+        public HashSet<string> EnabledModules { get; set; }
+
+        public bool EnabledAllModules { get; set; }
```
hikalkan:
```suggestion
        public bool IsEnabledForAllModules { get; set; }
```

framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/AbpCliCoreModule.cs
```diff
@@ -71,6 +72,11 @@ public override void ConfigureServices(ServiceConfigurationContext context)
                 options.Generators[AngularServiceProxyGenerator.Name] = typeof(AngularServiceProxyGenerator);
                 options.Generators[CSharpServiceProxyGenerator.Name] = typeof(CSharpServiceProxyGenerator);
             });
+
+            Configure<DynamicJavaScriptProxyOptions>(options =>
+            {
+                options.EnabledAllModules = true;
```
hikalkan: I didn't understand why CLI enables it for all modules? Isn't that a server-side option?

framework/src/Volo.Abp.Http/Volo/Abp/Http/ProxyScripting/Generators/JQuery/JQueryProxyScriptGenerator.cs
```diff
@@ -19,24 +20,51 @@ public class JQueryProxyScriptGenerator : IProxyScriptGenerator, ITransientDepen
...
         }
 
+        private bool ShouldCreateModuleScript(KeyValuePair<string, ModuleApiDescriptionModel> module)
+        {
+            if (_dynamicJavaScriptProxyOptions.EnabledAllModules)
+            {
+                return true;
+            }
+
+            if (_dynamicJavaScriptProxyOptions.EnabledModules.Any(m => m.Equals(module.Key, StringComparison.CurrentCultureIgnoreCase)))
```
hikalkan: I think we can make `_dynamicJavaScriptProxyOptions.EnabledModules` a HashSet<string> for a faster comparison. Also, we can force to be case-sensitive to make it even faster.

modules/account/src/Volo.Abp.Account.HttpApi/Volo/Abp/Account/ProfileController.cs
```diff
@@ -1,13 +1,14 @@
...
-namespace Volo.Abp.Identity
+namespace Volo.Abp.Account
 {
-    [RemoteService(Name = IdentityRemoteServiceConsts.RemoteServiceName)]
-    [Area("identity")]
+    [RemoteService(Name = AccountRemoteServiceConsts.RemoteServiceName)]
+    [Area("account")]
     [ControllerName("Profile")]
-    [Route("/api/identity/my-profile")]
+    [Route("/api/account/my-profile")]
```
hikalkan, line 10: I think that's a breaking change. Can you create an issue with a `breaking-change` label to announce that.

hikalkan (changes requested):

hikalkan:
We are also moving the profile service from identity to account. I think you're doing that because the UI is in the account part, right? Please create an annoucement about that breaking change and the reason. Can you also make same for the abp-commercial? Thanks.

modules/account/src/Volo.Abp.Account.HttpApi/Volo/Abp/Account/ProfileController.cs
realLiangshiwei: https://github.com/abpframework/abp/issues/10419

hikalkan (approved):

PR #10381, opened 2021-10-20
Improve cms-kit-pro comments search section

EngincanV:
> Fix cms-kit-pro comments search panel. it is not good ordered.

comments


cotur (changes requested):
I think, if we put `StartDate` & `EndDate` in one column instead of two, the search button can be placed in the same row.

Date inputs does not require two different column.

col-1 -> StartDate & EndDate
col-2 -> Username
col-3 -> EntityType
col-4 -> Search Button

EngincanV:
> I think, if we put `StartDate` & `EndDate` in one column instead of two, the search button can be placed in the same row.
> 
> Date inputs does not require two different column.
> 
> col-1 -> StartDate & EndDate col-2 -> Username col-3 -> EntityType col-4 -> Search Button

comments-style

cotur (approved):

PR #10389, opened 2021-10-21
fix select2 if modal doesn't exist

totpero:
When I try to use Select2ScriptContributor without bootstrap-modal I get Cannot read properties of undefined (reading 'Constructor') error.


maliming:
hi

We will remove this file in 5.0, so I move your changes to `rel-4.4`.

https://github.com/abpframework/abp/pull/10390

PR #10445, opened 2021-10-27
Try seek blob stream position to begin

JadynWong:
Use the file system or azure blob the blob provider, the Position of the returned stream is at the end.

### DIFF
--- a/framework/src/Volo.Abp.BlobStoring/Volo/Abp/BlobStoring/BlobContainer.cs
+++ b/framework/src/Volo.Abp.BlobStoring/Volo/Abp/BlobStoring/BlobContainer.cs
@@ -217,18 +217,10 @@ namespace Volo.Abp.BlobStoring
 
         protected virtual Stream TrySeekStreamToBegin(Stream stream)
         {
-            if (stream == null)
+            if (stream != null && stream.CanSeek)
             {
-                return stream;
+                stream.Seek(0, SeekOrigin.Begin);
             }
-
-            if (!stream.CanSeek)
-            {
-                return stream;
-            }
-
-            stream.Seek(0, SeekOrigin.Begin);
-
             return stream;
         }
     }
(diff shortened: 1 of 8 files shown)


maliming:
We should change `blob provider` instead of  `CmsKit`.

cotur (changes requested):
Hi @JadynWong,

as @maliming said, we need to upgrade the `Blob Providers` instead of updating it at the `Cms-Kit` module.

framework/src/Volo.Abp.BlobStoring/Volo/Abp/BlobStoring/BlobContainer.cs
```diff
@@ -214,5 +214,22 @@ public class BlobContainer : IBlobContainer
...
+            }
+
+            if (!stream.CanSeek)
+            {
+                return stream;
+            }
+
+            stream.Seek(0, SeekOrigin.Begin);
+
+            return stream;
```
rqx110: can simplified：
```suggestion            
            if (stream != null && stream.CanSeek)
            {
                stream.Seek(0, SeekOrigin.Begin);
            }

            return stream;
```

maliming:
We need to change the `Stream`  for all `BlobProviders`.

https://github.com/abpframework/abp/blob/dev/framework/src/Volo.Abp.BlobStoring.FileSystem/Volo/Abp/BlobStoring/FileSystem/FileSystemBlobProvider.cs#L79
https://github.com/abpframework/abp/blob/dev/framework/src/Volo.Abp.BlobStoring.Azure/Volo/Abp/BlobStoring/Azure/AzureBlobProvider.cs#L72
https://github.com/abpframework/abp/blob/dev/modules/blob-storing-database/src/Volo.Abp.BlobStoring.Database.Domain/Volo/Abp/BlobStoring/Database/DatabaseBlobProvider.cs#L122

JadynWong:
Hi @maliming 

I have some questions.
Now I try to handle it in `BlobContainer`'s `GetOrNullAsync`. Users usually use the `IBlobContainer` to manipulate the blob, and the `BlobProvider` only as an implementation.
Do we need to move the `TrySeekStreamToBegin` method to the `BlobProviderBase`, and handle it in each Provider?

maliming:
We are using `MemoryStream`, we can change it directly. 

No need for the `TrySeekStreamToBegin` method

JadynWong:
`DatabaseBlobProvider` using `byte[]` does not require it, it is default at start position

maliming (approved):

cotur (approved):

PR #10490, opened 2021-11-01
Automatically set TenantId while creating a new entity object.

hikalkan:
Previously ABP leaves it to the developer to set the right `TenantId` for a multi-tenant entity on object creation. With this change, it automatically sets the value from the current context.


lprichar (approved):
lgtm!

maliming:
The constructor of the base class is executed first, and then the subclass will set the `TenantId`.
Does this still fit your design?

image

maliming (approved):

dongfo:
you only set TenantId on Entity constructor  but ,constructor   of Entity<TKey>   do not inherit base()，their TenantId will not be set!

PR #10533, opened 2021-11-05
Update package.json

realLiangshiwei:
https://support.abp.io/QA/Questions/2091/v5-ABP-Suite-gives-error-on-startup-after-upgrading-to-Beta3


realLiangshiwei:
Duplicate of https://github.com/abpframework/abp/pull/10531

PR #10541, opened 2021-11-05
GetListByNormalizedRoleNameAsync does not return the expected result

muratyuceer:
GetListByNormalizedRoleNameAsync function does not return the expected result because the datafilters is not applied


muratyuceer:
Does the same issue happen in entityframework repository? If not, why them not have same logic.

hikalkan:
@muratyuceer EF data filtering relies on EF Core's global query filters, so we don't need a manual filtering.

realLiangshiwei:
Hi @muratyuceer,

I don't think the method will return the unexpected result, it finally queries the user and will be filtered

hikalkan:
I fixed this by https://github.com/abpframework/abp/pull/10614

Thanks @muratyuceer

Issue #10545, opened 2021-11-07
Add multitenancy to the "Tenant" entity


1d0n7kn0w:
I want to implement IMultiTenant to the "Tenant" entity. Is that possible?
I don't want to add the property "TenantId" by extending the entity, because I will lose all the mechanisms that abp.io does in the background to manage IMultiTenant entities

maliming:
This is not as designed, but you can download the source code of the module to customize it.

1d0n7kn0w:
@maliming making the "Tenant" class partial can be a solution, what do you think?

maliming:
I am not sure, you can give it a try.

PR #10548, opened 2021-11-08
Make the "Tenant" class partial

1d0n7kn0w:
use case : implementing IMultiTenant, thus allowing sub-tenancy
see #10545


maliming:
hi @1d0n7kn0w 

I think you can download the source code to change the **Tenant** class. This is not our design. Thanks.

1d0n7kn0w:
@maliming ok sorry but can you at least explain to me, what would be the impact of making the "Tenant" class partial on design?

1d0n7kn0w:
Hello @hikalkan , would like your opinion on this
Thanks

PR #10604, opened 2021-11-11
Update Localization.md

wosperry:
The original description is not easy to understand, add more details


maliming:
hi

I don't think these are needed. Thanks

PR #10636, opened 2021-11-17
Update Setting-Management.md

korolei:
Added 'ITransientDependency' interface to CustomSettingProvider class, so the following exception will not be thrown:
ComponentNotRegisteredException: The requested service 'CustomSettingProvider' has not been registered.


maliming (approved):

maliming:
Thanks @korolei

PR #10658, opened 2021-11-18
Log errors and not using if the current MongoDB does not support transactions.

maliming:
```ps1
[11:01:10 ERR] The current MongoDB database does not support transactions, All operations will be performed in non-transactions. This may cause errors.
[11:01:10 ERR] Standalone servers do not support transactions.
System.NotSupportedException: Standalone servers do not support transactions.
   at Mongo.MongoDB.MyUnitOfWorkMongoDbContextProvider`1.IsSupportedTransactions(MongoClient client) in
```

### DIFF
--- a/framework/src/Volo.Abp.MongoDB/Volo/Abp/Uow/MongoDB/UnitOfWorkMongoDbContextProvider.cs
+++ b/framework/src/Volo.Abp.MongoDB/Volo/Abp/Uow/MongoDB/UnitOfWorkMongoDbContextProvider.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.DependencyInjection;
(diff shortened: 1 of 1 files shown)


framework/src/Volo.Abp.MongoDB/Volo/Abp/Uow/MongoDB/UnitOfWorkMongoDbContextProvider.cs
```diff
@@ -184,7 +184,18 @@ private TMongoDbContext CreateDbContext(IUnitOfWork unitOfWork, MongoUrl mongoUr
                     session.AdvanceOperationTime(new BsonTimestamp(unitOfWork.Options.Timeout.Value));
                 }
 
-                session.StartTransaction();
+                try
+                {
+                    session.StartTransaction();
+                }
+                catch (Exception e)
```
hikalkan: Catch `NotSupportedException`, not all exceptions.

hikalkan (changes requested):

framework/src/Volo.Abp.MongoDB/Volo/Abp/Uow/MongoDB/UnitOfWorkMongoDbContextProvider.cs
maliming: https://github.com/abpframework/abp/pull/10658/commits/aa01142d8793bbea326626b6696b4a586060f23f

hikalkan (approved):

Issue #10642, opened 2021-11-17
[question]how can I use a abp-cli preview version,


xyfy:
I want to use some new option in the cli preview version

for example:
`abp generate-proxy -t csharp -url https://localhost:44302/`

xyfy:
I used it in a solution witch is using abp 5.0-rc1

xyfy:
I used `abp switch-to-preview`  then use `abp generate-proxy -t csharp -url https://localhost:44302/`

xyfy:
![image](https://user-images.githubusercontent.com/7838147/142157219-ad080dd7-2544-499a-b4cc-a129118d3f19.png)

xyfy:
any one can help me ?

EngincanV:
Can you update the CLI version to v5.0.0-rc.1 and re-try it?

```bash
dotnet tool update Volo.Abp.Cli -g --version 5.0.0-rc.1
```

xyfy:
Thanks ,it works !

PR #10659, opened 2021-11-18
Update CLI.md

xyfy:
Add a special version cli install  #10642


EngincanV:
Thanks for your contribution @xyfy, but if we specify the version in CLI documentation then we would need to change the document in every new version.

Issue #6954, opened 2021-01-01
EF Core Eager loading related entities with default generic repository does not work.


tsupakit:
Hi abp team, 
I was trying to follow the documentation of Entity Framework Core [Eager Loading / Load With Details](https://docs.abp.io/en/abp/latest/Entity-Framework-Core#eager-loading-load-with-details).
Unfortunately, it seems did not work at all.
My entities were designed and configured almost the same as in the [example](https://docs.abp.io/en/abp/latest/Entity-Framework-Core#loading-related-entities)
As it states in the documentation that

> GetAsync and FindAsync gets includeDetails with default value is true.

The result that I got from both methods was not include my related entity.
I even tried explicitly pass parameter `includeDetails: true` also did not work.
My workaround was to create a custom repository and use the Include() of entity framework.

ABP CLI : 4.0.2
Database Provider : EF Core

realLiangshiwei:
Do create a IncludeDetails extension method for the IQueryable<TEntity> for each aggregate root which has sub collections

See https://docs.abp.io/en/abp/latest/Best-Practices/Entity-Framework-Core-Integration#repository-implementation

tsupakit:
@realLiangshiwei Thanks for a suggestion.
Does this mean to use this feature of the default generic repository we also need to implement extension method explicitly?

realLiangshiwei:
Yes. All modules of ABP do this. https://github.com/abpframework/abp/blob/dev/modules/identity/src/Volo.Abp.Identity.EntityFrameworkCore/Volo/Abp/Identity/EntityFrameworkCore/IdentityDbContextModelBuilderExtensions.cs

PR #10710, opened 2021-11-22
Default implementation for WithDetails (EfCore)

puschie286:
provide a default implementation for WithDetails ( and all functions that have the 'includeDetails' parameter )
first (basic) implementation for automatic include of navigation properties for the EntityFramework.

would be great if someone can provide an implementation for MongoDB

related to: https://github.com/abpframework/abp/issues/6954


puschie286:
push

hikalkan:
Hi @puschie286 this may have performance side effect for existing applications. While the idea seems good, it is better to leave this to the developer.

puschie286:
@hikalkan well, you could still provide your own implementation if it effect performance.

I think it is very unusually to have a function and parameter that doesnt do anything by default ( which is not even mentioned in documentation )

PR #10713, opened 2021-11-23
Log errors and not using if the current database does not support transactions.

maliming:
From https://support.abp.io/QA/Questions/2176

eg: SQLite(`SqliteConnection does not support nested transactions`.).


framework/src/Volo.Abp.EntityFrameworkCore/Volo/Abp/Uow/EntityFrameworkCore/UnitOfWorkDbContextProvider.cs
```diff
@@ -170,19 +172,34 @@ private TDbContext CreateDbContextWithTransaction(IUnitOfWork unitOfWork)
...
+                    unitOfWork.AddTransactionApi(
+                        transactionApiKey,
+                        new EfCoreTransactionApi(
+                            dbtransaction,
+                            dbContext,
+                            _cancellationTokenProvider
+                        )
+                    );
+                }
+                catch (Exception e)
```
NecatiMeral:
```csharp
catch (Exception e) when (e is InvalidOperationException || e is NotSupportedException)
{
    Logger.LogError(TransactionsNotSupportedErrorMessage);
    Logger.LogException(e);
}
```

NecatiMeral: or maybe (to reduce code duplication)
```csharp
catch (Exception e) when (CatchAndLogTransactionNotSupportedException(e))
{
}

bool CatchAndLogTransactionNotSupportedException(Exception e)
{
  if (e is InvalidOperationException || e is NotSupportedException)
  {
      Logger.LogError(TransactionsNotSupportedErrorMessage);
      Logger.LogException(e);
  
      return true;
  }
  return false;
}
```

framework/src/Volo.Abp.EntityFrameworkCore/Volo/Abp/Uow/EntityFrameworkCore/UnitOfWorkDbContextProvider.cs
```diff
@@ -18,6 +18,8 @@ namespace Volo.Abp.Uow.EntityFrameworkCore
     public class UnitOfWorkDbContextProvider<TDbContext> : IDbContextProvider<TDbContext>
         where TDbContext : IEfCoreDbContext
     {
+        private const string TransactionsNotSupportedErrorMessage = "The current database does not support transactions, All operations will be performed in non-transactions, This may cause errors.";
```
hikalkan:
```suggestion
        private const string TransactionsNotSupportedErrorMessage = "Current database does not support transactions. Your database may remain in an inconsistent state in an error case.";
```

framework/src/Volo.Abp.EntityFrameworkCore/Volo/Abp/Uow/EntityFrameworkCore/UnitOfWorkDbContextProvider.cs
hikalkan: I agree with `when` usage.

hikalkan (approved):

PR #10737, opened 2021-11-25
Fix softdelete cause full update

puschie286:
directly call SetDeletionAuditProperties instead of mark all properties as modified.

Please check if this cause any side effects related to MultiTenant operations.


cotur (approved):

maliming:
hi
I will apply the changes to the **5.0** branch. Thanks @puschie286

maliming:
Hi

I will move this to 5.0 branch.

PR #10972, opened 2021-12-16
Update AbpApplicationConfigurationController.cs

karakule006:
In some projects, we may need to Customize AbpApplicationConfigurationController controllers. So it has to be virtual.


realLiangshiwei (approved):

realLiangshiwei:
Thanks @karakule006

karakule006:
Thanks.

PR #10981, opened 2021-12-17
Added EnsureExistsAsync for optimal checking

sh-erfan:
Checking for an entity existence is currently possible through FindAsync or GetAsync methods, which are not efficient enough when we do don't need the actual value.
Instead, we can use AnyAsync to generate faster and efficient check.
We don't need any true/false result. It is just a fast check functionality on repository.


framework/src/Volo.Abp.Ddd.Domain/Volo/Abp/Domain/Repositories/RepositoryExtensions.cs
```diff
@@ -47,6 +47,33 @@ public static class RepositoryExtensions
...
 
+
+    public static async Task EnsureExistsAsync<TEntity, TKey>(
+           this IRepository<TEntity, TKey> repository,
+           TKey id,
+           CancellationToken cancellationToken = default
+       )
+           where TEntity : class, IEntity<TKey>
+    {
+        if (await repository.AnyAsync(x => x.Id.Equals(id), cancellationToken))
```
rqx110: do you mean 
```suggestion
        if (!await repository.AnyAsync(x => x.Id.Equals(id), cancellationToken))
```
?
if, also need change below method

sh-erfan: Yes. I forgot that :)

maliming:
> we do don't need the actual value. But we need to know if it exists,

Can you share such a case?

sh-erfan:
> > we do don't need the actual value. But we need to know if it exists,
> 
> Can you share such a case?

```
public class GuardService: ITransientDependency
    {
        private readonly ICategoryRepository _categoryRepository;


        public GuardService(ICategoryRepository categoryRepository)
        {
            _categoryRepository = categoryRepository;
        }

        public Task EnsureCategoryExistsAsync(Guid id)
        {
            return _categoryRepository.EnsureExistsAsync(id);
        }
    }
```

Then in domain services, inject GuardService and use as follow:

```
public class TodoItemManager : DomainService
    {

        private readonly ITodoItemRepository _todoItemRepository;
        private readonly GuardService _guard;

        public TodoItemManager(ITodoItemRepository todoItemRepository, GuardService guard)
        {
            _todoItemRepository = todoItemRepository;
            _guard = guard;
        }

        public async Task CreateAsync(string name, Guid categoryId)
        {
            await _guard.EnsureCategoryExistsAsync(categoryId);
            Check.NotNullOrWhiteSpace(name, nameof(name));
            return new TodoItem(GuidGenerator.Create(), name, categoryId);
        }
    }
```

maliming:
I think you can create such an extension method in your project, it doesn't seem to be in common use in ABP.

sh-erfan:
> I think you can create such an extension method in your project, it doesn't seem to be in common use in ABP.

I agree. 
But don't you think the same about GetAsync(...) method, which can also be a project based extension? Because it is using FindAsync method. One might need to override this method in order to throw his/her own exception.
Example:

`await _bookRepo.GetAsync(x => x.AuthorName == "Jack")`

if no such book exists, the error is something like this: 
_"There is no such an entity given id. Entity type: {entityType.FullName}"_
A general message, even there is no check for **Id** in specified expression.

hikalkan:
I agree that this can be use cases. I request to additions to this PR:

- [x] Document the methods. I think we can add a small section under this: https://docs.abp.io/en/abp/latest/Repositories#soft-hard-delete
- [x] Write unit tests please.

maliming:
@sh-erfan 

I can do that if you don't have much time.

maliming (approved):

sh-erfan:
> @sh-erfan
> 
> I can do that if you don't have much time.

Thanks for finishing it

gdlcf88:
I think the better method names are:
* CheckExistAsync
* ShouldExistAsync
* MustExistAsync
* ExistOrThrowAsync

Since the name `EnsureExistsAsync` looks like "create the entity if it doesn't exist".

PR #10989, opened 2021-12-19
Fixed a l10n issue on confirm dialog

### LINKED ISSUE #10986: the Localization didn't work, just show default language
* Your **ABP Framework version**.
5.0.0
MVC
EF Core
1. use CLI to create a app like: abp new MyTodo -t app
2. Login with default admin account, change language to other language, eg: Chinese
3. go to Administration/Identity management/Roles, click the action/**Permissions**  in the list.
4. Then popup a window, **uncheck** the checkbox named **Grant all permissions**, but don't save, click Cancel button, then popup another window with question: Are you sure? 
5. The text in the button just show text in English as the screenshot P-1, not the chosen language. It should display the language I chosen. But in version 4.4.4, it just shows fine as the screenshot P-2.

P-1(v5.0.0)
image
P-2(v4.4.4)
image

6. By the way, the **margin** among lines seems too large somehow as the screenshot P-3, but in version 4.4.4, it's just perfect as the screenshot P-4, can make some adjustment?
P-3(v5.0.0)
image
P-4(v4.4.4)
image

iyilm4z:
Fixes #10986


realLiangshiwei:
We can fix it in 5.0, I will handle it, thanks.

PR #11014, opened 2021-12-21
When IWebHostEnvironment is not injected, hostGetHostingEnvironment will return an empty one instead of throwing exception

yekalkan:


framework/src/Volo.Abp.AspNetCore/Microsoft/Extensions/DependencyInjection/AbpAspNetCoreServiceCollectionExtensions.cs
```diff
@@ -1,11 +1,21 @@
...
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.Hosting;
 
 namespace Microsoft.Extensions.DependencyInjection
 {
     public static class AbpAspNetCoreServiceCollectionExtensions
     {
         public static IWebHostEnvironment GetHostingEnvironment(this IServiceCollection services)
         {
+            if (!services.Any(d => d.ServiceType == typeof(IWebHostEnvironment)))
```
hikalkan: This code searches the array 2 times (one in GetSingletonInstance). Use this instead:

`return services.GetSingletonInstanceOrNull<IWebHostEnvironment>() ?? new EmptyHostingEnvironment()...`;

hikalkan (approved):

PR #11112, opened 2021-12-29
Fixed sweetalert2 button l10n issue

JadynWong:
Hi @realLiangshiwei 

Other message types also display the button. It uses the default confirmButtonText 'Ok'. 
We should replace the localized text in the `default` configuration

https://sweetalert2.github.io/#confirmButtonText

image


realLiangshiwei:
Hi,

We have solved the problem: https://github.com/abpframework/abp/pull/10990

JadynWong:
Sorry rebase error. 

But #10990 is only work for `confirm` type.

realLiangshiwei:
You can right, we should change the `default` configuration, I will handle it. thanks

Issue #11111, opened 2021-12-29
AbpBlobStoring Throw exception System.ArgumentNullException:“Value cannot be null. Arg_ParamName_Name”


BrettSun:
AbpBlobStoring

when provider is not used Throw exception System.ArgumentNullException:“Value cannot be null. Arg_ParamName_Name”

More explicit exception information

"BLOB Container provider is not used. Example: Configure to use the File System storage provider by default"

maliming:
**BLOB Container provider is not used. Example: Configure to use the File System storage provider by default**

maliming:
Move to #11115

PR #11115, opened 2021-12-29
when provider is not used Throw exception Value cannot be null. Arg_ParamName_Name

BrettSun:
#11111 
AbpBlobStoring
when provider is not used Throw exception System.ArgumentNullException:“Value cannot be null. Arg_ParamName_Name”
More explicit exception information
"BLOB Container provider is not used. Example: Configure to use the File System storage provider by default"


realLiangshiwei:
Move to: https://github.com/abpframework/abp/pull/11118

Thanks a lot @BrettSun

PR #11136, opened 2021-12-30
Editing the Quick Start Docs

braim23:
Fixed some bits and small details. added articles such as "a", "an" and "the" when needed and deleting them when they weren't needed.

For example: There is one source code that we'll talk about and it is specific. So we'll say:
"You can find the* source code of the completed application".

Also, when we say:

"Go to the `blabla.cs` file"

'file' is a noun, and since we used a noun, we are supposed to add an article 'the'. Same goes if we say 'class', 'project', etc. instead of 'file'.

But when we say:
"Go to `blabla.cs`"

We didn't use a noun here, so we don't need to add 'the' at all, and this explains why I removed some "the"s.
image

### DIFF
--- a/docs/en/Tutorials/Todo/Index.md
+++ b/docs/en/Tutorials/Todo/Index.md
@@ -8,11 +8,11 @@
 }
 ````
 
-This is a single-part, quick-start tutorial to build a simple todo application with the ABP Framework. Here, a screenshot from the final application:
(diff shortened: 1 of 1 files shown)


hikalkan (approved):

hikalkan:
Thanks for the fixes and the explanation.

braim23:
> Thanks for the fixes and the explanation.

You're more than welcome :D can't wait to do even more

PR #11197, opened 2022-01-05
Update ConfigureVirtualFileSystem to better support VS debugging of a Docker Image

whytem:
The Blazor Server application template currently throws an exception and fails to start if a Visual Studio user attempts to add Docker Support to the Blazor project and then attempts to Debug from VS.

The cause is that Visual Studio mounts the source files at the "/src" folder within the Docker container by default while the default Dockerfile created by Visual Studio will run the application from the /app folder within the container. 

As the hostingEnvironment.IsDevelopment() variable will remain true in Debugging mode, the implicitly assumed paths in the ConfigureVirtualFileSystem method become incorrect and an unhandled DirectoryNotExists exception is thrown.

The below changes do not address the issue but allow a user to more quickly identify the cause of the exception, and correct the source path as required to address (or eliminate the code) as preferred.

Alternative approaches could be to amend the if statement to prevent execution of embedded replacement when debugging in a Docker container:

`if (hostingEnvironment.IsDevelopment() && Environment.GetEnvironmentVariable("DOTNET_RUNNING_IN_CONTAINER") != "true");`

However, this would cause file updates to not be applied until the docker container is rebuilt. Alternatively the new sourceDirectory string created could be updated to the default source mount location used by VS when running a debug session within a container if this could be made robust enough

```
if (Environment.GetEnvironmentVariable("DOTNET_RUNNING_IN_CONTAINER") == "true")
                    sourceDirectory = String.Format("{0}src{0}src{0}", Path.DirectorySeparatorChar);
```

### DIFF
--- a/templates/app/aspnet-core/src/MyCompanyName.MyProjectName.Blazor.Server/MyProjectNameBlazorModule.cs
+++ b/templates/app/aspnet-core/src/MyCompanyName.MyProjectName.Blazor.Server/MyProjectNameBlazorModule.cs
@@ -144,27 +144,31 @@ public class MyProjectNameBlazorModule : AbpModule
     {
         if (hostingEnvironment.IsDevelopment())
         {
+            //Update to path of where source folder is mounted if debugging in a Docker container
+            string sourceDirectory = Path.Combine(hostingEnvironment.ContentRootPath, $"..{ Path.DirectorySeparatorChar}");
+
             Configure<AbpVirtualFileSystemOptions>(options =>
             {
                 //<TEMPLATE-REMOVE>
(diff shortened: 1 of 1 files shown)


maliming:
> The cause is that Visual Studio mounts the source files at the "/src" folder within the Docker container by default while the default Dockerfile created by Visual Studio will run the application from the /app folder within the container.


Maybe you can consider changing this behavior.

whytem:
> > The cause is that Visual Studio mounts the source files at the "/src" folder within the Docker container by default while the default Dockerfile created by Visual Studio will run the application from the /app folder within the container.
> 
> Maybe you can consider changing this behavior.

Sure - it's possible to work around a number of different ways once you understand the issue. It's just not ideal that users encounter an undocumented exception on launch when running a "getting started" template which requires code or configuration modifications to address.

PR #11209, opened 2022-01-06
Publish events on UOW completed if the outbox is unavailable.

### LINKED ISSUE #11208: Distributed events unexpectedly published in advance if outbox is unavailable
https://github.com/abpframework/abp/commit/eae8d6340fe2a63cc6ccb78cf3675044d02edd02 fixed a similar problem of the local distributed event bus.

But for distributed event buses that implement `DistributedEventBusBase`, the problem is still there **if app developers do not configure any OutboxConfig**:
image

gdlcf88:
https://github.com/abpframework/abp/issues/11208

### DIFF
--- a/framework/src/Volo.Abp.EventBus/Volo/Abp/EventBus/Distributed/DistributedEventBusBase.cs
+++ b/framework/src/Volo.Abp.EventBus/Volo/Abp/EventBus/Distributed/DistributedEventBusBase.cs
@@ -73,6 +73,16 @@ public abstract class DistributedEventBusBase : EventBusBase, IDistributedEventB
             {
                 return;
             }
+
+            if (UnitOfWorkManager.Current != null)
+            {
+                UnitOfWorkManager.Current.OnCompleted(async () =>
+                {
+                    await PublishToEventBusAsync(eventType, eventData);
+                });
+
+                return;
+            }
         }
 
         await PublishToEventBusAsync(eventType, eventData);


framework/src/Volo.Abp.EventBus/Volo/Abp/EventBus/Distributed/DistributedEventBusBase.cs
```diff
@@ -73,6 +73,16 @@ public override Task PublishAsync(Type eventType, object eventData, bool onUnitO
...
+
+            if (UnitOfWorkManager.Current != null)
+            {
+                UnitOfWorkManager.Current.OnCompleted(async () =>
+                {
+                    await PublishToEventBusAsync(eventType, eventData);
+                });
+
+                return;
+            }
```
hikalkan, line 85: I have two questions for this implementation:

1. Why we are publishing on UOW complete if `onUnitOfWorkComplete` parameter is `false`? (if it was `true`, then it will be catched by the first if condition and return).
2. If we want to publish on UOW complete, why not using `AddToUnitOfWork` method like made above?

Sorry, I may forgot the details (actually, I opened the solution, but couldn't figure out it).

@maliming maybe you remember these.

gdlcf88: The `UnitOfWorkEventPublisher` publishes the "outbox using" distributed events on UOW complete. Please note that UOW is not really completed now:
```csharp
    public async Task PublishDistributedEventsAsync(IEnumerable<UnitOfWorkEventRecord> distributedEvents)
    {
        foreach (var distributedEvent in distributedEvents)
        {
            await _distributedEventBus.PublishAsync(
                distributedEvent.EventType,
                distributedEvent.EventData,
                onUnitOfWorkComplete: false,
                useOutbox: distributedEvent.UseOutbox
            );
        }
    }
```

But the `PublishAsync` method in DistributedEventBusBase:
```csharp
    public async Task PublishAsync(
        Type eventType,
        object eventData,
        bool onUnitOfWorkComplete = true,
        bool useOutbox = true)
    {
        if (onUnitOfWorkComplete && UnitOfWorkManager.Current != null)
        {
            AddToUnitOfWork(
                UnitOfWorkManager.Current,
                new UnitOfWorkEventRecord(eventType, eventData, EventOrderGenerator.GetNext(), useOutbox)
            );
            return;
        }

        if (useOutbox)
        {
            // App developers may not have configured an outboxConfig, so the `AddToOutboxAsync` returns false.
            if (await AddToOutboxAsync(eventType, eventData))
            {
                return;
            }
        }

        // Then the event is published immediately before the UOW invokes "SaveChangesAsync" and "CommitTransactionsAsync".
        await PublishToEventBusAsync(eventType, eventData);
    }
```

> Why we are publishing on UOW complete if onUnitOfWorkComplete parameter is false?

If you approve the changes in this PR:
1. We can fix the above problem.
2. For developers who use `onUnitOfWorkComplete = false` and `useOutbox = true` manually. If there is no outboxConfig, publishing on UOW complete is the last resort since it is better than (at least not worse than) publishing immediately. The former is more of a "simulated outbox" (just like ABP 4.0+ did).

maliming: We can output warring log if `useOutbox = true` and there is no `outboxConfig`.

gdlcf88: Module developers do not know whether an app has configured an outbox. Events publishing in modules will still face this problem if we only log the warning. Event handlers may get these events before the UOW completes. What if the UOW has rolled back?

maliming: We can consider adding a new parameter(`fallbackToUnitOfWorkComplete`).

hikalkan: > We can output warring log if `useOutbox = true` and there is no `outboxConfig`.

`useOutbox` is true by default and ignored if no outbox is configured. We don't expect the developer (that publish the event) sets this to `false` unless there is a very good reason. So, we don't force outbox is configured if `useOutbox` is true. Not a warning, but a debug/trace log can be written.

I will check the code again to understand why this PR is necessary.

gdlcf88: I provided another implementation: https://github.com/abpframework/abp/pull/12380

PR #11216, opened 2022-01-06
Handle the exception in `InboxProcessor` and `OutboxSender`.

### LINKED ISSUE #11210: Any handler throwing block the whole inbox queue
https://github.com/abpframework/abp/blob/9c3b34067524b852c9eaf497728409d6f5e08ec4/framework/src/Volo.Abp.EventBus/Volo/Abp/EventBus/Distributed/InboxProcessor.cs#L93-L118

According to the above design, any event handler throws will block the whole queue:

d0fea7de23ec241a506178d606dfb9a

This is not reliable for the production environment.

maliming:
Resolve #11210

### DIFF
--- a/framework/src/Volo.Abp.EventBus/Volo/Abp/EventBus/Distributed/InboxProcessor.cs
+++ b/framework/src/Volo.Abp.EventBus/Volo/Abp/EventBus/Distributed/InboxProcessor.cs
@@ -102,18 +102,26 @@ namespace Volo.Abp.EventBus.Boxes
 
                         foreach (var waitingEvent in waitingEvents)
                         {
-                            using (var uow = UnitOfWorkManager.Begin(isTransactional: true, requiresNew: true))
+                            try
                             {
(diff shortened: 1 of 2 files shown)


hikalkan:
It is not acceptable to skip an event and continue to publish/consume the next one. Event publish/consume order can be important in some cases. A true solution can be providing a failure policy, by designing a system that the final developer decided what to do on that case (a few times retry, re-queue event with a delay, ignore / drop message, etc. which can depend on the event data).

So, currently,

* if developers want to ignore errors on event handler, they can easily use try-catch, log and ignore the error.
* if they want to requeue, they can catch the error and republish the message with a custom data attached to count the error count for example
* outbox failures should be related to eventbus (e.g. rabbitmq) connections, and waiting and retrying should be enough, instead of ignoring errors and going to the next event.

If you want to think on that and find a complete solution, you're welcome. Otherwise, please keep the issue #11210 open, so we think on it later.

PR #11232, opened 2022-01-07
Enhancement Trigger Event Handler

### LINKED ISSUE #11030: [Enhancement] Replace EventBus Reflection with compiled Expression
EventBus module used Reflection to [trigger event-handlers](https://github.com/abpframework/abp/blob/ab6eedd52a662d430615ed7c80b12aa976594f55/framework/src/Volo.Abp.EventBus/Volo/Abp/EventBus/EventBusBase.cs#L196)

In an event-driven system a single user request can cause multiple event invocation and each event can have multiple handler so it can lead to a considerable number of reflection calls per request which can impact performance of application.

I just wanted to know can I work on replacing event-bus reflections codes with compiled expression?

realLiangshiwei:
Resolve https://github.com/abpframework/abp/issues/11030

```csharp
public class EmailSendEventData
{
}

public class EmailSendEventHandler : IDistributedEventHandler<EmailSendEventData>, ITransientDependency
{
    public Task HandleEventAsync(EmailSendEventData eventData)
    {
        return Task.CompletedTask;
    }
}

for (var i = 0; i < 10000; i++)
{
    await bus.PublishAsync(new EmailSendEventData());
}
```
Before 65ms
After 54ms

### DIFF
--- a/framework/src/Volo.Abp.EventBus/Volo/Abp/EventBus/EventHandlerInvoker.cs
+++ b/framework/src/Volo.Abp.EventBus/Volo/Abp/EventBus/EventHandlerInvoker.cs
@@ -8,33 +8,43 @@ namespace Volo.Abp.EventBus;
 
 public class EventHandlerInvoker : IEventHandlerInvoker, ISingletonDependency
 {
-    private readonly ConcurrentDictionary<string, IEventHandlerMethodExecutor> _cache;
+    private readonly ConcurrentDictionary<string, EventHandlerInvokerCacheItem> _cache;
 
     public EventHandlerInvoker()
     {
-        _cache = new ConcurrentDictionary<string, IEventHandlerMethodExecutor>();
+        _cache = new ConcurrentDictionary<string, EventHandlerInvokerCacheItem>();
     }
 
     public async Task InvokeAsync(IEventHandler eventHandler, object eventData, Type eventType)
     {
-        var notAnEventHandler = true;
-        if (typeof(ILocalEventHandler<>).MakeGenericType(eventType).IsInstanceOfType(eventHandler))
+        var cacheItem = _cache.GetOrAdd($"{eventHandler.GetType().FullName}-{eventType.FullName}", _ =>
         {
-            var eventHandlerCall = _cache.GetOrAdd($"{typeof(LocalEventHandlerMethodExecutor<>).FullName}{eventHandler.GetType().FullName}-{eventType.FullName}",
-                (_) => (IEventHandlerMethodExecutor)Activator.CreateInstance(typeof(LocalEventHandlerMethodExecutor<>).MakeGenericType(eventType)));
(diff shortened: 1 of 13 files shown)


framework/src/Volo.Abp.EventBus/Volo/Abp/EventBus/EventHandlerInvoker.cs
```diff
@@ -9,18 +10,26 @@ namespace Volo.Abp.EventBus;
...
+
+        var executor = _executorCache.GetOrAdd(key, _ => EventHandlerMethodExecutor.Create(GetHandleEventMethodInfo(handleType, eventType), handleType.GetTypeInfo()));
 
         return executor.ExecuteAsync(eventHandler, new[] { eventData });
     }
+
+    private static MethodInfo GetHandleEventMethodInfo(Type handleType, Type eventType)
+    {
+        var methods = handleType.GetMethods().Where(x => x.Name == EventHandlerMethodName).ToArray();
+        return methods.Length == 1 ? methods.First() : methods.FirstOrDefault(x => x.GetParameters().Any(param => param.ParameterType == eventType));
```
zeinali0: I'm not sure but I think in inheritance this could select wrong method to execute.
Like this use-case:
```
public class BaseClass{}
public class Foo : BaseClass 
{
}
public class Bar: BaseClass
{
}

public class FooBarEventHandler : ILocalEventHandler<BaseClass>
{
    public Task HandleEventAsync(BaseClass eventData)
    {
        Dosomthing();
        if(eventData is Foo foo)
        {
             HandleEventAsync(foo);
        }
    }

    private Task HandleEventAsync(Foo eventData)
    {
         return Task.CompletedTask;
    }
}
```
This would directly execute `HandleEventAsync(Foo eventData)` if the `eventType` be `Foo`  while it should execute ` HandleEventAsync(BaseClass eventData)`

realLiangshiwei: Hi, 
Did you try to test it?

Use `handlerFactories.EventType` instead of the `parameter type` when triggering handlers, it will work as expected

https://github.com/abpframework/abp/blob/e3e1779de6df5d26f01cdc8e99ac9cbcb3d24d3c/framework/src/Volo.Abp.EventBus/Volo/Abp/EventBus/EventBusBase.cs#L139

You can check this unit test

https://github.com/abpframework/abp/blob/e3e1779de6df5d26f01cdc8e99ac9cbcb3d24d3c/framework/test/Volo.Abp.EventBus.Tests/Volo/Abp/EventBus/Local/EventBus_MultipleHandle_Test.cs#L12

zeinali0: Sorry my mistake, I was triggering with `eventType` instead of `handlerFactories.EventType`

maliming (approved):

PR #11243, opened 2022-01-09
Batch publish events from outbox to the event bus

### LINKED ISSUE #10035: Optimization: Batch publish events from outbox to the event bus
We can implement this for those supports batch message publishing. It would be much more performant.

When I searched for it, I see that RabbitMQ has a batch publish operation. However, I saw this document: https://www.rabbitmq.com/tutorials/tutorial-seven-dotnet.html As I understand, publishing messages to RabbitMQ is not so safe! It seems we should wait for confirmation in order to delete event from outbox. Can you investigate for that?

realLiangshiwei:
Resolve https://github.com/abpframework/abp/issues/10035

```csharp
Configure<AbpEventBusBoxesOptions>(options =>
{
    options.OutboxPublishInBatch = true;
});
```

Batch publishing 1000 events

#### RabbitMQ  
Before: 50080ms
After: 547ms

#### Kafka
Before: 15303ms 
After: 475ms

#### Azure (Remote Service)
Before: 116157ms
After: 1534ms

#### Rebus(RabbitMQ transport)
Before: 26094ms 
After: 560ms


hikalkan:
Thanks @realLiangshiwei I will think on that in the next milestone.

hikalkan:
@realLiangshiwei can you please resolve the conflict.

framework/src/Volo.Abp.EventBus/Volo/Abp/EventBus/Distributed/AbpEventBusBoxesOptions.cs
```diff
@@ -34,6 +34,11 @@ public class AbpEventBusBoxesOptions
     /// Default: 2 hours
     /// </summary>
     public TimeSpan WaitTimeToDeleteProcessedInboxEvents { get; set; }
+    
+    /// <summary>
+    /// Default: false
+    /// </summary>
+    public bool OutboxPublishInBatch { get; set; }
```
hikalkan: Rename to `BatchPublishOutboxEvents` and make it `true` by default.

hikalkan (changes requested):

hikalkan (approved):

hikalkan:
Thanks!

PR #11268, opened 2022-01-11
Angular UI: Escaped HTML element chars for the extensible table

mehmet-erim:


npm/ng-packs/packages/core/src/lib/utils/string-utils.ts
```diff
@@ -21,3 +21,7 @@ export function interpolate(text: string, params: string[]) {
     .replace(/(['"]?\{\s*(\d+)\s*\}['"]?)/g, (_, match, digit) => params[digit] ?? match)
     .replace(/\s+/g, ' ');
 }
+
+export function escapeHtmlChars(value: string) {
+  return value && value.replace(/</g, '&lt;').replace(/>/g, '&gt;');
+}
```
maliming: https://github.com/abpframework/abp/blob/dev/npm/packs/core/src/abp.js#L691

mehmet-erim: Thanks!

bnymncoskuner (approved):

PR #11371, opened 2022-01-19
Remove `EnableLegacyTimestampBehavior`.

### LINKED ISSUE #11357: Postgres timestamp with time zone after updating to the 5.1.1 packages
Hello,

After updating my project to the latest packages and creating a new migration, all the datetime columns in my PostGres database are converted to "timestamp with time zone" (previously they were "timestamp without time zone").

I assume this is because the new npgsql 6.0 is used, but I see no  mentioning in the release notes about this breaking change?

Geert Veenstra

maliming:
**This option should be controlled by the developer.** See https://github.com/abpframework/eShopOnAbp/pull/65

Resolve #11357


realLiangshiwei (approved):

JadynWong:
Is it better to post a break change announcement？🤔

Or add to Abp-5.0 Migration-Guides.

---

Done

https://github.com/abpframework/abp/pull/11438

PR #11402, opened 2022-01-21
Fix permission management model problem for app-nolayers template

EngincanV:
app-nolayers


EngincanV:
Moved https://github.com/abpframework/abp/pull/11403

PR #11417, opened 2022-01-23
Implemented `IsActive` flag on `IUserData`

### LINKED ISSUE #11416: `IUserData` - Supporting new `IsActive` property
Hey folks,

I've been messing around with the lastest release. 
I very like the `IsActive` property but modules cannot make use of the `IsActive` property yet since `IUserData` doesn't provide this value (yet).

Is this by design, so you can tell a alternative way of getting this information in modules or can it be added from your perspective?

I'm open for implementing and contributing this change if desired.

* Your **ABP Framework version**: 5.1.2
* ~Your **User Interface** type (Angular/MVC/React... etc.) if the issue is related to a specific UI~
* ~Your database provider(EF Core/MongoDB)~
* ~Exception message and **stack trace** if available (check the logs).~
* ~Steps needed to **reproduce** the problem.~

NecatiMeral:
Fixes #11416

### WARNING
There are some migrations (in the CmsKit-module) which have dangerous changes. I tried to locate the root cause and I cannot find the underlying entities whose tables will get dropped.


hikalkan:
Thanks @NecatiMeral for your contribution. I will review and think about that.

modules/users/src/Volo.Abp.Users.EntityFrameworkCore/Volo/Abp/Users/EntityFrameworkCore/AbpUsersDbContextModelCreatingExtensions.cs
```diff
@@ -16,5 +16,6 @@ public static void ConfigureAbpUser<TUser>(this EntityTypeBuilder<TUser> b)
         b.Property(u => u.EmailConfirmed).HasDefaultValue(false).HasColumnName(nameof(IUser.EmailConfirmed));
         b.Property(u => u.PhoneNumber).HasMaxLength(AbpUserConsts.MaxPhoneNumberLength).HasColumnName(nameof(IUser.PhoneNumber));
         b.Property(u => u.PhoneNumberConfirmed).HasDefaultValue(false).HasColumnName(nameof(IUser.PhoneNumberConfirmed));
+        b.Property(u => u.IsActive).HasDefaultValue(true).HasColumnName(nameof(IUser.IsActive));
```
maliming: This will cause a problem.
https://github.com/dotnet/efcore/issues/15323#issuecomment-482093707

maliming (approved):

PR #11446, opened 2022-01-25
Fixes #11445 , bug when generating links for tenant subdomains, the u…

### LINKED ISSUE #11445: bug in reset password link when tenant resolving with subdomains
Abp version 5.0.1
UI:  Angular 
DB provider: EF Core

The Problem:
when using tenant subdomain resolving, and trying to reset the user password on some tenant 
the generated link becomes  tenantlocalhost:44350 instead of tenant.localhost:44350, ie missing the dot 
image

I believe it's caused by line 82 in AppUrlProvider, and then replacing the string with the dot at line 89, which should only be used in case of host tenant not in all cases.

https://github.com/abpframework/abp/blob/89559ae2dcaadb28f684478af9a2843184165a26/framework/src/Volo.Abp.UI.Navigation/Volo/Abp/Ui/Navigation/Urls/AppUrlProvider.cs#L82

aozien:
#11445 fix, now we're removing the dot in host tenant's case only.


maliming:
hi
Move to https://github.com/abpframework/abp/pull/11447
Thanks

PR #11460, opened 2022-01-26
Fix PermissionManagementModal style

JadynWong:
image

https://getbootstrap.com/docs/5.1/utilities/spacing/#notation

Bootstrap5 does not have a class for `pl-*`, it should be `ps-*`.


maliming:
Move to 5.1. https://github.com/abpframework/abp/pull/11462

PR #11551, opened 2022-02-10
Compatible with API Versioning.

### LINKED ISSUE #497: API Versioning test, finalize & document
We should check and understand ABP's current API versioning system (I had built a system based on appservices), think the missing parts if there are. We want to document the current status. Also, think on different scenarios (like if we don't want to use appservices and/or we write API controllers, how we should go) including client scenarios (javascript and c# clients with and without abp proxies). We should prepare a complete guide for developers who want to implement API versioning in their solutions.

maliming:
Resolve #497 

https://github.com/abpframework/abp/blob/dev/docs/en/API/API-Versioning.md

image

image


realLiangshiwei (approved):

hikalkan (approved):

hikalkan:
Thanks!

PR #11693, opened 2022-02-24
Modified saving problems of users, organizations, and roles

CaiJinHao:
When the role exists under the organization. It will always load all characters. This modification fixes this problem


maliming:
**Test failed for the solution: ../modules/identity**

Issue #11836, opened 2022-03-08
Document the distributed lock


hikalkan:
The document should explain the following topics:

* What is distributed locking, why we may need it, in brief.
* ABP's distributed locking abstraction package and its usage.
* Explanation and usage of the current implementation (https://github.com/abpframework/abp/tree/dev/framework/src/Volo.Abp.DistributedLocking), how to use redis provider for this implementation, give link to its original documents for more

Note: This is a shipped feature, so document it in `rel-5.2` branch please.

malikmasis:
@hikalkan made the last changes and completed it.

PR #11867, opened 2022-03-09
Fixes 11836 documentation the distributed lock

malikmasis:
closes abpframework/abp#11836


docs/en/Distributed-Lock.md
```diff
@@ -0,0 +1,72 @@
...
+````
+
+Also, you should add your Redis configuration in appsetting.json
+````json
+"Redis": {
+    "Configuration": "127.0.0.1"
+}
+````
+
+> To use in APB, you should download the below NuGet package.
```
EngincanV: I don't think this and the below lines for installing the package are necessary. Maybe we can create an **Installation** section as in [Redis-Cache documentation](https://docs.abp.io/en/abp/latest/Redis-Cache#installation) and add the following ABP CLI command to the section as an installation command.

```bash
abp add-package Volo.Abp.DistributedLocking
```

docs/en/Distributed-Lock.md
```diff
@@ -0,0 +1,72 @@
...
+````
+
+````csharp
+using Volo.Abp.DistributedLocking; 
+namespace AbpDemo
+{
+    public class MyService : ITransientDependency
+    {
+        private readonly IAbpDistributedLock _distributedLock;
+		public MyService(IAbpDistributedLock distributedLock)
```
EngincanV: Can you remove the spaces in here?

docs/en/Distributed-Lock.md
```diff
@@ -0,0 +1,72 @@
...
+        {
+            _distributedLock = distributedLock;
+        }
+
+        await using (var handle = await _distributedLock.TryAcquireAsync("NameOfLock"))
```
EngincanV: It can be better to write this code inside of a method to prevent possible confusion.

docs/en/Distributed-Lock.md
```diff
@@ -0,0 +1,72 @@
+# Distributed Lock
+Distributed locks are very useful to manage many processes that request the same object.
```
EngincanV: It can be better to explain more "What is Distributed Locking and which problems that it solves?".

malikmasis:
Thanks for your advice @EngincanV 
I've made changes according to these bits of advice.

docs/en/Distributed-Lock.md
```diff
@@ -0,0 +1,81 @@
...
+On another hand, it will protect from more times triggered hence you will provide more efficiency.
+To summarize, It's used to handle conflict between these requests. Once obtaining anyone it, the others need to wait till give up free.
+
+# Providers
+Distributed locks system provides an abstraction that can be implemented by any vendor/provider.
+
+ * `MedallionAbpDistributedLock`
+ABP depends on [DistributedLock.Core](https://www.nuget.org/packages/DistributedLock.Core) library which provides a distributed locking system for concurrency control in a distributed environment. There are [many distributed lock providers](https://github.com/madelson/DistributedLock#implementations) including Redis, SqlServer and ZooKeeper. You may use the one you want. Here, I will show the Redis provider.
+
+This provider contains a method named TryAcquireAsync and this method returns null if the lock could not be handled.
```
rqx110:
```suggestion
This provider contains a method named `TryAcquireAsync` and this method returns null if the lock could not be handled.
```

docs/en/Distributed-Lock.md
```diff
@@ -0,0 +1,81 @@
...
+> Name is a mandatory field. It keeps the locked provider name.
```
rqx110:
```suggestion
> `Name` is a mandatory field. It keeps the locked provider name.
```

docs/en/Distributed-Lock.md
```diff
@@ -0,0 +1,81 @@
...
+> Timeout is set as default. If it fells deadlock and doesn't work properly you can use define the time to kill it.
```
rqx110:
```suggestion
> `Timeout` is set as default. If it fells deadlock and doesn't work properly you can use define the time to kill it.
```

docs/en/Distributed-Lock.md
```diff
@@ -0,0 +1,81 @@
...
+> CancellationToken is set as default. It enables cooperative cancellation between threads, thread pool work items, or Task objects
```
rqx110:
```suggestion
> `CancellationToken` is set as default. It enables cooperative cancellation between threads, thread pool work items, or Task objects
```

malikmasis:
Thanks @rqx110,
I've applied your advice.
FYI.

hikalkan (approved):

Issue #10208, opened 2021-10-04
CmsKit - Editor Implementation for Page Style & Scripts


enisn:
[Ace editor](https://github.com/ajaxorg/ace) semms good for editing css & javascript.

--- 

![image](https://user-images.githubusercontent.com/23705418/135806430-656e7277-f776-44d7-9007-76d1bec8dab4.png)

PR #11885, opened 2022-03-10
feat(cms-kit): using codemirror as global resources editor

rqx110:
In this PR:
* using codemirror as global resources editor in cms-kit module:
image
image

* also update the `abp/codemirror` package: copy `mode`, `theme`, `addon` folder to output

### DIFF
--- a/modules/cms-kit/src/Volo.CmsKit.Admin.Web/Pages/CmsKit/GlobalResources/Index.cshtml
+++ b/modules/cms-kit/src/Volo.CmsKit.Admin.Web/Pages/CmsKit/GlobalResources/Index.cshtml
@@ -2,6 +2,7 @@
 
 @using Microsoft.AspNetCore.Mvc.Localization
 @using Volo.Abp.AspNetCore.Mvc.UI.Layout
+@using Volo.Abp.AspNetCore.Mvc.UI.Packages.Codemirror
 @using Volo.CmsKit.Admin.Web.Pages.CmsKit.GlobalResources
 @using Volo.CmsKit.Admin.Web.Menus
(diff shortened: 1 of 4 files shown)


enisn:
I've proposed ace editor before page-specific script & styles: https://github.com/abpframework/abp/issues/10208

modules/cms-kit/src/Volo.CmsKit.Admin.Web/Pages/CmsKit/GlobalResources/Index.cshtml
```diff
@@ -17,8 +18,15 @@
...
 
+@section styles{
+    <abp-style type="typeof(CodemirrorStyleContributor)" />
+}
+
 @section scripts {
     <abp-script-bundle>
+        <abp-script type="typeof(CodemirrorScriptContributor )"/>
+        <abp-script src="/libs/codemirror/mode/css/css.js"/>
+        <abp-script src="/libs/codemirror/mode/javascript/javascript.js"/>
```
enisn, line 29: Why they're added here manually instead of adding to ScriptContributor ? @rqx110

rqx110: assume any people who wants to use `codemirror`(reference by `ScriptContributor`) in other place, but they don't need 'js' or 'css' mode.   Maybe they just want 'csharp' or 'java'. If we add all "mode"s to `ScriptContributor`,  it will cause final bundle size lager.  So leave this option(add manually) to developer is better.

enisn (approved):

PR #11893, opened 2022-03-10
Manage abandon message when exception occured

GerardoGrecoPorini:
We have noticed that if a receiver application is unable to process the message for some reason, the message is not sent back to the queue and, after the max number of retries, the message set to completed.
We tried to set the property `AutoCompleteMessages ` to false but in this case the message is never set as completed.

With this pull request we should have fix this behavior and causes Service Bus to unlock the message within the queue and make it available to be received again. After max number of retries the message will be moved to the Dead Letter Queue.


gdlcf88:
https://github.com/abpframework/abp/issues/23412

PR #11901, opened 2022-03-11
Add option to disable service bus

wocar:
The main reason for these changes is integration testing

### DIFF
--- a/framework/src/Volo.Abp.EventBus.Azure/Volo/Abp/EventBus/Azure/AbpAzureEventBusOptions.cs
+++ b/framework/src/Volo.Abp.EventBus.Azure/Volo/Abp/EventBus/Azure/AbpAzureEventBusOptions.cs
@@ -7,4 +7,6 @@ public class AbpAzureEventBusOptions
     public string SubscriberName { get; set; }
 
     public string TopicName { get; set; }
+
+    public bool IsServiceBusDisabled { get; set; }
 }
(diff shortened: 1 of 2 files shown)


maliming:
Thanks @wocar

maliming (approved):

maliming:
hi @wocar 

Are you interested in abstracting this disable/enable option?

So other providers (RabbitMQ, Kafka) will also support this feature

rqx110:
how about rename `IsServiceBusDisabled` to `Disabled` or `IsDisabled`? @maliming

maliming:
> how about rename `IsServiceBusDisabled` to `Disabled` or `IsDisabled`? @maliming

https://github.com/abpframework/abp/blob/dev/framework/src/Volo.Abp.BackgroundJobs.Abstractions/Volo/Abp/BackgroundJobs/AbpBackgroundJobOptions.cs#L15

wocar:
Yeah I followed that naming convention.

wocar:
So I'm not quite sure where should we abstract this?

My problem is that when im using WebApplicationFactory to do integration testing, the Azure service actually gets initialized. 

Maybe it is just better to add a property to each `Options` class in every module?

wocar:
@maliming

maliming (approved):

maliming:
ok, Merged. Thanks again.

Issue #11950, opened 2022-03-14
Environment interface that is in @abp/core could has custom properties.


mahmut-gundogdu:
The environment interface that is in @abp/core could have custom properties. 

This is the Environment.ts
![image](https://user-images.githubusercontent.com/2217899/158184826-459c803e-5fc4-45cf-aa1b-572d3c56ab98.png)

For now, the developer could implement changes like that.
![image](https://user-images.githubusercontent.com/2217899/158186026-b65f75d3-3146-45b6-8c48-4346438d9bb3.png)

I think we can add extendible things to the Environment Interface.

PR #11951, opened 2022-03-14
feat: user can set a variable on the environment without extending

mahmut-gundogdu:
suggestion to  #11950

### DIFF
--- a/npm/ng-packs/packages/core/src/lib/models/environment.ts
+++ b/npm/ng-packs/packages/core/src/lib/models/environment.ts
@@ -10,6 +10,9 @@ export interface Environment {
   oAuthConfig: AuthConfig;
   production: boolean;
   remoteEnv?: RemoteEnv;
+  customSettings?:{ 
+    [key:string]: any;
+  }
 }
 
 export interface ApplicationInfo {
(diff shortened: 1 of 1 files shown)


rqx110:
why not just:
```ts
[key:string]: any;
```
instead of
```ts
customSettings?:{ 
    [key:string]: any;
  }
```

mahmut-gundogdu:
> why not just:
> 
> ```ts
> [key:string]: any;
> ```
> 
> instead of
> 
> ```ts
> customSettings?:{ 
>     [key:string]: any;
>   }
> ```

because I thought grouping the custom settings was a "good idea." but yes, we can add root

rqx110:
in my opnion, with `[key:string]: any;`, the end developer can define their favorite key(group) name. leave options to developer is better.

mahmut-gundogdu:
> in my opnion, with `[key:string]: any;`, the end developer can define their favorite key(group) name. leave options to developer is better.

Yes, you are right. It is a better idea. I am converting. Thank you for your feedback.

mahmut-gundogdu:
> in my opnion, with `[key:string]: any;`, the end developer can define their favorite key(group) name. leave options to developer is better.

What do you think "any vs unknown" in this case?

`[key: string]: unknown;
`
or 
`[key: string]: any;
`
I thought any is  (a little bit) 'unsafe'

rqx110:
> What do you think "any vs unknown" in this case?

sorry, I have no idea on this

muhammedaltug (approved):

PR #11983, opened 2022-03-16
Audit-Logging event enhancements

### LINKED ISSUE #11982: Audit-Logging event enhancements
In distributed system cases where Audit-Logging needs to be written by using events instead of direct database connection, there is a huge overhaul about mapping AuditLogInfo to an Eto. We can simplify this process.

gterdem:
`AuditLogOptions` now has _PublishEvent_ option defaults to `true` that publishes an integration event (eto). 

This enhacement is done for the distributed system use cases when an application (or microservice) needs to write the audit-logs using events.

Closes #11982


framework/src/Volo.Abp.Auditing/Volo/Abp/Auditing/AuditLogActionInfo.cs
```diff
@@ -17,7 +16,7 @@ public class AuditLogActionInfo : IHasExtraProperties
 
     public int ExecutionDuration { get; set; }
 
-    public ExtraPropertyDictionary ExtraProperties { get; }
+    public ExtraPropertyDictionary ExtraProperties { get; set; }
```
enisn, line 19: We generally use ExtraPropertyDictionary without a setter. That prevents replacing existing values by mistake, developers will be only able to modify the existing collection instead of setting the entire collection with a new one

enisn: This should be done like this:
```csharp
entity.ExtraProperties =
{
    { "Title" , "hello-world" }
}
```

instead of this:
```csharp
entity.ExtraProperties = new ExtraPropertyDictionary
{
    { "Title" , "hello-world" }
}
```

gterdem:
We decided **not to merge** this PR. The main reason is the **design flaw**. 
- Event publishing is an option. This doesn't prevent writing to the database. The correct approach should be to prevent database saving and only publish an event so the event handler can save it to the related database.
- Secondly, event publishing occurs in the `AuditLogging.Domain` layer. That will require all the publisher services depending on the domain layer of the audit logging. This can be prevented by adding a new library something like `Audit.Logging.EventPublisher` which doesn't seem to be necessary when the end-users can easily implement their own `IAuditingStore`.

We have written [Extracting a Module as a Microservice guide](https://docs.abp.io/en/commercial/latest/startup-templates/microservice/extracting-module-as-microservice) using the AuditLogging module as a sample.

PR #11991, opened 2022-03-17
Add configuration for ServiceBusAdministrationClient

GerardoGrecoPorini:
We have noticed that the configuration of the object ServiceBusAdministrationClient was missing. With this pull request we should have fix the issue.


maliming (approved):

maliming:
Thanks @GerardoGrecoPorini

PR #12084, opened 2022-03-25
Add OpenIddict module.

maliming:
https://github.com/abpframework/abp/blob/dev/docs/en/Modules/OpenIddict.md

**I can merge these commits into one to beautify the commit.**


modules/openiddict/src/Volo.Abp.OpenIddict.Domain.Shared/Volo/Abp/OpenIddict/Localization/OpenIddict/zh-Hans.json
```diff
@@ -0,0 +1,27 @@
+{
+  "culture": "zh-Hans",
+  "texts": {
+    "UserLockedOut": "登录失败,用户账户已被锁定.请稍后再试.",
+    "InvalidUserNameOrPassword": "用户名或密码错误!",
+    "LoginIsNotAllowed": "无法登录!妳的账号未激活或者需要验证邮箱地址/手机号.",
```
rqx110:
```suggestion
    "LoginIsNotAllowed": "无法登录!你的账号未激活或者需要验证邮箱地址/手机号.",
```

modules/openiddict/src/Volo.Abp.OpenIddict.Domain.Shared/Volo/Abp/OpenIddict/Localization/OpenIddict/zh-Hans.json
```diff
@@ -0,0 +1,27 @@
...
+    "InvalidUsername": "用户名或密码错误!",
+    "InvalidAuthenticatorCode": "验证码无效!",
+    "TheTargetUserIsNotLinkedToYou": "目标用户未和妳有关联!",
```
rqx110:
```suggestion
    "TheTargetUserIsNotLinkedToYou": "目标用户未和你有关联!",
```

hikalkan:
Sorry, it was impossible to make a review on GitHub for a PR with 700+ files changed :)

So, I made my comments and changes as a commit with some comments: https://github.com/abpframework/abp/pull/12084/commits/cbdf333e66e7f8b9ceab43c8f9144800252aa5ee
Please review them, take the necessary action if needed, then remove my comments.

maliming:
```
/* TODO: Should we use distributed locking here?
 * Because, multiple instances of the application may work in parallel, and only one
 * of them should work in a time. If you can't obtain the lock, it is good to wait for a longer time.
 * I suggest checking https://github.com/abpframework/abp/blob/dev/framework/src/Volo.Abp.EventBus/Volo/Abp/EventBus/Distributed/InboxProcessor.cs
 * for a good implementation, then apply a similar principle here.
 */
```

The [BackgroundJobWorker](
https://github.com/abpframework/abp/blob/e3e1779de6df5d26f01cdc8e99ac9cbcb3d24d3c/framework/src/Volo.Abp.BackgroundJobs/Volo/Abp/BackgroundJobs/BackgroundJobWorker.cs#L42) has implemented similar feature.

maliming:
The cache feature has been implemented.

maliming:
**DisableAccessTokenEncryption** has been removed, I updated the documentation to explain it.

https://github.com/abpframework/abp/blob/Module-OpenIddict/docs/en/Modules/OpenIddict.md#token-encryption

kevinchalet:
👋🏻 

@hikalkan sent me the link to this PR so I took a (very) brief look 😄 

Three things worth noting:
  - For token validation, the OpenIddict validation handler is the recommended option, as it uses a more recent JWT stack and comes with additional security checks that are not implemented by the MSFT JWT bearer handler (e.g token type validation, to ensure identity tokens are not treated as valid access tokens).

  - <del>I see you're trying to localize error messages eventually returned by OpenIddict: I did that in OpenIddict 3.0 beta4 (we even supported Turkish!) and it didn't end well for the reasons explained in this post: https://kevinchalet.com/2020/11/17/introducing-openiddict-3-0-s-first-release-candidate-version/ </del>

  - In `OpenIddictSupportedLoginModel.cs`, you use a dangerous approach that consists in faking an "OpenIddict request" and injecting it in the OpenIddict context to force it to consider the HTTP request as an authorization request. It's not 100% clear to me what this class does but I suspect it's basically a login page that needs to be aware of certain OIDC parameters initially sent as part of the authorization request. If so, you may want to consider a different approach: in your authorization controller, when calling `Challenge("scheme of your cookie handler")`, add the parameters you need to flow from the authorization endpoint to the login page to `AuthenticationProperties` and customize the `RedirectToLogin` event to add them to the query string: https://github.com/dotnet/aspnetcore/blob/main/src/Security/Authentication/Cookies/src/CookieAuthenticationHandler.cs#L474. Then, you'll be able to extract them without even using any OpenIddict type in your login stuff.

Cheers.

kevinchalet:
In your most recent commit, I see you implemented wildcard support for (post_logout_)redirect_uris. As it goes against the OIDC specification - that requires using simple string comparisons - you need to be extremely careful when implementing that. I see two potential issues:

  - I don't see `context.Reject()` ever called in your handlers so I don't think invalid redirect_uris will be blocked (I suspect your handlers will basically no-op).
  - When using wildcard support, it seems a single domain is considered valid for all clients.

It's important to get this part right as these two issues can basically ruin the entire security of your authorization server, as it allows stealing authorization codes/access tokens by redirecting the user agent to an untrusted domain (or to the trusted redirect_uri of a different client).

maliming:
Thank you so much @kevinchalet, You made a great project.

I will make some changes based on your review, thank you for following ABP.

hikalkan (approved):

beriniwlew:
Amazing work, keep it up!

PR #12114, opened 2022-03-28
Update Docs for Documentation GitHub

simix-alessandrokantousian:
## Task

Fixes: #12114 

## Update

- Documentation Docs Module: This branch added an update regarding the MongoDB part on how to configure the project part and add the collection with the data.

### DIFF
--- a/docs/en/Modules/Docs.md
+++ b/docs/en/Modules/Docs.md
@@ -88,6 +88,11 @@ Or you can also manually install nuget package to each project:
 
   `Install-Package Volo.Docs.Web`
 
+* If you use in your project MongoDb Database:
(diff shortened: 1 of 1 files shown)


maliming:
hi

I think these explanations are redundant, we have them in other documents.

simix-markus:
Hi, our main point is the complete json to make it work with MongoDb, wich there is no documentation and isn't equivalent to the SQL sample.
This wasn't trivial to us to figure it out, and we open a support ticket, since the documentation only cover [SQL](https://docs.abp.io/en/abp/latest/Modules/Docs#sample-project-record-for-github), and the schema are different.
The documentation says to put all GitHub* properties inside ExtraProperties, but in MongoDb there is no ExtraProperties, and the GitHub* properties should be in the root level of the document. We can't find this information anywhere before our ticket.
This is our first PR, feel free to simplify and remodel, but we see the MongoDb json record sample as new and relevant information.
Thanks.

PR #12268, opened 2022-04-15
Run worker in an independent thread.

### LINKED ISSUE #12255: Background Worker in continuous mode is blocking the services
# Context

Hello,

I want to use a background worker to write an UDP server specific to our application. However, I have faced 2 issues during the implementation. I will write here a minimal example of code to illustrate it.

# Preparation

Download a template from https://abp.io/get-started. Options: Application / Angular / Entity Framework Core / SQL Server None.
The actual version is 5.2.
Open the `aspnet-core` solution
In Domain, create a dummy continuous Background Worker class
```
using System.Net;
using System.Net.Sockets;
using System.Threading;
using System.Threading.Tasks;
using Microsoft.Extensions.Logging;
using Volo.Abp.BackgroundWorkers;

namespace TestBackgroundWorker;

public class DummyUdpServerBackgroundWorker : BackgroundWorkerBase
{
    public override async Task StartAsync(CancellationToken cancellationToken = default)
    {
        using var udpServer = new UdpClient();
        var destinationIp = IPAddress.Any;
        var serverPort = 5151;
        var endPoint = new IPEndPoint(destinationIp, serverPort);
        udpServer.Client.Bind(endPoint);


        while (!StoppingToken.IsCancellationRequested)
        {
            var result = await udpServer.ReceiveAsync(StoppingToken);
            Logger.LogDebug($"{result.Buffer.Length} bytes received");
        }
    }

}
```

Open `DomainModule.cs`
Add code from https://docs.abp.io/en/abp/latest/Background-Workers#register-background-worker
```
[DependsOn(typeof(AbpBackgroundWorkersModule))]
public class MyModule : AbpModule
{
    public override Task OnApplicationInitializationAsync(
        ApplicationInitializationContext context)
    {
        context.AddBackgroundWorkerAsync<DummyUdpServerBackgroundWorker >();
    }
}
```

Here is the first issue: the code does not compile, because there is an error in the documentation of version 5.2
```
not all code paths return a value
```

`OnApplicationInitializationAsync` is not declared as aync in code, and therefore should return a `Task`.

Now, change the code a little bit:
```
[DependsOn(typeof(AbpBackgroundWorkersModule))]
public class MyModule : AbpModule
{
    public override async Task OnApplicationInitializationAsync(
        ApplicationInitializationContext context)
    {
        await context.AddBackgroundWorkerAsync<DummyUdpServerBackgroundWorker >();
    }
}
```
Add `async` and `await`.

# Reproduce issue

Now run the `.HttpApi.Host` project. The Background Job runs, but the API does not answer to REST requests.

maliming:
Resolve #12255

Just like document says:

> Background workers are simple independent threads in the application running in the background.

### DIFF
--- a/framework/src/Volo.Abp.BackgroundWorkers/Volo/Abp/BackgroundWorkers/AbpBackgroundWorkersModule.cs
+++ b/framework/src/Volo.Abp.BackgroundWorkers/Volo/Abp/BackgroundWorkers/AbpBackgroundWorkersModule.cs
@@ -1,4 +1,6 @@
-﻿using System.Threading.Tasks;
+﻿using System;
+using System.Threading;
+using System.Threading.Tasks;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Options;
 using Volo.Abp.Modularity;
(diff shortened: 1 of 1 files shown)


realLiangshiwei (approved):

hikalkan:
StartAsync method of the background worker should not do all the work as in the #12255. It should just start the worker. It can create a new thread and stop the thread later in the stopasync method if necessary. Or it can start a timer like `PeriodicBackgroundWorkerBase`.

Also, even if I accept this PR, calling `IBackgroundWorkerManager.StartAsync` is a separate thread doesn't solve the problem. Because it starts all the workers. If the first worker's Start method doesn't return, how the others will start?

If you want, you can create a `SelfThreadBackgroundWorkerBase` that starts/stops the thread and provides an abstract `DoWork` method to be implemented by the end-developer (it is not so simple, you should also care about stopping). I am closing this PR.

hikalkan:
Can you please delete the branch?

PR #12330, opened 2022-04-22
proxy generation api versioning imp

### LINKED ISSUE #11969: Angular: Service Proxies does not support API versioning
https://github.com/abpframework/abp/blob/dev/docs/en/API/API-Versioning.md

image

muhammedaltug:
resolves #11969

### DIFF
--- a/npm/ng-packs/packages/schematics/src/models/method.ts
+++ b/npm/ng-packs/packages/schematics/src/models/method.ts
@@ -1,5 +1,6 @@
 import { eBindingSourceId, eMethodModifier } from '../enums';
-import { camel } from '../utils/text';
+import { camel, camelizeHyphen } from '../utils/text';
+import { getParamName } from '../utils/methods';
 import { ParameterInBody } from './api-definition';
 import { Property } from './model';
 import { Omissible } from './util';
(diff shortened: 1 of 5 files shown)


npm/ng-packs/packages/schematics/src/models/method.ts
```diff
@@ -50,12 +51,12 @@ export class Body {
       ? shouldQuote(paramName)
         ? `${descriptorName}['${paramName}']`
         : `${descriptorName}.${paramName}`
-      : nameOnMethod;
+      : camelizeHyphen(nameOnMethod);
```
mahmut-gundogdu: two intended Ternary operators is not well for readability. Maybe we should change "If" statement

npm/ng-packs/packages/schematics/src/utils/service.ts
```diff
@@ -93,7 +99,41 @@ export function createActionToSignatureMapper() {
 }
 
 function getMethodNameFromAction(action: Action): string {
-  return action.uniqueName.split('Async')[0];
+  return action.uniqueName.replace('Async', '');
```
mahmut-gundogdu: maybe it should be    return action.uniqueName.replace(new RegExp('Async$'), '');  (replace Async at end of line)

mahmut-gundogdu (approved):

PR #12353, opened 2022-04-25
GDPR: Add Volo.Abp.Gdpr.Abstractions package to framework

EngincanV:


framework/src/Volo.Abp.Gdpr.Abstractions/FodyWeavers.xml
```diff
@@ -0,0 +1,3 @@
+﻿<Weavers xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="FodyWeavers.xsd">
+  <ConfigureAwait />
```
hikalkan:
```suggestion
  <ConfigureAwait ContinueOnCapturedContext="false" />
```

framework/src/Volo.Abp.Gdpr.Abstractions/Volo/Abp/Gdpr/GdprUserDataDeleteRequestedEto.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System;
+
+namespace Volo.Abp.Gdpr;
+
+[Serializable]
+public class GdprUserDataDeleteRequestedEto
```
hikalkan:
```suggestion
public class GdprUserDataDeletionRequestedEto
```

framework/src/Volo.Abp.Gdpr.Abstractions/Volo/Abp/Gdpr/GdprUserDataRequestEventHandler.cs
```diff
@@ -0,0 +1,41 @@
...
+    {
+        ServiceScopeFactory = serviceScopeFactory;
+        EventBus = eventBus;
+    }
+    
+    public async Task HandleEventAsync(GdprUserDataRequestedEto eventData)
+    {
+        using (var scope = ServiceScopeFactory.CreateScope())
+        {
+            var gdprDataProviders = scope.ServiceProvider.GetServices<IGdprUserDataProvider>().ToList();
```
hikalkan: I, again, suggest to remove `IGdprUserDataProvider` interface and let the modules to handle the events themselves. I have two reasons:

1. This is a partial implementation and should not be included in the Abstractions package.
2. Using `IDistributedEventBus` makes the package depending on the `Volo.Abp.EventBus` package, which has a lot of direct and indirect dependencies. These dependencies will bring a lot of unnecessary dependencies in the client application, like Blazor WASM. If we make the changes I suggest, this package will have no dependency except the Volo.Abp.Core package.

hikalkan (changes requested):

hikalkan (approved):

hikalkan:
Add this new package to https://github.com/abpframework/abp/blob/dev/nupkg/common.ps1
Otherwise, it won't be published.
Also, talk to @yekalkan to add it to abp.io website to be able to install with abp cli.

PR #12354, opened 2022-04-25
Performance bug fixes and new features for Azure Event Bus

LuisPignataro:
The original code used a delay in the execution thread. This caused an unnecessary delay in the processing of messages.
The use of a message processor per Topic is allowed, and the control of the number of messages that can be processed simultaneously.
Added an attribute to set the Azure eventbus Topic to have more control of the configuration


LuisPignataro:
I found some potential issues that require further testing. Soon I will upload the changes with more improvements

PR #12358, opened 2022-04-26
Introduce `AbpRequestSizeLimitMiddleware `

### LINKED ISSUE #12348: Introduce `AbpRequestSizeLimitAttribute `
From https://github.com/abpframework/abp/issues/12336#issuecomment-1108171660

maliming:
Resolve #12348

```cs

if (MultiTenancyConsts.IsEnabled)
{
    // Before the multi-tenancy middleware.
    app.UseAbpRequestSizeLimit();
    app.UseMultiTenancy();
}

public class FileController : AbpControllerBase, IRemoteService
{
    [HttpPost]
    [AbpRequestSizeLimit(5)]
    public Task<IActionResult> Upload(IFormFile file)
    {
        return Task.FromResult<IActionResult>(Content("test"));
    }
}

```

### DIFF
--- a/framework/src/Volo.Abp.AspNetCore/Microsoft/AspNetCore/Builder/AbpApplicationBuilderExtensions.cs
+++ b/framework/src/Volo.Abp.AspNetCore/Microsoft/AspNetCore/Builder/AbpApplicationBuilderExtensions.cs
@@ -7,6 +7,7 @@ using Microsoft.Extensions.Hosting;
 using Volo.Abp;
 using Volo.Abp.AspNetCore.Auditing;
 using Volo.Abp.AspNetCore.ExceptionHandling;
+using Volo.Abp.AspNetCore.RequestSizeLimit;
 using Volo.Abp.AspNetCore.Security;
 using Volo.Abp.AspNetCore.Security.Claims;
 using Volo.Abp.AspNetCore.Tracing;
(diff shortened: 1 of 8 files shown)


hikalkan (approved):

hikalkan:
I have a few questions:

1. Why we are putting this into if `MultiTenancyConsts.IsEnabled`? Isn't it needed if multi-tenancy disabled? If so, will `AbpRequestSizeLimit` attribute will work when multi-tenancy disabled? If not, that's a big problem. Assume that I've created a module with using this attribute. Any application that is not multi-tenant won't use it. Right?

2. Why is the standard `RequestSizeLimit` not working. Can you please write the reason.

3. Where should we document this?

maliming:
The multi-tenant middleware(`FormTenantResolveContributor`) will read the request body. After that `RequestSizeLimit` won't work because the request stream has already been read.

I will update the `multi-tenant ` document.

JadynWong:
Can we use `RequestSizeLimitAttribute` directly if `RequestSizeLimit` is handled only when multi-tenancy is enabled?

By default aspnetcore handles `RequestSizeLimit`, abp handles it when multi-tenant is enabled.

image

maliming:
> Can we use `RequestSizeLimitAttribute` directly if `RequestSizeLimit` is handled only when multi-tenancy is enabled?
> 
> 
> 
> By default aspnetcore handles `RequestSizeLimit`, abp handles it when multi-tenant is enabled.
> 
> 
> 
> image
> 
> 

Looks good.

hikalkan:
@JadynWong that is much better than introducing a new attribute that can't work in a single-tenant environment. I agree. But, we may also resolve the issue by fixing the problem with multi-tenancy.

hikalkan:
I think we can remove `FormTenantResolveContributor` from the contributors list by default and remove from the multi-tenancy documentation. I believe no one uses it. We deprecate it and remove from the framework in 6.0.
So, we can rollback this PR in that case.

PR #12362, opened 2022-04-26
support url param in abp generate-proxy command

muhammedaltug:


framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/ServiceProxying/Angular/AngularServiceProxyGenerator.cs
```diff
@@ -48,7 +48,7 @@ public override async Task GenerateProxyAsync(GenerateProxyArgs args)
         var apiName = args.ApiName ?? defaultValue;
         var source = args.Source ?? defaultValue;
         var target = args.Target ?? defaultValue;
-
+        var url = args.Url ?? "''";
```
enisn: How about using `string.Empty` here? Is there any special case for using empty char here?

```suggestion
        var url = args.Url ?? string.Empty;
```

mahmut-gundogdu (approved):

PR #12370, opened 2022-04-26
Setting management localization fix

enisn:
- Closes https://github.com/volosoft/volo/issues/9482


enisn:
Since key is here: https://github.com/abpframework/abp/blob/dev/modules/setting-management/src/Volo.Abp.SettingManagement.Domain.Shared/Volo/Abp/SettingManagement/Localization/Resources/AbpSettingManagement/en.json#L21

realLiangshiwei (approved):

PR #12377, opened 2022-04-27
Add validation rules for this property.

### LINKED ISSUE #12282: [Feature Request] abp generate-proxy: also generate DTO constraint constants
It would be great if `abp generate-proxy` (`-t ng` in my case) also generates DTO constraint constants, such as max length of string fields, so we don't have to defined them in two different places.

maliming:
Resolve #12282

image


JadynWong:
Hi, Should we also consider handling the dynamic validation attribute provided by abp?

https://github.com/abpframework/abp/tree/e3e1779de6df5d26f01cdc8e99ac9cbcb3d24d3c/framework/src/Volo.Abp.Core/Volo/Abp/Validation

maliming:
The Dynamic attributes are inherit from the standard attributes. so I think It will work.

hikalkan (approved):

PR #12380, opened 2022-04-27
Fall back to publish after the UOW is completed if the outbox is unavailable

### LINKED ISSUE #11208: Distributed events unexpectedly published in advance if outbox is unavailable

gdlcf88:
Another way to resolve https://github.com/abpframework/abp/issues/11208.

Alternatives to https://github.com/abpframework/abp/pull/11209.

### DIFF
--- a/framework/src/Volo.Abp.EventBus/Volo/Abp/EventBus/Distributed/AbpDistributedEventBusOptionsExtensions.cs
+++ b/framework/src/Volo.Abp.EventBus/Volo/Abp/EventBus/Distributed/AbpDistributedEventBusOptionsExtensions.cs
@@ -1,0 +1,18 @@
+﻿using System;
+using System.Linq;
+
+namespace Volo.Abp.EventBus.Distributed;
+
+public static class AbpDistributedEventBusOptionsExtensions
+{
+    public static OutboxConfig GetOutbox(this AbpDistributedEventBusOptions options, Type eventType)
+    {
+        return options.Outboxes.Values.FirstOrDefault(outboxConfig =>
+            outboxConfig.Selector == null || outboxConfig.Selector(eventType));
+    }
+
+    public static bool OutboxExist(this AbpDistributedEventBusOptions options, Type eventType)
+    {
+        return options.GetOutbox(eventType) is not null;
+    }
+}
--- a/framework/src/Volo.Abp.EventBus/Volo/Abp/EventBus/UnitOfWorkEventPublisher.cs
+++ b/framework/src/Volo.Abp.EventBus/Volo/Abp/EventBus/UnitOfWorkEventPublisher.cs
@@ -1,5 +1,6 @@
 ﻿using System.Collections.Generic;
 using System.Threading.Tasks;
+using Microsoft.Extensions.Options;
 using Volo.Abp.DependencyInjection;
 using Volo.Abp.EventBus.Distributed;
 using Volo.Abp.EventBus.Local;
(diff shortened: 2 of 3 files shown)


framework/src/Volo.Abp.EventBus/Volo/Abp/EventBus/Distributed/AbpDistributedEventBusOptionsExtensions.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
+using System.Linq;
+
+namespace Volo.Abp.EventBus.Distributed;
+
+public static class AbpDistributedEventBusOptionsExtensions
+{
+    public static OutboxConfig GetOutbox(this AbpDistributedEventBusOptions options, Type eventType)
```
hikalkan, line 8: Better to name `FindOutbox` since we use `Get` for method those don't return null, and `Find` for those may return `null`.
Also, to make it more clear, add [CanBeNull] attributed on top of this method.

framework/src/Volo.Abp.EventBus/Volo/Abp/EventBus/UnitOfWorkEventPublisher.cs
```diff
@@ -37,12 +44,44 @@ public async Task PublishDistributedEventsAsync(IEnumerable<UnitOfWorkEventRecor
...
+                        onUnitOfWorkComplete: false,
+                        useOutbox: true
+                    );
+                }
+                else
+                {
+                    // Fall back to publish after the UOW is completed if the outbox is unavailable.
+                    if (_unitOfWorkManager.Current != null)
+                    {
+                        _unitOfWorkManager.Current.OnCompleted(async() => {
```
hikalkan, line 66: Now, I understand what you are trying to do. You want to publish event after the UOW completed, transaction is committed. But this is changing the behavior and it is against our design. We intentionally not doing this like that.

(1) With our current design; if we can't publish the event, the transaction is rolled back. If we publish the event, but can't commit the transaction, then we publish an unnecessary event.

(2) If we accept this PR; If we can't publish the event, the transaction is not rolled back. Then we have the operation is done, but we miss the event. Also, the end-user gets an error since we don't hide the exception (if we can't publish the event), but the operation is actually successful. We don't want that.

There is a tradeoff here we choose the option (1). If you don't want the tradeoff, the only solution is to configure the outbox.

So, we want to go with the option (1) and not want to change to option (2). So, sorry, I won't accept this PR.

The only solution we may accept is to provide a framework option to change the behavior for who want it (while keeping the current behavior as default).

hikalkan, line 66: Now, I understand what you are trying to do. You want to publish event after the UOW completed, transaction is committed. But this is changing the behavior and it is against our design. We intentionally not doing this like that.

(1) With our current design; if we can't publish the event, the transaction is rolled back. If we publish the event, but can't commit the transaction, then we publish an unnecessary event.

(2) If we accept this PR; If we can't publish the event, the transaction is not rolled back. We don't want that.

There is a tradeoff here we choose the option (1). If you don't want the tradeoff, the only solution is to configure the outbox.

So, we want to go with the option (1) and not want to change to option (2). So, sorry, I won't accept this PR.

The only solution we may accept is to provide a framework option to change the behavior for who want it (while keeping the current behavior as default).

hikalkan:
Now, I understand what you are trying to do. You want to publish event after the UOW completed, transaction is committed. But this is changing the behavior and it is against our design. We intentionally not doing this like that.

(1) With our current design; if we can't publish the event, the transaction is rolled back. If we publish the event, but can't commit the transaction, then we publish an unnecessary event.

(2) If we accept this PR; If we can't publish the event, the transaction is not rolled back. Then we have the operation is done, but we miss the event. Also, the end-user gets an error since we don't hide the exception (if we can't publish the event), but the operation is actually successful. We don't want that.

There is a tradeoff here we choose the option (1). If you don't want the tradeoff, the only solution is to configure the outbox.

So, we want to go with the option (1) and not want to change to option (2). So, sorry, I won't accept this PR.

The only solution we may accept is to provide a framework option to change the behavior for who want it (while keeping the current behavior as default).

gdlcf88:
> Also, the end-user gets an error since we don't hide the exception (if we can't publish the event), but the operation is actually successful.

It is indeed a problem that existed before ABP 5.0.

PR #12384, opened 2022-04-27
Revert "Publish local events when uow completed."

hikalkan:
Reverts abpframework/abp#11125

#11125 was a mistake and it breaks how we designed it with ABP 5.0 and declared in the [blog post](https://blog.abp.io/abp/ABP-IO-Platform-5.0-RC-1-Has-Been-Released):

image

Distributed event handlers must be executed in the same unit of work where we publish the events.


hikalkan:
I closed this since we will fix it in 5.2.x.

PR #12400, opened 2022-04-28
Cli: Configure Docker Files

yekalkan:


framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/ProjectBuilding/Building/Steps/TemplateCodeDeleteStep.cs
```diff
@@ -13,6 +13,8 @@ public override void Execute(ProjectBuildContext context)
                 file.Name.EndsWith(".cshtml") ||
                 file.Name.EndsWith(".json") ||
                 file.Name.EndsWith(".gitignore") ||
+                file.Name.EndsWith(".yml") ||
+                file.Name.EndsWith(".ps1") ||
```
gterdem, line 17: We need to add .json also:
https://github.com/volosoft/volo/blob/fe5640e781607bd2419816b307c3f2e0ab8a25a8/abp/templates/app-pro/aspnet-core/etc/docker/dynamic-env.json

If this this json file is not valid, remote environment feature is not working for angular.

framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/ProjectBuilding/Templates/App/AppNoLayersTemplateBase.cs
```diff
@@ -102,4 +95,35 @@ public override IEnumerable<ProjectBuildPipelineStep> GetCustomSteps(ProjectBuil
 
         return steps;
     }
+
+    private void ConfigureDockerFiles(ProjectBuildContext context, List<ProjectBuildPipelineStep> steps)
+    {
```
gterdem, line 100: We can remove [dynamic-env.json](https://github.com/volosoft/volo/blob/fe5640e781607bd2419816b307c3f2e0ab8a25a8/abp/templates/app-pro/aspnet-core/etc/docker/dynamic-env.json) if the solution is **not angular**.

gterdem (approved):

framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/ProjectBuilding/Building/Steps/TemplateCodeDeleteStep.cs
yekalkan: It is already added. (line 14)

framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/ProjectBuilding/Templates/App/AppNoLayersTemplateBase.cs
yekalkan: Already done.

PR #12406, opened 2022-04-29
Fix typos

MichelZ:


EngincanV:
Thanks for your contribution @MichelZ. If it's possible can you do the same changes in the **rel-5.0** branch and create a new PR? Because, otherwise if we merge this PR, your changes will only be seen from v5.3+ documentation.

MichelZ:
Sure, see #12407

EngincanV:
> Sure, see #12407

Thanks 👍

PR #12510, opened 2022-05-10
Fix the `ValueEquals` bug.

### LINKED ISSUE #12509: ValueObject .ValueEquals has a bug:It will misjudge that two different objects are equal
ABP Framework version:5.1.4
```
public class MyV: ValueObject {
   private string[] _strs;
   public MyV(string[] strs)
   {
        _strs = strs;
   }
   protected override IEnumerable<object> GetAtomicValues()
   {
        return this._strs.ToList();
   }
 }

 MyV v1 = new MyV(new string[] { "1", "2", "3" });
 MyV v2 = new MyV(new string[] { "1", "2" });

 v1.ValueEquals(v2) //return true , should be false
```

The reason for this problem is that it  calls ` MoveNext` twice:
```
  while (thisValues.MoveNext() && otherValues.MoveNext())
   {
        if (ReferenceEquals(thisValues.Current, null) ^
             ReferenceEquals(otherValues.Current, null))
        {
            ...
 return !thisValues.MoveNext() && !otherValues.MoveNext();
```

maliming:
Resolve #12509

### DIFF
--- a/framework/src/Volo.Abp.Ddd.Domain/Volo/Abp/Domain/Values/ValueObject.cs
+++ b/framework/src/Volo.Abp.Ddd.Domain/Volo/Abp/Domain/Values/ValueObject.cs
@@ -16,21 +16,29 @@ public abstract class ValueObject
             return false;
         }
 
-        ValueObject other = (ValueObject)obj;
+        var other = (ValueObject)obj;
 
-        IEnumerator<object> thisValues = GetAtomicValues().GetEnumerator();
-        IEnumerator<object> otherValues = other.GetAtomicValues().GetEnumerator();
+        var thisValues = GetAtomicValues().GetEnumerator();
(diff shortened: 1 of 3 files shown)


hikalkan (approved):

thefristloser:
I just checked the code.
The bug still exists.
test case:
```
 public class MyV : ValueObject
 {
     private string[] _strs;
     public MyV(string[] strs)
     {
         _strs = strs;
    }
    protected override IEnumerable<object> GetAtomicValues()
    {
        return this._strs.ToList();
   }
 }
MyV v1 = new MyV(new string[] { "1" });
MyV v2 = new MyV(new string[] {  });
v1.ValueEquals(v2)  //  return true. should be false
```
I've already said the reason for this bug.#12509  .
But you don't really understand.
Why can't such a simple thing be done well，I don't understand.

maliming:
Updating...

PR #12534, opened 2022-05-11
Openiddict minor revisions

hikalkan:
I found some wrong/unnecessary dependencies. I fixed some of them, please check them and write me if there was a reason for adding them.

I also have some notes. Please you consider and fix these:

- [x] Shouldn't OpenIddict.Server be a dependency of Volo.Abp.OpenIddict.AspNetCore? Currently we are depending from Volo.Abp.OpenIddict.Domain seems wrong to me (the code parts using the OpenIddict.Server package seems UI/Web-related).
- [x] Volo.Abp.OpenIddict.Domain.Shared should not depend on Volo.Abp.UI. AbpUiResource can be added in the UI layer of the module. See https://github.com/abpframework/abp/blob/dev/templates/module/aspnet-core/src/MyCompanyName.MyProjectName.Domain.Shared/MyProjectNameDomainSharedModule.cs#L27 and https://github.com/abpframework/abp/blob/dev/templates/module/aspnet-core/src/MyCompanyName.MyProjectName.HttpApi/MyProjectNameHttpApiModule.cs#L29 to understand how you can add AbpUiResource later in the UI/web layer.

One improvement note:

- [x] Use distributed locking in TokenCleanupBackgroundWorker to ensure only one instance does the work when we run multiple instances of the application.

And one problem with the demo:

- [x] In demo, I run server, api and client.mvc. I login from client.mvc (throught the server), it works and I see the claims. When I open the server app (https://localhost:44301/), I see it is logged in. But, when I click a menu item (ex: try to enter the tenant management page), I am getting error: "You are not authenticated! You should sign in to perform this operation." Only the setting page is accessable. It is same even if I directly login to the server application. When I check database, I see that the AbpPermissionGrants table is properly filled.

You can do all these work in this PR.

And one question:

- Does it work with our dynamic/static proxy system? I suppose it should be since we are using the old IdentityModel library.


maliming:
> Does it work with our dynamic/static proxy system? I suppose it should be since we are using the old IdentityModel library.

The `IdentityModel` is a wrapper class library for protocols, independent of `Identity Server` and `OpenIddict`. They implement protocols, and `IdentityModel` consumes protocols.

PR #12553, opened 2022-05-13
Document how to check global features in MVC UI.

maliming:


hikalkan:
Hi,

Currently, we don't provide any function to check. The developer, herself, works with arrays. At least, we should provide the following API.

`isEnabled` function to check if a feature is enabled or not:

````js
if (abp.globalFeatures.isEnabled(`Ecommerce.Subscription`))
{
  ...
}
````

Also, I didn't understand the purpose of `abp.globalFeatures.moduleEnabledFeatures`. I suppose it is grouping features by modules. But I believe that it is useless. In the UI side, we just need to `abp.globalFeatures.isEnabled` to check if a feature is enabled or not. If `moduleEnabledFeatures` is not used anywhere, we can drop it in JavaScript side, also from the endpoint.

**Comment by maliming** [comment]
hi

I did the changes.

https://github.com/abpframework/abp/pull/12553/commits/cecc2c1baa53675906bbd96ac27e476783821bad

File: npm/packs/core/src/abp.js [hunk 1]
```diff
@@ -773,4 +773,14 @@ var abp = abp || {};
         return abp.features.values[name];
     };
     
+    /* GLOBAL FEATURES *************************************************/
+
+    abp.globalFeatures = abp.globalFeatures || {};
+
+    abp.globalFeatures.enabledFeatures = abp.globalFeatures.enabledFeatures || {};
```
hikalkan:
```suggestion
    abp.globalFeatures.enabledFeatures = abp.globalFeatures.enabledFeatures || [];
```

Shouldn't be like that?

hikalkan (changes requested):

npm/packs/core/src/abp.js
maliming: You're right.

hikalkan (approved):

PR #12576, opened 2022-05-13
Added checkers for integers, floats and decimal

j0nimost:
Added new static methods for the `Check` class;

- Positive
- Range

### DIFF
--- a/framework/src/Volo.Abp.Core/Volo/Abp/Check.cs
+++ b/framework/src/Volo.Abp.Core/Volo/Abp/Check.cs
@@ -162,4 +162,182 @@ public static class Check
 
         return value;
     }
+
+    public static Int16 NotNegativeOrZero(
+        Int16 type,
+        [InvokerParameterName][NotNull] string parameterName)
+    {
+        if(type == 0)
+        {
+            throw new ArgumentException($"{parameterName} is equal to zero");
+        }
+        else if(type < 0)
+        {
+            throw new ArgumentException($"{parameterName} is less than zero");
+        }
+        return type;
+    }
+
+    public static Int32 NotNegativeOrZero(
+        Int32 type,
+        [InvokerParameterName][NotNull] string parameterName)
+    {
+        if (type == 0)
+        {
+            throw new ArgumentException($"{parameterName} is equal to zero");
+        }
+        else if (type < 0)
+        {
+            throw new ArgumentException($"{parameterName} is less than zero");
+        }
+        return type;
+    }
+
+    public static Int64 NotNegativeOrZero(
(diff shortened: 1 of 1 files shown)


framework/src/Volo.Abp.Core/Volo/Abp/Check.cs
```diff
@@ -162,4 +162,182 @@ public static ICollection<T> NotNullOrEmpty<T>(ICollection<T> value, [InvokerPar
 
         return value;
     }
+
+    public static Int16 NotNegativeOrZero(
```
hikalkan: Please rename all to `Positive`

framework/src/Volo.Abp.Core/Volo/Abp/Check.cs
```diff
@@ -162,4 +162,182 @@ public static ICollection<T> NotNullOrEmpty<T>(ICollection<T> value, [InvokerPar
...
+        else if (type < 0)
+        {
+            throw new ArgumentException($"{parameterName} is less than zero");
+        }
+        return type;
+    }
+
+
+
+    public static Int16 WithinRange(
```
hikalkan: Please rename all to `Range`

hikalkan (changes requested):
Thanks for your contribution. Please rename the methods as I suggested. Also, add unit tests for these methods.

rqx110:
rename param name `type` to `value`?
e.g. `Int16 type`  --> `Int16 value`

framework/src/Volo.Abp.Core/Volo/Abp/Check.cs
j0nimost: I have implemented the change  in this commit: [aa08d0d](https://github.com/abpframework/abp/pull/12576/commits/aa08d0d5ad86e298e433c15c26a82a36f8a173b9)

framework/src/Volo.Abp.Core/Volo/Abp/Check.cs
j0nimost: I have implemented the change  in this commit: [aa08d0d](https://github.com/abpframework/abp/pull/12576/commits/aa08d0d5ad86e298e433c15c26a82a36f8a173b9)

j0nimost:
> rename param name `type` to `value`? e.g. `Int16 type` --> `Int16 value`

I have renamed the property names for all in this commit:  [daab524](https://github.com/abpframework/abp/pull/12576/commits/daab524943f7e6bad8259fa6a90ee833495236f1)

hikalkan (approved):

PR #12602, opened 2022-05-16
Added checker for Struct types

j0nimost:
Added checker for structs, this applies to;
- Guids
- DateTimes
- and any other struct type


j0nimost:
Hi @maliming, I have a question can this feature include the Nullable struct checkers;

My proposal is here:

```C#
    /// <summary>
    /// Checks if value of type T (struct) is empty, default or null
    /// </summary>
    /// <typeparam name="T"></typeparam>
    /// <param name="value"></param>
    /// <param name="parameterName"></param>
    /// <returns></returns>
    /// <exception cref="ArgumentException"></exception>
    public static T NotDefaultorNull<T>(
        Nullable<T> value,
        string parameterName)
        where T : struct
    {
        if (value.Equals(default(T)))
        {
            throw new ArgumentException($"{parameterName} has a default value or is empty!", parameterName);
        }
        else if (value.HasValue == false)
        {
            throw new ArgumentException($"{parameterName} is null!", parameterName);
        }
        return (T)value;
    }
```

j0nimost:
@maliming  My reasoning behind this logic is in scenarios like this;

```csharp
// case 1
Guid? guid1 = null;
//case 2
Guid? guid2 = new Guid();
//case 3
Guid guid3 = new Guid();
```

A Guid can be nullable, empty or default so this checker validates either of them

maliming (approved):

PR #12633, opened 2022-05-17
add no-app-layer to remove-lock-files script

mahmut-gundogdu:


npm/scripts/remove-lock-files.ts
```diff
@@ -8,6 +8,8 @@ export async function removeLockFiles() {
     '../../templates/app/angular',
     '../../templates/app/react-native',
     '../../templates/module/angular',
+    '../../templates/app-nolayers/angular',
+    '../../templates/app-nolayers/react-native',
```
muhammedaltug, line 12: react-native directory doesn't exist in app-no-layers

Issue #12659, opened 2022-05-19
Rebus: Triggers events twice (handler is invoked twice)


brunobertechini:
* Your **ABP Framework version**.: 5.2.2
```
<PackageReference Include="Volo.Abp.EventBus.Rebus" Version="5.2.2" />
<PackageReference Include="Rebus.RabbitMq" Version="7.3.5" />
```

Hello there, I just created a new solution (5.2.2) and added the configuration for Rebus.RabbitMq to test the integration. 

I believe I am facing the same problem of https://github.com/abpframework/abp/issues/11431 (although not sure it is because of the same reason) of my handler being called twice when using Rebus+RabbitMq.

I have no special configuration. A simple event and a simple handler:

```
PreConfigure<AbpRebusEventBusOptions>(options =>
        {
            var queueName = "eventbus";
            options.InputQueueName = queueName;
            options.Configurer = rebusConfigurer =>
            {
                rebusConfigurer.Transport(t => t.UseRabbitMq("amqp://localhost", queueName));
                // rebusConfigurer.Subscriptions(s => s.UseJsonFile(@"subscriptions.json"));
            };
        });
```

```
public class SampleEvent1 : IEvent
{
    [Required]
    public string MyStringProp { get; set; }
    
    [Required]
    public int? MyIntProp { get; set; }
    
    [Required]
    public bool? MyBoolProp { get; set; }
    
    [Required]
    public DateTime? MyDateTimeProp { get; set; }
}
```

```
public class SampleEvent1Handler : IDistributedEventHandler<SampleEvent1>, ITransientDependency
{
    private readonly ILogger<SampleEvent1Handler> _logger;

    public SampleEvent1Handler(ILogger<SampleEvent1Handler> logger)
    {
        _logger = logger;
    }    
    
    public Task HandleEventAsync(SampleEvent1 eventData)
    {
        _logger.LogInformation($"Received Event {eventData.GetType().FullName}");
        
        Thread.Sleep(10000);
        
        _logger.LogInformation($"Finished processing Event {eventData.GetType().FullName}");
        
        return Task.CompletedTask;
    }
}
```

brunobertechini:
I have tested the same handler with `Volo.Abp.RabbitMQ` (`AbpEventBusRabbitMqModule`) only, and it works perfectly fine (handler is invoked only once).

realLiangshiwei:
Hi, @brunobertechini 

I could not reproduce the problem, can you provide a project?

[...]

PR #12670, opened 2022-05-20
Rebus: Fix the problem of calling Handler multiple times

realLiangshiwei:
Resolves: https://github.com/abpframework/abp/issues/12659


maliming (approved):

brunobertechini:
Thank you @realLiangshiwei, I will give it a try soon.

PR #12692, opened 2022-05-21
External user update on external login on the client side

### LINKED ISSUE #12203: Create/update user account after login with external authentication server
Consider this scenario: Our client application (can be Angular, Blazor WASM or tiered MVC UI) uses OpenIdConnect flow to login to the application, but not uses our own backend. For example, it uses a Keycloak server. In this case, the user is redirected to Keycloak, enters username and password, and redirected back to the client application. Our backend has no knowledge on this process, and there is a possibility that the user's account is not available in our local database (in the Identity module's Users table). In this case, we would like to create this user as an [external user](https://github.com/abpframework/abp/blob/dev/modules/identity/src/Volo.Abp.Identity.Domain/Volo/Abp/Identity/IdentityUser.cs#L71) in our local database. In this way, we can see this user in the Users page and set necessary permissions and make other user-related operations in our system. The user will continue to use the external auth server to login, but we will have her profile information (like Id, Username, Email, Name, Surname) in our local database.

For this purpose;

* Add REST API endpoint (and corresponding application service method) to the Identity module to create/update current user from external login provider. This endpoint will get the external provider name to be more generic (LDAP, OpenIdConnect, etc.). The endpoint should be authorized ([Authorize]) and work only with the current user for security (doesn't get user id as parameter!). This endpoint will query the external auth server to obtain user information, create or update the user in the local database.
* The REST API will be called by the client application, just after the authentication flow (typically OpenId Connect or resource owner password flow) completed, but before the first request is done. For MVC UI, as I can see, this can be done just in the `options.Events.OnUserInformationReceived` option of the [AddAbpOpenIdConnect](https://github.com/abpframework/abp/blob/e8a810ba4bd94bd1d4ca447ec4d8645354c75d1d/framework/src/Volo.Abp.AspNetCore.Authentication.OpenIdConnect/Microsoft/Extensions/DependencyInjection/AbpOpenIdConnectExtensions.cs#L23) method. But, we should check if there is a better way. For other UI options, we should find a proper point.

hikalkan:
Resolves #12203.

When `AddAbpOpenIdConnect` is used, ABP can automatically request to a URL on the server to inform the server, so it can create or update the user in the local database. The Identity module provides such an endpoint.
See `OpenIdLocalUserCreationClientOptions` for the configuration options. It is disabled by default, but can be enabled for such scenarios.


hikalkan:
@enisn PR is draft, because I haven't implemented the endpoint in the Identity module. Please implement it just like I did for the Identity.Pro module (PR 10354).

Also, please fully test this with Keyloak (you can temporary use the app template - without commiting)

enisn (approved):

enisn:
Seems good, I'll merge after I try with keycloak

ikappas:
@hikalkan @enisn Thank you for all your work!

I was reviewing this PR, as we need to implement this feature in an upcoming project, and have run into a potential issue.

We need to parse an store user data accessed via user claims exclusively available after the call to user info endpoint via the `GetClaimsFromUserInfoEndpoint` option. 

As a result, the user is created before this call is ever made and is missing this information. Also, any `options.ClaimActions.MapJsonKey` directives execute after `OnTokenValidated` finishes.

Am I missing something? Your feedback is greatly appreciated.

iddelacruz:
@enisn I'm trying to use abp with keycloak but I am seeking the good solutions becouse I want to continue using permissions and the other good modules. Can you help me or give me any documentation? Regards.

PR #12694, opened 2022-05-21
fixed statement.

nicksoftware:
I am Assuming that the statement  :
   "You generally don't need to this low level ICurrentPrincipalAccessor service and directly work with the ICurrentUser explained above." 
meant  : 
   "You generally don't need to use this low level ICurrentPrincipalAccessor service just directly work with the ICurrentUser explained above."


braim23 (approved):

braim23:
Thanks for bringing this up @hnicolus 😄

PR #12719, opened 2022-05-24
Update hangfire to 1.7.29

realLiangshiwei:


maliming (approved):

wnumwwfpps:
Can we update Hangfire? It has been a while since the last update, and the latest version is as follows:
https://www.hangfire.io/blog/2023/05/26/hangfire-1.8.2.html
@maliming

maliming:
hi @wnumwwfpps 
No problem, We will do it.

PR #12852, opened 2022-06-01
Using the event boxes orderly

### LINKED ISSUE #12851: Making the event box selectors orderly
Outboxes and inboxes with a `null` Selector should have the lowest priority as they are the fallback ones.

gdlcf88:
Resolve #12851


hikalkan:
I agree that the null selector should be evaluated as the last selector. However, I am not sure if OrderBy changes orders of the others. Ideally, non-null values should be tried without re-ordering, then the `null` ones should be tried later with their original order. @realLiangshiwei please check if that works as I expect.

hikalkan:
I agree that the null selector should be evaluated as the last selector. However, I am not sure if OrderBy changes orders of the others. Ideally, non-null values should be tried without re-ordering, then the `null` ones should be tried later with their original order. @realLiangshiwei please check if that works as I expect.

realLiangshiwei (approved):

realLiangshiwei:
It works as expect

Issue #12632, opened 2022-05-17
Update packages to latest


realLiangshiwei:
Update dependencies to the latest

PR #12861, opened 2022-06-02
Update package `Fody` to latest version

roc916:
Related with #12632

NuGet link:

[Fody](https://www.nuget.org/packages/Fody/)


maliming:
hi

This upgrade doesn't make sense at the moment. We actually need ConfigureAwait.Fody.

Thanks.

PR #12865, opened 2022-06-02
Solves performance issues. Adds support for topic per event

LuisPignataro:
Solves performance issues. 
Add simultaneous message processing, with configuration.
MessageId for duplicate detection
Adds support for topic per event


LuisPignataro:
Hello @maliming , I am going to integrate the code from version 7 into these changes. Is there something you think should be improved? Or any suggestion? I am using this code in production and it works great.

maliming:
hi

I will test it as soon as possible. Thanks

maliming:
hi @LuisPignataro 

I think we can replace `Task.Factory.StartNew` with `AsyncHelper`.

LuisPignataro:
Hi @maliming I'm not using Task.Factory.StartNew, I even did a search just in case.
I do not understand what you're referring to?

maliming:
hi 

I think we can keep the `AsyncHelper` changes. Other changes are left to the developers.

PR #12896, opened 2022-06-04
Introduce Integration Service Infrastructure

### LINKED ISSUE #12470: Integration Services
Application Services are ideal to be consumed by UI and other clients. We, sometimes, are also using an application service of a module X from a module Y to integrate these modules (or microservices).

There are some problems of using application services for inter-module (or inter-microservice) calls:

* They are specifically designed for UI scenarios (use cases), but inter-module calls have different requirements (different data (DTO) needs, different use cases, they differently evolve by the time, etc).
* We generally don't want to apply authorization for inter-module/service calls. But, if we use an existing authorized app service endpoint, we have to disable authorization conditionally (ugly) or have to authenticate microservices as clients.
* We may need different level of caching and other cross-cutting concerns.
* Most app services might be designed to be used by a user (`ICurrentUser.Id` is available), so requires a kind of user-session, so it becomes problem while consuming from a client without an active user interaction.

You may find other reasons too...

So, my offer is to introduce `IIntegrationService` (implemented by the integration service classes) that is specifically designed for inter-module/service calls. They will have similar defaults with app services (like auto [unitofwork](https://docs.abp.io/en/abp/latest/Unit-Of-Work), validation, and authorization if you want) and they can also be exposed as [conventional API controllers](https://docs.abp.io/en/abp/latest/API/Auto-API-Controllers).

Default HTTP API URL prefix for such services can be `/internal-api/` or `/integration-api/` instead of `/api/`. In this way, we can easily block request against these endpoints and safely consume them internally in our microservice cluster for inter-microservice calls.

Integration service classes can be located inside the `YourProject.Application` project of your solution (and their contracts in `YourProject.Application.Contracts`), if you don't want to create additional .NET projects in your solution. But, if you want (I think it would be much better for maintainability), you can create `YourProject.IntegrationServices` and `YourProject.IntegrationServices.Contracts` projects in the solution.

hikalkan:
Resolves #12470

- [x] Define `IntegrationServiceAttribute`. We can use it on application services to mark them as integration service.
- [x] Change default URL prefix to `/integration-api/` for integration services.
- [x] Implement `AbpAuditingOptions.IsEnabledForIntegrationServices` to allow us to enable/disable audit logging for these services globally. Calls to integration services are not audit-logged by default.


framework/src/Volo.Abp.AspNetCore/Volo/Abp/AspNetCore/Auditing/AbpAuditingMiddleware.cs
```diff
@@ -94,8 +94,23 @@ public async Task InvokeAsync(HttpContext context, RequestDelegate next)
...
     {
-        return context.Request.Path.Value != null &&
-               AspNetCoreAuditingOptions.IgnoredUrls.Any(x => context.Request.Path.Value.StartsWith(x));
+        if (context.Request.Path.Value == null)
+        {
+            return false;
+        }
+        
+        if (!AuditingOptions.IsEnabledForIntegrationServices && 
+            context.Request.Path.Value.StartsWith($"/{AbpAspNetCoreConsts.DefaultIntegrationServiceApiPrefix}/"))
```
rqx110, line 103: this maybe a problem, because the router prefix of `IntegrationServiceApi` is configurable

hikalkan: I know and you are right.

If we set another prefix while configuring conventional controllers for application services, then we should also add that URL to the `AbpAspNetCoreAuditingOptions.IgnoredUrls`. ABP can't make it automatic, because a conventional controller configuration may contain both regular application and integration services those share a single `ApiRoutePrefix`.

I think it is best to remove the `ApiRoutePrefix` option since it may break other things. Thanks for your comment.

Done with https://github.com/abpframework/abp/pull/12896/commits/c56473eeb7b02efdca7fd48f7fdabd3cbb9ee1d7

maliming (approved):

PR #13026, opened 2022-06-16
Identity Server Index Page

enisn:
Now the index page doesn't have the same menu as the application and shows a list of configured clients.

image


gterdem (approved):

ebicoglu (approved):

vijay2358:
i do see it calling repo directly  in index.cs can we use service or identity server services or abp service call

enisn:
It's included in templates and you can change it however you want. 

But, In that case; the authentication server can't be separated into different services or modules, so we did it like that. But normally it's not a best practice you're right.

PR #13036, opened 2022-06-17
Add `ClientUri ` and `LogoUri ` to `OpenIddictApplication`.

maliming:


enisn (approved):

enisn:
How can I seed ClientUri via using ApplicationManager?

I've tried a couple of ways but it `ClientUri` column is always null.

1. I've injected `IAbpApplicationManager`
2. Created a new `AbpApplicationDescriptor` and set `ClientUri` parameter.
3. Passed that instance as a parameter of `IAbpApplicationManager.CreateAsync` method.
4. Result:
image

---

@maliming Also you can check this commit: https://github.com/abpframework/abp/commit/5e3bfecd8bdb7e6d14fdfa4531c4589d56eab288

maliming:
ok, I will check.

enisn:
Ok, it works, forgot the drop database, it doesn't update existing entities currently, maybe  we can add an update logic later

PR #13042, opened 2022-06-18
refactor OnAfterRenderAsync of AbpCrudPageBase

wangj90:
In the previous code, the OnAfterRenderAsync method of the base class is called only when firstRender is true, it should be changed to OnAfterRenderAsync method of the base class needs to be called at any time


maliming (approved):

maliming:
Thanks @wangj90

PR #13093, opened 2022-06-23
Remove `Logout` page.

maliming:
https://user-images.githubusercontent.com/6908465/175236507-9ba03336-4208-434b-9133-acec753272a1.mp4


gterdem:
So we _remove_ the **SignedOut** view that informs the user about the signout and redirects back to the application. 

I think it is okay. If we do require not to redirect immediately, we can update it again later.

gterdem (approved):

PR #13174, opened 2022-06-29
fix DefaultFromAddress localization

icagatay:
setting-management module fix DefaultFromAddress and DefaultFromDisplayName localization from tr


maliming (approved):

maliming:
Thanks @icagatay

PR #13200, opened 2022-07-02
Add missing `Microsoft.Bcl.AsyncInterfaces` package dependency

### LINKED ISSUE #13199: Failed to use IAbpDistributedLock due to the AutoMapper upgraded
Before https://github.com/abpframework/abp/issues/12144, projects that both depend on the `Volo.Abp.DistributedLocking.Abstractions` and `Volo.Abp.AutoMapper` can correctly use:
```CSharp
await using (var handle = await _distributedLock.TryAcquireAsync("lock1"))
{
    handle.ShouldNotBeNull();
}
```

Due to the AutoMapper 11 being upgraded to netstandard2.1 but the `Volo.Abp.DistributedLocking.Abstractions` is still netstandard2.0, the `Microsoft.Bcl.AsyncInterfaces` dependency cannot be shared anymore, so it causes:
2d6c6ccd523b0c46f6136011896069b

Step to reproduce:
1. Create an ABP 5.3.1 app.
2. Use IAbpDistributedLock in the non-host layer (since the host project depends on the Autofac module which is netstandard2.0 and explicitly depends on `Microsoft.Bcl.AsyncInterfaces`).
3. Try to build the solution.

You can also reproduce it by removing the Autofac module dependency in `Volo.Abp.DistributedLocking.Abstractions.Tests`:
image

gdlcf88:
Fix #13199

### DIFF
--- a/framework/src/Volo.Abp.DistributedLocking.Abstractions/Volo.Abp.DistributedLocking.Abstractions.csproj
+++ b/framework/src/Volo.Abp.DistributedLocking.Abstractions/Volo.Abp.DistributedLocking.Abstractions.csproj
@@ -16,6 +16,7 @@
 
     <ItemGroup>
         <ProjectReference Include="..\Volo.Abp.Core\Volo.Abp.Core.csproj" />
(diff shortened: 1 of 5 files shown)


maliming:
Move to https://github.com/abpframework/abp/pull/13205

PR #13243, opened 2022-07-05
Add leptonx-lite theme package version to Directory.Build.props and update the versions in templates

EngincanV:
To update leptonx-lite package versions easily, I've added a new entry to **Directory.Build.props** file and also updated the package versions in templates.


maliming:
hi

We can't use the value of `Directory.Build.props` in templates. 

There is no `Directory.Build.props` after the project is created.

PR #13279, opened 2022-07-10
Cms kit editor issue 10970

malikmasis:
closes https://github.com/volosoft/volo/issues/10970

### DIFF
--- a/modules/cms-kit/src/Volo.CmsKit.Admin.Web/Pages/CmsKit/BlogPosts/create.js
+++ b/modules/cms-kit/src/Volo.CmsKit.Admin.Web/Pages/CmsKit/BlogPosts/create.js
@@ -257,14 +257,15 @@ $(function () {
             initEditor(item);
         });
     }
-
+    
+    var editor;
     function initEditor(element) {
         var $editorContainer = $(element);
         var inputName = $editorContainer.data('input-id');
         var $editorInput = $('#' + inputName);
         var initialValue = $editorInput.val();
 
-        var editor = new toastui.Editor({
+        editor = new toastui.Editor({
             el: $editorContainer[0],
             usageStatistics: false,
             useCommandShortcut: true,
@@ -333,6 +334,11 @@ $(function () {
         });
     }
 
+    $('#GeneratedWidgetText').click(function (e) {
+        var txt = $('#GeneratedWidgetText').val();
+        editor.insertText(txt);
+    });
+    
     function createAddWidgetButton() {
         const button = document.createElement('button');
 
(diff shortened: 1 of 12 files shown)


modules/cms-kit/src/Volo.CmsKit.Admin.Web/Pages/CmsKit/BlogPosts/create.js
```diff
@@ -257,14 +257,15 @@ $(function () {
             initEditor(item);
         });
     }
-
+    
+    var editor;
```
enisn, line 255: we can create an array for editors because a page can contain more than one editor. In that case, this code works for only the last initialized editor.

enisn (changes requested):

modules/cms-kit/src/Volo.CmsKit.Admin.Web/Pages/CmsKit/Contents/addWidgetModal.js
```diff
@@ -48,16 +43,8 @@ $(function () {
...
-                    var fixedData = "<div>" + updatedText + "</div>";
-                    var innerHtml = $("#ContentEditor")[0].innerHTML;
-                    var replacedInnerHtml = innerHtml.replace('<div><br></div>', fixedData);
-                    $("#ContentEditor")[0].innerHTML = replacedInnerHtml;
-                }
-                else {
-                    $('.ProseMirror div').contents()[0].data = contentEditorText + updatedText;
-                }
+                $('#GeneratedWidgetText').val(updatedText);
+                $("#GeneratedWidgetText").trigger("click");
```
enisn: We can use 'change' event instead of 'click'. 

In that case, it's tricky and hard to understand what happens. At the first time, it feels like user clicks on it.

enisn:
```suggestion
                $("#GeneratedWidgetText").trigger("change");
```

modules/cms-kit/src/Volo.CmsKit.Admin.Web/Pages/CmsKit/Pages/create.js
```diff
@@ -171,6 +166,11 @@ $(function () {
         });
     }
 
+    $('#GeneratedWidgetText').click(function () {
```
enisn: After making it 'change' instead of 'click':

```suggestion
    $('#GeneratedWidgetText').on('change', function () {
```

enisn (changes requested):

enisn (approved):

PR #13287, opened 2022-07-12
Update Getting-Started-AspNetCore-Application.md

pasestars:
Add builder.Services.ReplaceConfiguration(builder.Configuration);

### DIFF
--- a/docs/en/Getting-Started-AspNetCore-Application.md
+++ b/docs/en/Getting-Started-AspNetCore-Application.md
@@ -77,6 +77,8 @@ using BasicAspNetCoreApplication;
 
 var builder = WebApplication.CreateBuilder(args);
 
+builder.Services.ReplaceConfiguration(builder.Configuration);
+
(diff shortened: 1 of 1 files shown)


maliming:
hi

The `AddApplicationAsync` method already did this.

https://github.com/abpframework/abp/blob/rel-5.3/framework/src/Volo.Abp.AspNetCore/Microsoft/Extensions/DependencyInjection/WebApplicationBuilderExtensions.cs#L19

pasestars:
I use from this in Program.cs and had an error.
image
image

maliming:
Please use the extension method of `WebApplicationBuilder`.

pasestars:
But in this example do that, and in doc had an error.
image
I just follow this [tutorial](https://docs.abp.io/en/abp/latest/Getting-Started-AspNetCore-Application)

maliming:
We will check and update the samples and documents.

maliming:
Move to https://github.com/abpframework/abp/pull/13289

PR #13357, opened 2022-07-21
Split JSON into `Newtonsoft` & `SystemTextJson` packages. 

### LINKED ISSUE #6095: Can we enable to set non-public properties on System.Text.Json
System.Text.Json doesn't set non-public properties by default on deserialization.
Is there an option to enable to set them (https://github.com/dotnet/runtime/issues/29743 ?)? If so, it is better to enable it.

### LINKED ISSUE #12124: Separate Volo.Abp.Json into 2 packages (Newtonsoft.Json and STJ)
For now, Volo.Abp.Json depends on Newtonsoft.Json
It will transfer the dependency to each project whether Newtonsoft.Json is used or not.

I think we should separate the codes about Newtonsoft.Json from Volo.Abp.Json into a new package Volo.Abp.Json.Newtonsoft.

### LINKED ISSUE #12338: [suggestion] AbpJsonOptions  add a string[] property for datetime formats?
when client send params have different date time format, I can not successful receipt.

maliming:
Resolve #12124 Resolve #6095 Resolve #12338

Since **SystemTextJson** supports more custom features in net 7, we no longer need the hybrid Json feature, [Check the docs to see the changes](https://github.com/abpframework/abp/blob/json/docs/en/JSON.md#configuration)

https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/custom-contracts


NuGet.Config
```diff
@@ -3,5 +3,6 @@
     <packageSources>
         <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
         <add key="BlazoriseMyGet" value="https://www.myget.org/F/blazorise/api/v3/index.json" />
+		<add key="automapperdev" value="https://www.myget.org/F/automapperdev/api/v3/index.json" />
```
PMExtra: Indent with spaces rather than tabs please.

framework/src/Volo.Abp.AspNetCore.Mvc/Volo/Abp/AspNetCore/Mvc/Json/MvcCoreBuilderExtensions.cs
```diff
@@ -1,28 +1,31 @@
...
-using Microsoft.Extensions.ObjectPool;
-using Volo.Abp.Json;
+using Volo.Abp.Json.SystemTextJson;
+using Volo.Abp.Json.SystemTextJson.JsonConverters;
 
 namespace Volo.Abp.AspNetCore.Mvc.Json;
 
 public static class MvcCoreBuilderExtensions
 {
     public static IMvcCoreBuilder AddAbpHybridJson(this IMvcCoreBuilder builder)
```
rqx110: since we are remove hybrid json, rename this function name to something else? e.g.  `AddAbpJson`

hikalkan (approved):

PR #13359, opened 2022-07-21
Move "host features" to the settings page

realLiangshiwei:
image


roc916:
Whether the content of the dialog should be moved to the settings page?

maliming (approved):

PR #13390, opened 2022-07-23
Dapr integration

fahadashiq12:
Want to integrate with .net 6.0, what things to keep in mind while making version compatibility for using complete abp functionalities.


fahadashiq12:
This is just for asking purpose for test project.

PR #13405, opened 2022-07-25
fix ro translated

totpero:
some translation fixes


EngincanV (approved):

EngincanV:
Thanks for your contribution @totpero.

PR #13429, opened 2022-07-27
chore(test): disable auto generator launchSettings.json

netcore-jroger:


maliming (approved):

maliming:
Thanks @netcore-jroger

PR #13468, opened 2022-07-29
Added Blog Module Localization

salihozkara:
closes https://github.com/volosoft/volo/issues/11259
image


EngincanV:
We'll make this change in the `rel-6.0` branch to make it available in the rc.2 release.

PR #13472, opened 2022-07-29
Cli: Make adding new empty module to solution faster

yekalkan:
Also, i've added a new option to `abp new` command:

`--skip-installing-libs`

This makes application/module creation done in a few seconds.


rqx110:
suggest the name of the option:
`--skip-install`

refer to angular cli: `ng new angular13-demo --skip-install`
image

realLiangshiwei (approved):

Issue #13274, opened 2022-07-08
CMS kit not render html content instead showing html code


juchunwei:
### Documentation

Please check the official documentation before asking questions: https://docs.abp.io

### GitHub Issues

* Your **ABP Framework version**.  5.3.1
* Your **User Interface** type (Angular/MVC/React... etc.) if the issue is related to a specific UI: MVC
* Your database provider(EF Core/MongoDB) : EF Core
* Exception message and **stack trace** if available (check the logs).
* Steps needed to **reproduce** the problem.

I added a page with these code,

![image](https://user-images.githubusercontent.com/22266738/177911059-2626c0ab-d0ea-4668-8b70-754c3eee34b9.png)

the page showing like this, it's not render the html content.

![image](https://user-images.githubusercontent.com/22266738/177911153-c83e2a01-99bc-4238-aa75-b245b48ae8a2.png)
![image](https://user-images.githubusercontent.com/22266738/177911200-93478e75-1b7e-4479-a2e8-a481a94c756a.png)

How can I render the html content? Thanks!

maliming:
Steps needed to reproduce the problem.

juchunwei:
@maliming 

1. Create a new page in CMS.  And add the Html code `<div class="text-danger">1233</div>`.  I want to show the 1233 in red. See the first image.
2. Then create a menu and link to the page. 
3. Open the page.  it's showing the html code but not 1233 in red. (the second image). 
4. This code works: `Markdown.ToHtml(page.Content, MarkdownPipeline);`

maliming:
@enisn 

What do you think?

enisn:
This might be related to https://github.com/abpframework/abp/pull/13082/files#diff-4e842adbbd7affc52ac0f2ed434255da13feeae68e6d739ab53732e8f84097b0R75

I'll check if this happens in earlier versions.

enisn:
It happens here at https://github.com/abpframework/abp/blob/dev/modules/cms-kit/src/Volo.CmsKit.Public.Web/Renderers/MarkdownToHtmlRenderer.cs#L27

Maybe we can make it optional

enisn:
Hi @juchunwei 

`allowHtmlTags` parameter has been added to `IMarkdownToHtmlRenderer`, and it'll be true by default for all Pages & Blog Posts excluding Comments.

This change will be shipped in v6.0

juchunwei:
@enisn Greate! Thanks!

enisn:
[...]

PR #13494, opened 2022-08-01
Introduce CmsKitRenderer Options

enisn:
Closes https://github.com/abpframework/abp/issues/13274


malikmasis (approved):

enisn:
This option manages xss protection system-wide. So It's not a good solution for now

PR #13496, opened 2022-08-01
Added filter for the `Slug` on Page

malikmasis:
closes https://github.com/volosoft/volo/issues/11267


modules/cms-kit/src/Volo.CmsKit.EntityFrameworkCore/Volo/CmsKit/Pages/EfCorePageRepository.cs
```diff
@@ -39,7 +39,7 @@ public EfCorePageRepository(IDbContextProvider<ICmsKitDbContext> dbContextProvid
         return await (await GetDbSetAsync()).WhereIf(
                 !filter.IsNullOrWhiteSpace(),
                 x =>
-                    x.Title.Contains(filter))
+                    x.Title.Contains(filter) || x.Slug.Contains(filter))
```
enisn: how about making the `Title` query case insensitive?

enisn (approved):

PR #13522, opened 2022-08-02
removed pages prefix

### LINKED ISSUE #12803: CMS Kit - Remove `/pages` prefix from pages
When a page created, is has default `/pages/{page-slug}`.

Only `/{page-slug}` can be used as page url.

malikmasis:
closes https://github.com/abpframework/abp/issues/12803


enisn:
image

We can update tooltips too

enisn:
I can't find any issue for this.

@gizemmutukurt Can you check this too? There shouldn't be a problem with existing module URLs

enisn (approved):

gizemmutukurt (changes requested):
The error in isuue will be fixed, it has been determined that it is not related to this issue

https://github.com/abpframework/abp/issues/13723

enisn:
> I added a page named "source" and when I try to navigate to it I get an error.
> 
> image
> 
> image

I think content is null in that case. Markdown rendering can be disabled if the content is null.

gizemmutukurt (approved):

enisn:
It doesn't break anything at the moment.

PR #13558, opened 2022-08-04
Added tag search by tag name

salihozkara:


modules/cms-kit/src/Volo.CmsKit.Domain.Shared/Volo/CmsKit/Tags/TagConsts.cs
```diff
@@ -4,7 +4,7 @@ namespace Volo.CmsKit.Tags;
 
 public static class TagConsts
 {
-    public static int MaxEntityTypeLength { get; set; } = CmsEntityConsts.MaxEntityTypeLength;
+    public static int MaxEntityTypeLength = CmsEntityConsts.MaxEntityTypeLength;
 
-    public static int MaxNameLength { get; set; } = 32;
+    public const int MaxNameLength = 32;
```
enisn: If we make it consts, there will be no chance to change it later. And If a developer customized this length, then it's a **breaking-change**. We can leave it as static

enisn (changes requested):

enisn (approved):

PR #13566, opened 2022-08-04
Docs module improvements 12661

### LINKED ISSUE #12661: Add "Project" filter to Docs module admin UI
Add "Project" filter (optional) as the first filter item:

image

Also;

- File name filter is unnecessary since Name does a similar thing with full path support. Remove this filter.
- Move these filters under to "Advanced filters" just like we do for other pages: Format, creation time, Last update, Last significant update, Cache time.
- Version, Format and language code textboxes can be smaller (they are too long, but the data is mostly very short, like `en` for language).

malikmasis:
closes https://github.com/abpframework/abp/issues/12661


enisn (approved):
LGTM, @berkansasmaz can you check too before merging?

berkansasmaz (approved):

berkansasmaz:
LGTM, we can start the @gizemmutukurt testing process.

gizemmutukurt:
The design should be improved.
image

Example:
image

gizemmutukurt:
text  should be corrected

image

malikmasis:
The above example's elements' sizes are similar but here elements' sizes are different, even so, I tried to make them better.
Also, I've corrected typos.
image

PR #13591, opened 2022-08-07
Fixed inconsistent modifier orders in solution

beriniwlew:


hikalkan:
@maliming we have [.editorconfig file](https://github.com/abpframework/abp/blob/dev/.editorconfig). Also, we are using Rider/Resharp that suggest the modifier sorting. Can you try to understand why we have inconsistent usage. Beyond this PR/issue, I also remember that the editor is suggesting to change already true sortings.

rqx110:
this PR go against rider suggestions
Rider/Resharp suggests:
- `override async Task<T>` not `async override Task<T>`
- `public sealed class`  not `sealed public class`
- `static async Task` not `async static Task`
- `static readonly` not `readonly static`

roc916:
> this PR go against rider suggestions Rider/Resharp suggests:
> 
> * `override async Task<T>` not `async override Task<T>`
> * `public sealed class`  not `sealed public class`
> * `static async Task` not `async static Task`
> * `static readonly` not `readonly static`

I agree, we should modify [.editorconfig#L123](https://github.com/abpframework/abp/blob/dev/.editorconfig#L123).

Exported from Rider: 
```
csharp_preferred_modifier_order = public, private, protected, internal, new, static, abstract, virtual, sealed, readonly, override, extern, unsafe, volatile, async, required:suggestion
```

enisn:
This is default Microsoft's order:
`public, private, protected, internal, static, extern, new, virtual, abstract, sealed, override, readonly, unsafe, volatile, async`
https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0036#csharp_preferred_modifier_order

beriniwlew:
Do you want me to close this PR?  I might not have time to work on this.   Perhaps we should create a new issue for it.

PR #13597, opened 2022-08-07
Error when install front libs,not found leptomX-rc

washyn:
when the packages are installed with yarn, it shows that the package does not exist and throws an error, in the version of the template without layers it is like this ""@abp/aspnetcore.mvc.ui.theme.leptonxlite": "~1.0.0-rc .1"" and everything works fine


EngincanV:
Hi @washyn, thanks for reporting the problem. We'll make the related changes within the day. We need to make the changes in the `rel-6.0` branch, to make it available in the next version(**v6.0-rc.2**). So, I close the PR. Thanks, again.

PR #13626, opened 2022-08-10
Upgrade to NET 7.0.

### LINKED ISSUE #13336: Upgrade to .NET 7.0
We will upgrade the framework, modules and all other components to .NET 7.0. Current latest preview is [Preview 7](https://devblogs.microsoft.com/dotnet/announcing-dotnet-7-preview-6/). However, we don't have to use this. I think we can wait for the next preview (probably will be released in August). We should be sure that all our dependencies can properly work with .NET 7.0.

maliming:
Resolve #13336


maliming:
https://github.com/dotnet/aspnetcore/issues/42212

hikalkan (approved):

PR #13635, opened 2022-08-10
CMS Kit: Set a page as the "home page"

### LINKED ISSUE #13350: CMS Kit: Set a page as the "home page"
We can be able to set a page (in the admin UI) as the "home page". Home page should be shown when developer hasn't created a home page in the application. I mean the developer can create a regular, development-time home page (like `/Pages/Index.cshtml`). In this case, the development-time home page is shown.

malikmasis:
closes https://github.com/abpframework/abp/issues/13350


modules/cms-kit/src/Volo.CmsKit.Public.Web/DefaultHomePageMiddleware.cs
```diff
@@ -0,0 +1,28 @@
...
+    private readonly IPagePublicAppService _pagePublicAppService;
+
+    public DefaultHomePageMiddleware(RequestDelegate next, IPagePublicAppService pagePublicAppService)
+    {
+        _next = next;
+        _pagePublicAppService = pagePublicAppService;
+    }
+    public async Task InvokeAsync(HttpContext httpContext)
+    {
+        var page = await _pagePublicAppService.FindDefaultHomePageAsync();
```
enisn, line 20: Making a database query for each request is not acceptable for the final version of this feature. We should consider using caching here.

malikmasis: `FindDefaultHomePageAsync` already uses caching

enisn: https://github.com/abpframework/abp/pull/13635/files#diff-b0973506765b84539d2a37f37e02274bc1d2f438b00810fad8e843ba99f2224dR42

Right

enisn (approved):

malikmasis:
image

PR #13657, opened 2022-08-11
Added _v to affect to the application immediately

### LINKED ISSUE #13620: Global Resources doesn't affect to the application immediately
Currently, the style endpoint is `/cms-kit/global-resources/style` and the browser caches the response of that URL. Even if you change it, the browser doesn't get the latest style and applies a cached one.

It might me something like that `/cms-kit/global-resources/style?v=123456abcdef`.

Adding a token will prevent caching even if the content is changed.

image

malikmasis:
closes https://github.com/abpframework/abp/issues/13620
image

### DIFF
--- a/modules/cms-kit/src/Volo.CmsKit.Public.Application.Contracts/Volo/CmsKit/Public/GlobalResources/GlobalResourceDto.cs
+++ b/modules/cms-kit/src/Volo.CmsKit.Public.Application.Contracts/Volo/CmsKit/Public/GlobalResources/GlobalResourceDto.cs
@@ -1,8 +1,13 @@
-﻿namespace Volo.CmsKit.Public.GlobalResources;
+﻿using System;
 
+namespace Volo.CmsKit.Public.GlobalResources;
+
+[Serializable]
 public class GlobalResourceDto
 {
     public string Name { get; set; }
-    
+
     public string Value { get; set; }
+
+    public DateTime? LastModificationTime { get; set; }
 }
--- a/modules/cms-kit/src/Volo.CmsKit.Public.Application.Contracts/Volo/CmsKit/Public/GlobalResources/GlobalResourceDto.cs
+++ b/modules/cms-kit/src/Volo.CmsKit.Public.Application.Contracts/Volo/CmsKit/Public/GlobalResources/GlobalResourceDto.cs
@@ -1,13 +1,12 @@
 ﻿using System;
+using Volo.Abp.Application.Dtos;
 
 namespace Volo.CmsKit.Public.GlobalResources;
 
 [Serializable]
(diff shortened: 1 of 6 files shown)


enisn:
It'll create a new Guid always when this component is rendered. Can we use the last updated timestamp for global resource or something like that? 

In that case, with the current way, the feature loses caching ability completely

malikmasis:
> It'll create a new Guid always when this component is rendered. Can we use the last updated timestamp for global resource or something like that?
> 
> In that case, with the current way, the feature loses caching ability completely

Updated

modules/cms-kit/src/Volo.CmsKit.Public.Web/Pages/CmsKit/Shared/Components/GlobalResources/Style/GlobalStyleViewComponent.cs
```diff
@@ -1,13 +1,33 @@
...
+    }
+
+    [BindProperty(SupportsGet = true)]
+    public DateTime? LastModificationTime { get; set; }
+
     public async Task<IViewComponentResult> InvokeAsync()
     {
-        return View("~/Pages/CmsKit/Shared/Components/GlobalResources/Style/Default.cshtml");
+        var lastModificationTime = (await GlobalResourcePublicAppService.GetGlobalStyleAsync())?.LastModificationTime;
+        var lastModificationTimeTimestamp = (long)(lastModificationTime.HasValue ? lastModificationTime.Value .Subtract(DateTime.UnixEpoch).TotalSeconds: 0);
```
rqx110: reformat this line?

modules/cms-kit/src/Volo.CmsKit.Public.Application.Contracts/Volo/CmsKit/Public/GlobalResources/GlobalResourceDto.cs
```diff
@@ -1,8 +1,13 @@
...
+
+[Serializable]
 public class GlobalResourceDto
 {
     public string Name { get; set; }
-    
+
     public string Value { get; set; }
+
+    public DateTime? LastModificationTime { get; set; }
```
enisn: We can use inheriting from `AuditedEntityDto` instead of defining `LastModificationTime` manually.

```diff
- public class GlobalResourceDto
+ public class GlobalResourceDto : AuditedEntityDto
{
    public string Name { get; set; }

    public string Value { get; set; }

-    public DateTime? LastModificationTime { get; set; }
}
```

enisn (changes requested):

enisn (approved):

enisn (approved):

PR #13662, opened 2022-08-12
Introduce `AddIdentityServerCookieAuthentication` to `AbpIdentityServerBuilderOptions`.

maliming:


modules/identityserver/src/Volo.Abp.IdentityServer.Domain/Volo/Abp/IdentityServer/AbpIdentityServerDomainModule.cs
```diff
@@ -110,6 +105,37 @@ private static void AddIdentityServer(IServiceCollection services)
...
+            .AddPluggableServices()
+            .AddValidators()
+            .AddResponseGenerators()
+            .AddDefaultSecretParsers()
+            .AddDefaultSecretValidators();
+
+        if (abpIdentityServerBuilderOptions.AddIdentityServerCookieAuthentication)
+        {
+            identityServerBuilder.AddCookieAuthentication();
+        }
```
gterdem, line 131: I am not sure if it is a good idea to add cookie authentication options at the **Domain** layer.

maliming: Identity Server also will do this by default.

https://github.com/IdentityServer/IdentityServer4/blob/3ff3b46698f48f164ab1b54d124125d63439f9d0/src/IdentityServer4/src/Configuration/DependencyInjection/IdentityServerServiceCollectionExtensions.cs#L40

gterdem (approved):

PR #13718, opened 2022-08-17
Cms-kit tag issues 13417

### LINKED ISSUE #13417: CmsKit TagFeature Problems
- [ ] When the tag length exceeds the max length, we must specify what the error is in the error message.
- [ ] When the number of tags exceeds the maximum number of tags, we must specify what the error is in the error message.
- [ ] If an empty tag goes, we must specify what the error is in the error message. for example "CSharp,,Ef"
- [ ] If the Tags go null via DTO, we can accept the tags as empty instead of giving an error.

malikmasis:
fixes https://github.com/abpframework/abp/issues/13417


modules/cms-kit/src/Volo.CmsKit.Admin.Application.Contracts/Volo/CmsKit/Admin/Tags/EntityTagSetDto.cs
```diff
@@ -1,15 +1,41 @@
...
+        var l = validationContext.GetRequiredService<IStringLocalizer<CmsKitResource>>();
+
+        foreach (var tag in Tags)
+        {
+            if (tag.Length > TagConsts.MaxNameLength)
+            {
+                yield return new ValidationResult(
+                    l[
+                        "MaxTagLengthExceptionMessage",
+                        nameof(tag),
```
rqx110:
```suggestion
                        tag,
```

modules/cms-kit/src/Volo.CmsKit.Admin.Application.Contracts/Volo/CmsKit/Admin/Tags/EntityTagSetDto.cs
```diff
@@ -1,15 +1,41 @@
...
+                        TagConsts.MaxNameLength,
+                        typeof(EntityTagSetDto).FullName
```
rqx110: really need this full name in Validation message?

malikmasis: I think it seems better to find the error easily.

modules/cms-kit/src/Volo.CmsKit.Domain.Shared/Volo/CmsKit/Localization/Resources/en.json
```diff
@@ -206,6 +206,7 @@
     "Feature:ReactionEnable": "Enabled reaction",
     "Feature:ReactionEnableDescription": "Enable reaction in the application.",
     "Feature:TagEnable": "Enabled taging",
-    "Feature:TagEnableDescription": "Enable taging in the application."
+    "Feature:TagEnableDescription": "Enable taging in the application.",
+    "MaxTagLengthExceptionMessage": "{0} length must be equal to or lower than {1}. {2}"
```
rqx110: I think the message can be improved, e.g.
"MaxTagNameLengthExceptionMessage": "Can't set more than {1} characters for a tag: {0}"

modules/cms-kit/src/Volo.CmsKit.Admin.Application.Contracts/Volo/CmsKit/Admin/Tags/EntityTagSetDto.cs
rqx110: I don't think the end-user need this detail info.

modules/cms-kit/src/Volo.CmsKit.Domain.Shared/Volo/CmsKit/Localization/Resources/en.json
enisn: We can use AspNetCore validation resource for this message instead of defining new one

enisn: [AbpValidationResource](https://github.com/abpframework/abp/blob/dev/framework/src/Volo.Abp.Validation/Volo/Abp/Validation/Localization/AbpValidationResource.cs) already has a message for the same validation. It's better to use that:

https://github.com/abpframework/abp/blob/dev/framework/src/Volo.Abp.Validation/Volo/Abp/Validation/Localization/en.json#L31

malikmasis: Thanks @enisn

enisn (approved):

PR #13720, opened 2022-08-17
Update Previews.md

roos-robert:
Clarification for new users (such as me) that to move over to the preview of a release the CLI needs to be upgraded to the latest RC as well.


hikalkan (approved):

hikalkan:
Thanks, that will be good.

PR #13724, opened 2022-08-17
Fix ConcurrencyStamp consistency issue for IdentityRole and IdentityUser

olicooper:
Noticed that updated Roles have a different ConcurrencyStamp format.


maliming:
Move to 6.0 https://github.com/abpframework/abp/pull/13727

Issue #13761, opened 2022-08-21
Install distributed locking to the startup templates with proper configurations


hikalkan:
Application startup template has different options. In some options (e.g. separate auth server, tiered, ...) we are installing Redis packages and configuring it by default.
I think we can configure distributed locking too, because some ABP components are using it and we should configure it for [clustered deployment](https://docs.abp.io/en/abp/latest/Deployment/Clustered-Environment#configuring-a-distributed-lock-provider).

So, basically, we should install and configure distributed locking with redis provider ([see documentation](https://docs.abp.io/en/abp/latest/Distributed-Locking#configuring-a-provider) to learn how to do) for each application that we configure Redis, in the startup templates.

PR #13763, opened 2022-08-22
Configure distributed locking

realLiangshiwei:
Resolves: https://github.com/abpframework/abp/issues/13761


maliming (approved):

hikalkan:
@maliming @realLiangshiwei this PR should also add `AbpDistributedLockingModule` to `DependsOn` for all projects you add project reference for `Volo.Abp.DistributedLocking` package. Please fix this. Also fix for the volo side.

PR #13806, opened 2022-08-24
Localization v7

fahadashiq12:
Localization comes in the context of i18n mostly where accessible things comes handy for example users who have disability option, eye sightedness and a lot of more issues can come into it and we don't have such requirement in asp.io if yes can we contribute for it?


fahadashiq12:
This is not something we should go for this site, it all depends.

PR #13810, opened 2022-08-25
Hide the Add button if the widget is not defined

malikmasis:


malikmasis:
Closed because of the wrong merged branch

PR #13822, opened 2022-08-25
Add LeptonX pro replace angular docs.

### LINKED ISSUE #565: Docs module should support to add a branch list
To read docs from given branch

mahmut-gundogdu:
[Resolves #565](https://github.com/volosoft/lepton/issues/565)

Questions:

- [ ] Is the location of the file right? or should I move to X?
- [ ] What is the parent menu of these docs?
- [ ] Is there any suggestion about the content of the docs?


ismcagdas:
- Location of the document is correct. 
- We can place the document under the right menu item when it is completed.

We need to explain each component briefly similar to https://github.com/abpframework/abp/pull/13898/files.

PR #13843, opened 2022-08-26
fix `User.IsActive` and `IsLockedOut` is invalid when authorization

roc916:
Reference link:

[AbpSignInManager.cs#L80](https://github.com/abpframework/abp/blob/dev/modules/identity/src/Volo.Abp.Identity.AspNetCore/Volo/Abp/Identity/AspNetCore/AbpSignInManager.cs#L80)

[aspnetcore/Identity/SignInManager.cs#L379](https://github.com/dotnet/aspnetcore/blob/main/src/Identity/Core/src/SignInManager.cs#L379)


maliming (approved):

maliming:
Thanks @roc916

PR #13876, opened 2022-08-30
Hide ABP schemas

### LINKED ISSUE #13874: HideAbpEndpoints() and schemas
Calling method HideAbpEndpoints() hides ABP endpoints.

What about ABP schemas ? There is a lot of them. 

How to hide them too ?

realLiangshiwei:
Resolves: https://github.com/abpframework/abp/issues/13874


EngincanV (commented):
I think we should exclude the **Volo.Abp.Http.** namespace or don't remove schemas at all. What do you think @realLiangshiwei ?

image

> I've added a sample application service (with one method implemented) and the above error appeared.

realLiangshiwei:
Yes, it's a problem, we should leave the choice to the developers

PR #13893, opened 2022-09-01
Only get application configuration when blazor wasm startup.

maliming:


hikalkan:
Will be done like https://github.com/abpframework/abp/commit/f6a3530eaf2f48bcb39ec532e46ab807c0955569

PR #13903, opened 2022-09-01
Implement layout hooks for blazor

### LINKED ISSUE #6261: Layout Hooks for the Blazor UI
We can create a layout hooks system for the Blazor UI as similar to the [MVC UI Layout Hooks](https://docs.abp.io/en/abp/latest/UI/AspNetCore/Layout-Hooks).
This is not so essential since the `index.html` is already in the application side.

EngincanV:
Resolves #6261 

---

⚠️ *LayoutHookInfo.cs*, *LayoutHookViewModel.cs*, *LayoutHooks.cs*, *AbpLayoutHookOptions.cs* classes have been moved under the **Volo.Abp.Ui.LayoutHooks** namespace.

### DIFF
--- a/framework/src/Volo.Abp.AspNetCore.Components.Web.Theming/Layout/StandardLayouts.cs
+++ b/framework/src/Volo.Abp.AspNetCore.Components.Web.Theming/Layout/StandardLayouts.cs
@@ -2,7 +2,5 @@
 
 public static class StandardLayouts
 {
-    public const string Main = "Main";
-    public const string TopMenu = "TopMenu";
-    public const string SideMenu = "SideMenu";
+    public const string Application = "Application";
 }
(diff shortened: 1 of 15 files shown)


maliming:
`Build failed for the solution: ../modules/cms-kit`

framework/src/Volo.Abp.AspNetCore.Components.Web.Theming/Layout/StandardLayouts.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace Volo.Abp.AspNetCore.Components.Web.Theming.Layout;
+
+public static class StandardLayouts
+{
+    public const string Main = "Main";
+    public const string TopMenu = "TopMenu";
+    public const string SideMenu = "SideMenu";
```
enisn: We shouldn't use it here. It's a theme-specific option. We can move `TopMenu` and `SideMenu` option to the LeptonX project.

Also `Main` should be renamed as `Application` and this file can contain only 1 layout since blazor has only Application layout

enisn (changes requested):

enisn (approved):

Issue #14000, opened 2022-09-12
Cofigure Bundling case insensitive


totpero:
Hi,

Related to https://github.com/abpframework/abp/issues/13612#issuecomment-1238756990 , I want to replace [index.js](https://github.com/abpframework/abp/blob/dev/modules/identity/src/Volo.Abp.Identity.Web/Pages/Identity/Roles/index.js) file from Roles module.

### The problem:
The script file path included in bundle is lowercase in [this code](https://github.com/abpframework/abp/blob/dev/modules/identity/src/Volo.Abp.Identity.Web/Pages/Identity/Roles/Index.cshtml#L28): src="/Pages/**identity/roles**/index.js"  and the folders names from this path is uppercase. 

I create the same path in my project like in your documentation [example](https://docs.abp.io/en/abp/latest/UI/AspNetCore/Customization-User-Interface#example-add-a-javascript-file-for-a-specific-page) and in my configuration I add this:

```
Configure<AbpBundlingOptions>(options =>
{
    options.ScriptBundles
        .Configure(
            typeof(Volo.Abp.Identity.Web.Pages.Identity.Roles.IndexModel).FullName,
            bundleConfig =>
            {
                bundleConfig.AddFiles("/Pages/Identity/Roles/index.js");
            });
});
```
The only difference is that I set the path with first folder letter uppercase and the result is that the file is added two times in the bundle:
![image](https://user-images.githubusercontent.com/1710257/189638870-6a7639b5-b7cf-46cf-8896-452efb8df327.png)

### The Solution:
In my case is simple, I follow your name used in bundle to replace the file.

1. I think  `bundleConfig.AddFiles(..` need to compare the file path using lowercase because you can't have two folders with the same name.
2. Or you need to use standard convention for the folders and change this [line](https://github.com/abpframework/abp/blob/dev/modules/identity/src/Volo.Abp.Identity.Web/Pages/Identity/Roles/Index.cshtml#L28) from: `/Pages/identity/roles/index.js` to: `/Pages/Identity/Roles/index.js`.

realLiangshiwei:
> Overriding a static embedded resource (like JavaScript, Css or image files) of a module is pretty easy. Just place a file in the same path in your solution and let the [Virtual File System](https://docs.abp.io/en/abp/latest/Virtual-File-System) to handle it.

https://docs.abp.io/en/abp/latest/UI/AspNetCore/Customization-User-Interface#overriding-static-resources

totpero:
[...]

PR #14003, opened 2022-09-13
Standard folder name

totpero:
Standard folder name related to #14000


realLiangshiwei:
Thanks, 
See: https://github.com/abpframework/abp/issues/14000#issuecomment-1244977776

PR #14017, opened 2022-09-14
Replace all date range inputs to daterange picker mvc] #13793

### LINKED ISSUE #13793: Replace all date range inputs to daterange picker [MVC]
Use the daterange picker that we used in the [EventHub](https://www.openeventhub.com) project in all the ABP modules.

Add this daterangepicker https://github.com/abpframework/abp/blob/dev/npm/packs/aspnetcore.mvc.ui.theme.shared/package.json

- JavaScript file: https://github.com/abpframework/eventhub/blob/main/src/EventHub.Web/Pages/Index.cshtml#L27
- CSS file: https://github.com/abpframework/eventhub/blob/main/src/EventHub.Web/Pages/Index.cshtml#L20
- Initialization code: https://github.com/abpframework/eventhub/blob/main/src/EventHub.Web/Pages/Index.js#L38

> after we change all these items, we need to make the same for Angular and Blazor. 
   so please note down which projects you made change 
   and make changes only in 1 Pull Request, don't create multiple PRs for each project

Here's the list of the PRO modules to be updated:
https://github.com/abpframework/abp/tree/dev/modules

Check all these projects _(if you see date start - date end replace it)_

- [x] [cms-kit](https://github.com/abpframework/abp/tree/dev/modules/cms-kit)
- [x] [docs](https://github.com/abpframework/abp/tree/dev/modules/docs)

image

salihozkara:
Resolves #13793


rqx110:
since we are using `DateRangePicker` component, why still using `SingleDatePicker` style in `cms-kit` and `doc` module?

gizemmutukurt:
The error I get in the doc module I run in the volo repo is as follows
image

salihozkara:
> since we are using `DateRangePicker` component, why still using `SingleDatePicker` style in `cms-kit` and `doc` module?

because the start and end dates are left optional, for example, only the end date is left as selectable.

gizemmutukurt (approved):

ebicoglu:
> 



> since we are using `DateRangePicker` component, why still using `SingleDatePicker` style in `cms-kit` and `doc` module?

as @salihozkara said, when the start & end date are optional, we cannot use DateRangePicker. Because [this tool](https://www.daterangepicker.com/) doesn't have an optional date selection.

enisn:
> The error I get in the doc module I run in the volo repo is as follows image

`abp install-libs` should solve that

gizemmutukurt:
> > The error I get in the doc module I run in the volo repo is as follows image
> 
> `abp install-libs` should solve that

Since two folders(moment and bootstrap-daterangepicker) are missing in the libs file under wwwroot, the abp install-libs command was not enough to fix the error. When the time comes, @salih will add the relevant folders.

ebicoglu (approved):

ebicoglu:
fix the conflicts

salihozkara:
> > The error I get in the doc module I run in the volo repo is as follows image
> 
> `abp install-libs` should solve that

[libs.zip](https://github.com/abpframework/abp/files/9601852/libs.zip) You can get rid of the error by throwing the two folders under the libs folder in the zip under the wwwroot/libs folder under the web project you are running.

PR #14040, opened 2022-09-16
cms-kit: Fix typo on `BlogFeatureAppService.cs`

EngincanV:


enisn:
It's a breaking change since it's a virtual method that can be overridden before.

enisn (approved):

PR #14164, opened 2022-09-28
Improve Dapr integration.

### LINKED ISSUE #14024: Dapr integration improvements
- [x] Remove `AbpAspNetCoreMvcDaprEventBusOptions.Contributors` and allow users to use the standard `[Topic(...)]` attribute on their controller actions. You can check the [[DaprEndpointRouteBuilderExtensions](https://github.com/dapr/dotnet-sdk/blob/master/src/Dapr.AspNetCore/DaprEndpointRouteBuilderExtensions.cs)](https://github.com/dapr/dotnet-sdk/blob/master/src/Dapr.AspNetCore/DaprEndpointRouteBuilderExtensions.cs) class to understand how you can find the routes. Our endpoint (`AbpAspNetCoreMvcDaprPubSubController`) should only handle the events those haven't registered with the `Topic` attribute.
- [x] `IDaprSerializer` has some unused methods, should we remove them?

Wait for https://github.com/abpframework/abp/pull/14015

maliming:
Resolve #14024

- [x] Remove `AbpAspNetCoreMvcDaprEventBusOptions.Contributors` and allow users to use the standard `[Topic(...)]` attribute on their controller actions. 
- [x]  Remove some unused methods in `IDaprSerializer `.
- [x]  <del>Change some methods to async. this causes **it is not possible to add `DaprClient` to DI by default, we always need to use `IAbpDaprClientFactory` to create it.**</del>


hikalkan:
> Change some methods to async. this causes it is not possible to add DaprClient to DI by default

I am not agree on that. I intentionally made all sync (I was aware about the results). My argument is that: Dapr's official document suggests it: https://learn.microsoft.com/en-us/dotnet/architecture/dapr-for-net-developers/getting-started#add-dapr-service-invocation

So, we don't want to break the natural usage and force ABP's services. My argument here is similar to use the Topic attribute instead of introducing a different infrastructure and configuration.

docs/en/Dapr/Index.md
```diff
@@ -299,8 +281,7 @@ public class MyController : AbpController
 {
     [HttpPost("/stock-changed")]
     [Topic("pubsub", "StockChanged")]
-    public async Task<IActionResult> TestRouteAsync(
-        [FromBody] StockCountChangedEto model)
+    public async Task<IActionResult> TestRouteAsync([FromBody] AbpDaprSubscriptionRequest<StockCountChangedEto> model)
```
hikalkan: Why should we introduce a class like `AbpDaprSubscriptionRequest`. It changes the normal usage and make the code ABP-dependent. I just want to understand if we must do something like that.

maliming:
> > Change some methods to async. this causes it is not possible to add DaprClient to DI by default
> 
> I am not agree on that. I intentionally made all sync (I was aware about the results). My argument is that: Dapr's official document suggests it: https://learn.microsoft.com/en-us/dotnet/architecture/dapr-for-net-developers/getting-started#add-dapr-service-invocation
> 
> So, we don't want to break the natural usage and force ABP's services. My argument here is similar to use the Topic attribute instead of introducing a different infrastructure and configuration.

Reverted.

docs/en/Dapr/Index.md
maliming: The Dapr request JSON is below, The `data` is our `Eto`.
```json
{
  "data": {
    "changedDate": "2022-11-07T09:34:12.2124673+08:00",
    "product": "Product 0"
  },
  "datacontenttype": "application/json",
  "id": "9eeca469-1e1a-4775-90e9-36fc73e5c960",
  "pubsubname": "test-pubsub",
  "source": "dapr-publisher",
  "specversion": "1.0",
  "time": "2022-11-07T09:34:12+08:00",
  "topic": "StockCountChanged",
  "traceid": "00-6098bef76cf1f691c42d9bc20e9912b4-db7c76d8bb993198-01",
  "traceparent": "00-6098bef76cf1f691c42d9bc20e9912b4-db7c76d8bb993198-01",
  "tracestate": "",
  "type": "com.dapr.event.sent"
}
```

```cs
public class AbpDaprSubscriptionRequest<T>
    where T : class
{
    public string PubSubName { get; set; }

    public string Topic { get; set; }

    public T Data { get; set; }
}
```

maliming:
```cs
public class AbpDaprSubscriptionRequest<T>
    where T : class
{
    public string PubSubName { get; set; }

    public string Topic { get; set; }

    public T Data { get; set; }
}
``

hikalkan (approved):

PR #14182, opened 2022-09-29
Cli: Check latest version one time per day

yekalkan:


framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/CliPaths.cs
```diff
@@ -10,6 +10,7 @@ public static class CliPaths
     public static string Log => Path.Combine(AbpRootPath, "cli", "logs");
     public static string Root => Path.Combine(AbpRootPath, "cli");
     public static string AccessToken => Path.Combine(AbpRootPath, "cli", "access-token.bin");
+    public static string Memory => Path.Combine(AbpRootPath, "cli", "memory.bin");
```
ebicoglu: don't save this setting in global file in the computer like "c:\users\alper\.cli\memory.bin"
save this in the same folder of the ABP CLI installation directory.
so that people can access the new version if they re-install the CLI.
each new file like "memory.bin" is an extra information for us to remember and maintain.

ebicoglu (approved):

PR #14197, opened 2022-10-01
Introduce `IHasEntityVersion` and `EntitySynchronizer`

### LINKED ISSUE #14196: Entity synchronizers and a new EntityVersion audit property
These are my personal opinion. I am unsure if it is a good idea. More suggestions are welcome. 🌹 

## Entity Synchronizers

Entity synchronizers are in high-frequency demand for distributed app and module development.

https://github.com/abpframework/abp/blob/48c52625f4c4df007f04d5ac6368b07411aa7521/modules/blogging/src/Volo.Blogging.Domain/Volo/Blogging/Users/BlogUserSynchronizer.cs#L9-L41

It subscribes entity-updated events from the ABP's distributed event bus and updates the local user entities.

I'm making an `EntitySynchronizer<TEntity, TKey, TExternalEntityEto>` abstract class (and an `EntitySynchronizer<TEntity, TExternalEntityEto>` for entities with composite PK, of course) to help.

The normal usage is like this:
```CSharp
public class BlogUserSynchronizer : EntitySynchronizer<BlogUser, Guid, UserEto>, ITransientDependency
{
    public BlogUserSynchronizer(IObjectMapper objectMapper, IRepository<BlogUser, Guid> repository) :
        base(objectMapper, repository)
    {
    }
}
```

ETO to entity mapping should be configured manually. We can use the Automapper's profile or override `MapToEntityAsync` in the BlogUserSynchronizer class.

By default, it synchronizes on the remote entities created/updated/deleted. You can set `IgnoreEntityCreatedEvent`, `IgnoreEntityUpdatedEvent`, and `IgnoreEntityDeletedEvent` of the base class to `true` to skip handling if you want.

## `EntityVersion` Audit Property

I define an `IHasEntityVersion` interface to introduce a new audit property.

Please see https://github.com/abpframework/abp/pull/11216#issuecomment-1006843999. If we break the strong ordering of the ABP event boxes, older EntityEto may be used to update local entities by mistake.

We can check the ETO's `EntityVersion` to resolve this problem. The synchronizer will skip handling if it is less than or equal to the local entity's version.

gdlcf88:
Resolve #14196

This PR includes the following features:

- Introduces `IHasEntityVersion` to add auto-versioning to entity classes.
- Introduces `EntitySynchronizer` as a base class to automatically sync an entity's properties from a source entity.


hikalkan:
Thanks @gdlcf88 the implementation looks pretty nice.

One change I want from you: We may not want to add versioning to our entity. I think versioning should be optional. If entity implements it, then it should be considered. But, if not implemented, it should just skip the version check.

Also, it would be great if you can write a simple document to explain how to use these system and how to establish an end-to-end entity synchronization using that. We will enhance the document, but an initial one would be good.

gdlcf88:
@hikalkan - I have finished the base class refactoring and docs.

BTW, I changed the base class name from `ExternalEntitySynchronizer` to `EntitySynchronizer`. Please tell me if that is not a good idea, and I can revert the change.

hikalkan:
>BTW, I changed the base class name from ExternalEntitySynchronizer to EntitySynchronizer. Please tell me if that is not a good idea, and I can revert the change.

No problem, I think it is a good change.

hikalkan (approved):

hikalkan:
Thanks @gdlcf88 for your great contribution. I will test it more and complete the documentation.

hikalkan:
I've revised and documented: https://github.com/abpframework/abp/pull/15134

PR #14207, opened 2022-10-03
Cli: Check login before suite command

yekalkan:


TronActive:
@yekalkan would this be the reason I cannot see the modules under the modules tab after the latest rc for 6?

yekalkan:
@joe704la No. But modules page will be fixed with the stable release.

framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/Commands/SuiteCommand.cs
```diff
@@ -42,17 +44,28 @@ public class SuiteCommand : IConsoleCommand, ITransientDependency
...
         _nuGetService = nuGetService;
+        _authService = authService;
         _cliHttpClientFactory = cliHttpClientFactory;
         Logger = NullLogger<SuiteCommand>.Instance;
     }
 
     public async Task ExecuteAsync(CommandLineArgs commandLineArgs)
     {
+#if !DEBUG    
+        var loginInfo = await _authService.GetLoginInfoAsync();
```
ebicoglu, line 61: check the `null` case
image

ebicoglu (approved):

PR #14243, opened 2022-10-06
dynamic form added nullable boolean render

blackWins:
```
public bool NotNullBoolean { get; set; }
public bool? NullableBoolean { get; set; }
```

image


maliming:
hi

I think the `Indeterminate` is more suitable.

https://getbootstrap.com/docs/5.0/forms/checks-radios/#indeterminate

image

blackWins:
sometime we use dynamic form render filter area, we want boolean type can set null 
image
like this ( `null` , `true` , `false` )
image

maliming:
We can add an option to determine the result(select or checkbox)

blackWins:
Use option to determine the result(select or checkbox) is not necessary

When use nullable `Boolean` there are three states ( `true`,`false` ,`null`) , using select should be more intuitive and convenient.

If the user only wants `true` or `false` , the checkbox is the best choice

blackWins:
code like this , use `[BooleanSelector]` Attribute
```
        public bool BooleanDefault { get; set; }

        [BooleanSelector(emptyText:"Null",falseText:"False",trueText:"True")]
        public bool? NullableBoolean { get; set; }

        [BooleanSelector]
        public bool NotNullBoolean { get; set; }
```
UI like this
image

PR #14441, opened 2022-10-25
Fix: Localization book types format

Dennis19075:
The Texts dictionary in ABP Enterprise has the correct format and ABP Framework has not.


maliming:
hi

I remember this is no problem, and I have a commit related to it.

https://github.com/abpframework/abp-samples/commit/241d9b9c60922574eb4433ab82172dd4b0c3a12f

Dennis19075:
Good morning.

Ok, it was just an observation. Because I invested a lot of time searching why the Book Type looks different. For people with no experience, this would be really helpful.

Sent with a Spark

PR #14453, opened 2022-10-26
Remove Closure Allocation from CurrentTenant, performance improve.

CrazyBaran:
I want to open discussion about some performance trick for our Abp framework. I right now made some exercises about .net performance and I found out that you do closure allocation in very commonly used CurrentTenant class. Right now I prepare some proof of concept which eliminate this unnecessary closure allocation . Question is if such performance improvements is something which I can contribute? I could made benchmark for us, how many allocation we save in this approach. If somebody would like to go deep into this topic I would suggest excellent articles from @danielmarbach.
[PerformanceTricksAzureSDK#be-aware-of-closure-allocations](https://github.com/danielmarbach/PerformanceTricksAzureSDK#be-aware-of-closure-allocations)
[pipeline-and-closure-allocations](https://particular.net/blog/pipeline-and-closure-allocations)
Give me a hint if I should go deeper with that.

@danielmarbach PS: Of course thank you for brilliant lecture on .NET Developer Days in Warsaw :)

Issue #541, opened 2018-10-10
Error monitoring - Sentry


bruno-garcia:
I couldn't find any support for a error monitoring tool.

Giving a green light from the maintainers, I'd be happy to PR support to [Sentry](https://github.com/getsentry/sentry) which is an open source error tracking tool.

Sentry has a SaaS [sentry.io](https://sentry.io/) which includes a free tier and also [can run on-premise](https://github.com/getsentry/onpremise).

There's a [new .NET SDK for Sentry which supports ASP.NET Core.](https://github.com/getsentry/sentry-dotnet)

hikalkan:
It would be better to do it after we complete this: https://github.com/abpframework/abp/issues/371
Follow it and send PR after we resolve it.

natiki:
> It would be better to do it after we complete this: #371
Agreed as the list of trackers is pretty much endless.... 

https://exceptionless.com/
https://elmah.io/
https://www.datadoghq.com/
https://stackify.com/

to name a few. Most have free tiers. So the ability to add providers via #371 will be the way to go.

bruno-garcia:
Which ones are open sourcr and can be self hosted free of charge?

Issue #371, opened 2018-07-23
ExceptionHandling package


hikalkan:
Used as a central subsystem to write handled & unhandled exceptions. We can then store or publish these exceptions or register to listen exceptions.

hikalkan:
Closing this. Opened a new issue: https://github.com/abpframework/abp/issues/3328

PR #14474, opened 2022-10-27
Sentry Instrumentation into Abp

CrazyBaran:
tl;dr;
Sometimes there is need for deep dive with Instrumentation of application to found out, what cause the problem. Automatic Sentry integration give only look into Query compiling, Database connection, Query Execution and, if we proper implement HttpClient in Abp, into HttpClient request. I would like to had full view of application also with abp dependencies.

I had hope that, @maliming didn't kill me for over excited to abp, after [Closure Allocation](https://github.com/abpframework/abp/pull/14453) but worth to try. I start develop Sentry Integration which could go deep with Instrumentation into abp framework. This is solid IT business feature, which I feel it is worthy to integrate. It could give very robust data for commercial users which look for good maintained system. My experience from asp.net zero told me that it is deserving to try. I have no longer access to my account to asp.net zero, but I remember that good integration with Relic Instrumentation, repair significant problems with couple of hard queries in our old friend asp.net boilerplate. I like open here discussion about Sentry, cause this is a stack, which I working right now. I had hope that it would be great cooperation. 
PS. I need wrote this anyway so maybe I could help other people.

Connected Issue which I found:
[https://github.com/abpframework/abp/issues/541](https://github.com/abpframework/abp/issues/541)
[https://github.com/abpframework/abp/issues/371](https://github.com/abpframework/abp/issues/371)

### DIFF
--- a/framework/Volo.Abp.sln
+++ b/framework/Volo.Abp.sln
@@ -431,6 +431,8 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Volo.Abp.DistributedLocking
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Volo.Abp.Instrumentation.Sentry", "src\Volo.Abp.Instrumentation.Sentry\Volo.Abp.Instrumentation.Sentry.csproj", "{5D8A62C7-CF33-4553-B1FE-BC3FF5F0ECB8}"
 EndProject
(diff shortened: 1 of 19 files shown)


wub:
Thanks for this @CrazyBaran. We'd definitely be interested in using something like this - currently, Sentry + ABP doesn't work very well.

CrazyBaran:
Sorry for delay, I had still in mind to get this done, but I had storm in my work right now.

PMExtra:
@maliming Have you reviewed this?

Issue #13946, opened 2022-09-07
Angular UI: External Localization Infrastructure Changes


hikalkan:
**Please implement this just after https://github.com/abpframework/abp/pull/13845 merged**

Localizations are separated from the Application Configuration Endpoint. The Angular UI should adapt the changes.

### What should be done

* Use the new endpoint (example: `/api/abp/application-localization?cultureName=tr`) to obtain the localization texts. You can call that endpoint just after the current `/api/abp/application-configuration` endpoint. The new endpoint also returns some information about the resources, like `baseResourceNames`. Base resource names will be used to determine localization values. So, the Angular UI should implement the inheritance logic in the UI.
* Set `?includeLocalizationResources=false` querystring parameter while consuming the current `/api/abp/application-configuration` endpoint. This will exclude localizations (`localization.values` will come empty).

Note that you normally don't deal with URLs and use the service proxies.

### Localization endpoint example result

````json
{
    "resources": {
        "TestApp": {
            "texts": {
                "TEST": "Bu bir denemedir 7",
                "TEST2": "This is the second test v3",
                "PermissionDynamic1": "PermissionDynamic1 localized value 2"
            },
            "baseResources": ["AbpValidation"]
        }
    }
}
````

Full result: [localization-result.zip](https://github.com/abpframework/abp/files/9503811/localization-result.zip)

### Inheritance logic

I've implemented the inheritance logic for MVC UI JavaScript side: https://github.com/abpframework/abp/pull/13845/files#diff-523599b6c73c637ed5f2b9da2e6977e37e75cd599bda945c288d6e10480fc7ff (find `abp.js` file in the PR if link doesn't work)
It basically try to get a localization text from base resources, if it can't find in the original resource. Base resource check must be RECURSIVE (it should check bases, then their bases if not found, and so on), like I've done.

PR #14480, opened 2022-10-28
No longer include parent cultures in `ApplicationLocalizationConfigurationDto`.

### LINKED ISSUE #14413: Same localization key/value pairs in `ApplicationLocalizationConfigurationDto`.
https://support.abp.io/QA/Questions/3907/same-localization-keyvalue-pairs-in-ApplicationLocalizationConfigurationDto

maliming:
Resolve #14413

### DIFF
--- a/framework/src/Volo.Abp.AspNetCore.Mvc/Volo/Abp/AspNetCore/Mvc/ApplicationConfigurations/AbpApplicationConfigurationAppService.cs
+++ b/framework/src/Volo.Abp.AspNetCore.Mvc/Volo/Abp/AspNetCore/Mvc/ApplicationConfigurations/AbpApplicationConfigurationAppService.cs
@@ -231,7 +231,7 @@ public class AbpApplicationConfigurationAppService : ApplicationService, IAbpApp
 
                 if (localizer != null)
                 {
-                    foreach (var localizedString in await localizer.GetAllStringsAsync())
+                    foreach (var localizedString in await localizer.GetAllStringsAsync(includeParentCultures: false))
                     {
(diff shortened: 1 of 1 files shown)


hikalkan:
I will check this. Did you think the side effects? Does everything work as expected with this?

maliming:
I checked MVC and Blazor will work. But I'm not sure about the angular, We should check it.

https://github.com/abpframework/abp/blob/9d2902b335ca089be6ae07242d4470472de39cf5/framework/src/Volo.Abp.AspNetCore.Components.WebAssembly/Volo/Abp/AspNetCore/Components/WebAssembly/WebAssemblyCachedApplicationConfigurationClient.cs#L47

https://github.com/abpframework/abp/blob/9d2902b335ca089be6ae07242d4470472de39cf5/framework/src/Volo.Abp.AspNetCore.Mvc.Client/Volo/Abp/AspNetCore/Mvc/Client/MvcCachedApplicationConfigurationClient.cs#L79

hikalkan:
TL;TR; We don't need to this change, 7.0 already optimizes the inherited values and much more.

Localization in the Application Configuration Endpoint does exists only for backward compatibility. It will no longer be used. We'd created a separate endpoint for localization (https://github.com/abpframework/abp/pull/13845). We set `IncludeLocalizationResources` to `false` from our user interfaces (Blazor and MVC is done, Angular is in dev (https://github.com/abpframework/abp/issues/13946)).

You couldn't see that as braking change, because MVC and Blazor already doesn't use it. However, this PR's change is an important breaking change for existing clients and breaks backward compatibility. I am closing that.

PR #14502, opened 2022-10-31
dependency injection replacement refactor

linch90:
Hi, there is a scene, I defined a `IMyFilter` interface, and implement it with several filter classes in my module.
```csharp
[ExposeServices(typeof(IMyFilter))]
public class MyFilterA : IMyFilter
{   
}

[ExposeServices(typeof(IMyFilter))]
public class MyFilterB : IMyFilter
{   
}

[ExposeServices(typeof(IMyFilter))]
public class MyFilterC : IMyFilter
{   
}
```
Every filter class has its own duty.

In the project that reference the module, it has a minor different implementation for `MyFilterB`, such as
```csharp
[ExposeServices(typeof(IMyFilter))]
public class MyFilterBOverride : MyFilterB
{
     // override the implementation of the base type: MyFilterB
}
```
In current abp framework, it use default `IServiceCollection.Replace(ServiceDescriptor descriptor)` method. It may replace the `MyFilterA`, `MyFilterB` or `MyFilterC`. This is not the result I want.

What I want is, the `MyFilterBOverride` just override the `MyFilterB`, not others.

### DIFF
--- a/framework/src/Volo.Abp.Core/Microsoft/Extensions/DependencyInjection/ServiceCollectionCommonExtensions.cs
+++ b/framework/src/Volo.Abp.Core/Microsoft/Extensions/DependencyInjection/ServiceCollectionCommonExtensions.cs
@@ -2,6 +2,7 @@ using System;
 using System.Linq;
 using System.Reflection;
 using JetBrains.Annotations;
+using Microsoft.Extensions.DependencyInjection.Extensions;
 using Volo.Abp;
 using Volo.Abp.Reflection;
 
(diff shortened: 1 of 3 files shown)


maliming:
How about this way?

```cs
[ExposeServices(typeof(MyFilterB))]
public class MyFilterBOverride : MyFilterB
{

}
```

linch90:
@maliming Hi, I tried your way, revert the `services.ReplaceBaseServiceOrDefault(serviceDescriptor);` to `services.Replace(serviceDescriptor);` and use
```csharp
[ExposeServices(typeof(MyFilterB))]
public class MyFilterBOverride : MyFilterB
{
}
```
then I run the test `Should_Replace_Base_Implementation_By_ReplaceService`:
image
The MyFilterB is not replaced by MyFilterBOverride.

maliming:
We also need to deal with such cases.

```cs
public class MyFilterB_A: MyFilterB
{

}

public class MyFilterBOverride : MyFilterB_A
{

}
```

I think it can be easily implemented by directly changing `IServiceCollection` manually.

Especially during `ConfigureServices` many situations are unforeseen.

linch90:
@maliming Yes, you are right, this can not deal with all the situation.

PR #14528, opened 2022-11-03
Made Container protected to allow reuse

### LINKED ISSUE #14517: Make _container in BlobContainer protected and methods virtual.
https://github.com/abpframework/abp/blob/3427ba632fd8f3947ca65db40a0119674d05e5fc/framework/src/Volo.Abp.BlobStoring/Volo/Abp/BlobStoring/BlobContainer.cs#L13

Right now I made custom BlobProvider for Azure to had possibility to create SasToken rather to push all file throw application. From documentation we had information that if we like to Customize we should override AzureBlobProvider.
https://docs.abp.io/en/abp/latest/Blob-Storing-Azure#other-services
but cause this _container is private we cannot really Customize BlobContainer, after customization of Provider, so it is hard to reuse code and add new methods to BlobContainer. Let me know if my understanding is proper.  tl;dr; I would like to extend BlobContainer with new methods.
Walkaround is of course again use blobContainerFactory.Create<TContainer>(); in constructor of custom blob, but as you could understand it is little weird solution.

CrazyBaran:
Resolve #14517
I don't see right now reason to made method virtual, if underling Container had virtual methods.
@maliming Up

### DIFF
--- a/framework/src/Volo.Abp.BlobStoring/Volo/Abp/BlobStoring/BlobContainer.cs
+++ b/framework/src/Volo.Abp.BlobStoring/Volo/Abp/BlobStoring/BlobContainer.cs
@@ -10,11 +10,11 @@ namespace Volo.Abp.BlobStoring;
 public class BlobContainer<TContainer> : IBlobContainer<TContainer>
     where TContainer : class
 {
-    private readonly IBlobContainer _container;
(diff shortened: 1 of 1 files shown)


maliming (approved):

maliming:
Thanks @CrazyBaran

PR #14561, opened 2022-11-07
Angular UI: External Localization Infrastructure Changes

### LINKED ISSUE #13946: Angular UI: External Localization Infrastructure Changes

mahmut-gundogdu:
Resolves #13946

### DIFF
--- a/npm/ng-packs/packages/core/src/lib/services/config-state.service.ts
+++ b/npm/ng-packs/packages/core/src/lib/services/config-state.service.ts
@@ -13,7 +13,8 @@ import { InternalStore } from '../utils/internal-store-utils';
 })
 export class ConfigStateService {
   private readonly store = new InternalStore({} as ApplicationConfigurationDto);
-
+  private includeLocalizationResources = false;
+  
   get createOnUpdateStream() {
     return this.store.sliceUpdate;
   }
@@ -26,7 +27,7 @@ export class ConfigStateService {
 
   private initUpdateStream() {
     this.updateSubject
-      .pipe(switchMap(() => this.abpConfigService.get()))
+      .pipe(switchMap(() => this.abpConfigService.get({ includeLocalizationResources:this.includeLocalizationResources})))
       .subscribe(res => this.store.set(res));
   }
 
--- a/npm/ng-packs/packages/core/src/lib/services/config-state.service.ts
+++ b/npm/ng-packs/packages/core/src/lib/services/config-state.service.ts
@@ -1,7 +1,8 @@
 import { Injectable } from '@angular/core';
-import { Observable, Subject } from 'rxjs';
-import { map, switchMap, take } from 'rxjs/operators';
+import { Observable, Subject,of } from 'rxjs';
+import { map, switchMap, take, } from 'rxjs/operators';
 import { AbpApplicationConfigurationService } from '../proxy/volo/abp/asp-net-core/mvc/application-configurations/abp-application-configuration.service';
+import { AbpApplicationLocalizationService } from '../proxy/volo/abp/asp-net-core/mvc/application-configurations/abp-application-localization.service';
 import {
   ApplicationConfigurationDto,
   ApplicationGlobalFeatureConfigurationDto,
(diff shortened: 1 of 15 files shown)


npm/ng-packs/packages/core/src/lib/services/config-state.service.ts
```diff
@@ -1,7 +1,8 @@
 import { Injectable } from '@angular/core';
-import { Observable, Subject } from 'rxjs';
-import { map, switchMap, take } from 'rxjs/operators';
+import { Observable, Subject,of } from 'rxjs';
+import { map, switchMap, take, } from 'rxjs/operators';
```
muhammedaltug: please reformat file

npm/ng-packs/packages/core/src/lib/services/config-state.service.ts
```diff
@@ -13,21 +14,46 @@ import { InternalStore } from '../utils/internal-store-utils';
...
+        switchMap(() =>
+          this.abpConfigService.get({
+            includeLocalizationResources: this.includeLocalizationResources,
+          }),
+        ),
+      )
+      .pipe(switchMap(appState => this.getLocalizationAndCombineWithAppState(appState)))
+      .subscribe(res => {
+        return this.store.set(res);
+      });
```
muhammedaltug: unnecessary return

npm/ng-packs/packages/core/src/lib/services/localization.service.ts
```diff
@@ -250,3 +264,43 @@ export class LocalizationService {
...
+  const item = source[baseResourceName];
+
+  if (item.baseResources.length === 0) {
+    return item;
+  }
+
+  return item.baseResources.reduce((acc,baseResource) => {
+    const baseItem = recursivelyMergeBaseResources(baseResource, source);
+    const texts = { ...baseItem.texts, ...item.texts };
+    return  {...acc,texts}
```
muhammedaltug: please reformat file

npm/ng-packs/packages/core/src/lib/services/localization.service.ts
```diff
@@ -250,3 +264,43 @@ export class LocalizationService {
...
+  },item);
+  
+};
+
+function mergeResourcesWithBaseResource(resource: ResourceDto):ResourceDto {
+  const entities = Object.keys(resource).map((key) => {
+    const newValue =  recursivelyMergeBaseResources(key,resource)
+    return [key,newValue]
+  });
+  return entities.reduce((acc,[key,value]) => ({...acc,[key]:value}),{})
```
muhammedaltug: please reformat file

muhammedaltug (approved):

PR #14597, opened 2022-11-10
Set tenant admin's password from host side

malikmasis:
closes https://github.com/volosoft/volo/issues/12067


modules/identity/src/Volo.Abp.Identity.EntityFrameworkCore/Volo/Abp/Identity/EntityFrameworkCore/EfCoreIdentityUserRepository.cs
```diff
@@ -301,4 +302,13 @@ public override async Task<IQueryable<IdentityUser>> WithDetailsAsync()
     {
         return (await GetQueryableAsync()).IncludeDetails();
     }
+
+    public virtual async Task<IdentityUser> GetUserByTenantIdAndUserNameAsync(Guid tenantId, [NotNull] string userName, bool includeDetails = true, CancellationToken cancellationToken = default)
```
hikalkan: * You've declared, but not used the `includeDetails` parameter. See other similar methods to understand how to use.
* If a method may return null, then its name should start with `Find`, not `Get`
* A good naming can be `FindByTenantIdAndUserNameAsync`. Don't use `User`, we are already in the user context.

hikalkan (changes requested):

hikalkan (approved):

modules/identity/src/Volo.Abp.Identity.Domain/Volo/Abp/Identity/IIdentityUserRepository.cs
```diff
@@ -102,4 +102,11 @@ public interface IIdentityUserRepository : IBasicRepository<IdentityUser, Guid>
         bool? notActive = null,
         CancellationToken cancellationToken = default
     );
+
+    Task<IdentityUser> FindByTenantIdAndUserNameAsync(
+        Guid tenantId,
```
hikalkan: Make the `tenantId` nullable, so the method can be used for host users too.

hikalkan (changes requested):

hikalkan (approved):

Issue #4496, opened 2020-06-25
Introduce the DynamicStringLengthAttribute


hikalkan:
`StringLengthAttribute` gets constant values as max and min lengths. Example:

````csharp
public class PersonDto
{
    [Required]
    [StringLength(64)]
    public string Name { get; set; }
}
````

So, if you want to determine the length on runtime (for example, you want to set from a configuration file) that is not possible, because of the nature of the attributes in .net.

Introducing the `DynamicStringLengthAttribute` which can determine the value from a static property of a class.

### Example Usage

1. Create a class to define the static property:

````csharp
public static class PersonConstraints
{
    public static int MaxNameLength { get; set; } = 64;
}
````

2. Use the `DynamicStringLengthAttribute` on your DTO:

````csharp
public class PersonDto
{
    [Required]
    [DynamicStringLength(typeof(PersonConstraints), nameof(PersonConstraints.MaxNameLength))]
    public string Name { get; set; }
}
````

3. You can set the `PersonConstraints.MaxNameLength` before its first usage, generally in the `PreConfigureServices` method of your module:

````csharp
public override void PreConfigureServices(ServiceConfigurationContext context)
{
    PersonConstraints.MaxNameLength = 128;
}
````

It is still not possible to change the value after application initialization, but this makes possible to implement https://github.com/abpframework/abp/issues/4139 to us. Also, you can use this feature in your applications, while you generally may not need it.

PR #14648, opened 2022-11-14
Password show hiden feature 

malikmasis:
closes https://github.com/volosoft/volo/issues/10076


modules/tenant-management/src/Volo.Abp.TenantManagement.Domain.Shared/Volo/Abp/TenantManagement/TenantConsts.cs
```diff
@@ -6,4 +6,8 @@ public static class TenantConsts
     /// Default value: 64
     /// </summary>
     public static int MaxNameLength { get; set; } = 64;
+
+    public const int MaxPasswordLength = 128; 
+    
+    public const int AdminEmailAddress = 256;
```
rqx110: rename to `MaxAdminEmailLength` is better. and why here use `const` not `static`?
or can we reuse `AbpUserConsts.MaxEmailLength` ?:https://github.com/abpframework/abp/blob/48c02d787503356fcc56b9234cb7f11e23a5a8be/modules/users/src/Volo.Abp.Users.Domain.Shared/Volo/Abp/Users/AbpUserConsts.cs#L23
just like:
https://github.com/abpframework/abp/blob/48c02d787503356fcc56b9234cb7f11e23a5a8be/modules/identity/src/Volo.Abp.Identity.Domain.Shared/Volo/Abp/Identity/IdentityUserConsts.cs#L15

modules/tenant-management/src/Volo.Abp.TenantManagement.Blazor/Pages/TenantManagement/TenantManagement.razor.cs
```diff
@@ -127,4 +126,9 @@ protected override ValueTask SetTableColumnsAsync()
 
         return base.SetTableColumnsAsync();
     }
+
+    protected virtual void ShowHidePassword(bool isShowPassword)
```
rqx110: no need parameter `isShowPassword` here.
simple:
```csharp
protected virtual void ShowHidePassword()
{
    IsShowPassword = !IsShowPassword;
}
```

modules/identity/src/Volo.Abp.Identity.Blazor/Pages/Identity/UserManagement.razor.cs
```diff
@@ -228,16 +229,21 @@ protected override ValueTask SetToolbarItemsAsync()
         return base.SetToolbarItemsAsync();
     }
 
-    protected virtual void ChangePasswordTextRole(TextRole? textRole)
+    protected virtual void ChangePasswordTextRole(TextRole? textRole, bool? isShowPassword = null)
```
rqx110: no need `isShowPassword` parameter here. Can use `IsShowPassword` property directly in this function

modules/tenant-management/src/Volo.Abp.TenantManagement.Domain.Shared/Volo/Abp/TenantManagement/TenantConsts.cs
malikmasis: 1. I need to the const value because it should be changed
2. Attributes accept just const value

modules/tenant-management/src/Volo.Abp.TenantManagement.Blazor/Pages/TenantManagement/TenantManagement.razor.cs
malikmasis: It's simpler but not a safe behaviour

modules/identity/src/Volo.Abp.Identity.Blazor/Pages/Identity/UserManagement.razor.cs
malikmasis: It's simpler but not a safe behaviour

modules/tenant-management/src/Volo.Abp.TenantManagement.Domain.Shared/Volo/Abp/TenantManagement/TenantConsts.cs
rqx110: const value will not allow developer to change outside of module.
use `DynamicStringLengthAttibute`, see https://github.com/abpframework/abp/issues/4496

https://github.com/abpframework/abp/blob/5ef9707adc458851c3bfe0c03063811c2596e8c5/modules/tenant-management/src/Volo.Abp.TenantManagement.Web/Pages/TenantManagement/Tenants/CreateModal.cshtml.cs#L40

modules/tenant-management/src/Volo.Abp.TenantManagement.Blazor/Pages/TenantManagement/TenantManagement.razor.cs
rqx110: sorry, can you explain what's mean of **not a safe behaviour**

malikmasis: yw, to avoid multiple instances

modules/account/src/Volo.Abp.Account.Web/Pages/Account/Login.js
```diff
@@ -0,0 +1,22 @@
+$(function () {
+    $("#PasswordVisibilityButton").click(function (e) {
```
enisn, line 2: Using class instead of id might be more useful than this. In that way the page supports multiple password fields

modules/identity/src/Volo.Abp.Identity.Blazor/Pages/Identity/UserManagement.razor.cs
```diff
@@ -35,12 +35,13 @@ public partial class UserManagement
     protected string CreateModalSelectedTab = DefaultSelectedTab;
 
     protected string EditModalSelectedTab = DefaultSelectedTab;
+    protected bool IsShowPassword { get; set; }
```
enisn: The name of the property is confusing. It can be `IsPasswordShown` or `ShowPassword`

modules/identity/src/Volo.Abp.Identity.Blazor/Pages/Identity/UserManagement.razor.cs
enisn: It seems `isShowPassword` isn't necessary as a method parameter. That value is already known in the class.

enisn (changes requested):

gizemmutukurt (approved):

enisn (approved):

PR #14655, opened 2022-11-14
Croatian translations

plukic:
Croatian translations based on this link:
https://docs.abp.io/en/abp/2.5/Contribution/Localization-Text-Files


hikalkan:
Thanks @plukic for your contribution. We will check it in a short time.

maliming (approved):

PR #14726, opened 2022-11-18
Introduce DeleteDirectAsync method for IRepository.

### LINKED ISSUE #14725: Introduce DeleteDirectAsync method for IRepository
EF Core 7 has a new `ExecuteDeleteAsync` extension method: https://learn.microsoft.com/en-us/ef/core/what-is-new/ef-core-7.0/whatsnew#basic-executedelete-examples

We can create a new method on the `IRepository<TEntity>` interface and implement it for EF Core and MongoDB.

````csharp
    Task DeleteDirectAsync(
        [NotNull] Expression<Func<TEntity, bool>> predicate,
    );
````

Add a summary on top of the method like:

>Deletes all entities those fit to the given predicate. It directly deletes entities from database, without fetching them. Some features (like soft-delete, multi-tenancy and audit logging) won't work, so use this method carefully when you need it. Use the `DeleteAsync` method if you need to these features.

Note that: For the ORMs doesn't support direct delete, we can fallback to the [existing DeleteAsync method](https://github.com/abpframework/abp/blob/dev/framework/src/Volo.Abp.Ddd.Domain/Volo/Abp/Domain/Repositories/IRepository.cs#L66) (set `autoSave` to true if you fallback).

### PR DESCRIPTION
**Pull request by maliming**
Resolve #14725

### CODE REVIEW

File: framework/src/Volo.Abp.MongoDB/Volo/Abp/Domain/Repositories/MongoDB/MongoDbRepository.cs [hunk 1]
```diff
@@ -484,6 +484,11 @@ public async override Task<long> GetCountAsync(CancellationToken cancellationTok
         await DeleteManyAsync(entities, autoSave, cancellationToken);
     }
 
+    public override async Task DeleteDirectAsync(Expression<Func<TEntity, bool>> predicate, CancellationToken cancellationToken = default)
+    {
+        await DeleteAsync(predicate, true, cancellationToken);
```
hikalkan: Does't MongoDB have a way of doing that? If it has, please also properly implement it for MongoDB too.

hikalkan (approved):

PR #14842, opened 2022-11-28
Introduce `IAbpHostEnvironment` and reduce cache durations for development environment.

### LINKED ISSUE #13952: Reduce cache durations for development environment
In development environment, we can consider to disable caching or reducing cache durations for some points, like the `MvcCachedApplicationConfigurationClient` class. Such caches are good for production, but may confuse developers (they may not see result of some changes immediately). Create a list of such points we can change, then implement it.

maliming:
Resolve #13952

The logic of `IAbpHostEnvironment` is:

1. It will use the value from `AbpApplicationCreationOptions`.
```cs
AbpApplicationFactory.Create<IndependentEmptyModule>(options =>
{
   options.Environment = Environments.Staging;
})
```

2. If `options.Environment` is null, It will use `IWebHostEnvironment` and `IWebAssemblyHostEnvironment` values when you use ASP NET Core or Blazor wasm
3. If `environment` is still null value, It will use `Production` as default.


framework/src/Volo.Abp.Core/Volo/Abp/AbpApplicationBase.cs
```diff
@@ -61,6 +66,15 @@ public abstract class AbpApplicationBase : IAbpApplication
         {
             ConfigureServices();
         }
+
+        if (options.Environment.IsNullOrWhiteSpace())
+        {
+            var abpHostEnvironment = services.GetSingletonInstance<IAbpHostEnvironment>();
+            if (abpHostEnvironment.EnvironmentName.IsNullOrWhiteSpace())
+            {
+                abpHostEnvironment.EnvironmentName = Environments.Production;
```
hikalkan: It always comes as Production, with this setup. Because `options.SkipConfigureServices` is always `true`, so the aspnet core module, for example, will set the configuration later. Test without manually setting it, you will see it won't get from aspnet core.

framework/src/Volo.Abp.IdentityModel/Volo/Abp/IdentityModel/IdentityModelAuthenticationService.cs
```diff
@@ -97,11 +100,12 @@ public virtual async Task<string> GetAccessTokenAsync(IdentityClientConfiguratio
...
             tokenCacheItem = new IdentityModelTokenCacheItem(tokenResponse.AccessToken);
-            await TokenCache.SetAsync(cacheKey, tokenCacheItem,
-                new DistributedCacheEntryOptions
-                {
-                    AbsoluteExpirationRelativeToNow = TimeSpan.FromSeconds(configuration.CacheAbsoluteExpiration)
-                });
+            await TokenCache.SetAsync(cacheKey, tokenCacheItem, new DistributedCacheEntryOptions
+            {
+                AbsoluteExpirationRelativeToNow = AbpHostEnvironment.IsDevelopment()
+                    ? TimeSpan.Zero
```
hikalkan: Instead of zero, we may set a value like 5 seconds.

framework/src/Volo.Abp.IdentityModel/Volo/Abp/IdentityModel/IdentityModelAuthenticationService.cs
```diff
@@ -146,7 +150,9 @@ protected virtual async Task<IdentityModelDiscoveryDocumentCacheItem> GetDiscove
             await DiscoveryDocumentCache.SetAsync(tokenEndpointUrlCacheKey, discoveryDocumentCacheItem,
                 new DistributedCacheEntryOptions
                 {
-                    AbsoluteExpirationRelativeToNow = TimeSpan.FromSeconds(configuration.CacheAbsoluteExpiration)
+                    AbsoluteExpirationRelativeToNow = AbpHostEnvironment.IsDevelopment()
+                        ? TimeSpan.Zero
```
hikalkan: Instead of zero, we may set a value like 5 seconds.

framework/src/Volo.Abp.AspNetCore.Mvc.Client/Volo/Abp/AspNetCore/Mvc/Client/AbpAspNetCoreMvcClientModule.cs
```diff
@@ -9,5 +11,16 @@ namespace Volo.Abp.AspNetCore.Mvc.Client;
...
-    
+    public override void ConfigureServices(ServiceConfigurationContext context)
+    {
+        var abpHostEnvironment = context.Services.GetAbpHostEnvironment();
+        if (abpHostEnvironment.IsDevelopment())
+        {
+            Configure<AbpAspNetCoreMvcClientCacheOptions>(options =>
+            {
+                options.TenantConfigurationCacheAbsoluteExpiration = TimeSpan.Zero;
+                options.ApplicationConfigurationDtoCacheAbsoluteExpiration = TimeSpan.Zero;
```
hikalkan: Instead of zero, we may set a value like 5 seconds. Otherwise, it will make request to server too frequently, even multiple times in a single HTTP request.

hikalkan (changes requested):

hikalkan (approved):

Issue #14716, opened 2022-11-17
Use EF Core 7 ExecuteDelete for IRepository.DeleteAsync with predicate


hikalkan:
EF Core 7 has a new `ExecuteDeleteAsync` extension method: https://learn.microsoft.com/en-us/ef/core/what-is-new/ef-core-7.0/whatsnew#basic-executedelete-examples

We can use it in `EfCoreRepository.DeleteAsync` method:

````csharp
    Task DeleteAsync(
        [NotNull] Expression<Func<TEntity, bool>> predicate,
        bool autoSave = false,
        CancellationToken cancellationToken = default
    );
````

Here: https://github.com/abpframework/abp/blob/dev/framework/src/Volo.Abp.EntityFrameworkCore/Volo/Abp/Domain/Repositories/EntityFrameworkCore/EfCoreRepository.cs#L290-L305

This can be a breaking change for some applications (because it won't track the deleted entities in change tracker anymore).
 So, it would be better to implement it with ABP 7.0.

maliming:
> because it won't track the deleted entities in change tracker anymore

This breaks the soft delete and auditing...

hikalkan:
Hmmm.. OK, then it would be better to introduce a new method that explicitly explains it breaks these. I will create another issue.

hikalkan:
Moved to https://github.com/abpframework/abp/issues/14725

PR #14853, opened 2022-11-28
Bulk Operations article

enisn:


braim23 (approved):

maliming:
BTW `ExecuteDeleteAsync` breaks the soft delete and auditing..., because it won't track the deleted entities in change tracker anymore.

https://github.com/abpframework/abp/issues/14716

enisn:
> BTW `ExecuteDeleteAsync` breaks the soft delete and auditing..., because it won't track the deleted entities in change tracker anymore.
> 
> #14716

Good point. I'll add a section about it

docs/en/Community-Articles/2022-11-28-ef-bulk-operations/POST.md
```diff
@@ -0,0 +1,59 @@
...
+
+
+## Bulk Operations in ABP Framework
+
+ABP Framework supports Entity Framework Core as its ORM. So, you can use these new methods in your repository classes in your ABP applications. But there is another feature that the ABP Framework provides: `IEfCoreBulkOperationProvider`. You can visit the [documentation](https://docs.abp.io/en/abp/latest/Entity-Framework-Core#customize-bulk-operations) to learn more about it. You always have full control of the execution of bulk operations in your ABP applications. When you call the `InsertMany`, `DeleteMany` or `UpdateMany` methods of the repository, that provider will be invoked and you can customize all the queries from one point. You don't mostly need to implement & customize that interface, but it's good to know that you can do it if you need. 
+
+## Usage in Repositories
+
+In most of the real world case scenarios, you need custom queries to execute bulk operations. You can't manage all the cases by overriding `IEfCoreBulkOperationProvider` and you can't even identify the entities in this provider because it's not aware of the properties of the entities. The best way to use this new feature is inside the repositories.
```
hikalkan: You should completely remove `IEfCoreBulkOperationProvider` from this article. It has a very different purpose. First of all, it requires that you have a list of entities (the entity objects already fetched from database), but the new ExecuteXxx methods works without fetching. As second, `IEfCoreBulkOperationProvider` works with a list of arbitrary entities, doesn't force that they fits to a predicate, but the new ExecuteXxx methods works with entities with the same predicate/query-conditions. While they may seem similar with naming, they totally works different and can't be used as alternative to each other.

If you want, you can mention about `IEfCoreBulkOperationProvider` as a last section of this article, but noticing the difference.

docs/en/Community-Articles/2022-11-28-ef-bulk-operations/POST.md
```diff
@@ -0,0 +1,59 @@
...
+
+        // No need for an action for bulk inserting.
+        await base.InsertManyAsync(books); 
+    }
+}
+```
+
+> There is no need to take an action for bulk inserting. You can use the `InsertManyAsync` method of the repository instead of creating a new method for it if you don't have custom logic. It'll use a new bulk inserting feature automatically since it's available in EF Core 7.0.
+
+> `ExecuteDeleteAsync` breaks the soft delete and auditing since deleted entities are removed from the change tracker. ABP can't handle it. So, you have to be careful while using it.
```
hikalkan: Deleted entities are never retrieved to change tracker. So, "...are removed" is not right.

```suggestion
> If you use `ExecuteDeleteAsync` or `ExecuteUpdateAsync`, then ABP's soft delete and auditing features can not work. Because these ABP features work with EF Core's change tracking system and these new methods doesn't work with the change tracking system. So, use them carefully.
```

docs/en/Community-Articles/2022-11-28-ef-bulk-operations/POST.md
```diff
@@ -0,0 +1,59 @@
...
+            .Where(x => x.IsListed && x.PublishedOn.Year <= 2022)
+            .ExecuteUpdate(s => s.SetProperty(x => x.IsListed, x => false));
+    }
+
+    public async Task DeleteOldBooksAsync()
+    {
+        var dbSet = await GetDbSetAsync();
+
+        dbSet.Where(x => x.PublishedOn.Year <= 2000)
+            .ExecuteDelete();
```
hikalkan: Why not use the async method `ExecuteDeleteAsync`?

docs/en/Community-Articles/2022-11-28-ef-bulk-operations/POST.md
```diff
@@ -0,0 +1,59 @@
...
+    {
+    }
+
+    public async Task UpdateListingAsync()
+    {
+        var dbSet = await GetDbSetAsync();
+
+        dbSet
+            .Where(x => x.IsListed && x.PublishedOn.Year <= 2022)
+            .ExecuteUpdate(s => s.SetProperty(x => x.IsListed, x => false));
```
hikalkan: Why not use the async method `ExecuteUpdateAsync`?

docs/en/Community-Articles/2022-11-28-ef-bulk-operations/POST.md
```diff
@@ -0,0 +1,59 @@
...
+
+    public async Task DeleteOldBooksAsync()
+    {
+        var dbSet = await GetDbSetAsync();
+
+        dbSet.Where(x => x.PublishedOn.Year <= 2000)
+            .ExecuteDelete();
+    }
+
+    public async Task ImportAsync(List<Book> books)
```
hikalkan: I think this example is not necessary and confuses developers. `InsertManyAsync` is already public and can be directly called out of the repository. The section below ("There is no need to take an action for bulk inserting...") may remain.

hikalkan (changes requested):

hikalkan (approved):

PR #14861, opened 2022-11-28
add DisplayName for IdentityRoleCreateOrUpdateDtoBase Name property

wangj90:
add DisplayName for IdentityRoleCreateOrUpdateDtoBase Name property.
because the field name displayed when validate is Name, which can cause ambiguity.
Also consistent with the tenant-management: https://github.com/abpframework/abp/blob/b7311f23bedf7110141466d7b7f070875ce265bf/modules/tenant-management/src/Volo.Abp.TenantManagement.Application.Contracts/Volo/Abp/TenantManagement/TenantCreateOrUpdateDtoBase.cs#L11


wangj90:
the validation looks like this:
image

wangj90:
@realLiangshiwei @maliming what's the problem?

maliming (approved):

Issue #14551, opened 2022-11-05
Permission not granted if no child is selected (UI)


nev-21:
### Documentation

https://docs.abp.io/en/abp/latest/Authorization#child-permissions

```
var authorManagement = myGroup.AddPermission("Author_Management");
authorManagement.AddChild("Author_Management_Create_Books");
authorManagement.AddChild("Author_Management_Edit_Books");
authorManagement.AddChild("Author_Management_Delete_Books");
```
> -GetListAsync and GetAsync will be available to users if they have Author_Management permission is granted.
> -Other methods require additional permissions.

### GitHub Issues

* **ABP Framework 6.0.0**.
* Angular 14.2.8
* EF Core

### Steps to reproduce

1. Create a simple app with a permission with childs (You can also use the abp online demo).
2. Check only the parent permission.
3. See the number of granted permissions is not including the permission.
4. Save and open again
5. The permission was not granted (saved)

Example: see Languages permission with and without child, after click Save it is not saved.

![image](https://user-images.githubusercontent.com/89469783/200143855-fe9c0b66-c131-4e6b-adbf-d63c3ddaa8d3.png)

It doesn't recognize the permission, the number should be (3).

![image](https://user-images.githubusercontent.com/89469783/200143868-c5f4aafe-f5cb-43d5-8bef-e44b3b42c289.png)

After click on save, it is not saved.
![image](https://user-images.githubusercontent.com/89469783/200143900-63b40297-6034-4172-ad8c-0b36de8708ec.png)

phil-enfom:
Hi, 
Also have this issue on ABP 7.0 with Angular UI
If we want to just give just readonly access to a user we would only tick the parent permission but on save this is lost.

Can I have an update on this please.

Phil

phil-enfom:
@maliming this has been resolved by #14942

PR #14910, opened 2022-12-01
Refactor the Permission Tree component's parent-child check feature.

mahmut-gundogdu:
Resolves https://github.com/volosoft/volo/issues/12751

Test steps.
Create abp app.
Go to identity/ Role or User.
Create new one.
Open permission modal.
Control all-parent-child related click. 

Expected behavior:
When user mark as check a parent, all children should be checked
when user mark as uncheck  a parent ,all children should be unchecked  
when user mark as check a child, parent should be checked.
when user mark as uncheck a last checked child, parent should be unchecked.

### DIFF
--- a/npm/ng-packs/packages/permission-management/src/lib/components/permission-management.component.html
+++ b/npm/ng-packs/packages/permission-management/src/lib/components/permission-management.component.html
@@ -75,7 +75,7 @@
                 [attr.id]="permission.name"
                 class="form-check-input"
                 [disabled]="isGrantedByOtherProviderName(permission.grantedProviders)"
-                (click)="onClickCheckbox(permission, permissionCheckbox.value)"
+                (click)="onClickCheckbox(permission, permissionCheckbox.checked)"
               />
               <label class="form-check-label" [attr.for]="permission.name"
                 >{{ permission.displayName }}
--- a/npm/ng-packs/packages/permission-management/src/lib/components/permission-management.component.ts
+++ b/npm/ng-packs/packages/permission-management/src/lib/components/permission-management.component.ts
@@ -156,36 +156,33 @@ export class PermissionManagementComponent
     return false;
   }
 
-  onClickCheckbox(clickedPermission: PermissionGrantInfoDto, value) {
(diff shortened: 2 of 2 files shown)


gizemmutukurt (approved):

phil-enfom:
Hi @mahmut-gundogdu
The mentioned behaviour does not align with with what I posted on the Commercial Support Site.

I would like the behaviour for just the parent permission to be able to be selected which is possible in the MVC frontend but not the Angular.

If you tick the parent only the parent should activate, if you tick a child it should also tick the parent.

I am using the parent permission to grant read access to information then giving child permissions for Create, Edit, and Delete.

At the moment when you tick just the parent the system does not acknowledge the permssion has been ticked in the counts on the left and does not send it to the backend. If you do the same in the MVC frontend it is acknowledged and sent correctly.

Otherwise what is the point of having a permission that can not be granted on its own?

I would then have to add a Read Permission for every module as well as create edit and delete in which Create, Edit and Delete would also be dependant on.....

This current change would still end up with behaviour different from the MVC Frontend

phil-enfom:
See #14551 and https://support.abp.io/QA/Questions/4111/Permission-not-granted-if-no-child-is-selected-UI---Repost-of-Github-Issue-14551

phil-enfom:
**Test from Blazor Server**

Open Permissions:
image

Click Parent, Count changes to 1
image

Tick Child,m Count changes to 2
image

Untick Child, Count Decrements to 1
image

Tick Child then untick parent, Child unticks and Count Decrements to 0
image

mahmut-gundogdu:
> See #14551 and https://support.abp.io/QA/Questions/4111/Permission-not-granted-if-no-child-is-selected-UI---Repost-of-Github-Issue-14551

yes, I made mistake. I have fixed. could you check the pr ? https://github.com/abpframework/abp/pull/14942

mahmut-gundogdu:
I miss understand the behaviors of permissions-modal. The issue handled this branch again.https://github.com/abpframework/abp/pull/14942

PR #14943, opened 2022-12-05
Update OpenIddict-Blazor.md

omer-repo:
There were 3 "MyApplication**Blazor**Module.cs" points that should be "MyApplication**Host**Module.cs"


maliming (approved):

maliming:
Thanks @omer-repo

Issue #15069, opened 2022-12-12
Changing the password from Admin panel deletes the password from database


Mirza233:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

When editing user settings as admin and entering an invalid password, the password is deleted from the DB (see attachment).

![Screenshot 2022-12-12 110122](https://user-images.githubusercontent.com/53795269/207017265-715069a6-5b96-4961-be05-679cdc192e60.png)

### Reproduction Steps

- In the main menu select **Administration/Identity management/Users**.
- Create a new user with all the valid data.
- After the user is created, click on the **Actions** button and in the dropdown menu, select the **Edit** option.
- In the **Password** field in the modal enter **123** and click **Save**.
- Verify that an Error modal has popped up saying the password must be at least 6 characters long, etc.
- Click Ok

### Expected behavior

Password wasn't changed

### Actual behavior

Password got removed (set to null) from the database and the selected user cannot log in anymore.

### Version

latest (6.0.2)

### User Interface

Common (Default)

### Database Provider

MongoDB

### Tiered or separate authentication server

None (Default)

### Operation System

Windows (Default)

### Other information

In _IdentityUserAppService.cs_ **UpdateAsync** method, there is the following block of code:

    if (!input.Password.IsNullOrEmpty())
    {
        (await UserManager.RemovePasswordAsync(user)).CheckErrors();
        (await UserManager.AddPasswordAsync(user, input.Password)).CheckErrors();
    }

The problem occurs when **RemovePasswordAsync()** passes the **CheckErrors()** validation, but the **AddPasswordAsync()** doesn't. 
A possible fix would consist of modifying the block above as follows

    if (!input.Password.IsNullOrEmpty() && ValidatePassword(input.Password))
    {
        (await UserManager.RemovePasswordAsync(user)).CheckErrors();
        (await UserManager.AddPasswordAsync(user, input.Password)).CheckErrors();
    }

Where **ValidatePassword()** would throw an exception if the password doesn't pass the validation (similarly as **CheckErrors()**).

maliming:
hi @Mirza233 

Please use Mongodb that support the `Transaction` and remove

https://github.com/abpframework/abp/blob/883a0b1211af751d702a3421cefed26fc8429db3/templates/app/aspnet-core/src/MyCompanyName.MyProjectName.MongoDB/MongoDb/MyProjectNameMongoDbModule.cs#L37

PR #15072, opened 2022-12-12
Fix unexpected delete password issue.

roc916:
Resolve https://github.com/abpframework/abp/issues/15069


maliming:
hi @roc916 
We can call `CheckPasswordAsync` before updating it.

https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.identity.usermanager-1.checkpasswordasync?view=aspnetcore-7.0

roc916:
hi @maliming `UpdatePasswordHash` has `validatePassword` parameter.  https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.identity.usermanager-1.updatepasswordhash?view=aspnetcore-7.0

maliming:
The problem with #15069 is a transaction problem. The transaction should roll back.

PR #15085, opened 2022-12-12
disable tree style loading with token

mahmut-gundogdu:
We are moved style to lazyLoadservice. If the user want to stop inject styling, it can do that with a token.


npm/ng-packs/packages/components/tree/src/lib/disable-tree-style-loading.token.ts
```diff
@@ -0,0 +1,5 @@
+import { InjectionToken } from '@angular/core';
+
+export const DISABLE_TREE_STYLE_LADING_TOKEN = new InjectionToken<boolean>(
```
muhammedaltug: please fix typo `DISABLE_TREE_STYLE_LOADING_TOKEN`

muhammedaltug (approved):

Issue #15114, opened 2022-12-14
Cli: show the current cli version


yekalkan:
Show the version in the first info line : [13:56:30 INF] ABP CLI 7.0.0-rc.4 (https://abp.io)

PR #15128, opened 2022-12-15
Cli: show the current cli version

yekalkan:
resolves https://github.com/abpframework/abp/issues/15114

This way we can see the cli version when someone send us their CLI logs. And we don't send request to anywhere, so it doesn't slow CLI down much.

image


ebicoglu:
I removed` https://abp.io` text. anyone who's using this CLI knows the address

PR #15359, opened 2023-01-09
Angular: using AuthService instead of OAuthService

rqx110:


mahmut-gundogdu (approved):
Yes you are absolutely right, it should be like that. Nice catch.

PR #15344, opened 2023-01-05
Update login button

ozermzn:
Loading design for login button has been added.
Update login button #15344

PR #15375, opened 2023-01-09
Feat/ng loading button

### LINKED ISSUE #15263: Add loading button that on the home page.
When user click the Login button that on the home page (when the user did not authenticated), the button should be disabled and it shown a loading indicator. the process are finished. Loading will be removed.

The base branch should be rel-7.0 but if you solve the issue with another branch, we can fix that. No problem. just do it  :)

here  the code.
https://github.com/abpframework/abp/blob/e04173c8dbc81b80f9af06d05a15862464680b43/templates/app/angular/src/app/home/home.component.html#L13

also it will add another templates. after the feature is work as well.

You can use  AbpLoading directive
https://docs.abp.io/en/abp/latest/UI/Angular/Loading-Directive

Docs to how to install development env. 
https://github.com/abpframework/abp/blob/dev/docs/en/Contribution/How-to-Contribute-abp.io-as-a-frontend-developer.md

If you have any question about the issue. Please contact to me.

masum-ulu:
close #15263

### DIFF
--- a/docs/en/Migration-Guides/Abp-7_0.md
+++ b/docs/en/Migration-Guides/Abp-7_0.md
@@ -140,7 +140,7 @@ export class AppModule {}
 
 ```
 ## Lepton X Google-Font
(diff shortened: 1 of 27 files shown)


masum-ulu:
Sorry it is already done with #15344

Issue #15404, opened 2023-01-13
Automapper doesn't work when add maps with `validate: true`


gdlcf88:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

See https://github.com/abpframework/abp/pull/15405

### Version

7.0.0

### User Interface

Common (Default)

### Database Provider

EF Core (Default)

### Tiered or separate authentication server

None (Default)

### Operation System

Windows (Default)

maliming:
See https://github.com/abpframework/abp/pull/15405#issuecomment-1382647493

PR #15405, opened 2023-01-13
Add UTs to reproduce the Automapper's `AddMaps` method problem

gdlcf88:
#15404


maliming:
If you want to use `MemberList.Source` please add `.ForSourceMember(x => x.ExtraProperties, x => x.DoNotValidate())`

https://github.com/AutoMapper/AutoMapper/issues/1556#issuecomment-234586193

PR #15419, opened 2023-01-16
Convert Simplified Chinese to Traditional Chinese

HyleonOu:
[Convert Simplified Chinese to Traditional Chinese](https://github.com/abpframework/abp/commit/60816503a60c529c59de17ed8b6f15f6091c415f)


maliming (approved):

maliming:
Thanks @HyleonOu

Issue #15381, opened 2023-01-11
Different display of language


rqx110:
all templates are created with abp cli 7.0

Angular:
![image](https://user-images.githubusercontent.com/1829209/211689745-5e27ad91-2f2b-49bd-9e3d-4d85fb484cc7.png)

Blazor:
![image](https://user-images.githubusercontent.com/1829209/211691815-7e35c62a-3ab3-451d-84c1-a8250ced3123.png)

Blazor-Server:
![image](https://user-images.githubusercontent.com/1829209/211692309-61306f5c-d2e7-4069-8b87-a51648e09d75.png)

MVC
![image](https://user-images.githubusercontent.com/1829209/211693338-356cc8ef-dd74-4c43-a65c-7a3cd2d75428.png)

also, the order of the login info and language is different

maliming:
It will be resolved in the next patch version.

hikalkan:
Angular should also use the "EN" syntax. "admin" should always be in the most right side.
Thanks for reporting this. We will be working on it.

rqx110:
if we use short name, some languages will have same display.
e.g.
![image](https://user-images.githubusercontent.com/1829209/211968756-24ea87b2-7df3-47f0-bc45-06d950f275b5.png)

mahmut-gundogdu:
<img width="695" alt="image" src="https://user-images.githubusercontent.com/2217899/214371813-fc9695c6-fdf8-4dcc-a7b5-b6d8095d4d3f.png">
 Angular is ready to review.

mahmut-gundogdu:
Angular twoLetterISOLanguageName have been added, it will release next version

rqx110:
> "admin" should always be in the most right side.

@mahmut-gundogdu

mahmut-gundogdu:
> Angular twoLetterISOLanguageName have been added, it will release next version

Thank you. I am fixing on Angular

mahmut-gundogdu:
I did it.  Thank you for response.
<img width="178" alt="image" src="https://user-images.githubusercontent.com/2217899/214782693-c5855dc4-5c5e-4307-836f-dba7052b1bfb.png">

enisn:
It's done on blazor-server too and it'll be released in the next release of LeptonX, So I'm closing this issue

PR #15435, opened 2023-01-17
Add TwoLetterISOLanguageName LanguageInfo

enisn:
Related to https://github.com/abpframework/abp/issues/15381

The purpose of this change is to provide TwoLetterISOLanguageName to non-CSharp clients.

image


maliming:
CultureInfo has many names, we can consider using the original name `TwoLetterISOLanguageName` to prevent ambiguity.

image

enisn:
You're right,

I've used the original name in json:
image

maliming (approved):

PR #15514, opened 2023-01-26
Fix EfCoreBlogPostRepository & EfCoreBlogRepository to use ICmsKitDbC…

andmattia:
Related to [ReplaceDbContext-causes-an-error](https://support.abp.io/QA/Questions/1290/microservice-service-template-ReplaceDbContext-causes-an-error).

I move this to repository to ICmsKitDbContext


maliming (approved):

maliming:
Thanks @andmattia

PR #14514, opened 2022-11-02
Feature management reset 14470

malikmasis:
closes https://github.com/abpframework/abp/issues/14470

image

PR #15524, opened 2023-01-27
Use providers instead of direct repository access for resetting features

enisn:
Deleting from the database is not an option for this feature. Also, data still remains in the cache when we remove it with Repository. So, this PR just uses Providers to reset features.

---

Firstly appeared in https://github.com/abpframework/abp/pull/14514

---

Closes https://github.com/volosoft/volo/issues/13278


enisn:
CI Pipeline says tests can't be built but I can't reproduce the test build error:
image

modules/feature-management/src/Volo.Abp.FeatureManagement.Domain/Volo/Abp/FeatureManagement/FeatureManager.cs
```diff
@@ -215,4 +215,33 @@ await using (await providers[0].HandleContextAsync(providerName, providerKey))
 
         return featureNameValueWithGrantedProvider;
     }
+
+    public async Task DeleteAsync(string providerName, string providerKey)
```
enisn:
```suggestion
    public virtual async Task DeleteAsync(string providerName, string providerKey)
```

maliming (approved):

gizemmutukurt (approved):

PR #15540, opened 2023-01-29
#15539 aded  localization languaje for select2 for

washyn:
aded  localization languaje for select2 to display in different languages


enisn (approved):
Thanks for you contribution

PR #15556, opened 2023-01-30
Describe all abp database tables in a new document 

salihozkara:
Resolves https://github.com/volosoft/volo/issues/3581

### DIFF
--- a/docs/en/Modules/Database-Tables.md
+++ b/docs/en/Modules/Database-Tables.md
@@ -6,1150 +6,570 @@ This documentation describes all database tables and their purposes. You can rea
 
 ### AbpAuditLogs
 
-AbpAuditLogs is used to store audit logs.
-
-#### Description
-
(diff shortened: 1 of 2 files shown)


EngincanV (commented):
I've read the documentation and I have some suggestions (@salihozkara):

1. In each description of tables, you used a word something like: "*allows to manage and track the ... effectively.*" This is repeated over almost every section. I think this is not good, you can check the descriptions for all table sections and change them accordingly.
2. **Uses** and **Used By** sections should be removed. (if you think these sections should not be removed, then it can stay as it is, but don't use both words, select one of them and stick with it along the documentation)
3. The single lines before each *Description* section are not needed since the first paragraph of the descriptions is almost the same. (we don't need to repeat it, so you can remove  the *Description* sections from the documentation and start describing the related table starting from the first line of the *Description* section.)
4. There are some paragraphs that repeat the previous one, like:

```md
For example, you can use the ApplicationName, UserId, ExecutionTime columns to filter the audit logs by `ApplicationName`, `UserId`, `ExecutionTime` columns to filter the audit logs by application name, user id, and execution time respectively, so that you can easily track the actions performed in the application.
```

For this kind of sentence, please revise it and remove the redundant parts.

EngincanV (approved):

PR #15573, opened 2023-02-01
Refactor `IExceptionToErrorInfoConverter` namespace and namings

### LINKED ISSUE #15557: Consider changing `IExceptionToErrorInfoConverter` namespace
Hi,

The `IExceptionToErrorInfoConverter` interface and it's implementation are in the `Volo.Abp.ExceptionHandling` package (which targets netstandard2.0).
It's misleading that the components are inside the `Volo.Abp.AspNetCore.ExceptionHandling` namespace since it can be used on all platforms (wpf, winforms, you name it).

I can contribute a PR which refactors the namespace if it's ok. I  just wanted to discuss this idea first.

NecatiMeral:
Fixes #15557 

## Summary

Refactored `IExceptionToErrorInfoConverter` and related classes:
* Changed namespace from `Volo.Abp.AspNetCore.ExceptionHandling` to `Volo.Abp.ExceptionHandling`
* Renamed classes where appropriate
* Renamed option properties
  * kept compatibility for now by using aliases for previous option names; can remove compatibility layer if wanted

## Further considerations

We could inject `IOptions<AbpExceptionHandlingOptions>` into `DefaultExceptionToErrorInfoConverter` so the configured default values could be used. Currently every caller manually injects the options on the `Convert` method.

We could change the class like this:

```csharp
    protected AbpExceptionLocalizationOptions LocalizationOptions { get; }

    public DefaultExceptionToErrorInfoConverter(
        IOptions<AbpExceptionLocalizationOptions> localizationOptions,
        IOptions<AbpExceptionHandlingOptions> handlingOptions,
        /* omitted for readability */
    )

    protected virtual AbpExceptionHandlingOptions CreateDefaultOptions()
    {
        return new AbpExceptionHandlingOptions
        {
            IncludeDetails = HandlingOptions.IncludeDetails,
            IncludeStackTrace = HandlingOptions.IncludeStackTrace
        };
    }
```

### DIFF
--- a/docs/en/Exception-Handling.md
+++ b/docs/en/Exception-Handling.md
@@ -328,13 +328,12 @@ You can also throw these type of exceptions in your code (although it's rarely n
 ````csharp
 Configure<AbpExceptionHandlingOptions>(options =>
 {
-    options.SendExceptionsDetailsToClients = true;
-    options.SendStackTraceToClients = false;
+    options.IncludeDetails = true;
+    options.IncludeStackTrace = false;
 });
 ````
 
 Here, a list of the options you can configure:
 
-* `SendExceptionsDetailsToClients` (default: `false`): You can enable or disable sending exception details to the client.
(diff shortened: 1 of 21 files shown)


NecatiMeral:
The CI tests fail at permission-management efc. Can't explain why the test seems to be flaky.

PR #15589, opened 2023-02-03
Fixed duplicated Email MessageId header

huseyinalan:
message.MessageId setter already adding MessageId to Headers

### Description
MessageId property setter already adding MessageId to Headers and it causes duplicated header error
RFC 5322 standard doesn't allow duplicated headers

image
line: 818

### Checklist

### How to test it?

Send an test email to google smtp


maliming (approved):

maliming:
Thanks @huseyinalan

PR #15596, opened 2023-02-03
add visible$ property to nav items

### LINKED ISSUE #10204: Add visible$ property to ABP.Nav
add visible$ property to ABP.Nav which accepts Array<Observable<boolean>> type.

bariscanyilmaz:
### Description

Resolves #10204 

Added visible$ property ot abp.nav

### Checklist


mahmut-gundogdu (changes requested):
why we have two property ? We should decided that. It might be one property. There is an example in leptonX

mahmut-gundogdu (approved):

muhammedaltug (approved):

Issue #12118, opened 2022-03-29
Provide a proper way to update menu in Blazor without refreshing


enisn:
Blazor menu can't be updated until the page is refreshed. 

We can provide an API to update it or just register MenuViewModel into the dependency container and it can be replaced and changes will be applied to ui easily.

---

_Originally posted here: https://support.abp.io/QA/Questions/2778/Blazor-Change-Menu-Display-Name-dynamically_

sturlath:
@enisn just saw this issue that is similar to [my issue ](https://support.abp.io/QA/Questions/4144/Blazor-WASM-needs-two-loads-F5-to-show-side-menus) except that ours just happens when logging in for the first time and no dynamic loading going on. Was some change done regarding this to prepare for this task that we missed in the update?

sturlath:
@enisn can I understand you 👍 as there is an issue with 6.0.1?

enisn:
It might be related to `IsExpandedPropertyName` bindable property.

If the node binding context (nodes in viewmodel) doesn't have a property named `IsExpanded`, expand state couldn't be kept.

Can you provide a simple code example to reproduce the issue?

tvddev:
@hikalkan - is it possible to get this in 7.1 it has been open almost 1 year and is really core for a modern broswer app.

RelevanceRo:
@sturlath how did you solve your problem with the menu not rendering when logging in, only after a refresh? Mine still raging. Wasn't able to solve it even with a support from ABP. Same config as you: Blazor WASM.

Thanks a bunch for any tips!

Catalin

enisn:
There is an API for each theme that allows you replace one single menu item with your custom component:
https://github.com/abpframework/abp/blob/dev/docs/en/UI/Blazor/Basic-Theme.md#overriding-the-menu-item

---

```csharp
context.Menu.Items.Add(
  new ApplicationMenuItem("Custom.1", "My Custom Menu", "#")
    .UseComponent(typeof(MyMenuItemComponent)));
```

PR #15611, opened 2023-02-06
Add custom menu component support to Basic Theme

enisn:
### Description

Resolves part of https://github.com/abpframework/abp/issues/12118 (write the related issue number if available)

Since defining a custom component is related to a theme, this operation can't be planned without a theme dependency. So, the component code will be different according to different themes. Each theme should implement this logic itself if possible.

custom-component-basic-theme

### Checklist

- [x] I documented it (or no need to document or I will create a separate documentation issue) _Will be done after all themes implemented_

### How to test it?

UseComponent() method can be used in MenuContributor in a project that uses the basic theme.

```csharp
context.Menu.Items.Add(
  new ApplicationMenuItem("Custom.1", "My Custom Menu", "#")
    .UseComponent(typeof(MyMenuComponent)));
```

### DIFF
--- a/modules/basic-theme/src/Volo.Abp.AspNetCore.Components.Web.BasicTheme/Themes/Basic/FirstLevelNavMenuItem.razor
+++ b/modules/basic-theme/src/Volo.Abp.AspNetCore.Components.Web.BasicTheme/Themes/Basic/FirstLevelNavMenuItem.razor
@@ -1,13 +1,19 @@
-﻿@using Volo.Abp.UI.Navigation
+﻿@using Volo.Abp.AspNetCore.Components.Web.BasicTheme.Navigation;
+@using Volo.Abp.UI.Navigation
 @{
     var elementId = MenuItem.ElementId ?? "MenuItem_" + MenuItem.Name.Replace(".", "_");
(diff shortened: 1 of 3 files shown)


EngincanV (approved):

modules/basic-theme/src/Volo.Abp.AspNetCore.Components.Web.BasicTheme/Themes/Basic/FirstLevelNavMenuItem.razor
```diff
@@ -1,13 +1,19 @@
...
+    var customComponentType = MenuItem.GetComponentTypeOrDefault();
 }
 @if (MenuItem.IsLeaf)
 {
-    if (MenuItem.Url != null)
+    if (customComponentType != null)
+    {
+        <DynamicComponent Type="@customComponentType" />
+    }
+    else if (url != null)
```
rqx110, line 16: is this mistake? the `url` will never `null`.
revert to use `MenuItem.Url != null` ?

enisn: Done in https://github.com/abpframework/abp/pull/15625

PR #15664, opened 2023-02-09
change flag-icon-css to flag-icons

bariscanyilmaz:
### Description

Resolves volosoft/lepton#754

flag-icon-css package renamed to flag-icons
all its class names and package names are updated in all files


rqx110:
why we not delete the `flag-icon-css` package completely (means the whole `flag-icon-css` folder)?
if we want keep the `flag-icon-css` package, pls also keep the `abp.resourcemapping.js` of `flag-icon-css`

bariscanyilmaz:
> why we not delete the `flag-icon-css` package completely (means the whole `flag-icon-css` folder)? if we want keep the `flag-icon-css` package, pls also keep the `abp.resourcemapping.js` of `flag-icon-css`

We can't delete flag-icons-css because there are components using it. Also, Flag-icons-css package was renamed to flag-icons so we renamed too.  

Why you want to keep resource mapping of flag-icon-css our aim is replacing them with the new package

rqx110:
got it

mahmut-gundogdu (approved):

PR #15692, opened 2023-02-14
Cli: Enhancements related with starting ABP Suite

yekalkan:
Resolves https://github.com/volosoft/volo/issues/13480

- Suite won't try to start if it is already running and Cli will open Suite in the browser directly.
- Suite won't start if the Port is in usage by another process and show an error text.

@gizemmutukurt 
- You will test the scenarios above. 
- You will use local cli build in the related branch. 
- You don't need to test Suite's functionality.

### DIFF
--- a/framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/Commands/Services/SuiteAppSettingsService.cs
+++ b/framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/Commands/Services/SuiteAppSettingsService.cs
@@ -1,0 +1,73 @@
+﻿using System;
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+using Newtonsoft.Json.Linq;
+using Volo.Abp.Cli.Utils;
+using Volo.Abp.DependencyInjection;
+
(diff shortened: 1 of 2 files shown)


framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/Commands/Services/SuiteAppSettingsService.cs
```diff
@@ -0,0 +1,96 @@
...
+        {
+            return DefaultPort;
+        }
+        
+        return Convert.ToInt32(url.Split(":").Last());
+    }
+
+    private string GetFilePathOrNull(string version)
+    {
+        var suiteVersion = version;
```
EngincanV: Instead of creating a new variable, we can use the `version` parameter.

EngincanV (approved):

gizemmutukurt:
Test is ok.
- [x] Suite won't try to start if it is already running and Cli will open Suite in the browser directly.
- [x] Suite won't start if the Port is in usage by another process and show an error text.

gizemmutukurt (approved):

Issue #15712, opened 2023-02-15
HideAbpEndpoints also hides endpoints that have a route parameter constraint


tinowilloeper:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

To only see my own endpoints I added the option HideAbpEndpoints() to AddAbpSwaggerGenWithOAuth. After doing that I checked the swagger document and some of my own endpoints were also hidden. These endpoints have a route parameter constraint like `[Route("{id:long}")]` or `[Route("internal/{id:guid}")]`

### Known Workarounds

**The reason:** 
In the Apply-method of AbpSwashbuckleDocumentFilter.cs the Templates will be selected with the constraint. The OpenApiPaths in the OpenApiDocument will not contain the constraint. So the RemoveAll-function removes also these endpoints.
`swaggerDoc
    .Paths
    .RemoveAll(path => !actionUrls.Contains(path.Key));
`

**The solution:** 
_Remove the route parameter constraints while selecting the actionUrls like this:_
`.Select(actionDescriptor => actionDescriptor.AttributeRouteInfo?.Template?
    .EnsureStartsWith('/')
    .Replace("?", "")
    .RemoveRouteParameterConstraint())
`

_And an extension method:_

```csharp
public static string RemoveRouteParameterConstraint(this string route)
{
    while (route.Contains(':'))
    {
        var start = route.IndexOf(':');
        var length = route.IndexOf('}', start) - start;
        route = route.Remove(start, length);
    }
    return route;
}
```

### Version

7.0.0

### User Interface

Common (Default)

### Database Provider

EF Core (Default)

### Tiered or separate authentication server

None (Default)

### Operation System

Linux

### Other information

https://github.com/abpframework/abp/blob/dev/framework/src/Volo.Abp.Swashbuckle/Volo/Abp/Swashbuckle/AbpSwashbuckleDocumentFilter.cs#L20

Sorry for the bad formatting.

EngincanV:
Hi, thanks for reporting. I'll be fixing it.

PR #15722, opened 2023-02-16
Remove route constraints before determining which endpoints to hide

EngincanV:
### Description

Resolves https://github.com/abpframework/abp/issues/15712

As mentioned in [this issue](https://github.com/abpframework/abp/issues/15712), when there is a route constraint in the route, it would be hidden because we didn't handle the route constraints. We need to remove the constraint part from the route and then check if's ABP's endpoint and hide according to that. 

### Checklist

- [X] I documented it (or no need to document or I will create a separate documentation issue) - **No need to document**.

### How to test it?

You can create a controller and change its routes as `{id:minlength(1)}` (or any other built-in constraint) and also follow [this documentation](https://docs.abp.io/en/abp/latest/API/Swagger-Integration#hide-abp-endpoints-on-swagger-ui) to enable hiding abp related endpoints. Then, when you open the swagger, you should be able to see your custom route in the open api documentation.

### DIFF
--- a/framework/src/Volo.Abp.Swashbuckle/Volo/Abp/Swashbuckle/AbpSwashbuckleDocumentFilter.cs
+++ b/framework/src/Volo.Abp.Swashbuckle/Volo/Abp/Swashbuckle/AbpSwashbuckleDocumentFilter.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text.RegularExpressions;
 using Microsoft.AspNetCore.Mvc.Abstractions;
 using Microsoft.OpenApi.Models;
 using Swashbuckle.AspNetCore.SwaggerGen;
@@ -10,6 +11,8 @@ namespace Volo.Abp.Swashbuckle;
 public class AbpSwashbuckleDocumentFilter : IDocumentFilter
 {
     protected virtual string[] ActionUrlPrefixes { get; set; } = new[] { "Volo." };
+
+    protected virtual string RegexConstraintPattern => @":regex\(([^()]*)\)";
     
     public virtual void Apply(OpenApiDocument swaggerDoc, DocumentFilterContext context)
     {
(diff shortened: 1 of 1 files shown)


framework/src/Volo.Abp.Swashbuckle/Volo/Abp/Swashbuckle/AbpSwashbuckleDocumentFilter.cs
```diff
@@ -17,12 +18,31 @@ public virtual void Apply(OpenApiDocument swaggerDoc, DocumentFilterContext cont
...
+        
+        var startIndex = route.IndexOf(":", StringComparison.Ordinal);
+        var endIndex = route.IndexOf("}", StringComparison.Ordinal);
+        
+        if (startIndex == -1 || endIndex == -1)
+        {
+            return route;
+        }
+
+        return route.Remove(startIndex, (endIndex - startIndex));
```
rqx110: the route constraints can use a Regular Expression, is it a problem? 
`[Route("users/{ssn:regex(^\\d{{3}}-\\d{{2}}-\\d{{4}}$)}]`
see https://learn.microsoft.com/en-us/aspnet/core/fundamentals/routing?view=aspnetcore-7.0#route-constraints

and if route is `[Route("users/{id:int:min(1)}/{age:int")]`, is it a problem?

EngincanV: I will cover these situations 👍

EngincanV: Done https://github.com/abpframework/abp/pull/15722/commits/2ea659709883e1e13b7c2b7c325283e89696fb47. Thanks @rqx110 🙏

rqx110 (approved):

realLiangshiwei (approved):

PR #15727, opened 2023-02-16
Capchta localization edited

onurpicakci:
Resolves https://github.com/volosoft/vs-internal/issues/1670


berkansasmaz (approved):

modules/cms-kit/src/Volo.CmsKit.Public.Web/Security/Captcha/SimpleMathsCaptchaGenerator.cs
```diff
@@ -74,7 +83,7 @@ public void Validate(Guid requestId, int value)
         var request = Session[requestId];
         if (request.Output.Result != value)
         {
-            throw new CaptchaException("The captcha code doesn't match text on the picture! Please try again.");
```
berkansasmaz, line 77: If we're not using a `CaptchaException`, why don't we remove it anymore?

berkansasmaz (approved):

PR #15800, opened 2023-02-24
cms-kit: Fix `TagHelper` name

JadynWong:
It should be `abp-style-bundle`


EngincanV:
@JadynWong if it's possible can you do the same changes in the `rel-7.1` branch and send a new PR? (Because it would be good to include this fix in the next patch release)  If you don't have time, please let me know, so I can fix it.

> I close this PR because it should be fixed in the `rel-7.1` branch.

JadynWong:
I will do it.

PR #15809, opened 2023-02-26
Improvements for permission-management modal

EngincanV:
### Description

Resolves https://github.com/volosoft/volo/issues/13428

We made some improvements on the Angular side for the permission management modal. For example, if all permissions are granted for a certain group, it becomes disabled for a different provider (if it's all granted for other providers). In this PR, I have done the same required changes for both MVC & Blazor UIs.

### Checklist

### How to test it?

You can grant all permissions for a role and then check the permission modal on the */Users* page.

### DIFF
--- a/modules/permission-management/src/Volo.Abp.PermissionManagement.Blazor/Components/PermissionManagementModal.razor
+++ b/modules/permission-management/src/Volo.Abp.PermissionManagement.Blazor/Components/PermissionManagementModal.razor
@@ -10,7 +10,9 @@
         <ModalBody MaxHeight="50">
             
             <Field>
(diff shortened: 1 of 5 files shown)


maliming (approved):

gizemmutukurt:
When I select Grant all permissions, select all should be checked.

d458b920-43d2-4d2c-8ffb-82fffdc0ed8c

gizemmutukurt:
It would be more appropriate to have the page design as in the screenshot I added in the issue.
image

EngincanV:
> It would be more appropriate to have the page design as in the screenshot I added in the issue. image

I think we can keep headers for each tab for both MVC & Blazor UIs.

gizemmutukurt (approved):

PR #15847, opened 2023-03-01
give more details about list service filtering

bariscanyilmaz:
### Description

TODO: Give more details about list service filtering.

### DIFF
--- a/docs/en/UI/Angular/List-Service.md
+++ b/docs/en/UI/Angular/List-Service.md
@@ -201,3 +201,38 @@ You may use observables in combination with [AsyncPipe](https://angular.io/guide
 
 <input type="text" name="search" [(ngModel)]="list.filter">
 ```
+ABP doesn't have a built-in filtering mechanism. You need to implement yourself and handle `filter` property in backend as seen in below. 
+
+```cs
+public class BookRequestDto : PagedAndSortedResultRequestDto
+{
+  public string Filter { get; set; }
+}
+```
+```cs
+public interface IBookAppService :
+    ICrudAppService< //Defines CRUD methods
+        BookDto, //Used to show books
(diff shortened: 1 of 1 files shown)


docs/en/UI/Angular/List-Service.md
```diff
@@ -201,3 +201,38 @@ You may use observables in combination with [AsyncPipe](https://angular.io/guide
...
+
+}
+```
+```cs
+//Override GetListAsync in BookAppService
+
+public override async Task<PagedResultDto<BookDto>> GetListAsync(BookRequestDto input)
+{
+  var query=await ReadOnlyRepository.GetQueryableAsync();
+  query.WhereIf(!string.IsNullOrEmpty(input.Filter),x=>x.Name.Contains(input.Filter));
```
rqx110:
```suggestion
  var queryable = await ReadOnlyRepository.GetQueryableAsync();
  var query = queryable.WhereIf(!string.IsNullOrEmpty(input.Filter), x => x.Name.Contains(input.Filter));
```

enisn (commented):
I'm not sure this much backend information for angular documentation. It's not the only way for filtering. Filtering can be achieved in many different ways.

Instead overriding `GetListAsync()`;
 - People can create AppService without using CrudAppService.
 - This kind of queries should be done in repository as a best practice.
 - We can't assume that entity will have **Name** property of entity is a book.

_**In my opinion,** we can write a sentence about filtering implementation should be done on backend side and put a link if there is a related documentation._

What do you think @hikalkan ?

hikalkan:
I completely agree with @enisn 
We should not write such opinionated documentation. That may lead developers thinking this is the only way to do it.

enisn (approved):

Issue #5676, opened 2020-10-02
Upgrade to .NET 5.0 rc1


hikalkan:
We want to move all the framework, modules, themes... to .NET 5.0 in a branch to see if we have any problem. If everything goes well, we can make this movement early (considering to skip the ABP version 3.3 and merge it to the version 4.0, then publish early previews).

hikalkan:
See https://docs.microsoft.com/en-us/dotnet/core/compatibility/3.1-5.0

maliming:
Incompatible packages:
https://www.nuget.org/packages/Microsoft.AspNetCore.Mvc.Versioning , https://github.com/microsoft/aspnet-api-versioning/issues/662

leonkosak:
@maliming `Change netstandard2.0 to netstandard2.1 for all EF Core projects.`
Why not directly to net5.0?

hikalkan:
@leonkosak because Microsoft does it like that. So, ask to them :) Maybe it has no meaning in practical (.net framework won't support it), but we also don't have problem with .netstandard. Maybe someone use with Xamarin :P
We try to target .netstandard wherever possible.

Xeevis:
@hikalkan I've been reading up on [blog post](https://devblogs.microsoft.com/dotnet/the-future-of-net-standard/) by Immo Landwerth, program manager for .NET at Microsoft. From it I've got these pointers
* Upgrade from `netstandardY.X` to `net5.0` gives access to more runtime features, language features, APIs and ensures platform portability (`netstandardY.X` implements platform-specific features that can't be statically validated to ensure code is truly portable) so there is certainly something to gain there
* `netstandard2.x` makes sense for reusable libraries, something you grab from NuGet to gain just certain functionality. But can portions of ABP even be considered reusable libraries working outside the ABP ecosystem (which may require `net5.0` anyway)?
* Landwerth suggests for all app components to target same `netX.Y` version because that way you can assume the same BCL features everywhere
* He also expressed expectation that widely used libraries will end up multi-targeting both `netstandard2.x` and `net5.0` as to not shoot oneself in the foot by forgoing latest platform features all the while still maintaining backwards compatibility

Considering these points I'd be inclined to just got for `net5.0` universally and if there really are reusable libraries buried within, just multi-target them. Eventually the "for-compatibility" `netstandard2.x` target will be dropped when more projects transition.

maxkol:
RC2 is already here
https://dotnet.microsoft.com/download/dotnet/5.0

[...]

PR #15878, opened 2023-03-04
Update AbpStringExtensions.ReplaceFirst, to spans

### LINKED ISSUE #15877: AbpStringExtensions.ReplaceFirst optimalisation using spans
### Description

The method AbpStringExtensions.ReplaceFirst uses two Substrings. This is one of the situations where spans are much more efficient.

Spans are harder to read. However, the name of the method is self-explanatory.

### Data

You can run my [Github benchmark example](https://github.com/merijng/ConsoleBenchmarkDotNet).

image

In total, the function is roughtly 20 percent faster. Also, the memory allocation is way less.

I also checked the other AbpStringExtensions methods. However, ReplaceFirst was the only method that uses multiple substrings for one string. if only one substring happens, there is no major performance gain. Perhaps performance gains are possible in other ways using spans, I have not investigated that.

### Analysis

Spans are faster in substrings because spans are lightweight objects that only store the starting and ending positions of a substring, whereas substrings create new strings in memory, leading to higher overhead and slower performance.

merijng:
### Description

Resolves #15877

you can see the benchmark in the issue (#15877).

### Checklist

### How to test it?

- To test the performance, use [my seperate repo](https://github.com/merijng/ConsoleBenchmarkDotNet). 
- Output and input strings remain the same, you can test that using the original unit tests (that still work).
15878

### Why is the code coverage lower?
Because the code coverage tool can't scan spans :(


berkansasmaz:
Thanks, @merijng 🎉 

I will review your PR as soon as possible.

berkansasmaz (changes requested):
I've reviewed your code and the performance improvement is really great, but there are a few problems. The `ReplaceFirst` method is defined in the framework, so no assumptions can be made, for example, the `ReplaceFirst` method can be called on a string with 5 characters or it can be called on a string with 50,000 characters. Since we cannot make such assumptions, we must be careful when using `stackalloc` otherwise a stack overflow exception may be encountered.

For example, a sentence in dotnet's documentation:
> Because the amount of memory available on the stack depends on the environment in which the code is executed, be conservative when you define the actual limit value.

_ref: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/stackalloc_

You can also check an article on the subject: https://vcsjones.dev/stackalloc/

If you want to perform repetitive tests, you can test the problem with the following data.

```charp
"Test".PadLeft(5000000, 't').ReplaceSpan("t".PadLeft(100000, 't').PadLeft(100000, 'e').PadLeft(300000, 't'), "XX", StringComparison.OrdinalIgnoreCase);
```

By the way, thank you very much for the test project you sent. 

---

Here, this problem can be solved by two methods:
1. As mentioned [in this document](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/stackalloc), with code similar to the following:

```charp
int length = 1000;
Span<byte> buffer = length <= 1024 ? stackalloc byte[length] : new byte[length];
```

2. By blending the old method with the new method:

```charp
int length = 1000;
if (length > 1024){
    ReplaceSubstring()
} else {
     ReplaceSpan()
}
```

When I look at the benchmark results in both methods, they are almost the same, so the first method seems more appropriate to me, so as not to overcomplicate the code.

But I need another pair of eyes to make sure I don't miss anything. @maliming, can you review this PR after the relevant changes?

merijng:
@berkansasmaz @maliming 

Just changed the commit, to the first option. I agree!

**By the way:**
Why is ABP on netstandard2.0? In netstandard2.1 (/net7), you can use string.Concat which is even faster. There is also no stackalloc necessary. Below an example+benchmark (you can add it to the console).

```
public static string ReplaceSpan2(this string str, string search, string replace, StringComparison comparisonType = StringComparison.Ordinal)
        {
            //Check.NotNull(str, nameof(str));
            int pos = str.IndexOf(search, comparisonType);
            if (pos < 0)
            {
                return str;
            }

            var sourceSpan = str.AsSpan();
            return string.Concat(sourceSpan.Slice(0, pos), replace, sourceSpan.Slice(pos + search.Length));
        }
```

image

maliming:
hi

> Why is ABP on netstandard2.0? In netstandard2.1 (/net7)

https://github.com/abpframework/abp/issues/5676#issuecomment-706354605

merijng:
@maliming 

First, the change as requested has been done.

Second, what I mean with my question is:

If you have planned netstandard2.1 in the near future, then spans can upgrade the performance of ABP in more ways (like shown in my previous post). Beside that: in netstandard2.1 code of spans is way cleaner.

If not: then I guess this is the only pull request whereas spans can upgrade ABP :(

**About compatibility with .NET Standard 2.1**
I see no reason to support .NET Framework and older versions of .NET, in ABP 7+?

Reference: 

- [Announcing .NET Standard 2.1](https://devblogs.microsoft.com/dotnet/announcing-net-standard-2-1/)
- [Compatiblity](https://learn.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-2-1)

maliming:
hi

We can discuss `.NET Standard` version in a new issue.

maliming (approved):

maliming (approved):

berkansasmaz (approved):

PR #15882, opened 2023-03-06
cms-kit: Fix global resource feature checking

JadynWong:
### Description

- When disabled Host/Tenant global resources feature, the pages will be thrown an exception.
- `GlobalResourcePublicAppService.GetGlobalStyleAsync` renders the `/cms-kit/global-resources/style` controller cache useless and the controller does not return a response cache.

screenshot1

### Changed
- If the global resources feature is disabled, the global resources hook will return blank content.
- Remove style timestamp.


enisn (approved):
Seems Ok at the moment ✅

Thanks for contribution 🎉

PR #15943, opened 2023-03-11
Add edit link in project page for all docs

andmattia:
Add link edit for all docs also for docs without collaborators

### Description

In case a doc doesn't have contributors edit link isn't show. Fix it add link always and contributors only contributors is greater than 0


berkansasmaz:
Thanks, @andmattia 🎉

I will review as soon as possible.

berkansasmaz (approved):

PR #15955, opened 2023-03-14
Add `LastPasswordChangeTime` to `IdenttiyUser` 

realLiangshiwei:
This is a breaking change because you need to update ef core migrations.


realLiangshiwei:
https://github.com/abpframework/abp/pull/15956

Issue #2640, opened 2020-01-15
abp v2.0.0,when FullAuditedAggregateRoot use owned entity types(ValueObject), Efcore throw error


yinchang0626:
When i add a ValueObject as owned by FullAuditedAggregateRoot.
Get a exception below
```csharp
System.InvalidOperationException: 'The type 'ISBN' cannot be marked as owned because a non-owned entity type with the same name already exists.'
```
I just reuse code of #2365 to report this error,but in v1.1.2 or pure aspnet mvc template propject, It works fine.
You can get sample code in
 [this for v1.1.2(PASS)](https://github.com/yinchang0626/abp/tree/feature/book_store_ownedType_112),
[this for v2.0.0(NG)](https://github.com/yinchang0626/abp/tree/feature/book_store_ownedType_200).

@hikalkan 
Suppose Abp support OwnedType of EFCore(#2278 ),I can't find anything to test this feature,
could I add those sample code to BookStore sample or add unitest ,
because of in DDD,ValueObject  is very common sense. We hope abp can fully support it for EFCore

maliming:
TRY

```
protected override void OnModelCreating(ModelBuilder builder)
{
	//Mark owned types, so base.OnModelCreating can handle them
	builder.Owned<ISBN>();

	base.OnModelCreating(builder);

	/* Configure the shared tables (with included modules) here */
	
	//...........
}
```

yinchang0626:
@maliming 
thank you for help , butthe error still exist
![image](https://user-images.githubusercontent.com/16490084/72404427-4921ed80-3790-11ea-8305-c5e4ba7dff0f.png)

maliming:
Please put the code(`builder.Owned<ISBN>();` `base.OnModelCreating(builder);`) in your `OnModelCreating` method.

https://github.com/yinchang0626/abp/blob/feature%2Fbook_store_ownedType_200/samples/BookStore/src/Acme.BookStore.EntityFrameworkCore/EntityFrameworkCore/BookStoreDbContext.cs#L36

yinchang0626:
@maliming 
Thank you very much,it works!!
This is related to the execution order.

Is possible move `base.OnModelCreating(builder);` to bottom of `protected override void OnModelCreating(ModelBuilder builder)`
```csharp
        protected override void OnModelCreating(ModelBuilder builder)
        {
            //builder.Owned<ISBN>();
            //builder.ConfigureBookStore();// move to bottom

[...]

PR #16089, opened 2023-03-27
change GetAsync to FindAsync for organization unit GetCodeOrDefaultAsync method

wangj90:
### Description

Maybe we shoudle use FindAsync instead of GetAsync because if we can't find the organization unit object based on the Id, an EntityNotFound exception message will be thrown. And there is obviously no need to throw an exception here.

### Checklist


maliming (approved):

maliming:
hi 

Why change the order of `OnModelCreating` method?

wangj90:
> hi
> 
> Why change the order of `OnModelCreating` method?

sorry，this commit should be another pr，i will delete it.

maliming:
It should not be changed. Is there any reason?

wangj90:
I will create a issue for the OnModelCreating order. Presumably the reason for this is that the IsOwned judgement should be called after it has been Owned。

wangj90:
Ok, i delete the OnModelCreating order commit。

maliming:
https://github.com/abpframework/abp/issues/2640

PR #16100, opened 2023-03-27
Add response cache attributes in controller of abp.

### LINKED ISSUE #16075: Add version number to ServiceProxyScript and ApplicationLocalizationScript for proper caching
I suggest adding a version number to the ServiceProxyScript and ApplicationLocalizationScript files to ensure proper caching by web browsers. This will prevent users from experiencing unwanted errors due to the browser caching the old version of the file. Adding a version number also ensures that the browser can properly cache the updated file when its contents change.

This update will improve the speed and reliability of the application by ensuring that the browser can quickly and accurately load the latest version of these scripts.

image

maliming:
Resolve #16075


maliming:
Using `ResponseCache` in the framework may cause unexpected behavior. We can replace the built-in controller when needed.

image

https://github.com/abpframework/abp/blob/MyAbpApplicationLocalizationScriptController/templates/app/aspnet-core/src/MyCompanyName.MyProjectName.Web/MyAbpApplicationLocalizationScriptController.cs#L12

PR #16173, opened 2023-04-04
fix card component lint errors

### LINKED ISSUE #16171: Fix card component lint errors
Card component causes lint errors during build-test-lint action. 

* HostBinding or HostListener rather than the `host` metadata
* Directive class names should end with one of these suffixes: "Directive"

bariscanyilmaz:
### Description

Resolves #16171 

TODO: add `Directive` suffix to directive components and use `HostBinding` instead of Host
### Checklist

### How to test it?

### DIFF
--- a/npm/ng-packs/packages/theme-shared/src/lib/components/card/card-body.component.ts
+++ b/npm/ng-packs/packages/theme-shared/src/lib/components/card/card-body.component.ts
@@ -7,7 +7,7 @@ import { Component, HostBinding, Input } from '@angular/core';
   </div>`,
 })
 export class CardBodyComponent {
-  @HostBinding('class') class = 'card-body';
+  @HostBinding('class') componentClass = 'card-body';
   @Input() cardBodyClass: string;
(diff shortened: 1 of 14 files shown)


npm/ng-packs/packages/theme-shared/src/lib/components/card/card-body.component.ts
```diff
@@ -1,16 +1,13 @@
...
-    </div>`,
-  host: {
-    class: 'card-body',
-  }
+  template: ` <div [ngClass]="cardBodyClass" [ngStyle]="cardBodyStyle">
+    <ng-content></ng-content>
+  </div>`,
 })
 export class CardBodyComponent {
+  @HostBinding('class') class = 'card-body';
```
rqx110: the variable `class` is the keyword of typescript, it is better to rename to other name?

rqx110 (approved):

masum-ulu (approved):

PR #16174, opened 2023-04-04
Cms: Allow/Disallow External URLs in Comments

### LINKED ISSUE #15772: Cms-Kit — Don't allow external links
CMS Kit comments section is good for visitor comments.
But sometimes, malicious users can submit advertisement links.
Add an option to CMS-Kit to enable/disable allowing external links.
The allowed link domains must be configurable, like github.com, abp.io, microsoft.com etc...
If the submitted link is out of the allowed list, warn the user and don't allow it.

Option name:   **- [ ] Don't allow external links**

Write how to use it in the docs => https://docs.abp.io/en/abp/latest/Modules/Cms-Kit/Comments

EngincanV:
### Description

Resolves https://github.com/abpframework/abp/issues/15772

With this PR, now developers can allow/disallow external URLs by comment widgets. I have implemented widget-specific settings, which means in a comment section external URLs can be allowed, and in another one, they could be disallowed. ü

We may consider making it a global option, please let me know what you think about that (@yekalkan), so I can work on it.

### Checklist

### DIFF
--- a/modules/cms-kit/src/Volo.CmsKit.Public.Web/Pages/CmsKit/Shared/Components/Commenting/Default.cshtml
+++ b/modules/cms-kit/src/Volo.CmsKit.Public.Web/Pages/CmsKit/Shared/Components/Commenting/Default.cshtml
@@ -32,7 +32,6 @@
           style="@(string.IsNullOrEmpty(repliedCommentId?.ToString() ?? "") ? "" : "display:none")">
         <form class="cms-comment-form">
             <input hidden value="@(repliedCommentId?.ToString() ?? "")" name="repliedCommentId" />
-            <input hidden asp-for="Input.AllowExternalUrls" name="allowExternalUrls" />
             <div class="row">
                 <div class="col">
                     <div class="mb-3 m-0">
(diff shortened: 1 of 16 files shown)


modules/cms-kit/src/Volo.CmsKit.Public.Web/Pages/CmsKit/Shared/Components/Commenting/Default.cshtml
```diff
@@ -32,6 +32,7 @@
           style="@(string.IsNullOrEmpty(repliedCommentId?.ToString() ?? "") ? "" : "display:none")">
         <form class="cms-comment-form">
             <input hidden value="@(repliedCommentId?.ToString() ?? "")" name="repliedCommentId" />
+            <input hidden asp-for="Input.AllowExternalUrls" name="allowExternalUrls" />
```
yekalkan: This doesn't seem to be safe enough, but i understand the reason behind it. You want to set AllowExternalUrls separately for each comment section. However, i don't think it is neccessary. You can set it in options by `EntityType`. Example code:

```
        Configure<CmsKitCommentOptions>(options =>
        {
            options.AllowedExternalUrls = new Dictionary<string, List<string>>
            {
                {
                    "Product",
                    new List<string>()
                    {
                        "https://abp.io/"
                    }
                }
            };
        });
```

With this implementation, you can check it on backend. And i think this will cover the real need. Comments on `Product` EntityTypes won't allow the external urls globally. You can ask @ebicoglu about this.

EngincanV: Yes, this makes more sense. I will update it according to your comment.

yekalkan (approved):

PR #16188, opened 2023-04-05
Fix ExtraProperties problem in Blazor

realLiangshiwei:
### Description

Resolves https://github.com/volosoft/volo/issues/13827

### Checklist


enisn (approved):
Works well! ✅

PR #16189, opened 2023-04-05
Commercial.abp.io - new testimonial localization 

onurpicakci:


EngincanV (commented):
There are some typos and grammar mistakes in the text. The text for English should be:

> "ABP Framework is a good framework but it needs time to understand the different layers, classes, and libraries it used (specially ABP). I spent a lot of time reading the code base, but ABP Commercial save us time to create the project specialty entities (AR) and the repository linked to each of them.
> I liked also the approach used in ABP is very mature, we know is based on DDD and monolith."

Please update the localizations according to the revised text above.

EngincanV (approved):

PR #16191, opened 2023-04-06
Add taghelper for caching scripts.

### LINKED ISSUE #16075: Add version number to ServiceProxyScript and ApplicationLocalizationScript for proper caching

maliming:
Resolves #16075

**TagHelper** will get js content and calculate the hash and write to the cache **every time**, and the controller will try to read the cache.

Before:
```html
<script src="~/Abp/ApplicationLocalizationScript?cultureName=@CultureInfo.CurrentUICulture.Name"></script>
<script src="~/Abp/ApplicationConfigurationScript"></script>
<script src="~/Abp/ServiceProxyScript"></script>
```

Now:

```html
<abp-application-localization-script culture-name="@CultureInfo.CurrentUICulture.Name" />
<abp-application-configuration-script />
<abp-service-proxy-script />
```

```html
<script src="/Abp/ApplicationLocalizationScript?cultureName=en-US&hash=76b828152119343c0457ffcfb046ed59"></script>
<script src="/Abp/ApplicationConfigurationScript?hash=8612ca794d94425b727dd1fe3e4c4189"></script>
<script src="/Abp/ServiceProxyScript?hash=b960c0f2593583133c6dc4ee7ddd4b36"></script>
    
```


framework/src/Volo.Abp.AspNetCore.Mvc.UI.Bootstrap/TagHelpers/CacheScripts/AbpApplicationConfigurationScriptTagHelperService.cs
```diff
@@ -0,0 +1,57 @@
...
+using Volo.Abp.Caching;
+using Volo.Abp.Json;
+
+namespace Volo.Abp.AspNetCore.Mvc.UI.Bootstrap.TagHelpers.CacheScripts;
+
+public class AbpApplicationConfigurationScriptTagHelperService : AbpTagHelperService<AbpApplicationConfigurationScriptTagHelper>
+{
+    protected IAbpApplicationConfigurationAppService AbpApplicationConfigurationAppService { get; }
+    protected IJsonSerializer JsonSerializer { get; }
+    protected IDistributedCache<AbpControllerScriptCacheItem> Cache { get; }
```
hikalkan, line 18: Do not use `IDistributedCache`. Distributed cache is not proper to this use case;

* It is an heavy operation for such a large data.
* We are calculating it in every request, so we don't need to share it with other application instances.
* It may conflict with different versions of the same application since they will use the same cache item (actually, here all applications are using the same cache item, which is worse).

This comment is valid for other scripts too...

We can use IMemoryCache. However, if we use a web farm, the request will probably go to another server, so will be no meaning to cache it. In this case, we may consider to not cache it and always re-calculate. In this case, it will make a double calculation. I couldn't find an ideal solution yet. We should re-think that.

framework/src/Volo.Abp.AspNetCore.Mvc.UI.Bootstrap/TagHelpers/CacheScripts/AbpApplicationLocalizationScriptTagHelperService.cs
```diff
@@ -0,0 +1,59 @@
...
+
+    protected virtual async Task<string> GetJsFileHash()
+    {
+        var dto = await AbpApplicationLocalizationAppService.GetAsync(new ApplicationLocalizationRequestDto
+        {
+            CultureName = TagHelper.CultureName,
+            OnlyDynamics = TagHelper.OnlyDynamics
+        });
+
+        var json = JsonSerializer.Serialize(dto, indented: true);
```
hikalkan, line 55: `indented` makes the file size larger. Use indented only for development environment. Set false for other environments.
This comment is valid for others too.

framework/src/Volo.Abp.AspNetCore.Mvc/Volo/Abp/AspNetCore/Mvc/AbpAspNetCoreMvcModule.cs
```diff
@@ -104,10 +106,29 @@ public override void ConfigureServices(ServiceConfigurationContext context)
...
+        var abpAspNetCoreMvcOptions = context.Services.ExecutePreConfiguredActions<AbpAspNetCoreMvcOptions>();
+
         var mvcCoreBuilder = context.Services.AddMvcCore(options =>
         {
             options.Filters.Add(new AbpAutoValidateAntiforgeryTokenAttribute());
+
+            options.CacheProfiles.Add(AbpApplicationLocalizationScriptController.GetCacheProfileName, new CacheProfile
+            {
+                Duration = abpAspNetCoreMvcOptions.ScriptCacheDuration,
+                VaryByQueryKeys = new []{ "hash" }
```
hikalkan, line 118: use `_v` instead of `hash` in querystring (everywhere). Hashing is our internal way to implement the versioning, and it is an implementation detail (I specifically indicated that in [my comment](https://github.com/abpframework/abp/issues/16075#issuecomment-1486740668)).

framework/src/Volo.Abp.AspNetCore.Mvc.UI.Bootstrap/TagHelpers/CacheScripts/AbpApplicationConfigurationScriptTagHelperService.cs
```diff
@@ -0,0 +1,57 @@
...
+
+    protected virtual async Task<string> GetJsFileHash()
+    {
+        var dto = await AbpApplicationConfigurationAppService.GetAsync(new ApplicationConfigurationRequestOptions
+        {
+            IncludeLocalizationResources = TagHelper.IncludeLocalizationResources
+        });
+
+        var json = JsonSerializer.Serialize(dto, indented: true);
+        await Cache.SetAsync(nameof(AbpApplicationConfigurationScriptController), new AbpControllerScriptCacheItem(json));
```
hikalkan, line 54: Application configuration script content depends on the current user, client and tenant. We can use the same key for all users.

framework/src/Volo.Abp.AspNetCore.Mvc/Volo/Abp/AspNetCore/Mvc/Localization/AbpApplicationLocalizationScriptController.cs
```diff
@@ -53,14 +58,19 @@ await LocalizationAppService.GetAsync(input)
         );
     }
 
-    protected virtual string CreateScript(ApplicationLocalizationDto localizationDto)
+    protected virtual async Task<string> CreateScript(ApplicationLocalizationRequestDto input)
     {
+        var json = (await ScriptCache.GetAsync(nameof(AbpApplicationLocalizationScriptController)))?.Script;
+        if (json.IsNullOrWhiteSpace())
```
hikalkan, line 64: While working with caches, we should always consider the scenario where the cache item is removed from cache for some reason. In this case, we should re-generate it (just like while we cached before)

hikalkan (changes requested):

hikalkan:
@maliming thank you for your great effort to implement this. However, with current state, it brings problems more than it solves. If you can think and come with a finer solution, we can re-consider it. Otherwise, I will try to think it on more. If we can't find a good implementaion, it is better to not do it.

Issue #3148, opened 2020-03-16
Add IRepository.GetAsync and IRepository.FindAsync methods those get predicate


hikalkan:
This simplifies to query a single entity that has a complex primary key.

Signatures:

````csharp
/// <summary>
/// Get a single entity by the given <paramref name="predicate"/>.
/// It returns null if no entity with the given <paramref name="predicate"/>.
/// It throws <see cref="InvalidOperationException"/> if there are multiple entities with the given <paramref name="predicate"/>.
/// </summary>
/// <param name="predicate">A condition to find the entity</param>
/// <param name="includeDetails">Set true to include all children of this entity</param>
/// <param name="cancellationToken">A <see cref="T:System.Threading.CancellationToken" /> to observe while waiting for the task to complete.</param>
Task<TEntity> FindAsync(
    [NotNull] Expression<Func<TEntity, bool>> predicate,
    bool includeDetails = true,
    CancellationToken cancellationToken = default
);

/// <summary>
/// Get a single entity by the given <paramref name="predicate"/>.
/// It throws <see cref="EntityNotFoundException"/> if there is no entity with the given <paramref name="predicate"/>.
/// It throws <see cref="InvalidOperationException"/> if there are multiple entities with the given <paramref name="predicate"/>.
/// </summary>
/// <param name="predicate">A condition to filter entities</param>
/// <param name="includeDetails">Set true to include all children of this entity</param>
/// <param name="cancellationToken">A <see cref="T:System.Threading.CancellationToken" /> to observe while waiting for the task to complete.</param>
Task<TEntity> GetAsync(
    [NotNull] Expression<Func<TEntity, bool>> predicate,
    bool includeDetails = true,
    CancellationToken cancellationToken = default
);

````

PR #16313, opened 2023-04-19
Blog Module - Member page created

onurpicakci:
Ekran Resmi 2023-04-19 12 16 07


gizemmutukurt:
The member profile doesn't look a bit good by design, so you can look at the sample from the adjacent community. It looks better there.
image

**Community Example:**
image

onurpicakci:
We can't add extra user information (job, company name, social accounts). If we want to include it, the blog module will depend on IdentityUser, and this is something we don't want. But I will edit the username part in the design.

gizemmutukurt:
latest view
image

gizemmutukurt (approved):

modules/blogging/src/Volo.Blogging.MongoDB/Volo/Blogging/Posts/MongoPostRepository.cs
```diff
@@ -58,5 +58,13 @@ public async Task<List<Post>> GetOrderedList(Guid blogId, bool @descending = fal
 
             return await query.OrderByDescending(x => x.CreationTime).ToListAsync(GetCancellationToken(cancellationToken));
         }
+
+        public async Task<List<Post>> GetListByUserIdAsync(Guid userId, CancellationToken cancellationToken = default)
+        {
+            var query = (await GetMongoQueryableAsync(cancellationToken)).Where(x => x.CreatorId == userId)
+                .OrderBy(x => x.CreationTime);
```
EngincanV: On the EF Core side, you are getting posts by newest, but in MongoDB implementation it's the opposite. So, please make the following change:

```suggestion
                .OrderByDescending(x => x.CreationTime);
```

ebicoglu: EFCore orders by CreationTime desc but this orders by CreationTime asc. If there's no special case, always order by CreationTime DESC

modules/blogging/src/Volo.Blogging.Web/Pages/Members/Index.cshtml
```diff
@@ -0,0 +1,96 @@
...
+                                                <span class="text-dark-200">@post.ReadCount.ToString() @L["Views"]</span>
+                                            </div>
+                                            <h3 class="post-title mb-3">
+                                                <a href="@Model.GetBlogPostUrl(post)">
+                                                    @post.Title
+                                                </a>
+                                            </h3>
+                                            <p class="post-desc">
+                                                <a href="@Model.GetBlogPostUrl(post)">
+                                                    @(post.Description.Length >= 150 ? post.Description.Substring(0, 150) + "..." : post.Description)
```
EngincanV: You can use the [`TruncateWithPostfix` extension method](https://github.com/abpframework/abp/blob/dev/framework/src/Volo.Abp.Core/System/AbpStringExtensions.cs#L525-L528):

```suggestion
post.Description.TruncateWithPostfix(150)
```

modules/blogging/src/Volo.Blogging.Web/Pages/Members/Index.cshtml.cs
```diff
@@ -0,0 +1,66 @@
...
+        _postAppService = postAppService;
+        _memberAppService = memberAppService;
+        _blogAppService = blogAppService;
+    }
+
+    public async Task<IActionResult> OnGetAsync(string userName)
+    {
+        User = await _memberAppService.GetAsync(userName);
+
+        if (User is null)
```
EngincanV, line 37: You are using the `GetAsync` method, and if it could not find a record, it should throw the `EntityNotFoundException`, so should not return `null`.

modules/blogging/src/Volo.Blogging.Application/Volo/Blogging/Members/MemberAppService.cs
```diff
@@ -0,0 +1,24 @@
...
+    private readonly IRepository<BlogUser, Guid> _userRepository;
+
+    public MemberAppService(IRepository<BlogUser, Guid> userRepository)
+    {
+        _userRepository = userRepository;
+    }
+    
+    public async Task<BlogUserDto> GetAsync(string username)
+    {
+        var user = await _userRepository.GetAsync(x => x.UserName == username);
```
ebicoglu: You must change the method name `GetAsync()` to `FindAsync()`.
Don't use `GetAsync()` because it'll throw an exception if no user is found in the DB.
See the IRepository `FindAsync` and `GetAsync` difference https://github.com/abpframework/abp/issues/3148

EngincanV (commented):
In addition to my comments, please also generate a proxy for `MemberAppService` and also generate static js client proxies for updated app-services. (@onurpicakci)

modules/blogging/src/Volo.Blogging.MongoDB/Volo/Blogging/Posts/MongoPostRepository.cs
EngincanV: Same with https://github.com/abpframework/abp/pull/16313#discussion_r1176109802

modules/blogging/src/Volo.Blogging.Web/Pages/Members/Index.cshtml.cs
EngincanV: Same with https://github.com/abpframework/abp/pull/16313#discussion_r1176116165

modules/blogging/src/Volo.Blogging.Application/Volo/Blogging/Members/MemberAppService.cs
```diff
@@ -0,0 +1,24 @@
...
+    public MemberAppService(IRepository<BlogUser, Guid> userRepository)
+    {
+        _userRepository = userRepository;
+    }
+    
+    public async Task<BlogUserDto> FindAsync(string username)
+    {
+        var user = await _userRepository.FindAsync(x => x.UserName == username);
+        
+        return ObjectMapper.Map<BlogUser, BlogUserDto>(user);
```
EngincanV: You can make a null check here, and return null if it's not finding the record, otherwise, map to the `BlogUserDto` as you do now. (because, it throws `NullReferenceException` if the user could not be found and try to map the null value to `BlogUserDto`.

EngincanV (changes requested):

EngincanV (approved):

ebicoglu (approved):

PR #16324, opened 2023-04-20
Fix Unitofwork items concurrency problem

### LINKED ISSUE #16323: Unitofwork items concurrency problem
https://support.abp.io/QA/Questions/4898

realLiangshiwei:
### Description

Resolves https://github.com/abpframework/abp/issues/16323

### Checklist

### DIFF
--- a/framework/src/Volo.Abp.Uow/Volo/Abp/Uow/UnitOfWorkExtensions.cs
+++ b/framework/src/Volo.Abp.Uow/Volo/Abp/Uow/UnitOfWorkExtensions.cs
@@ -25,7 +25,7 @@ public static class UnitOfWorkExtensions
         }
         else
         {
-            unitOfWork.Items.Add(key, value);
+            unitOfWork.Items.TryAdd(key, value);
         }
     }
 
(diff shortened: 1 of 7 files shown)


maliming (approved):

sturlath:
Why not 7.2 final?

framework/src/Volo.Abp.Uow/Volo/Abp/Uow/UnitOfWorkExtensions.cs
```diff
@@ -25,7 +25,7 @@ public static void AddItem<TValue>([NotNull] this IUnitOfWork unitOfWork, string
         }
```
hikalkan: We may always use `unitOfWork.Items[key] = value;` without the `if` statement. I don't know why it was designed like that (checking containskey), however, if we consider concurrency now, this code is being a problem:

````csharp
if (!unitOfWork.Items.ContainsKey(key))
        {
            unitOfWork.Items[key] = value;
        }
````

Because another thread may change the `Items` after `ContainsKey`, but before setting the value.

hikalkan (changes requested):

hikalkan:
Actually, Unit of Work was designed as non thread safe. A unit of work should be used by only one thread in a time, so it doesn't need to support concurrency.

framework/src/Volo.Abp.Uow/Volo/Abp/Uow/UnitOfWorkExtensions.cs
realLiangshiwei: OK, I undid the changes and removed the if statement, it should fix the problem

realLiangshiwei:
Ok,  we can close this if it's by design.

PR #16359, opened 2023-04-24
Add "Single Layer" option to Get Started tutorial

braim23:


docs/en/Getting-Started-Create-Solution-Single.md
```diff
@@ -0,0 +1,52 @@
...
+}
+````
+
+> This document assumes that you prefer to use **{{ UI_Value }}** as the UI framework and **{{ DB_Value }}** as the database provider. For other options, please change the preference on top of this document.
+
+## Create a New Project
+
+We will use the ABP CLI to create a new ABP project.
+
+> Alternatively, you can **create and download** projects from the [ABP Framework website](https://abp.io/get-started) by easily selecting all options from the page.
```
EngincanV: I think this can be misleading because on the get-started page, only the CLI command is generated according to the options. (the project is not generated)

Also, please make this change for other docs as well.

docs/en/Getting-Started-Create-Solution-Single.md
```diff
@@ -0,0 +1,52 @@
...
+{{ if DB == "Mongo" }}
+
+## MongoDB Transactions
+
+The [startup template](Startup-templates/Index.md) **disables** transactions in the `.MongoDB` project by default. If your MongoDB server supports transactions, you can enable it in the *YourProjectModule* class's `ConfigureMongoDB` method:
+
+  ```csharp
+Configure<AbpUnitOfWorkDefaultOptions>(options =>
+{
+	options.TransactionBehavior = UnitOfWorkTransactionBehavior.Disabled;
```
EngincanV: We are talking about enabling the transaction but the code snippet uses the default disabled behaviour, we can change this code block as follows:

```suggestion
	options.TransactionBehavior = UnitOfWorkTransactionBehavior.Enabled; //or UnitOfWorkTransactionBehavior.Auto
```

docs/en/Getting-Started-Running-Solution-Single.md
```diff
@@ -0,0 +1,98 @@
...
+
+Running the application is pretty straight-forward, you can run the application with any IDE that supports .NET or by running the `dotnet run` CLI command in the directory of your project: 
+
+{{else if UI=="Blazor"}}
+
+Running the application is pretty straight-forward, you just need to run the `TodoApp.Host` application with any IDE that supports .NET or by running the `dotnet run` CLI command in the directory of your project.
+
+> **Note:** The `host` application hosts and serves the `blazor` application. Therefore, you should run the `host` application only.
+
+After the application runs, open the application in your default browser:
```
EngincanV:
```suggestion
After the application runs, open the application in your default browser.
```

After this line, "After running the project, the index page should be seen as below:" will be shown in the next paragraph, so we should use *.* instead of *:*.

docs/en/Getting-Started-Running-Solution-Single.md
```diff
@@ -0,0 +1,98 @@
...
+
+You can explore and test your HTTP API with this UI. If it works, then we can run the Angular client application.
+
+You can run the application using the following (or `yarn start`) command:
+
+````bash
+npm start
+````
+
+This command takes time, but eventually runs and opens the application in your default browser:
```
EngincanV:
```suggestion
This command takes time, but eventually runs and opens the application in your default browser.
```

After this line, "After running the project, the index page should be seen as below:" will be shown in the next paragraph, so we should use *.* instead of *:*.

EngincanV (approved):

PR #16381, opened 2023-04-26
Correction to Add-Migration command

jrees:
### Description

In section 'Create a new Database Migration' I believe the command should be change from 'Add-Migration Created_Book_Entity' to 'Add-Migration Added_Authors'.

This appears to be a simple error missed when copying and pasting the previous section about adding a migration for creating the book entity, so there is nothing to test.


maliming (approved):

maliming:
Thanks @jrees

PR #16422, opened 2023-04-28
Owner and creator fluent authorization extensions

### LINKED ISSUE #16421: Introduce `IMayHaveOwner` and `IMustHaveOwner` for entities

Many types of entities belong to a user. An order has a customer ID, a book has an author ID, and a todo has an owner ID. We can call these IDs `OwnerId`.

The `OwnerId` is different from the `CreatorId`. The creator is the user who created the entity, but we can't determine that he owns the entity. For example, if the admin creates an order for a customer, the creator is the admin, but the owner is the customer.

If ABP provides shared `IMayHaveOwner` and `IMustHaveOwner` interfaces, we can create extension methods for owner authorization.

I want to create a PR for it.

### Additional context

_No response_

gdlcf88:
Resolve https://github.com/abpframework/abp/issues/16421.

This PR is based on https://github.com/abpframework/abp/pull/16420.

### DIFF
--- a/framework/src/Volo.Abp.Authorization.Abstractions/Volo/Abp/Authorization/RoleRequirement.cs
+++ b/framework/src/Volo.Abp.Authorization.Abstractions/Volo/Abp/Authorization/RoleRequirement.cs
@@ -1,0 +1,21 @@
+using JetBrains.Annotations;
+using Microsoft.AspNetCore.Authorization;
+
+namespace Volo.Abp.Authorization;
+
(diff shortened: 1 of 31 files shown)


maliming:
We can use `Policy-based authorization` to implement similar feature.

@hikalkan  What do you think?

https://learn.microsoft.com/en-us/aspnet/core/security/authorization/policies?view=aspnetcore-7.0

PR #16451, opened 2023-05-02
Use multiple target frameworks.

### LINKED ISSUE #16430: How long to update the target framework "netstandard2.1" version?
Please update this version target framework "netstandard2.1" as soon as possible because this new Microsoft version has more running speed and new class features.
Please replace or update all dll classes in source abp framwork .

maliming:
Resolve #16430

We can use the below code to use different features.
```cs
#if NET7_0
        var client = new HttpClient();
        client.DefaultRequestVersion = HttpVersion.Version20;
#elif NETSTANDARD2_1
        var client = new HttpClient();
#elif NETSTANDARD2_0
        var client = new HttpClient();
#endif
```


hikalkan (approved):

applebananamilk:
Whether to test against multiple frameworks?

PR #16496, opened 2023-05-05
Script nonce support for CSP

salihozkara:
### Description

If someone wants to use the script nonce value within the scope of CSP, arrangements have been made in the abp framework so that there is no problem in the bundle system.

### Checklist


salihozkara:
Here are some potential changes that may break existing code: 

- A new NonceScriptTagHelper has been added, replacing the script tag in HTML. If there is any C# code used inside the HTML tag, it may cause an error. For example, `<script @string.empty>` will no longer work.

EngincanV (commented):
Instead of introducing a new tag helper, can't we add a new property to the `abp-script` tag helper, as we did for *Defer*:

https://github.com/abpframework/abp/blob/c8199a2582f9e5302047b9c4360174e30e2e62a2/framework/src/Volo.Abp.AspNetCore.Mvc.UI.Bundling/Volo/Abp/AspNetCore/Mvc/UI/Bundling/TagHelpers/AbpScriptTagHelper.cs#L8-L9

*Adding an property*:

```csharp
 [HtmlAttributeName("nonce")] 
 public bool Nonce { get; set; } 
```

If you do like this, you would not need to add a new property to `AbpSecurityHeadersOptions` and would not need to create a new tag helper. 

---

Also, you would not need to inject the `IOptions<AbpSecurityHeadersOptions>` to `AbpTagHelperStyleService`.

What do you think @salihozkara?

salihozkara:
> Instead of introducing a new tag helper, can't we add a new property to the `abp-script` tag helper, as we did for _Defer_:
> 
> https://github.com/abpframework/abp/blob/c8199a2582f9e5302047b9c4360174e30e2e62a2/framework/src/Volo.Abp.AspNetCore.Mvc.UI.Bundling/Volo/Abp/AspNetCore/Mvc/UI/Bundling/TagHelpers/AbpScriptTagHelper.cs#L8-L9
> 
> _Adding an property_:
> 
> ```cs
>  [HtmlAttributeName("nonce")] 
>  public bool Nonce { get; set; } 
> ```
> 
> If you do like this, you would not need to add a new property to `AbpSecurityHeadersOptions` and would not need to create a new tag helper.
> 
> Also, you would not need to inject the `IOptions<AbpSecurityHeadersOptions>` to `AbpTagHelperStyleService`.
> 
> What do you think @salihozkara?

The purpose of creating a script tag helper is to automatically add a nonce value to all scripts when `UseContentSecurityPolicyNonce` option is activated. By using this tag helper, the scripts will recognize it as the source of the nonce value when the user option is enabled. Otherwise, it would be necessary to manually modify all script tags.

> Also, you would not need to inject the IOptions<AbpSecurityHeadersOptions> to AbpTagHelperStyleService.

I can remove the dependency and control it from the context. Ex: https://github.com/abpframework/abp/blob/5de1132c52e8a6f37be0b0ede4e15d359933a444/framework/src/Volo.Abp.AspNetCore.Mvc.UI.Bundling/Volo/Abp/AspNetCore/Mvc/UI/Bundling/TagHelpers/AbpTagHelperScriptService.cs#L52

framework/src/Volo.Abp.AspNetCore/Volo/Abp/AspNetCore/Security/AbpSecurityHeadersMiddleware.cs
```diff
@@ -27,13 +28,31 @@ public async Task InvokeAsync(HttpContext context, RequestDelegate next)
 
         /*The X-Frame-Options HTTP response header can be used to indicate whether or not a browser should be allowed to render a page in a <frame>, <iframe> or <object>. SAMEORIGIN makes it being displayed in a frame on the same origin as the page itself. The spec leaves it up to browser vendors to decide whether this option applies to the top level, the parent, or the whole chain*/
         AddHeader(context, "X-Frame-Options", "SAMEORIGIN");
+        
+        var contentSecurityPolicyValueDictionary = Options.Value.ContentSecurityPolicyValueDictionary.ToDictionary(x=>x.Key, x=>x.Value);
```
ebicoglu: move the line 32 to 40

framework/src/Volo.Abp.AspNetCore/Volo/Abp/AspNetCore/Security/AbpSecurityHeadersMiddleware.cs
```diff
@@ -27,13 +28,31 @@ public async Task InvokeAsync(HttpContext context, RequestDelegate next)
...
-        if (Options.Value.UseContentSecurityPolicyHeader)
+        if (Options.Value.UseContentSecurityPolicyHeader && !Options.Value.IgnoredUrls.Any(x => context.Request.Path.StartsWithSegments(x.EnsureStartsWith('/'))))
         {
+            if (Options.Value.UseContentSecurityPolicyNonce)
+            {
+                if (!context.Items.ContainsKey(AbpAspNetCoreConsts.ScriptNonceKey))
+                {
+                    var randomValue = Guid.NewGuid().ToString("N");
+                    context.Items.Add(AbpAspNetCoreConsts.ScriptNonceKey, randomValue);
+                    if (contentSecurityPolicyValueDictionary.TryGetValue("script-src", out var scriptSrc))
```
ebicoglu: use ABP's `AddIfNotContains` extension method

framework/src/Volo.Abp.AspNetCore/Volo/Abp/AspNetCore/Security/AbpSecurityHeadersMiddleware.cs
```diff
@@ -27,13 +28,31 @@ public async Task InvokeAsync(HttpContext context, RequestDelegate next)
 
         /*The X-Frame-Options HTTP response header can be used to indicate whether or not a browser should be allowed to render a page in a <frame>, <iframe> or <object>. SAMEORIGIN makes it being displayed in a frame on the same origin as the page itself. The spec leaves it up to browser vendors to decide whether this option applies to the top level, the parent, or the whole chain*/
         AddHeader(context, "X-Frame-Options", "SAMEORIGIN");
+        
+        var contentSecurityPolicyValueDictionary = Options.Value.ContentSecurityPolicyValueDictionary.ToDictionary(x=>x.Key, x=>x.Value);
 
-        if (Options.Value.UseContentSecurityPolicyHeader)
+        if (Options.Value.UseContentSecurityPolicyHeader && !Options.Value.IgnoredUrls.Any(x => context.Request.Path.StartsWithSegments(x.EnsureStartsWith('/'))))
```
ebicoglu: use if guards

ebicoglu: add extra if guard for only html document check for content type

ebicoglu (changes requested):
check the requested content: Only add a CSP header if it's plain HTML.
check this feature on 3rd party libraries that have UI for instance: `Swagger`, `Hangfire`

hikalkan (approved):

salihozkara:
Ignore Elsa example:

```cs
Configure<AbpSecurityHeadersOptions>(o =>
{
        o.UseContentSecurityPolicyHeader = true;
        o.UseContentSecurityPolicyScriptNonce = true;
        o.IgnoredScriptNonceSelectors.Add(context =>
        {
             var endpoint = context.GetEndpoint();
             return Task.FromResult(endpoint?.Metadata.GetMetadata<PageRouteMetadata>()?.RouteTemplate == "/{YOURHOSTPAGE}");
        });
});
```

framework/src/Volo.Abp.AspNetCore.Mvc.UI.Bundling/Volo/Abp/AspNetCore/Mvc/UI/Bundling/TagHelpers/AbpTagHelperStyleService.cs
```diff
@@ -8,19 +8,23 @@
 using Microsoft.AspNetCore.Razor.TagHelpers;
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Options;
+using Volo.Abp.AspNetCore.Security;
 
 namespace Volo.Abp.AspNetCore.Mvc.UI.Bundling.TagHelpers;
 
 public class AbpTagHelperStyleService : AbpTagHelperResourceService
 {
+    private readonly IOptions<AbpSecurityHeadersOptions> _securityHeadersOptions;
```
EngincanV: Here, you can use the class as follows:

```suggestion
    protected AbpSecurityHeadersOptions SecurityHeadersOptions { get; }
```

Then, change the [27th line](https://github.com/abpframework/abp/pull/16496/files#diff-59be621ab60b20a1194159c130e092748728730c1603afc2cde3a9379ca3474bR27) as follows:

```csharp
SecurityHeadersOptions = securityHeadersOptions.Value;
```

framework/src/Volo.Abp.AspNetCore/Volo/Abp/AspNetCore/Security/AbpSecurityHeadersOptions.cs
```diff
@@ -1,17 +1,29 @@
...
 namespace Volo.Abp.AspNetCore.Security;
 
 public class AbpSecurityHeadersOptions
 {
     public bool UseContentSecurityPolicyHeader { get; set; }
+    
+    public bool UseContentSecurityPolicyNonce { get; set; }
 
-    public string ContentSecurityPolicyValue { get; set; }
+    public Dictionary<string,IEnumerable<string>> ContentSecurityPolicyValues { get; }
```
EngincanV:
```suggestion
    public Dictionary<string, IEnumerable<string>> ContentSecurityPolicyValues { get; }
```

EngincanV (approved):

hikalkan (approved):

hikalkan (approved):

PR #16504, opened 2023-05-08
Check token expiration.

maliming:
https://support.abp.io/QA/Questions/5022/User-timeoutpermission-behaviour

The validity period of `access_token` is shorter than `Cookies` and the API request will be 401 after it expires. But the `ThreadCurrentPrincipal` will not be invalidated.

The `CheckTokenExpiration` method will check the expiration time of `access_token` and request the authserver every 1 minute(configurable) to ensure the `access_token` is valid.

Sign out cookies if the token expires or is no longer valid.


framework/src/Volo.Abp.AspNetCore/Microsoft/Extensions/DependencyInjection/CookieAuthenticationOptionsExtensions.cs
```diff
@@ -0,0 +1,96 @@
...
+    public static CookieAuthenticationOptions CheckTokenExpiration(this CookieAuthenticationOptions options, string oidcAuthenticationScheme = "oidc", TimeSpan? advance = null, TimeSpan? validationInterval = null)
+    {
+        advance ??= TimeSpan.FromMinutes(3);
+        validationInterval ??= TimeSpan.FromMinutes(1);
+        var originalHandler = options.Events.OnValidatePrincipal;
+        options.Events.OnValidatePrincipal = async principalContext =>
+        {
+            originalHandler?.Invoke(principalContext);
+
+            if (principalContext.Principal != null && principalContext.Principal.Identity != null && principalContext.Principal.Identity.IsAuthenticated)
```
hikalkan: Use more guard clauses and returns instead of nested "if"s. Currently, it is very hard to read the code.

framework/src/Volo.Abp.AspNetCore/Microsoft/Extensions/DependencyInjection/CookieAuthenticationOptionsExtensions.cs
```diff
@@ -0,0 +1,96 @@
...
+                if (DateTimeOffset.TryParseExact(tokenExpiresAt, "o", null, DateTimeStyles.RoundtripKind, out var expiresAt) && expiresAt < DateTimeOffset.UtcNow.Subtract(advance.Value))
+                {
+                    logger.LogInformation("The access_token is expired.");
+                    await SignOutAsync(principalContext);
+                    return;
+                }
+
+                if (principalContext.Properties.IssuedUtc != null)
+                {
+                    if (DateTimeOffset.UtcNow.Subtract(principalContext.Properties.IssuedUtc.Value) > validationInterval)
```
hikalkan: It will always check for every request, after the first validationInterval passes. Because, there is no system that records the last validation time. Or, I didn't undertand the code.

maliming: `principalContext.ShouldRenew = true;` will update the `IssuedUtc`.

The cookie will be renewed so the `IssuedUtc` is set as current utc.

framework/src/Volo.Abp.AspNetCore/Microsoft/Extensions/DependencyInjection/CookieAuthenticationOptionsExtensions.cs
maliming: ok

hikalkan (approved):

genoher:
It is possible to create a patch for version 6. Thanks.

PR #16513, opened 2023-05-09
docs.abp.io - misspelled title edit

onurpicakci:


EngincanV:
I think we don't need to rename the file. This is a common convention, see an example: 

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/tutorials/signalr/samples/6.x/SignalRChat/Pages/Index.cshtml.cs

Also, our sample projects follow this convention as well. For an example, see https://github.com/abpframework/abp-samples/blob/master/BookStore-Mvc-EfCore/src/Acme.BookStore.Web/Pages/Authors/Index.cshtml.cs

Issue #15601, opened 2023-02-04
Swagger UI Authentication Problem on isolated network


gterdem:
Original issue: https://github.com/abpframework/abp/issues/13922

This is not specific to k8s. It is related to isolated network.

Since Swagger UI authentication doesn't have an option to define meta-address (yet), both browser interactions (like login) and validation requests are done to the same endpoint.

Authorization flow when logging in should be done to real DNS with HTTPS .
`.well-known` endpoint should be done to internal network container using HTTP.

gterdem:
Related issue: https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/2604

gterdem:
Fixed via https://github.com/abpframework/abp/pull/16574

Requires template update to use new configuration.

PR #16574, opened 2023-05-16
Enhance Swagger to support OpenIdConnect

realLiangshiwei:
### Description

Related https://github.com/abpframework/abp/issues/15601

```csharp
context.Services.AddAbpSwaggerGenWithOidc(
      configuration["AuthServer:Authority"],
      flows: new[] { "authorization_code" },
      scopes: new[] { "MyProjectName" },
      discoveryEndpoint: "https://localhost:44305/.well-known/openid-configuration",
      setupAction: options =>
      {
          options.SwaggerDoc("v1", new OpenApiInfo { Title = "MyProjectName API", Version = "v1" });
          options.DocInclusionPredicate((docName, description) => true);
          options.CustomSchemaIds(type => type.FullName);
      });
```

image

### Checklist


framework/src/Volo.Abp.Swashbuckle/Microsoft/Extensions/DependencyInjection/AbpSwaggerGenServiceCollectionExtensions.cs
```diff
@@ -64,19 +62,64 @@ public static class AbpSwaggerGenServiceCollectionExtensions
...
+    }
+
+    public static IServiceCollection AddAbpSwaggerGenWithOidc(
+        this IServiceCollection services,
+        [NotNull] string authority,
+        Action<SwaggerGenOptions> setupAction = null,
+        List<string> flows = null,
+        string openIdConnectDiscoveryEndpoint = "/.well-known/openid-configuration")
+    {
+        var openIdConnectDiscoveryUrl = new Uri($"{authority.TrimEnd('/')}{openIdConnectDiscoveryEndpoint.EnsureStartsWith('/')}");
```
gterdem: If we use a different endpoint for metadata, it will end up something like:
`https://localhost:44301/https://myabp.com/.well-known/openid-configuration`

If null, it should be: `new Uri($"{authority.TrimEnd('/')}{openIdConnectDiscoveryEndpoint.EnsureStartsWith('/')}"`
otherwise it should be the new URL we pass, like: `https://myabp.com/.well-known/openid-configuration`

gterdem (changes requested):
The main problem here is, we don't pass the scopes so all the scopes are available under any API.
image

Each microservice should only have its own scope available.

gterdem (commented):
The Authority(issuer) and the MetaAddress(DiscoveryEndpoint) are the same.

We need to be able to set them differently. Intercept the Login and Logout requests from the browser to Authority (Issuer) and all the rest to MetaAddress(DiscoveryEndpoint).

gterdem:
Configured the request interception.

Result with real dns:

[swagger-auth-fixed.webm](https://github.com/abpframework/abp/assets/25867/ef34657f-05a4-4a98-aefa-331e785267b1)

gterdem (approved):

PR #16602, opened 2023-05-18
[Blazor wasm]`logout` if token expired.

maliming:
https://support.abp.io/QA/Questions/5071

https://github.com/abpframework/abp/assets/6908465/f97406d6-264e-4162-b09a-d1d52f153402


framework/src/Volo.Abp.AspNetCore.Components.WebAssembly/Volo/Abp/AspNetCore/Components/WebAssembly/AbpAspNetCoreComponentsWebAssemblyModule.cs
```diff
@@ -18,7 +19,8 @@ namespace Volo.Abp.AspNetCore.Components.WebAssembly;
 [DependsOn(
     typeof(AbpAspNetCoreMvcClientCommonModule),
     typeof(AbpUiModule),
-    typeof(AbpAspNetCoreComponentsWebModule)
+    typeof(AbpAspNetCoreComponentsWebModule),
+    typeof(AbpEventBusModule)
```
hikalkan: Why adding this to dependencies? Volo.Abp.AspNetCore.Components.WebAssembly.csproj has no such a project reference.

hikalkan (changes requested):

framework/src/Volo.Abp.AspNetCore.Components.WebAssembly/Volo/Abp/AspNetCore/Components/WebAssembly/AbpAspNetCoreComponentsWebAssemblyModule.cs
maliming: I add a `ClientProxyExceptionEventHandler` in this project.

maliming: I will also add the project reference.

hikalkan: If there is no project reference, but you can use the module (that defines ClientProxyExceptionEventHandler), it means you indirectly reference the project. In this case, you should have an indirect module dependency. So, you should not define the module dependency, it should also come indirectly.

hikalkan (approved):

PR #16607, opened 2023-05-18
Added support for binary-encoded CloudEvents

whytem:
The CloudEvent format adopted by Dapr permits events to be encoded using either "application/json" or "application/octet-stream" types as indicated by the "datacontenttype" field of the event.

When the event is encoded using the "application/octet-stream" type, the payload is a base64 string in the "data_base64" property and there is no "data" property. As such, the current controller code will fail with a KeyNotFoundException. This condition was occurring using redis configured as the Dapr StateStore.

The below fix checks for a "data_base64" property which is effectively mutually exclusive with a "data" property. If so, it decodes the payload which is an application/json encoded CloudEvent. See below:

**Example Request Body:**
{
    "data_base64": "eyJkYXRhIjp7ImF3YXJkZWRTdGFnZSI6IkNsb3NlZCIsImRlYWxOYW1lIjoiVGVzdERlYWwiLCJpZCI6IlJhbmRvbUlEIiwic3RhdHVzIjoiQWN0aXZlbHkgQmlkZGluZyJ9LCJkYXRhY29udGVudHR5cGUiOiJhcHBsaWNhdGlvbi9qc29uIiwiaWQiOiIzYTNmNmEwMC0zYmRkLTQ5MjAtOTk5Yy0zNzYzZjVmYmE2N2UiLCJwdWJzdWJuYW1lIjoia3Jlc3QtcHVic3ViIiwic291cmNlIjoia3Jlc3QtcG9sYXJpc3NlcnZpY2UiLCJzcGVjdmVyc2lvbiI6IjEuMCIsInRpbWUiOiIyMDIzLTA1LTE3VDIyOjQ0OjUwWiIsInRvcGljIjoiUG9sYXJpcy5EZWFsQ3JlYXRlZCIsInRyYWNlaWQiOiIwMC00YTg3NDc3MGJlYWJmMTY4N2FjZmMyYTcyOTEyYTlhNi1iYzliMzY0MjIxNGY1OTg5LTAxIiwidHJhY2VwYXJlbnQiOiIwMC00YTg3NDc3MGJlYWJmMTY4N2FjZmMyYTcyOTEyYTlhNi1iYzliMzY0MjIxNGY1OTg5LTAxIiwidHJhY2VzdGF0ZSI6IiIsInR5cGUiOiJjb20uZGFwci5ldmVudC5zZW50In0=",
    "datacontenttype": "application/octet-stream",
    "id": "4287a39c-1024-4a15-9603-d68105ad992f",
    "pubsubname": "pubsub",
    "source": "Dapr",
    "specversion": "1.0",
    "topic": "Polaris.DealCreated",
    "type": "com.dapr.event.sent"
}

**Decoded data_base64 contents **
{
    "data": {
        "awardedStage": "Closed",
        "dealName": "TestDeal",
        "id": "RandomID",
        "status": "Actively Bidding"
    },
    "datacontenttype": "application/json",
    "id": "3a3f6a00-3bdd-4920-999c-3763f5fba67e",
    "pubsubname": "pubsub",
    "source": "polarisservice",
    "specversion": "1.0",
    "time": "2023-05-17T22:44:50Z",
    "topic": "Polaris.DealCreated",
    "traceid": "00-4a874770beabf1687acfc2a72912a9a6-bc9b3642214f5989-01",
    "traceparent": "00-4a874770beabf1687acfc2a72912a9a6-bc9b3642214f5989-01",
    "tracestate": "",
    "type": "com.dapr.event.sent"
}

### DIFF
--- a/docs/en/Dapr/Index.md
+++ b/docs/en/Dapr/Index.md
@@ -209,7 +209,7 @@ ABP provides the following endpoints to receive events from Dapr:
 * `dapr/subscribe`: Dapr uses this endpoint to get a list of subscriptions from the application. ABP automatically returns all the subscriptions for your distributed event handler classes and custom controller actions with the `Topic` attribute.
 * `api/abp/dapr/event`: The unified endpoint to receive all the events from Dapr. ABP dispatches the events to your event handlers based on the topic name.
 
-> **Since ABP will call `MapSubscribeHandler` internally, you should not manually call it anymore.** You can use the `app.UseCloudEvents()` middleware in your ASP.NET Core pipeline if you want to support the [CloudEvents](https://cloudevents.io/) standard.
+> **Since ABP will call `MapSubscribeHandler` internally, you should not manually call it anymore.** You must also add the `app.UseCloudEvents()` middleware in your ASP.NET Core pipeline if you plan to support receiving [Raw pub/sub messages](https://docs.dapr.io/developing-applications/building-blocks/pubsub/pubsub-raw/) which can be configured using the AbpSubscribeOptions class.
 
 ### Usage
 
--- a/framework/src/Volo.Abp.AspNetCore.Mvc.Dapr.EventBus/Volo/Abp/AspNetCore/Mvc/Dapr/EventBus/AbpAspNetCoreMvcDaprEventBusModule.cs
+++ b/framework/src/Volo.Abp.AspNetCore.Mvc.Dapr.EventBus/Volo/Abp/AspNetCore/Mvc/Dapr/EventBus/AbpAspNetCoreMvcDaprEventBusModule.cs
@@ -52,7 +52,7 @@ public class AbpAspNetCoreMvcDaprEventBusModule : AbpModule
                                 Metadata = new AbpMetadata
                                 {
                                     {
-                                        AbpMetadata.RawPayload, "true"
+                                        AbpMetadata.RawPayload, subscribeOptions.EnableRawPayload.ToString().ToLower()
                                     }
                                 }
                             };
--- a/framework/src/Volo.Abp.AspNetCore.Mvc.Dapr.EventBus/Volo/Abp/AspNetCore/Mvc/Dapr/EventBus/Controllers/AbpAspNetCoreMvcDaprEventsController.cs
+++ b/framework/src/Volo.Abp.AspNetCore.Mvc.Dapr.EventBus/Volo/Abp/AspNetCore/Mvc/Dapr/EventBus/Controllers/AbpAspNetCoreMvcDaprEventsController.cs
@@ -21,6 +21,9 @@ public class AbpAspNetCoreMvcDaprEventsController : AbpController
         var daprSerializer = HttpContext.RequestServices.GetRequiredService<IDaprSerializer>();
         var body = (await JsonDocument.ParseAsync(HttpContext.Request.Body));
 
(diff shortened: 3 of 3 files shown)


whytem:
I'm looking into this more and I now think it may be a bug somewhere in the Event subscription process. i.e. it's unclear why the event is being delivered by the Dapr sidecar to the Abp application wrapped in a binary-encoded envelope as this does not appear to be standard behavior.

Subscribing to the same event in a separate test AspNet Core app (i.e. without using the Abp library) results in the event being delivered by the Dapr sidecar as expected directly in the JSON format. So that eliminates any nuances of the event itself and the specific PubSub implementation (redis in this case).

I'll see if I can figure out why but this fix likely is not required if we can eliminate what's causing Dapr to wrap the event in an additional binary-encoded envelope before delivery to the Abp event handler.

whytem:
The offending code appears to be the following from the ConfigureServices routine in AbpAspNetCoreMvcDaprEventBusModule.cs. This effectively results in every endpoint be configured to support receiving raw (instead of CloudEvent compliant JSON) event payloads. As a result, the Dapr sidecar encapsulates the event data received in a binary-encoded envelope as it cannot assume anything about the structure of the event body.

                            var subscription = new AbpSubscription
                            {
                                PubsubName = daprEventBusOptions.PubSubName,
                                Topic = eventName,
                                Route = AbpAspNetCoreMvcDaprPubSubConsts.DaprEventCallbackUrl,
                                Metadata = new AbpMetadata
                                {
                                    {
                                        **AbpMetadata.RawPayload, "true"**
                                    }
                                }
                            };
                            subscriptions.Add(subscription);

I believe this should default to the value in AbpSubscribeOptions.EnableRawPayload (i.e. false) rather than a hardcoded "true' value.

The changes in this pull request happen to work as the event payload in this case is a JSON-encoded cloud event but this should not be assumed. If support for Raw events is intended then the data should be extracted from the data_base64 field and passed to the registered event handler as a byte stream without performing any serialization or making any assumptions regarding the contents.

Let me know if I should delete this PR and log the above as an issue instead. Thanks

maliming:
hi

How can I reproduce this by https://github.com/abpframework/abp-samples/tree/master/Dapr/PubSub

whytem:
Hi maliming,

Thanks for reviewing. You can reproduce the same behavior I was experiencing in the sample project by removing the CloudEvents middleware which I hadn't noticed until reviewing the sample project (i.e. app.UseCloudEvents() in Program.cs:55 of the DaprSubscribe project). 

This middleware has the same effect as the code in the PR whereby it decapsulates any binary-encoded events before they reach the controller. As a result, adding documentation outlining the need for the middleware is likely a more robust solution than this PR if support for raw events by default is required.

However, I still think it makes more sense for support for raw events to be opt-in rather than configured by default for all events in the Abp module. Raw events are really only intended for a narrow use case where it's required to support receiving events published by a legacy (i.e. non-Dapr) system which can't be modified to adhere to CloudEvent spec. If the events are not subscribed to with the raw events flag then the Dapr sidecar will deliver them directly in the original JSON and the middleware (or suggested code in the PR) is not needed. This is also much more efficient and developer-friendly.

This change can be made by eliminating the addition on the "rawPayload" = "true" metadata in the ConfigureServices routine of the AbpAspNetCoreMvcDaprEventBusModule. Support for raw events can then be opted-in by adding Configure<AbpSubscriptionOptions>(o => o.EnableRawPayload = true) and the CloudEvents middleware for the limited set of users who might require it.

Thanks

PR #16617, opened 2023-05-22
Support to remove and replace controllers.

### LINKED ISSUE #16593: Replace default AbpController/Service DTOs instead of extending
Hi,
I want to customize `ProfileAppService` and replace the default DTOs of my own (not extending, because I am currently implementing a module and want to use my own profile model and split them into another table and keep the default `AbpUsers` table clean).

I have read #3132 and related issues but have not found any clues that can help me replace the DTOs. 
The only guide on DTOs is Extending Entities on the docs. So the DTOs can only be extended and cannot be replaced?

I've tried disable the `Volo.Abp.Account.ProfileController` in `HttpApi` configuration using `IgnoredControllersOnModelExclusion` but it still works. Removing `[DependsOn]` `AbpAccountApplicationModule` and `AbpAccountApplicationContractsModule` doensn't work either.

Now I want to:
- Disable both  `Volo.Abp.Account.ProfileController` and `Volo.Abp.Account.ProfileAppService` so I can keep using the endpoint `api/account/my-profile`.
Or
- Be able to replace the default DTOs of the above ABP controller and service (which I will have to reimplement because I don't want to use `IProfileAppService` interface with the default DTOs).

What should I do to achieve the above?

maliming:
Resolve #16593

**The controller has the same router but with different input and output parameters.**

### DIFF
--- a/framework/src/Volo.Abp.AspNetCore.Mvc/Volo/Abp/AspNetCore/Mvc/Conventions/AbpServiceConvention.cs
+++ b/framework/src/Volo.Abp.AspNetCore.Mvc/Volo/Abp/AspNetCore/Mvc/Conventions/AbpServiceConvention.cs
@@ -11,6 +11,7 @@ using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Logging.Abstractions;
 using Microsoft.Extensions.Options;
 using Volo.Abp.Application.Services;
+using Volo.Abp.AspNetCore.Controllers;
 using Volo.Abp.DependencyInjection;
 using Volo.Abp.GlobalFeatures;
 using Volo.Abp.Http;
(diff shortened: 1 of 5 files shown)


framework/src/Volo.Abp.AspNetCore.Mvc/Volo/Abp/AspNetCore/Mvc/Conventions/AbpServiceConvention.cs
```diff
@@ -99,9 +130,12 @@ protected virtual void RemoveDuplicateControllers(ApplicationModel application)
...
                     .ToArray();
 
-                controllerModelsToRemove.AddRange(exposedControllerModels);
-                Logger.LogInformation($"Removing the controller{(exposedControllerModels.Length > 1 ? "s" : "")} {exposeServicesAttr.ServiceTypes.Select(c => c.AssemblyQualifiedName).JoinAsString(", ")} from the application model since {(exposedControllerModels.Length > 1 ? "they are" : "it is")} replaced by the controller: {controllerModel.ControllerType.AssemblyQualifiedName}");
-                continue;
+                if (exposedControllerModels.Any())
+                {
+                    controllerModelsToRemove.AddRange(exposedControllerModels);
+                    Logger.LogInformation($"Removing the controller{(exposedControllerModels.Length > 1 ? "s" : "")} {exposeServicesAttr.ServiceTypes.Select(c => c.AssemblyQualifiedName).JoinAsString(", ")} from the application model since {(exposedControllerModels.Length > 1 ? "they are" : "it is")} replaced by the controller: {controllerModel.ControllerType.AssemblyQualifiedName}");
+                    continue;
```
rqx110: move `continue;` to outside of `if (exposedControllerModels.Any())` block?

hikalkan:
Please add a section to the documentation. Otherwise, people can't know how to use this feature.

You can add a last section here: https://docs.abp.io/en/abp/7.2/API/Auto-API-Controllers
Make it a separate PR and add me & @braim23 as reviewer.

I opened an issue for this: https://github.com/abpframework/abp/issues/16706

hikalkan (approved):

PR #16681, opened 2023-05-26
fix(): ensure new password is different from current password

tntwist:
### Description

This pr ensures that the new password always divers from the current password when changing passwords.
Currently its only validated in the mvc js. With this change its always validated.

### Checklist

- [x ] I fully tested it as developer / designer and created unit / integration tests

### How to test it?

There are three ways to test it.
1. Run the added unit test for profile app service.
2. Start HttpApi, authorize and make a request against the ChangePassword-Endpoint
3. Create demo project and change password in frontend.

Each method provides an error message saying "The new password should be different from the old password".


maliming (approved):

maliming:
I will update the pro side.

PR #16686, opened 2023-05-27
Fix file issue in dynamic form

salihozkara:
### Description
When using a file in dynamic form, its properties were input fields instead of input in the file type. I fixed this.

Example:
```cs
public class MyModal{
    public IFormFile File { get; set; }
    public IFormFileCollection Files { get; set; }
    public List<IFormFile> Files2 { get; set; }
    
    public List<FormFile> Files3 { get; set; }
    
    public FormFile File2 { get; set; }
}
```

Old : 
image

New : 
image

### Checklist


maliming (approved):

gdlcf88:
Nice job :)

PR #16711, opened 2023-05-30
Fix nightly build

mahmut-gundogdu:
When I wrote the code, God and I know. but now. Only god knows. Please review carefully. My brain is tired.


npm/ng-packs/packages/schematics/package.json
```diff
@@ -9,20 +9,20 @@
   "license": "MIT",
   "schematics": "./collection.json",
   "dependencies": {
-    "@angular-devkit/core": "~15.0.1",
-    "@angular-devkit/schematics": "~15.0.1",
-    "@angular/cli": "~15.0.1",
+    "@angular-devkit/core": "16.0.1",
+    "@angular-devkit/schematics": "16.0.1",
+    "@angular/cli": "16.0.1",
```
masum-ulu, line 14: *devkits* are fine but, @angular/cli package may can effect npm install after create project with CLI like this line https://github.com/abpframework/abp/blob/cce33573251553f817e7e88cafd5bd7f8285caf1/npm/ng-packs/packages/account-core/package.json#L13

Maybe not now but for the future versions it should not be conflict

npm/publish-ng.ps1
```diff
@@ -29,17 +28,16 @@ if ($IsPrerelease) {
 }
 
 $commands = (
-  $UpdateNgPacksCommand,
   "cd ng-packs\scripts",
   "yarn install",
  # $UpdateNgPacksUpdateVersionCommand,
```
masum-ulu: There is no variable for this command should we keep that ?

masum-ulu (changes requested):

masum-ulu (approved):

Issue #16724, opened 2023-05-31
Not support event choreography


MapleWithoutWords:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe the problem.

Because in certain special scenarios, I want to control the order of event processing. 

### Describe the solution you'd like

I want it to execute in the same order as shown in the figure below. And event handlers can be added at will

![image](https://github.com/abpframework/abp/assets/38368335/f1492bd6-f9ef-4897-9603-fe15ff5874a2)

hikalkan:
Distributed or Local events?

MapleWithoutWords:
> Distributed or Local events?

Yes, I think we can support that

MapleWithoutWords:
Maybe I can try to add this feature. @hikalkan

MapleWithoutWords:
I think distributed and local are the same, I just control the order in which it is executed

> Distributed or Local events?

maliming:
hi @MapleWithoutWords 

https://github.com/abpframework/abp/issues/16361

MapleWithoutWords:
@maliming I think local event support choreography help with program extension. It's a very good feature. 
Example:
 I have a feature that has a lot of logic, and one day I want to add caching to it, and if there's event orchestration, I can easily add functionality to it. I just need to add two event handlers, one to judge the cache and return on the first subscription, and one to add the cache at the end.

maliming:
https://github.com/abpframework/abp/pull/16544

MapleWithoutWords:
thank you

> #16544

PR #16728, opened 2023-06-01
feat:Add event choreography.And fix:issues-#16724

MapleWithoutWords:


berkansasmaz:
This PR seems invalid. See: https://github.com/abpframework/abp/issues/16724#issuecomment-1571248455

Still, thank you for your contribution 🎉

PR #16737, opened 2023-06-02
Fix RouteValues must not be null error

Sjhzjxc:
### Description

When I use `abp-route-*` in `abp-dynamic-form`,like this.
image
I got an error.
image
The solution comes from 
[aspnetcore/FormTagHelper.cs at v7.0.5 · dotnet/aspnetcore · GitHub](https://github.com/dotnet/aspnetcore/blob/v7.0.5/src/Mvc/Mvc.TagHelpers/src/FormTagHelper.cs#L122-L137)
[aspnetcore/FormActionTagHelper.cs at v7.0.5 · dotnet/aspnetcore · GitHub](https://github.com/dotnet/aspnetcore/blob/v7.0.5/src/Mvc/Mvc.TagHelpers/src/FormActionTagHelper.cs#L153-L168)

### Checklist

### How to test it?

use `abp-route-*` in `abp-dynamic-form` without `asp-all-route-data`.


EngincanV (approved):

EngincanV:
Thanks for your contribution @Sjhzjxc 🎉

PR #16795, opened 2023-06-09
Add `CorrelationId` to distributed events.

### LINKED ISSUE #16773: Add CorrelationId to distributed events
The purpose is to attach current correlation id to distributed events, so we can relate events with HTTP requests and can trace all the related activities.

- [x]  Attach the current correlation id (get using `ICorrelationIdProvider.Get()`) to distributed events as an additional property. This can be added to header of the message with provider's API. For example, RabbitMQ's `BasicPublish` method gets `IBasicProperties` which contains `Headers` while sending/receiving events. We can define a header and transfer the value with that header. Header name can be `X-Correlation-Id` as we already use it in HTTP headers.
- [x]  Add correlation id to `IncomingEventInfo` and `OutgoingEventInfo` as an extra property (to the `ExtraProperties` collection) with the same header name (`X-Correlation-Id`).
- [x]  Change current ICorrelationIdProvider implementations, so we can use the ambient context pattern (implemented with with `AsyncLocal<T>` internally) to set an arbitrary correlation id value with a using statement, just like we did at `ICurrentPrincipalAccessor.Change`. In this way, we can temporary override the current correlation id (it sets back to previous value after the using statement).
- [x]  We will use the `ICorrelationIdProvider.Change` method while executing an incoming event. In this way, if we get the correlation id value inside our distributed event handler, we can get the value that is attached to the event. Also, if we make an HTTP call to another microservice, it will use the same correlation id value that is originally attached to the event by the sender.
- [x]  We should also be able to get the correlation id value in global `DistributedEventSent` and `DistributedEventReceived` events. It should already work if we use `ICorrelationIdProvider.Change` before executing that local event (in the event handler, we can get from `ICorrelationIdProvider.Get`, no need to add value to the `DistributedEventSent` and `DistributedEventReceived` classes.

maliming:
Resolves #16773

This is breaking changes because the `ICorrelationIdProvider` will return **null** by default, but we will try to get and set it on `AbpCorrelationIdMiddleware`.


framework/src/Volo.Abp.AspNetCore/Volo/Abp/AspNetCore/Tracing/AbpCorrelationIdMiddleware.cs
```diff
@@ -20,16 +21,37 @@ public class AbpCorrelationIdMiddleware : IMiddleware, ITransientDependency
...
+    }
+
+    protected virtual string GetCorrelationIdFromRequest(HttpContext context)
+    {
+        string correlationId = context.Request.Headers[_options.HttpHeaderName];
+        if (correlationId.IsNullOrEmpty())
         {
-            CheckAndSetCorrelationIdOnResponse(context, _options, correlationId);
+            lock (context.Request.Headers)
+            {
```
hikalkan: Shouldn't we try to get it again before the next `if`, like `correlationId = context.Request.Headers[_options.HttpHeaderName];`?

hikalkan: However, I think `lock` is completely unnecessary. Because, an HTTP request is not multi-thread and there is no race condition.

hikalkan (approved):

hikalkan (approved):

hikalkan:
Thanks.

PR #16875, opened 2023-06-19
[WIP] Bulk entity events

### LINKED ISSUE #16107: Introduce bulk entity events

consider this scenario:
you have a separate algolia search server, and you want to sync data between your database and the algolia index.

the current solution is to implement `ILocalEventHandler<EntityChangedEventData<TEntity>>`.
this works when only a single entity is getting changed, but not for `InsertMany`,`UpdateMany`,`DeleteMany` since they will introduce the N+1 problem

I suggest that new events to get introduced:
- `IBulkEntityChangeEventHelper`
  ```csharp
    public interface IBulkEntityChangeEventHelper
    {
        void PublishEntityCreatedEvent(IEnumerable<object> entity);

        void PublishEntityUpdatedEvent(IEnumerable<object> entity);

        void PublishEntityDeletedEvent(IEnumerable<object> entity);
    }
  ```
- `BulkEntityEventData` -> contains a list of entities
- `BulkEntityChangedEventData` -> fired when many entities are changed
- `BulkEntityCreatedEventData` -> fired when many entities are created
- `BulkEntityUpdatedEventData` -> fired when many entities are updated
- `BulkEntityDeletedEventData` -> fired when many entities are deleted

and then we can add a configuration for users to override the behavior of `InsertMany`, `UpdateMany`,`DeleteMany` to control whether the old events `EntityChangedEventData` are fired alongside the new bulk ones or not

### Additional context

_No response_

ahmednfwela:
### Description

This is still a work in progress, early reviews are welcome.

Resolves #16107

This introduces new services and models/ETOs to handle bulk entity events

### Checklist

### How to test it?


framework/src/Volo.Abp.EntityFrameworkCore/Volo/Abp/EntityFrameworkCore/AbpDbContext.cs
```diff
@@ -298,6 +314,7 @@ private void PublishEventsForTrackedEntity(EntityEntry entry)
             case EntityState.Added:
                 ApplyAbpConceptsForAddedEntity(entry);
                 EntityChangeEventHelper.PublishEntityCreatedEvent(entry.Entity);
+                BulkEntityChangeEventHelper.PublishBulkEntityCreatedEvent(new List<object> { entry.Entity });
```
ahmednfwela, line 317: Maybe move this to `SaveChangesAsync` to be able to handle bulk events properly ?

PR #16922, opened 2023-06-21
fix csrf token sent in swagger

ahmednfwela:
### Description

Fixes https://support.abp.io/QA/Questions/3840/Keycloak-Swagger-Client-on-KeycloakDemo

Intercept whenever a request is made to `.well-known/openid-configuration` and get the `issuer` from there to exclude from adding verification token header.

### Checklist

### How to test it?

Reproduction steps in the linked issue.


ahmednfwela:
Note: I haven't tested this implementation when both the auth server and the swagger server are on the same url

PR #16979, opened 2023-06-26
Dynamic Setting Store implementation.

### LINKED ISSUE #16945: Dynamic Setting Store implementation
Just like we did for permissions and features, we also need to introduce dynamic setting store. The purpose is to collect all setting definitions in a single database. In this way, we can get all setting definitions from a single service in a microservice solution.

maliming:
### Description

Resolves #16945

**Breaking change: We need to update migrations for the setting management module.**


hikalkan (approved):

hikalkan:
Thanks 👍

PR #16987, opened 2023-06-28
Update UnitOfWorkExtensions.cs

helloxubo:
The condition is wrong.


maliming:
hi @helloxubo 

I moved this code to `7.0`. 

https://github.com/abpframework/abp/pull/16988

Thanks.

Issue #1893, opened 2019-10-11
ExternalLogin throw System.ArgumentNullException: Value cannot be null. (Parameter 'userName')


vincywindy:
When I try to use github login,the callback throw this exception:
```c#
System.ArgumentNullException: Value cannot be null. (Parameter 'userName')
   at Volo.Abp.Check.NotNull[T](T value, String parameterName)
   at Volo.Abp.Identity.IdentityUser..ctor(Guid id, String userName, String email, Nullable`1 tenantId)
   at Volo.Abp.Account.Web.Pages.Account.LoginModel.CreateExternalUserAsync(ExternalLoginInfo info)
   at Volo.Abp.Uow.UnitOfWorkInterceptor.InterceptAsync(IAbpMethodInvocation invocation)
   at Volo.Abp.Castle.DynamicProxy.CastleAbpInterceptorAdapter`1.ExecuteWithReturnValueAsync[T](IInvocation invocation, IInvocationProceedInfo proceedInfo)
   at Volo.Abp.Account.Web.Pages.Account.LoginModel.OnGetExternalLoginCallbackAsync(String returnUrl, String returnUrlHash, String remoteError)
   at Volo.Abp.Uow.UnitOfWorkInterceptor.InterceptAsync(IAbpMethodInvocation invocation)
   at Volo.Abp.Castle.DynamicProxy.CastleAbpInterceptorAdapter`1.ExecuteWithReturnValueAsync[T](IInvocation invocation, IInvocationProceedInfo proceedInfo)
   at Microsoft.AspNetCore.Mvc.RazorPages.Infrastructure.ExecutorFactory.GenericTaskHandlerMethod.Convert[T](Object taskAsObject)
   at Microsoft.AspNetCore.Mvc.RazorPages.Infrastructure.ExecutorFactory.GenericTaskHandlerMethod.Execute(Object receiver, Object[] arguments)
   at Microsoft.AspNetCore.Mvc.RazorPages.Infrastructure.PageActionInvoker.InvokeHandlerMethodAsync()
   at Microsoft.AspNetCore.Mvc.RazorPages.Infrastructure.PageActionInvoker.InvokeNextPageFilterAsync()
   at Microsoft.AspNetCore.Mvc.RazorPages.Infrastructure.PageActionInvoker.Rethrow(PageHandlerExecutedContext context)
[...]

PR #16999, opened 2023-07-01
Handle the external login info from GitHub

gdlcf88:


maliming:
You can configure your external provider.

https://github.com/abpframework/abp/issues/1893#issuecomment-541092999

PR #17039, opened 2023-07-05
Support module class type at AbpAspNetCoreIntegratedTestBase

hikalkan:
`AbpAspNetCoreIntegratedTestBase` was getting a generic type parameter, which should be a Startup class type.
With this PR, we will be able to pass an ABP module class type to that generic type parameter. It internally uses a startup class that simplifies the test application project. It also initializes ABP asynchronously (by the help of the AsyncHelper class).

It is backward compatible.


hikalkan:
@maliming can you please make the necessary change in the abp commercial startup template?

PR #17045, opened 2023-07-05
Move invoke code after setting OpenIdConnectOptions for override flexibility

huseyinalan:
### Description

Moved the invoke code after setting the Abp options to allow developers to override them.

While inspecting the codes, I noticed that there is an AccessDeniedPath property. After inspecting the Abp code, it seems to be doing the same thing, so I refactored it
image

### DIFF
--- a/framework/src/Volo.Abp.AspNetCore.Authentication.OpenIdConnect/Microsoft/Extensions/DependencyInjection/AbpOpenIdConnectExtensions.cs
+++ b/framework/src/Volo.Abp.AspNetCore.Authentication.OpenIdConnect/Microsoft/Extensions/DependencyInjection/AbpOpenIdConnectExtensions.cs
@@ -28,8 +28,6 @@ public static class AbpOpenIdConnectExtensions
         {
             options.ClaimActions.MapAbpClaimTypes();
 
-            configureOptions?.Invoke(options);
-
             options.Events ??= new OpenIdConnectEvents();
             var authorizationCodeReceived = options.Events.OnAuthorizationCodeReceived ?? (_ => Task.CompletedTask);
 
@@ -63,6 +61,8 @@ public static class AbpOpenIdConnectExtensions
                     logger?.LogException(ex);
                 }
             };
+
+            configureOptions?.Invoke(options);
         });
     }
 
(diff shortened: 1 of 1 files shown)


framework/src/Volo.Abp.AspNetCore.Authentication.OpenIdConnect/Microsoft/Extensions/DependencyInjection/AbpOpenIdConnectExtensions.cs
```diff
@@ -39,16 +37,7 @@ public static AuthenticationBuilder AddAbpOpenIdConnect(this AuthenticationBuild
...
-            options.Events.OnRemoteFailure = remoteFailureContext =>
-            {
-                if (remoteFailureContext.Failure is OpenIdConnectProtocolException &&
-                    remoteFailureContext.Failure.Message.Contains("access_denied"))
-                {
-                    remoteFailureContext.HandleResponse();
-                    remoteFailureContext.Response.Redirect($"{remoteFailureContext.Request.PathBase}/");
-                }
-                return Task.CompletedTask;
-            };
```
maliming, line 51: hi

Why you remove these lines?

huseyinalan: Hi,
I replaced with this one they do same thing.

```c#
options.AccessDeniedPath = "/";
``` 
You can check from this
[RemoteAuthenticationHandler.cs](https://github.com/dotnet/aspnetcore/blob/main/src/Security/Authentication/Core/src/RemoteAuthenticationHandler.cs)

maliming (approved):

framework/src/Volo.Abp.AspNetCore.Authentication.OpenIdConnect/Microsoft/Extensions/DependencyInjection/AbpOpenIdConnectExtensions.cs
maliming: Thanks for your explain.

PR #17049, opened 2023-07-06
Add extra properties to User Edit Modal

### LINKED ISSUE #16909: Add extra properties to User Edit and Create Modal
image

salihozkara:
### Description

Resolves #16909 

image

### Checklist


gizemmutukurt:
-  I just created the user and only entered the password during creation. But in terms of UI, I understand that this field should change when I change the password. Also, the time field appears in an independent timezone.
image


- It seems not right that these fields are filled when I do not update.
image

- I made a few incorrect entries, but the "Access failed count" value did not change.
image

ebicoglu:
Failed access count
Password update time
Lockout end time

salihozkara:
image

gizemmutukurt (approved):

ebicoglu (approved):

PR #17067, opened 2023-07-07
Register `ajax.reloadEx()` method to Datatables.

### LINKED ISSUE #17029: Consider hold current page when reloading data on datatable.
https://support.abp.io/QA/Questions/5355/Paging-issue-after-update-any-record

https://datatables.net/reference/api/ajax.reload()

> Reset (default action or true) or hold the current paging position (false). A full re-sort and re-filter is performed when this method is called, which is why the pagination reset is the default action.

`ajax.reload(null, false);` will keep the hold page.

----

Angular has the same behavior.

maliming:
Resolve #17029


realLiangshiwei (approved):

ebicoglu (approved):

ebicoglu:
@gizemmutukurt will test the functionality.
@maliming can you write the steps on how to test it

maliming:
@gizemmutukurt 

You can create more than ten roles and set the page size as 10 and jump to the second page. 

1. create a new role again on the second page. The page should be kept on the second page after creation.
2. delete some roles. The page will go to the first page if the count of roles is not enough on the second page.

gizemmutukurt (approved):

PR #17072, opened 2023-07-08
remove unnecessary name

Sjhzjxc:
### Description

ITemplateDefinitionContext\'s GetAll method does not require a name parameter
​
### Checklist


hikalkan:
This PR is not being compiled.
@maliming can you check this and see why there is such a parameter? It looks strange to me. Also, remember that we don't want to make breaking change.

maliming (approved):

PR #17110, opened 2023-07-13
CmsKit: Add most popular tags

salihozkara:
### Description

Added a new component that can show popular tags for an entity.

image

### Checklist

- [x] I documented it


gizemmutukurt:
There is a tag2 tag, which is used more than the mikroservice tag. but microservices appear to take precedence.

image
image

gizemmutukurt (approved):

ebicoglu (approved):

PR #17196, opened 2023-07-26
add schematics build on build:all

mahmut-gundogdu:
it cherry-picked from  https://github.com/abpframework/abp/pull/17179/


masum-ulu (approved):

masum-ulu:
MyGet was down that's why we removed, It'll discard

PR #17212, opened 2023-07-28
Add `AbpIdentityUserValidator`.

maliming:
This can prevent a new user's name use another existing user's email.

https://support.abp.io/QA/Questions/5482/Should-not-be-able-to-create-a-user-with-a-username-equal-to-the-email-address-of-another-user


gterdem (commented):
I strongly think “ YouCanNotUseAnotherUsersEmailAsYourUsername” is not a good localization and response. It is too much information and it is a security problem. And this way, lots of information can be gathered with brute force.
Something like “Invalid username” or just an error code to check the application logs should be enough, the details can be logged at the webserver.

gterdem (approved):
For security, this error and code should be used in multiple occasions. Ex, when user creation fails because of any other identity option validation.

maliming:
We are using the Identity's error messages for other occasions.

https://github.com/dotnet/aspnetcore/blob/v7.0.0/src/Identity/Extensions.Core/src/Resources.resx#L165

PR #17246, opened 2023-08-01
decreasing margin for form prop

Sinan997:
### Description

Resolves [#14794](https://github.com/volosoft/volo/issues/14794)

decreasing gap between checkbox

Screenshot_2
Screenshot_3

## But decreasing margin is affecting other form props

for example, previous ***new claim type*** form
Screenshot_5

after decreasing margin it will look like this
image

### DIFF
--- a/npm/ng-packs/packages/theme-shared/extensions/src/lib/components/extensible-form/extensible-form-prop.component.html
+++ b/npm/ng-packs/packages/theme-shared/extensions/src/lib/components/extensible-form/extensible-form-prop.component.html
@@ -7,7 +7,7 @@
     </ng-container>
   </ng-template>
 
-  <div [ngClass]="containerClassName" class="mb-3">
+  <div [ngClass]="containerClassName" class="mb-2">
     <ng-template ngSwitchCase="input">
       <ng-template [ngTemplateOutlet]="label"></ng-template>
       <input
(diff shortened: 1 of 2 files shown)


npm/ng-packs/packages/theme-shared/extensions/src/lib/components/extensible-form/extensible-form-prop.component.html
```diff
@@ -7,7 +7,7 @@
     </ng-container>
   </ng-template>
 
-  <div [ngClass]="containerClassName" class="mb-3">
+  <div [ngClass]="containerClassName" class="mb-2">
```
mahmut-gundogdu, line 10: why we have two times

mahmut-gundogdu (changes requested):

mahmut-gundogdu:
@armagan should review it

npm/ng-packs/packages/theme-shared/extensions/src/lib/components/extensible-form/extensible-form-prop.component.html
Sinan997: default value
image

if FormProp has className attribute
image

then it can be used

armgnunlu (approved):

masum-ulu (approved):

Sinan997:
this pr was branched from dev, it needs to be branched from rel-7.3 so moved to #17265

PR #17260, opened 2023-08-03
Add `InferPathParameterBindingSources` method to `AbpServiceConvention`.

maliming:
ASP NET Core will not set `{assignToId?}` binding source as `Path` by default. 
We have to specify `FromRoute` for `nullable assignToId` manually.

This pr automatically sets the binding source to `Path` for parameters if the controller inherits `AbpControllerBase` and `IApplicationService`.  

This is mainly designed for the `HttpApi` layer of the module.

```cs
[Route("api/identity/roles")]
public class TestHttpApiController : AbpControllerBase, IApplicationService
{
    [HttpDelete]
    [Route("{id}/assign/{assignToId?}")]
    public virtual Task DeleteAsync(Guid id, Guid? assignToId)
    {
        return Task.CompletedTask;
    }
}
```


maliming:
I think we can move the code to the `AbpServiceConvention` class.

PR #17265, opened 2023-08-03
decreasing margin for form prop

Sinan997:
### Description

Resolves [#14794](https://github.com/volosoft/volo/issues/14794)

decreasing gap between checkbox

Screenshot_2
Screenshot_3

## But decreasing margin is affecting other form props

for example, previous ***new claim type*** form
Screenshot_5

after decreasing margin it will look like this
image


mahmut-gundogdu:
Armagan should review it.

armgnunlu (approved):

gizemmutukurt (approved):

PR #17272, opened 2023-08-04
Add `UserRolePermissionValueProvider`.

maliming:
We will use `IUserRoleFinder` to get user roles and check the permissions.
In some cases, there is no role info in current claims.

```cs
await _permissionChecker.IsGrantedAsync(targetUserId, "MyPermission")
```


modules/identity/src/Volo.Abp.Identity.Domain/Volo/Abp/Identity/Permissions/UserRolePermissionValueProvider.cs
```diff
@@ -0,0 +1,98 @@
...
+        {
+            return PermissionGrantResult.Undefined;
+        }
+
+        if (!Guid.TryParse(userId, out var userGuidId))
+        {
+            return PermissionGrantResult.Undefined;
+        }
+
+        var userRoles = await UserRoleFinder.GetRoleNamesAsync(userGuidId);
```
hikalkan, line 38: `UserPermissionValueProvider` will return undefined if current user has no permission. Then `UserRolePermissionValueProvider` (this class) will work and try to find roles with a remote API call (for a microservice scenario). A remote API call is not good for permission check that should work fast.
Notice that it work always, not only when we want to check a user's permission with a temporary principle with userid claim. That's unacceptable.

hikalkan: Another problem is that: Now, `Volo.Abp.Identity.Domain` package should be added to every microservice needs to check a user's permission with id.

hikalkan: Another problem: When we create a temporary principle only with userid claim, some custom permission value providers may not work properly (if they are checking another claim). So, creating a temporary principle only with userid claim is a partial workaround solution. True implementation should be creating real principle of the target user. However, it requires much more work to be done and should be well-designed by considering possible use cases.

hikalkan (changes requested):

hikalkan:
My solution offer will be completely different:

Define `IPermissionIntegrationService` to the permission-management module (like `IIdentityUserIntegrationService` we did for the identity module)
In this way, if a module needs to check someone's permission, it can directly depend on the permission management integration service and use it. This is much simple.

PR #17324, opened 2023-08-08
Volo.Abp.Cli: get-source packages

alecsg77:
### Description
I added a new flag option (-p|--package) to `abp get-source` command to download the source code of the NPM/NuGet packages as an alternative to the module source code.

### Checklist

- [x] I tested it as developer
- [x] I documented it

### How to test it?

`abp get-source @abp/ng.theme.basic -p -o ng.theme.basic`


yekalkan:
Thank you for the contribution. However, this feature is not that simple because some internal configurations on our websites are needed for this to work since CLI downloads the packages/modules from our website. We don't plan this feature currently.

PR #17337, opened 2023-08-10
Add `PermissionIntegrationService` to permission-management module.

maliming:


modules/permission-management/src/Volo.Abp.PermissionManagement.Application.Contracts/Volo/Abp/PermissionManagement/Integration/IPermissionIntegrationService.cs
```diff
@@ -0,0 +1,11 @@
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Volo.Abp.Application.Services;
+
+namespace Volo.Abp.PermissionManagement.Integration;
+
+[IntegrationService]
+public interface IPermissionIntegrationService : IApplicationService
+{
+    Task<List<PermissionGrantOutput> > IsGrantedAsync(List<PermissionGrantInput> input);
```
hikalkan:
```suggestion
    Task<List<PermissionGrantOutput>> IsGrantedAsync(List<PermissionGrantInput> input);
```

hikalkan: Instead of directly returning a `List`, return a `ListResultDto`. This will be easier for us to add more properties to the output object without breaking existing clients.

modules/permission-management/src/Volo.Abp.PermissionManagement.Application.Contracts/Volo/Abp/PermissionManagement/Integration/PermissionGrantOutput.cs
```diff
@@ -0,0 +1,11 @@
+using System;
+using System.Collections.Generic;
+
+namespace Volo.Abp.PermissionManagement.Integration;
+
+public class PermissionGrantOutput
+{
+    public Guid UserId { get; set; }
```
hikalkan, line 8: Returning the `UserId` is not necessary since the client already knows it.

modules/permission-management/src/Volo.Abp.PermissionManagement.HttpApi.Client/ClientProxies/Volo/Abp/PermissionManagement/PermissionsClientProxy.Generated.cs
```diff
@@ -1,15 +1,17 @@
...
-using Volo.Abp.Http.Modeling;
 using Volo.Abp.DependencyInjection;
+using Volo.Abp.Http.Client;
 using Volo.Abp.Http.Client.ClientProxying;
+using Volo.Abp.Http.Modeling;
 using Volo.Abp.PermissionManagement;
 
 // ReSharper disable once CheckNamespace
-namespace Volo.Abp.PermissionManagement.ClientProxies;
+namespace Volo.Abp.PermissionManagement;
```
hikalkan, line 14: Isn't that a breaking change? Why we do that if so?

modules/permission-management/src/Volo.Abp.PermissionManagement.HttpApi/Volo/Abp/PermissionManagement/Integration/PermissionIntegrationController.cs
```diff
@@ -0,0 +1,27 @@
...
+public class PermissionIntegrationController: AbpControllerBase, IPermissionIntegrationService
+{
+    protected IPermissionIntegrationService PermissionIntegrationService { get; }
+
+    public PermissionIntegrationController(IPermissionIntegrationService permissionIntegrationService)
+    {
+        PermissionIntegrationService = permissionIntegrationService;
+    }
+
+    [HttpPost]
```
hikalkan: Why post? Shouldn't it be Get? We are not changing anything in the server.

hikalkan (changes requested):

modules/permission-management/src/Volo.Abp.PermissionManagement.HttpApi.Client/ClientProxies/Volo/Abp/PermissionManagement/PermissionsClientProxy.Generated.cs
maliming: We have changed some namespace rules. These proxy classes are old. 
We don't use them(`XXXClientProxy`) directly so changing the namespace won't be a breaking change.

modules/permission-management/src/Volo.Abp.PermissionManagement.Application.Contracts/Volo/Abp/PermissionManagement/Integration/PermissionGrantOutput.cs
maliming: We may check the permissions for multiple users, We have to use `UserId`.

modules/permission-management/src/Volo.Abp.PermissionManagement.HttpApi.Client/ClientProxies/Volo/Abp/PermissionManagement/PermissionsClientProxy.Generated.cs
maliming: We will re-generate proxies of all modules later.

hikalkan (approved):

PR #17380, opened 2023-08-16
Add separate methods to get role and usercount to improve performance

realLiangshiwei:
### Description

Add separate methods to get role and usercount to improve performance

### Checklist


modules/identity/src/Volo.Abp.Identity.EntityFrameworkCore/Volo/Abp/Identity/EntityFrameworkCore/EfCoreIdentityRoleRepository.cs
```diff
@@ -28,6 +28,30 @@ public EfCoreIdentityRoleRepository(IDbContextProvider<IIdentityDbContext> dbCon
             .FirstOrDefaultAsync(r => r.NormalizedName == normalizedRoleName, GetCancellationToken(cancellationToken));
     }
 
+    public async Task<List<IdentityRoleWithUserCount>> GetListWithUserCountAsync(
```
rqx110: make method virtual?

realLiangshiwei: > make method virtual?

Thanks, I will.

maliming (approved):

PR #17421, opened 2023-08-21
Use NoTracking for readonly repositories for EF core.

### LINKED ISSUE #597: Use NoTracking for readonly repositories for EF core

maliming:
### Description

Resolves #597

### Checklist


framework/src/Volo.Abp.Ddd.Domain/Microsoft/Extensions/DependencyInjection/ServiceCollectionRepositoryExtensions.cs
```diff
@@ -80,15 +81,29 @@ public static class ServiceCollectionRepositoryExtensions
...
         Type implementationType,
-        bool replaceExisting)
+        bool replaceExisting,
+        bool isReadOnlyRepository = false)
     {
+        var descriptor = ServiceDescriptor.Transient(serviceType, implementationType);
+
+        if (isReadOnlyRepository)
+        {
+            services.AddTransient(implementationType);
```
hikalkan:
```suggestion
            services.TryAddTransient(implementationType);
```

Try would be better here.

framework/src/Volo.Abp.Ddd.Domain/Microsoft/Extensions/DependencyInjection/ServiceCollectionRepositoryExtensions.cs
```diff
@@ -80,15 +81,29 @@ public static class ServiceCollectionRepositoryExtensions
         IServiceCollection services,
         Type serviceType,
         Type implementationType,
-        bool replaceExisting)
+        bool replaceExisting,
+        bool isReadOnlyRepository = false)
     {
+        var descriptor = ServiceDescriptor.Transient(serviceType, implementationType);
```
hikalkan: Instead of unnecessarily create a descriptor in case of readonly, move this code to else of the `isReadOnlyRepository` control. Here, just declare without assigning a value.

framework/src/Volo.Abp.Ddd.Domain/Volo/Abp/Domain/Repositories/BasicRepositoryBase.cs
```diff
@@ -34,6 +34,9 @@ public abstract class BasicRepositoryBase<TEntity> :
 
     public ICancellationTokenProvider CancellationTokenProvider => LazyServiceProvider.LazyGetService<ICancellationTokenProvider>(NullCancellationTokenProvider.Instance);
 
+
+    public bool IsReadOnly { get; set; } = false;
```
hikalkan: You've used `ObjectHelper.TrySetProperty` above. So, you can define the setter as `protected` here to prevent changing the value out of the repository class.

```suggestion
    public bool IsReadOnly { get; potected set; }
```

It is already false by default, no need to set to false explicitly.

framework/src/Volo.Abp.EntityFrameworkCore/Volo/Abp/Domain/Repositories/EntityFrameworkCore/EfCoreRepository.cs
```diff
@@ -37,30 +35,50 @@ async Task<DbContext> IEfCoreRepository<TEntity>.GetDbContextAsync()
...
             }
         }
+        else
+        {
+            dbContext = _dbContextProvider.GetDbContext();
+        }
+
+        if (IsReadOnly)
+        {
+            dbContext.ChangeTracker.QueryTrackingBehavior = QueryTrackingBehavior.NoTracking;
```
hikalkan: Different repositories share the same `DbContext` object in a unit of work. Setting `QueryTrackingBehavior` will effect all other repositories. So, do not make that.

framework/src/Volo.Abp.EntityFrameworkCore/Volo/Abp/Domain/Repositories/EntityFrameworkCore/EfCoreRepository.cs
```diff
@@ -37,30 +35,50 @@ async Task<DbContext> IEfCoreRepository<TEntity>.GetDbContextAsync()
...
             }
         }
+        else
+        {
+            dbContext = await _dbContextProvider.GetDbContextAsync();
+        }
+
+        if (IsReadOnly)
+        {
+            dbContext.ChangeTracker.QueryTrackingBehavior = QueryTrackingBehavior.NoTracking;
```
hikalkan: Different repositories share the same DbContext object in a unit of work. Setting QueryTrackingBehavior will effect all other repositories. So, do not make that.

hikalkan (changes requested):

hikalkan (approved):

hikalkan:
Add one more test please:

Begin a UOW, use a readonly repository and a regular repository, get different entitities from both of the repositories, change some values in these two entities in the same UOW, then complete UOW block. After UOW block, check if the entities were updated. The one you got from readonly repository **should not** be updated in the database (because it is not being tracked), but the one you got from the normal repository should be updated in the database.

PR #17434, opened 2023-08-21
Swagger with OIDC doesn't need full path

### LINKED ISSUE #17433: Swagger with OIDC doesn't need full path
The discovery endpoint is default for OIDC: `.well-known/openid-configuration`.

We can define this automatically instead of expecting users to provide it for different discovery endpoints.

gterdem:
### Description

Resolves #17433

### Checklist

### How to test it?

Sample project: https://github.com/abpframework/abp-samples/pull/246

### DIFF
--- a/framework/src/Volo.Abp.Swashbuckle/wwwroot/swagger/ui/abp.swagger.js
+++ b/framework/src/Volo.Abp.Swashbuckle/wwwroot/swagger/ui/abp.swagger.js
@@ -28,7 +28,16 @@ var abp = abp || {};
             }
             // Intercept .well-known request when the discoveryEndpoint is provided
             if (!firstRequest && oidcDiscoveryEndpoint.length !== 0 && request.url.includes(".well-known/openid-configuration")) {
-                request.url = oidcDiscoveryEndpoint;
+                if (oidcDiscoveryEndpoint.endsWith(".well-known/openid-configuration")) {
+                    request.url = oidcDiscoveryEndpoint;
+                    console.log(request.url);
+                    return;
+                }
+                if (!oidcDiscoveryEndpoint.endsWith("/")) {
+                    oidcDiscoveryEndpoint += "/"
+                }
+                request.url = oidcDiscoveryEndpoint + ".well-known/openid-configuration";
+                console.log(request.url);
             }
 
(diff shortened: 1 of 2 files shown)


framework/src/Volo.Abp.Swashbuckle/wwwroot/swagger/ui/abp.swagger.js
```diff
@@ -28,7 +28,16 @@ var abp = abp || {};
             }
             // Intercept .well-known request when the discoveryEndpoint is provided
             if (!firstRequest && oidcDiscoveryEndpoint.length !== 0 && request.url.includes(".well-known/openid-configuration")) {
-                request.url = oidcDiscoveryEndpoint;
+                if (oidcDiscoveryEndpoint.endsWith(".well-known/openid-configuration")) {
+                    request.url = oidcDiscoveryEndpoint;
+                    console.log(request.url);
```
realLiangshiwei, line 33: Should we use `abp.log.debug` instead of `console.log`?

rqx110: why not just remove it

realLiangshiwei: https://github.com/abpframework/abp/pull/17434/files#diff-383bbd9f9228466840c24c0c2c619e2ba1eb008ad0554a3e23a55fc1880b6a69R190

realLiangshiwei (approved):

framework/src/Volo.Abp.Swashbuckle/wwwroot/swagger/ui/abp.swagger.js
rqx110: could any one explain why can't remove the `console.log`?

gterdem: > could any one explain why can't remove the `console.log`?

So the users can see the URL and understand if they get an error because of malformed URL:
image

We encourage to check the browser logs for errors.

PR #17491, opened 2023-08-27
Introduce ` RepositoryInterceptor` and `Enable/DisableTracking()` extension methods.

### LINKED ISSUE #17487: Introduce IRepository.DisableTracking() extension method
Continues from https://github.com/abpframework/abp/issues/1848

We've implemented "no tracking by default" for readonly repositories. See https://github.com/abpframework/abp/issues/597 and see document: https://github.com/abpframework/abp/blob/fe95e8ddd0d236dc4003d96a1a90cb64f6f7c619/docs/en/Entity-Framework-Core.md#read-only-repositories

However, it doesn't solve everything. Because, most of the time, we are injecting a regular `IRepository<...>`, then in some method we only need querying, but in some other methods in the same application service, we need to modify objects.

One implementation propasal from me is to introduce `bool IRepository.IsChangeTrackingEnabled` property. We can set it to enable or disable change tracking for all get operations on this repository object. Usage:

* Set _repository.IsChangeTrackingEnabled to `false`.
* Query from the _repository object
* Set _repository.IsChangeTrackingEnabled to `true` **only if you need to enable it in the same request again** (if your application service method is just returning data, you don't need to re-enable it since the repository object already will be disposed after the request).

As a shortcut, we can introduce a `DisableTracking()` extension method:

````csharp
public async Task<List<TodoItemDto>> GetListAsync()
{
    using(_todoItemRepository.DisableTracking())
    {
        var items = await _todoItemRepository.GetListAsync(); //items are not tracked
    }

    var items2 = await _todoItemRepository.GetListAsync(); //items are tracked again
}
````

If you want to disable tracking for a complete method body, you may not use the `using` block:

````csharp
public async Task<List<TodoItemDto>> GetListAsync()
{
    _todoItemRepository.DisableTracking();

    var items = await _todoItemRepository.GetListAsync(); //items are not tracked
}
````

> USAGE WARNING: This approach is safe for typical application service / controller methods. However, if another methods of this service instance is called after `GetListAsync`, tracking will remain disabled. So, it is a best practice to always use the `using` statement if you are not sure that only one method of this service is called, then it is disposed. (so we won't mention this usage in docs)

### IMPLEMENTATION NOTES

* When we dispose the object returned from the `DisableTracking()` method, it should not directly set `IsChangeTrackingEnabled` to `true`, but `should` always set to the _previous_ value.
* We can also introduce `EnableTracking()` that works exactly opposite of `DisableTracking()`.
* It won't be implemented for MongoDB, since it doesn't have such a change tracking system. For MongoDB repositories, we can still use the same API, but it won't have any effect.

### Other Ideas

Introducing a `DisableEntityChangeTracking` attribute would be a more comfortable way (it disables tracking only for a specific method body or can be used on a class to disable for all methods in a class) in many cases:

````csharp
[DisableEntityChangeTracking]
public async Task<List<TodoItemDto>> GetListAsync()
{
    var items = await _todoItemRepository.GetListAsync(); //items are not tracked
}
````

That requires to implement ambient context pattern that works independently from repository objects (implemented with interception). So, instead of `bool IsChangeTrackingEnabled`, we can define nullable bool (`bool? IsChangeTrackingEnabled`) which is `null` by default. If it is `null`,  it uses the value in the ambient context. If it is also `null`, goes with default behavior.

`EnableEntityChangeTracking` attribute can also be nice to convert the behaviour vice verse (assume that a method is called from another method that disables tracking, but we want to enable tracking for the current method).

We could even define a convention that change tracking is disable for the application service / controller method starting with `Get` prefix in naming, but this will be a great breaking change. Even if we do it optional, it can be dangerous. So, change tracking should be disabled carefully, only if a performance difference occurs. For example, for a single entity querying methods, it doesn't worth to disable change tracking.

### Even other ideas

This feature has some overlaping with the https://github.com/abpframework/abp/issues/1848
Because making a repository _read-only_ automatically disables change tracking. It can be better to discard `IsReadOnly` property, and set `IsChangeTrackingEnabled` to `false` for read-only repositories. One drawback is that we can't throw exception if we call `UpdateAsync` method on a readonly repository (which is only possible if we cast it to `IRepository<...>`). But I think we can ignore that problem and cancel throwing exceptions in these cases. Agree?

### PR DESCRIPTION
**Pull request by maliming**
Added two extension methods so the developer can disable/enable entity tracking by these methods manually.

```
IDisposable DisableTracking(this IRepository repository)
IDisposable EnableTracking(this IRepository repository)
```

The `EnableEntityChangeTrackingAttribute` and `DisableEntityChangeTrackingAttribute` can be used by class/interface/method. The `RepositoryInterceptor` will get its `Enabled` property and change the current `Tracking` value by `IEntityChangeTrackingProvider`.

The `Repository` will check these conditions to decide whether to enable tracking.
* `IsChangeTrackingEnabled` (`null` by default. You can change it by `DisableTracking` or `EnableTracking`)
* The current `Tracking` value of `IEntityChangeTrackingProvider` . (It will be `null` if there is no `EntityChangeTrackingAttribute`)
* `true` by default.

Resolves #17487

### Checklist

### CODE REVIEW

File: framework/src/Volo.Abp.Ddd.Domain/Volo/Abp/Domain/Repositories/RepositoryHelper.cs [hunk 1]
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    private static bool AnyMethodHasEntityChangeTrackingAttribute(TypeInfo implementationType)
+    {
+        return implementationType
+            .GetMethods(BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic)
+            .Any(HasEntityChangeTrackingAttribute);
+    }
+
+    private static bool HasEntityChangeTrackingAttribute(MemberInfo methodInfo)
```
hikalkan:
```suggestion
    private static bool HasEntityChangeTrackingAttribute(MemberInfo memberInfo)
```

framework/src/Volo.Abp.Ddd.Domain/Volo/Abp/Domain/Repositories/RepositoryInterceptor.cs
```diff
@@ -0,0 +1,29 @@
...
+
+    public async override Task InterceptAsync(IAbpMethodInvocation invocation)
+    {
+        if (!RepositoryHelper.IsEntityChangeTrackingMethod(invocation.Method, out var changeTrackingAttribute))
+        {
+            await invocation.ProceedAsync();
+            return;
+        }
+
+        using (_entityChangeTrackingProvider.Change(changeTrackingAttribute?.Enabled))
```
hikalkan: Change only if `changeTrackingAttribute` is not `null`. Otherwise, you may change to `null` without intention.

framework/src/Volo.Abp.EntityFrameworkCore/Volo/Abp/Domain/Repositories/EntityFrameworkCore/EfCoreRepository.cs
```diff
@@ -107,7 +108,7 @@ public EfCoreRepository(IDbContextProvider<TDbContext> dbContextProvider)
 
     public async override Task<TEntity> InsertAsync(TEntity entity, bool autoSave = false, CancellationToken cancellationToken = default)
     {
-        CheckReadOnly();
+        CheckChangeTracking();
```
hikalkan: Inserting/updating/deleting an entity work even if you don't track changes for queried entities.

hikalkan (changes requested):

framework/src/Volo.Abp.Ddd.Domain/Volo/Abp/Domain/Repositories/EntityChangeTrackingAttribute.cs
```diff
@@ -0,0 +1,14 @@
+using System;
+
+namespace Volo.Abp.Domain.Repositories;
+
+[AttributeUsage(AttributeTargets.Method | AttributeTargets.Class | AttributeTargets.Interface)]
+public class EntityChangeTrackingAttribute : Attribute
```
hikalkan: We can make this class `abstract`.

framework/src/Volo.Abp.Ddd.Domain/Volo/Abp/Domain/Repositories/EntityChangeTrackingAttribute.cs
```diff
@@ -0,0 +1,14 @@
...
+{
+    public bool Enabled { get; set; }
```
hikalkan:
```suggestion
    public virtual bool IsEnabled { get; set; }
```

hikalkan (changes requested):

framework/src/Volo.Abp.Ddd.Domain/Volo/Abp/Domain/Repositories/RepositoryInterceptor.cs
maliming: The `changeTrackingAttribute` will not be `null` if `IsEntityChangeTrackingMethod` return `true`.

framework/src/Volo.Abp.EntityFrameworkCore/Volo/Abp/Domain/Repositories/EntityFrameworkCore/EfCoreRepository.cs
maliming: But the entity returned by the method or entity from parameters is `no tracking` after the method is called.

hikalkan: Where it is done? I didn't see any code part that detaches inserted/updated entities.

maliming: There might be code like this, So I think output warning logs are necessary.

```
using (var uow = uowManager.Begin())
{
    var author = await authorRepository.InsertAsync(new Author(Guid.NewGuid(), "Douglas"));

    // Entity still under tracking
    author.Name = "Douglas2";
    await uow.CompleteAsync();
}
```

hikalkan: So, in this example, `author` object is being tracked, and the name change is auto-saved, right? But the warning message says "Your changes may not be saved". Isn't that misleading the developer?

maliming: The code I share will not work if we disable entity tracking.
But developers may use such code, so I think we should output warnings when entity tracking is disabled.

hikalkan (approved):

PR #17497, opened 2023-08-29
Use `WebApplicationFactory` as the base class for Web integration testing.

### LINKED ISSUE #17484: Revisit ASP.NET Core Integration Test Infrastructure
We have a test base class for aspnet core: https://github.com/abpframework/abp/blob/dev/framework/src/Volo.Abp.AspNetCore.TestBase/Volo/Abp/AspNetCore/TestBase/AbpAspNetCoreIntegratedTestBase.cs

However, I see Microsoft made a lot of enhancements and introduces new ways of creating intergated tests. Please check them and try to understand what we can do for a more modern and integrated test experience:

https://learn.microsoft.com/en-us/aspnet/core/test/integration-tests?view=aspnetcore-7.0

maliming:
Resolve #17484


hikalkan (approved):

hikalkan:
I marked as breaking change since we made obsolete the old class and we may remove it in the future. So, we should mention that in the migration guide for 8.0.

PR #17532, opened 2023-09-01
Update if users collection is not empty

realLiangshiwei:


realLiangshiwei:
https://github.com/abpframework/abp/pull/17525/files

PR #17543, opened 2023-09-04
Disable logging for `HEAD` if `IsEnabledForGetRequests` is `false`.

### LINKED ISSUE #4010: Does the HEAD method access need to be recorded by audit log?
ABP sets `IsEnabledForGetRequests=false` by default to disable audit log for GET method.

The HEAD method is very similar to the GET method, and it is often used by heartbeat detectors. Maybe it does not have to be recorded.

maliming:
Resolve #4010


rqx110:
I am very confuse that the topic of this PR is about `HEAD` http method, but in code is about `Options` http method.

hikalkan (approved):

Issue #17529, opened 2023-09-01
Remove confirmation error during redirect to login


masum-ulu:
## Todo

- **Application directly go to login page without show any confirmation**
- **Only if developer specify during import theme-shared it'll shown**

PR #17544, opened 2023-09-04
AngularUI: Remove confirmation for 401 status code

masum-ulu:
### Description

**I separate PR because 7.4+ version error handler updated**

Resolves for 7.4 version #17529 

TODO: Remove confirmation modal for 401 status code, `this PR also resolves for commercial side`

### Checklist

### How to test it?
Send request before authenticate, it'll return 401 instead 403 and error handler will display a confirmation and it'll redirect to login. Try for password flow it works same for code flow

### DIFF
--- a/npm/ng-packs/packages/theme-shared/src/lib/services/status-code-error-handler.service.ts
+++ b/npm/ng-packs/packages/theme-shared/src/lib/services/status-code-error-handler.service.ts
@@ -77,12 +77,9 @@ export class StatusCodeErrorHandlerService implements CustomHttpErrorHandlerServ
 
     switch (this.status) {
       case 401:
+        this.authService.navigateToLogin();
+        break;
       case 404:
-        if (this.status === 401) {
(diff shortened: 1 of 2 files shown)


npm/ng-packs/packages/theme-shared/src/lib/services/status-code-error-handler.service.ts
```diff
@@ -78,15 +78,16 @@ export class StatusCodeErrorHandlerService implements CustomHttpErrorHandlerServ
     switch (this.status) {
       case 401:
       case 404:
+        if (this.status === 401) {
+          this.authService.navigateToLogin();
+          return;
+        }
+
```
rqx110: move these lines to `case 401:`

masum-ulu: This time I need to check twice for customError case

mahmut-gundogdu (approved):

PR #17584, opened 2023-09-07
Remove Volo.Abp.AspNetCore.Mvc.UI.Theme.Shared from Openiddict module

### LINKED ISSUE #17583: Remove Volo.Abp.AspNetCore.Mvc.UI.Theme.Shared from Openiddict module
Reference to `Volo.Abp.AspNetCore.Mvc.UI` instead of `Volo.Abp.AspNetCore.Mvc.UI.Theme.Shared`

realLiangshiwei:
### Description

Resolves https://github.com/abpframework/abp/issues/17583

### Checklist

### How to test it?

run the openiddict demo app


hikalkan:
Do we add Volo.Abp.AspNetCore.Mvc.UI.Theme.Shared to all modules? I think we do it. So, no problem to add here too. This is an abstraction package, not the Lepton theme.
So, I am canceling this.

PR #17597, opened 2023-09-10
Removed Acme.BookStore.Web in swagger ui in tiered mvc

barbarosyurttagul:
### Description

There is no open issue for this PR. 

`Tutorials Part 1 Auto API Controllers` explains how to launch the **swagger UI**. The markdown file has a condition depending on the UI to run the relevant project. However since we are talking about API controller, the only project we have to run to see the swagger UI that we work in our tutorial is `HttpApi.Host`. `Web `project has no relationship with API Controllers. If we run the Web project to see the Books controller endpoints, we can not.

### Checklist

- [ ] no need to document

### DIFF
--- a/docs/en/Tutorials/Part-1.md
+++ b/docs/en/Tutorials/Part-1.md
@@ -436,7 +436,7 @@ ABP can [**automagically**](../API/Auto-API-Controllers.md) configure your appli
 
 ### Swagger UI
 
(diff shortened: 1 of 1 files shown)


barbarosyurttagul:
I think the documentation is only for untiered architecture. Because of this I closed the PR.

PR #17669, opened 2023-09-19
Check Redis server when creating project.

### LINKED ISSUE #17473: ABP CLI: WARN for prerequirements
There are some prerequisites to be able to create, build and run the ABP startup templates.

For example https://docs.abp.io/en/abp/latest/Getting-Started-Setup-Environment?UI=MVC&DB=EF&Tiered=Yes

Requirements for that solution configuration:

* An IDE (e.g. [Visual Studio](https://visualstudio.microsoft.com/vs/)) that supports [.NET 7.0+](https://dotnet.microsoft.com/download/dotnet) development.
* [Node v16 or v18](https://nodejs.org/)

* [Yarn v1.20+ (not v2)](https://classic.yarnpkg.com/en/docs/install) [1](https://docs.abp.io/en/abp/latest/Getting-Started-Setup-Environment?UI=MVC&DB=EF&Tiered=Yes#f-yarn) or npm v6+ (already installed with Node)

* [Redis](https://redis.io/) (the startup solution uses the Redis as the [distributed cache](https://docs.abp.io/en/abp/latest/Caching)).

If users create such a solution (tiered, MVC, EF Core), we can check these requirements and WARN the user **before** and **after** creating the project.

We may skip the IDE check; maybe they won't use an IDE; however, without Node, solution creation will fail because it will not be able to install client-side packages. That list also needs to be included. For example, EF Core command line tools (`dotnet ef ...`) is also a requirement because we are creating the initial migrations.

So, these are required to create such a solution. We can show a colorful WARNING message (background = black, foreground = yellow) (message can be "WARNING: The following tools should be installed on this machine to be able to successfully create this solution. Otherwise, you may get errors on solution creation process.") and ask the user to either cancel the operation or continue BEFORE starting the solution creation process (if they continue, they will probably get an error, but let them go if they want).

There are also some requirements needed not to create the solution but to run it. For example, Redis, in this list, is such a requirement. We should show a colorful WARNING message (message can be "NOTICE: The following tools is required to run your solution.") AFTER creating the solution if we can not detect a local Redis instance.

EF Core, for example, can be auto-installed by ABP CLI, if not found in the computer.

So, with this issue;

- Auto-install some dependencies (EF Core is a good example, but installing Redis is not a good idea, and installing Node may not be easy/possible), so do your best.
- Show a WARNING for some missing dependencies (required to create the solution) BEFORE solution creation, and let the user to continue or cancel.
- Show a WARNING for some missing dependencies (required to run the solution) AFTER solution creation. No use action in that case, exit from CLI as normal.

Notice that dependencies may vary based on the solution preferences.

maliming:
Resolve #17473

```cs
Creating your project...
Project name: QA
Template: app
Tiered: yes
Output folder: test
Theme: LeptonXLite
Check out the documents at https://docs.abp.io/en/abp/latest/Startup-Templates/Application
'QA' has been successfully created to 'test'
NOTICE: The following tools is required to run your solution.
        * Redis is not installed or not running on your computer.
Creating initial migrations...
```

### DIFF
--- a/framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/ProjectBuilding/Templates/CheckRedisPreRequirements.cs
+++ b/framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/ProjectBuilding/Templates/CheckRedisPreRequirements.cs
@@ -9,7 +9,7 @@ public class CheckRedisPreRequirements : ProjectBuildPipelineStep
     public override void Execute(ProjectBuildContext context)
     {
         var modules = context.Files.Where(f => f.Name.EndsWith("Module.cs", StringComparison.OrdinalIgnoreCase));
-        if (modules.Any(module => !module.Content.Contains("Redis:Configuration")))
+        if (modules.Any(module => module.Content.Contains("Redis:Configuration")))
         {
             context.BuildArgs.ExtraProperties["PreRequirements:Redis"] = "true";
         }
--- a/framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/ProjectBuilding/Templates/CheckRedisPreRequirements.cs
+++ b/framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/ProjectBuilding/Templates/CheckRedisPreRequirements.cs
@@ -1,0 +1,16 @@
+using System.Linq;
+using Volo.Abp.Cli.ProjectBuilding.Building;
+
+namespace Volo.Abp.Cli.ProjectBuilding.Templates;
+
+public class CheckRedisPreRequirements : ProjectBuildPipelineStep
+{
+    public override void Execute(ProjectBuildContext context)
+    {
+        var modules = context.Files.Where(f => f.Name.EndsWith("Module.cs"));
+        if (modules.Any(module => !module.Content.Contains("Redis:Configuration")))
+        {
+            context.BuildArgs.ExtraProperties["PreRequirements:Redis"] = "true";
+        }
+    }
+}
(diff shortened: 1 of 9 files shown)


framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/ProjectBuilding/Templates/CheckRedisPreRequirements.cs
```diff
@@ -0,0 +1,17 @@
...
+using Volo.Abp.Cli.ProjectBuilding.Building;
+
+namespace Volo.Abp.Cli.ProjectBuilding.Templates;
+
+public class CheckRedisPreRequirements : ProjectBuildPipelineStep
+{
+    public override void Execute(ProjectBuildContext context)
+    {
+        var modules = context.Files.Where(f => f.Name.EndsWith("Module.cs", StringComparison.OrdinalIgnoreCase));
+        if (modules.Any(module => !module.Content.Contains("Redis:Configuration")))
```
EngincanV: There seems to be a problem here. If one of the module files does not have Redis configuration (for example, the domain shared module has not), then this *if block* will run, and the `PreRequirements:Redis` extra property will be set. 

Therefore, [it will check that Redis is installed and running](https://github.com/abpframework/abp/blob/6945764746cb8f3b3119a196ad4bb663876c2c5c/framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/Commands/NewCommand.cs#L138) after created the application even it might not need to. 

So, we should either update this line or the line below: https://github.com/abpframework/abp/blob/6945764746cb8f3b3119a196ad4bb663876c2c5c/framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/Commands/NewCommand.cs#L138

EngincanV (changes requested):

EngincanV (approved):

PR #17678, opened 2023-09-20
Use `api_version` as variable name for `api-version`.

maliming:


hikalkan:
1) Why?
2) Isn't that a breaking change?

maliming:
This seems to be a problem with the customer's project rather than our code, I'll check again,

PR #17699, opened 2023-09-21
Refactor modules' local user synchronization

### LINKED ISSUE #17167: `UserEto` is better to implement `IHasEntityVersion`

_No response_

After https://github.com/abpframework/abp/issues/16495 is merged, the [UserEto](https://github.com/abpframework/abp/blob/dev/modules/users/src/Volo.Abp.Users.Abstractions/Volo/Abp/Users/UserEto.cs) is better to implement `IHasEntityVersion`. So that we can remove user synchronizers (such as BlogUserSynchronizer) and use the [ABP Entity Synchronizer](https://docs.abp.io/en/abp/latest/Distributed-Event-Bus#entity-synchronizer).

### Additional context

_No response_

gdlcf88:
Resolve #17167


maliming:
build and test / build-test (pull_request) Failing after 360m

PR #17720, opened 2023-09-25
No case-sensitive filtering at the Users page

### LINKED ISSUE #17716: No case-sensitive filtering at the Users page
Filter users with no case sensitivity at the Users page.

Reported at https://github.com/volosoft/vs-internal/issues/2821

realLiangshiwei:
### Description

Resolves https://github.com/abpframework/abp/issues/17716

### Checklist

### How to test it?

run the `app` template

### DIFF
--- a/modules/identity/src/Volo.Abp.Identity.EntityFrameworkCore/Volo/Abp/Identity/EntityFrameworkCore/EfCoreIdentityUserRepository.cs
+++ b/modules/identity/src/Volo.Abp.Identity.EntityFrameworkCore/Volo/Abp/Identity/EntityFrameworkCore/EfCoreIdentityUserRepository.cs
@@ -174,7 +174,7 @@ public class EfCoreIdentityUserRepository : EfCoreRepository<IIdentityDbContext,
         DateTime? minModifitionTime = null,
         CancellationToken cancellationToken = default)
     {
-        var upperFilter = filter?.ToUpper();
+        var upperFilter = filter?.ToUpperInvariant();
         return await (await GetDbSetAsync())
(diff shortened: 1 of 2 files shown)


modules/identity/src/Volo.Abp.Identity.EntityFrameworkCore/Volo/Abp/Identity/EntityFrameworkCore/EfCoreIdentityUserRepository.cs
```diff
@@ -174,13 +174,14 @@ into gp
         DateTime? minModifitionTime = null,
         CancellationToken cancellationToken = default)
     {
+        var upperFilter = filter?.ToUpper();
```
EngincanV: I think we can use the `ToUpperInvariant()` to prevent character mismatches and pass [The Turkey Test](http://www.moserware.com/2008/02/does-your-code-pass-turkey-test.html).

EngincanV (commented):
Can you make the same change for the MongoDB side by considering my review comment? @RealLowis

modules/identity/src/Volo.Abp.Identity.EntityFrameworkCore/Volo/Abp/Identity/EntityFrameworkCore/EfCoreIdentityUserRepository.cs
realLiangshiwei: ok, good idea

EngincanV (approved):

PR #17734, opened 2023-09-26
Blazor - The DataGrid single button does not utilize a dropdown.

### LINKED ISSUE #17703: Blazor - The single button does not utilize a dropdown.

If there is only one button in EntityActions, it should be displayed as a standalone button.

https://support.abp.io/QA/Questions/1431/Blazor---Action-Button---DataGridEntityActionsColumn-with-single-EntityAction

You don't need to set the `EntityActions` property; instead, you can determine whether to use a dropdown based on the number of actions available. If there is only one action, it will be displayed as a standalone button without a dropdown.

### Additional context

image

image

zhaofenglee:
### Description

Resolves #17703 

### Checklist


enisn (approved):

framework/src/Volo.Abp.BlazoriseUI/Components/EntityActions.razor
```diff
@@ -1,6 +1,6 @@
 ﻿@typeparam TItem
 <CascadingValue Value="this" IsFixed="true">
-    @if ( Type == ActionType.Dropdown )
+    @if ( Actions.Count > 1 )
```
realLiangshiwei, line 3: I think we don't need to change this.

framework/src/Volo.Abp.BlazoriseUI/Components/EntityActions.razor.cs
```diff
@@ -40,7 +40,12 @@ public partial class EntityActions<TItem> : ComponentBase
 
     internal void AddAction(EntityAction<TItem> action)
     {
-        Actions.Add(action);
+        if (action.Visible)
```
realLiangshiwei, line 43: Please do not check `Visible` here, It should use the same logic of MVC UI
https://github.com/abpframework/abp/blob/dev/framework/src/Volo.Abp.AspNetCore.Mvc.UI.Theme.Shared/wwwroot/libs/abp/aspnetcore-mvc-ui-theme-shared/datatables/datatables-extensions.js#L201
https://github.com/abpframework/abp/blob/dev/framework/src/Volo.Abp.AspNetCore.Mvc.UI.Theme.Shared/wwwroot/libs/abp/aspnetcore-mvc-ui-theme-shared/datatables/datatables-extensions.js#L192

zhaofenglee: If you do not check the Visible property, the dropdown button may still be displayed even if the user does not have permission. Currently, ABP Suite generates CRUD code using the "Visible" control to determine whether it should be displayed.
https://github.com/abpframework/abp-samples/blob/dd2e83054a307d80b0fb54bd0490c8e3a5905e8d/BookStore-Blazor-EfCore/src/Acme.BookStore.Blazor/Pages/Books.razor#L43-L53

framework/src/Volo.Abp.BlazoriseUI/Components/EntityActions.razor
zhaofenglee: Not change this code will result in it not functioning as expected. This issue can be reproduced by debugging through the source code.

framework/src/Volo.Abp.BlazoriseUI/Components/EntityActions.razor.cs
realLiangshiwei: This is by design, MVC UI is also like this.

image
image

We should ensure the consistency of different UIs.

@enisn  what do you think?

zhaofenglee: This design is reasonable, and the PR can be closed.

PR #17738, opened 2023-09-26
Introduce MongoDbDatabaseEventHandlerBase

yekalkan:


framework/src/Volo.Abp.MongoDB/Volo/Abp/MongoDB/MongoDbDatabaseEventHandlerBase.cs
```diff
@@ -0,0 +1,58 @@
+﻿using System;
+using System.Threading.Tasks;
+using Volo.Abp.DependencyInjection;
+using Volo.Abp.EventBus.Distributed;
+using Volo.Abp.MultiTenancy;
+
+namespace Volo.Abp.MongoDB;
+
+public abstract class MongoDbDatabaseEventHandlerBase<TDbContext> :
```
hikalkan:
```suggestion
public abstract class MongoDatabaseMigrationEventHandlerBase<TDbContext> :
```

Also move to `Migrations` namespace just like [we did for EF Core](https://github.com/abpframework/abp/blob/d07935736877d9579b4824080596c82d705755a7/framework/src/Volo.Abp.EntityFrameworkCore/Volo/Abp/EntityFrameworkCore/Migrations/EfCoreDatabaseMigrationEventHandlerBase.cs).

hikalkan (changes requested):

hikalkan (approved):

PR #17758, opened 2023-09-28
Update the `Configuring OpenIddict` documentation

EngincanV:
Documentation updates for https://github.com/abpframework/abp/pull/17512

### DIFF
--- a/docs/en/Deployment/Configuring-OpenIddict.md
+++ b/docs/en/Deployment/Configuring-OpenIddict.md
@@ -2,61 +2,42 @@
 
 This document introduces how to configure `OpenIddict` in the `AuthServer` project.
 
-There are different configurations in the `AuthServer` project for `Development` and `Production` environment.
+There are different configurations in the `AuthServer` project for the `Development` and `Production` environments.
 
 ````csharp
 public override void PreConfigureServices(ServiceConfigurationContext context)
 {
     var hostingEnvironment = context.Services.GetHostingEnvironment();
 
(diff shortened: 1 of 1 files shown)


maliming (approved):

docs/en/Deployment/Configuring-OpenIddict.md
```diff
@@ -2,61 +2,42 @@
...
-            //...
-        });
+       PreConfigure<AbpOpenIddictAspNetCoreOptions>(options =>
+       {
+          options.AddDevelopmentEncryptionAndSigningCertificate = false;
+       });
+
+       PreConfigure<OpenIddictServerBuilder>(serverBuilder =>
+       {
+          serverBuilder.AddProductionEncryptionAndSigningCertificate("openiddict.pfx", "00000000-0000-0000-0000-000000000000");
```
kevinchalet, line 21: FYI, it's generally much better to use two different/separate certificates: one for signing and one for encrypting. It's not only better from a security perspective, but also the only one way to support different algorithms (e.g RSA for encryption and ECDSA for signing) 😃

kevinchalet: (note: my remark isn't strictly related to this PR, just thought it was a good idea to mention it here)

maliming: Thanks @kevinchalet 

We will mention this in the document. 👍

PR #17797, opened 2023-10-03
Remove abp package . it cause local ref problem in ABP

### LINKED ISSUE #16034: Testing Libraries does not work with local references.
Jest or tsconfig did not see local reference. for example.

If you change  config state service, validation-utils.spec.ts file couldn't see the changes because that is not pushed.

https://github.com/abpframework/abp/blob/dev/npm/ng-packs/packages/theme-shared/src/lib/tests/validation-utils.spec.ts 

Testing is hard when the local ref doesn't work.

mahmut-gundogdu:
### Description

Resolves #16034

When packages installed via npm.js test read from node_modules so it cause the local ref problem in Testing. So I have removed abp angular packages in package.json in ng-packs. it fixed . 

### Checklist

### How to test it?
- go to abp/npm/ng-packs
- remove node_modules (if exists)
-  run yarn install
- Go to a test file for example `validation-utils.spec.ts`
- change a dependency outside of that package like `config-state.service.ts`
expected: the test should detect and work with local changes.


npm/ng-packs/package.json
```diff
@@ -44,19 +44,6 @@
...
-    "@abp/ng.identity": "~7.4.0-rc.5",
-    "@abp/ng.oauth": "~7.4.0-rc.5",
-    "@abp/ng.permission-management": "~7.4.0-rc.5",
-    "@abp/ng.schematics": "~7.4.0-rc.5",
-    "@abp/ng.setting-management": "~7.4.0-rc.5",
-    "@abp/ng.tenant-management": "~7.4.0-rc.5",
-    "@abp/ng.theme.basic": "~7.4.0-rc.5",
-    "@abp/ng.theme.lepton-x": "^2.4.0-rc.4",
-    "@abp/ng.theme.shared": "~7.4.0-rc.5",
     "@abp/utils": "~7.4.0-rc.5",
```
masum-ulu, line 47: `"@abp/ng.theme.lepton-x": "^2.4.0-rc.5"` is missing. There is not local reference path for this package!

mahmut-gundogdu:
the PR cause different problem. So I have to close. It will fix otherway in the future.

PR #17812, opened 2023-10-06
Add `README.md` file to every new solution to explain how to run it.

### LINKED ISSUE #17807: Add README.MD file to every new solution to explain how to run it
# README.MD files in startup templates

We want to add a README.MD file for the startup templates that;

* Writes pre-requirements
* Write the essential configurations to be considered
* Explains the steps needed to do before running the solution
* Explains how to run the solution.
* Briefly explains the projects in that solution

Notice that the content of that file will be different based on the preferences while creating the solution, so it should be dynamically built/manipulated on template creation.

## Example

Here, an example I've tried and needed to do for the `abp new Acme.BookStore -csf -t app` command. This is the most simple example, and it will be different for others.

### Pre-requirements

* [.NET 7.0.8+ SDK](https://dotnet.microsoft.com/download/dotnet)
* [Node v18 or 20](https://nodejs.org/en)

### Configurations

The solution comes with a default configuration that works out of the box. However, you may consider to change the following configuration before running your solution:

* Check the `ConnectionStrings` in `appsettings.json` files under the `Acme.BookStore.Web` and `Acme.BookStore.DbMigrator` projects and change it if you need.

### Preparings for the first run

* Run `abp install-libs` command to install client-side package dependencies. This step is automatically done when you create a new solution with ABP CLI. However, you should run it yourself if you have first cloned this solution from your source control, or added a new client-side package dependency to your solution.
* Run `Acme.BookStore.DbMigrator` to create the initial database. This should be done in the first run. It is also needed if a new database migration is added to the solution later.

### Solution structure

This is a layered monolith application that consists of two applications:

* `/src/Acme.BookStore.DbMigrator`: A console application which applies the migrations and also seeds the initial data. It is useful on development as well as on production environment.
* `/src/Acme.BookStore.Web`: ASP.NET Core MVC / Razor Pages application that is the essential web application of the solution.

### Additional resources

You can see the following resources to learn more about your solution and the ABP Framework:

* [Web Application Development Tutorial](https://docs.abp.io/en/abp/latest/Tutorials/Part-1)
* [Application Startup Template Structure](https://docs.abp.io/en/abp/latest/Startup-Templates/Application)

maliming:
Resolves #17807

### DIFF
--- a/templates/app-nolayers/aspnet-core/README.md
+++ b/templates/app-nolayers/aspnet-core/README.md
@@ -1,6 +1,6 @@
 # MyCompanyName.MyProjectName
 
-## About This Solution
+## About this solution
 
 This is a minimalist, non-layered startup solution with the ABP Framework. All the fundamental ABP modules are already installed. 
 
@@ -15,7 +15,7 @@ The solution comes with a default configuration that works out of the box. Howev
 
 * Check the `ConnectionStrings` in `appsettings.json` files under the `MyCompanyName.MyProjectName` project and change it if you need.
 
-### Preparings for the first run
+### Before running the application
 
 * Run `abp install-libs` command on your solution folder to install client-side package dependencies. This step is automatically done when you create a new solution with ABP CLI. However, you should run it yourself if you have first cloned this solution from your source control, or added a new client-side package dependency to your solution.
 * The application needs to connect to a database. Run the following command in the `MyCompanyName.MyProjectName` directory:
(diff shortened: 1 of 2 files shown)


EngincanV (approved):
I have three small suggestions that you may wanna consider:

* It's not too important but there is a standardization problem on the headers in the documentation. For example, one of the headers is like `About This Solution`, and the other one is like `Preparings for the first run`.  We should select, one of these naming conventions and stick through the documentation. (I think the second one is better but it's up to you.)
* I think `Before running the application` might be a better header naming than `Preparings for the first run`.
* [As mentioned here](https://github.com/abpframework/abp/issues/17807#issuecomment-1748743160), we may add a link of the currently selected theme (basic or leptonx-lite), under the `Additional resources` section.

EngincanV (approved):

yekalkan (approved):

PR #17813, opened 2023-10-06
Async functions do not return a Task type. If the parameter is empty, an exception will be returned.

happtim:
### Description

Write my business code according to the test case. Discover that an exception is thrown when the parameter is empty. If the await keyword is not added to an asynchronous method, returning null will also throw an exception.

### Checklist


maliming (approved):

maliming:
Thanks @happtim

sisyphus1212:
@maliming are you ok?

Issue #17866, opened 2023-10-13
Should check MenuEnable feature in the CmsKitPublicMenuContributor


realLiangshiwei:
<img width="849" alt="image" src="https://github.com/abpframework/abp/assets/16813853/da86a6d7-a503-43e3-b038-14d3372ecfde">

PR #17867, opened 2023-10-13
Check MenuEnable feature in the CmsKitPublicMenuContributor

realLiangshiwei:
### Description

Resolves https://github.com/abpframework/abp/issues/17866

### Checklist

### DIFF
--- a/modules/cms-kit/src/Volo.CmsKit.Public.Web/Menus/CmsKitPublicMenuContributor.cs
+++ b/modules/cms-kit/src/Volo.CmsKit.Public.Web/Menus/CmsKitPublicMenuContributor.cs
@@ -25,7 +25,7 @@ public class CmsKitPublicMenuContributor : IMenuContributor
     private async Task ConfigureMainMenuAsync(MenuConfigurationContext context)
     {
         var featureChecker = context.ServiceProvider.GetRequiredService<IFeatureChecker>();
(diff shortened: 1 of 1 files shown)


modules/cms-kit/src/Volo.CmsKit.Public.Web/Menus/CmsKitPublicMenuContributor.cs
```diff
@@ -22,7 +24,8 @@ public async Task ConfigureMenuAsync(MenuConfigurationContext context)
 
     private async Task ConfigureMainMenuAsync(MenuConfigurationContext context)
     {
-        if (GlobalFeatureManager.Instance.IsEnabled<MenuFeature>())
+        var featureChecker = context.ServiceProvider.GetRequiredService<IFeatureChecker>();
+        if (GlobalFeatureManager.Instance.IsEnabled<MenuFeature>() && featureChecker.IsEnabledAsync(CmsKitFeatures.MenuEnable))
```
enisn: I think there is a missing `await` usage
```suggestion
        if (GlobalFeatureManager.Instance.IsEnabled<MenuFeature>() && await featureChecker.IsEnabledAsync(CmsKitFeatures.MenuEnable))
```

realLiangshiwei: you are right

enisn (approved):

PR #17870, opened 2023-10-13
Enable `Central Package Management`.

### LINKED ISSUE #16559: Using central package management

https://learn.microsoft.com/en-us/nuget/consume-packages/Central-Package-Management

> Starting with [NuGet 6.2](https://learn.microsoft.com/en-us/nuget/release-notes/nuget-6.2), you can centrally manage your dependencies in your projects with the addition of a Directory.Packages.props file and an MSBuild property.

abp currently uses `Directory.Build.props` to manage most nuget packages. 

I think we can improve the package management experience by using centralized package management. This also applies to templates, where developers can more easily check and manage all abp package versions.

### Additional context

_No response_

maliming:
Resolves #16559

All package versions of the **framework and modules** will be defined on the [Directory.Packages.props](https://github.com/abpframework/abp/blob/dev/Directory.Packages.props) file.

[We will disable this feature in startup templates](https://github.com/abpframework/abp/blob/dev/templates/Directory.Packages.props#L3).


hikalkan (approved):

htve:
We also need to enable it in the template project

PR #17912, opened 2023-10-18
Downgrade ImageSharp version to `1.0.4`

enisn:
### Description

Resolves https://github.com/abpframework/abp/issues/17906

Issue #17906, opened 2023-10-17
CMS Kit: Fix missing method exception problem


EngincanV:
![image](https://github.com/abpframework/abp/assets/43685404/105ce79b-109a-4300-8898-91452fc46025)

Update the `SixLabors.ImageSharp.Drawing` package version and make the related changes:

https://github.com/abpframework/abp/blob/6ecdeb991188f5b9be52393315a9c001adfe2a6b/Directory.Packages.props#L141-L142

enisn:
Still a problem for CMS Kit

enisn:
https://github.com/abpframework/abp/pull/17916#issuecomment-1767912201

PR #17916, opened 2023-10-18
Revert "Downgrade ImageSharp version to `1.0.4`"

enisn:
Reverts abpframework/abp#17912

Only CMSKit should use `1.0.4` version


maliming (approved):

maliming:
We can override the package version in cms-kit.

https://learn.microsoft.com/en-us/nuget/consume-packages/Central-Package-Management#overriding-package-versions

enisn:
> We can override the package version in cms-kit.
> 
> https://learn.microsoft.com/en-us/nuget/consume-packages/Central-Package-Management#overriding-package-versions

@salihozkara will work on this issue: https://github.com/abpframework/abp/issues/17906

We can upgrade CmsKit usage to the latest version

PR #17950, opened 2023-10-22
Show AppService endpoints in Swagger UI of module template

jflaga:
### Description

When creating a project using the module template, the `HttpApi.Host` module does not have configuration for creating Auto API Controllers. 

To fix that, I added this configuration in `MyProjectNameHttpApiHostModule.cs`:

``` csharp
Configure<AbpAspNetCoreMvcOptions>(options =>
{
    options.ConventionalControllers.Create(typeof(MyProjectNameApplicationModule).Assembly);
});
```


maliming:
hi

You can define your `controllers` in `MyCompanyName.MyProjectName.HttpApi` project.

PR #17960, opened 2023-10-24
Add `AddMarginBottomClass` to remove the `mb` class.

### LINKED ISSUE #17948: Remove mb-3 that surrounds abp-input mvc tag helper

As seen here https://bootstrap-taghelpers.abp.io/Components/FormElements
The abp-input tag helper surrounds the input with a div that has mb-3 on it. 

I suggest removing the div altogether so we can put the margins that suit our designs.

### Additional context

How can I download the code to my project and modify the tag helper?

maliming:
Resolve #17948


myesn:
Please consider default to false, because I don't think a hidden element should have side effects:
```csharp
public bool AddMarginBottomClass  { get; set; } = false;
```

maliming:
> Please consider default to false, because I don't think a hidden element should have side effects:
> 
> ```cs
> public bool AddMarginBottomClass  { get; set; } = false;
> ```

We don't want to make a breaking change.

EngincanV (approved):

PR #17990, opened 2023-10-26
Add email/username changed event

salihozkara:
### Description

When changing email/username, there may be scenarios where the old email/username is needed. For this reason, when the email/username changes, publish a distributed event containing the old email/username.

#### Breaking Change
Added IDistributedEventBus to IdentityUserManager's constructor

### Checklist


modules/identity/src/Volo.Abp.Identity.Domain/Volo/Abp/Identity/IdentityUserManager.cs
```diff
@@ -285,4 +289,43 @@ public virtual async Task ResetRecoveryCodesAsync(IdentityUser user)
...
+
+    public async override Task<IdentityResult> SetEmailAsync(IdentityUser user, string email)
+    {
+        var oldMail = user.Email;
+        
+        var result = await base.SetEmailAsync(user, email);
+        
+        result.CheckErrors();
+        
+        await DistributedEventBus.PublishAsync(
```
maliming: We can consider checking the `oldValue != currentValue` and then publish the event.

maliming (approved):

Issue #17989, opened 2023-10-26
Upgrade SixLabors.ImageSharp the latest


realLiangshiwei:
https://support.abp.io/QA/Questions/6048/Issue-after-updating-to-740-and-my-custom-LoginModel-throwing-DI-issue

TronActive:
This will be a big help

huyqta:
Is this fixing on latest version? I've just upgraded to 7.4.0 today and still met SixLabors.ImageSharp error.

PR #17999, opened 2023-10-27
Upgrade SixLabors.ImageSharp the latest

realLiangshiwei:
### Description

Resolves https://github.com/abpframework/abp/issues/17989

### Checklist


EngincanV (approved):

TronActive:
@realLiangshiwei, a heads up with this update is it will break the commercial version. More specifically, in the AccountAppService, it breaks DI because of the SixLabors.ImageSharp.Formats.Jpeg.JpegEncoder has changed. There were a few other changes that needed to be fixed as well. I don't need the functionality to add and change the profile image, so I just overrode the ImageCompressor and ImageSharpImageCompressorContributor to get around this and removed that functionality.

PR #18014, opened 2023-10-29
Angular Data Table Column Extensions enhance

linch90:
### Description

Add `ColumnVisible` to EntityProp for show/hide the whole table column, as the `Visible` only show or hide the cell content, not the whole column.

The `ColumnVisible` is a predicate that to decide if the entity column should be displayed or not.
It signaturer is `getInjected => boolean`, you can use the `getInjected` to inject the service you want.

### DIFF
--- a/npm/ng-packs/packages/theme-shared/extensions/src/lib/components/extensible-table/extensible-table.component.html
+++ b/npm/ng-packs/packages/theme-shared/extensions/src/lib/components/extensible-table/extensible-table.component.html
@@ -24,23 +24,24 @@
 
   <ng-container *ngFor="let prop of propList; let i = index; trackBy: trackByFn">
     <ngx-datatable-column
+      *ngIf="prop.columnVisible(getInjected)"
       [width]="columnWidths[i + 1] || 200"
-      [name]="prop.displayName | abpLocalization "
+      [name]="prop.displayName | abpLocalization"
       [prop]="prop.name"
       [sortable]="prop.sortable"
     >
-    <ng-template ngx-datatable-header-template let-column="column" >
-      <span     
-      *ngIf="prop.tooltip; else elseBlock"
-      [ngbTooltip]="prop.tooltip | abpLocalization" 
-      container="body"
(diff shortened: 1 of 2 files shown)


npm/ng-packs/packages/theme-shared/extensions/src/lib/components/extensible-table/extensible-table.component.html
```diff
@@ -24,23 +24,24 @@
 
   <ng-container *ngFor="let prop of propList; let i = index; trackBy: trackByFn">
     <ngx-datatable-column
+      *ngIf="prop.columnVisible(getInjected)"
```
mahmut-gundogdu: I would use Abp visible.  instead of if. It can work with observable and promise

https://github.com/abpframework/abp/blob/dev/npm/ng-packs/packages/theme-shared/src/lib/directives/visible.directive.ts

npm/ng-packs/packages/theme-shared/extensions/src/lib/models/entity-props.ts
```diff
@@ -31,6 +31,7 @@ export class EntityProp<R = any> extends Prop<R> {
   readonly component?: Type<any>;
   readonly enumList?: Array<ABP.Option<any>>;
   readonly tooltip?: string;
+  readonly columnVisible: ColumnPredicate;
```
mahmut-gundogdu, line 34: options.visible, already exists. what is the different ? I think, we can use that.

npm/ng-packs/packages/theme-shared/extensions/src/lib/components/extensible-table/extensible-table.component.html
linch90: Hi, changes has been applied.

mahmut-gundogdu (approved):

masum-ulu (approved):

masum-ulu:
Thanks for your contribution @linch90

PR #18030, opened 2023-10-30
Show error message when open modal fails in mvc.

genoher:
### Description

When an error occurs when opening a modal, there is no record, unless you open the browser console.

I have modified the default behavior of the ModalManager class so that when it loads the information from a modal, if an error occurs, it displays an error message.

### Checklist

- [ ] I fully tested it as developer without creating unit testing

### How to test it?

To test it, the easiest thing is to throw an exception in the OnGetAsync method of a page that is used as a modal. EditModal.cshtml


genoher:
Related question

https://support.abp.io/QA/Questions/3285/How-can-I-show-an-exception-that-occurs-in-application-services-in-OnGetAsync-method-in-the-modal

maliming (approved):

maliming:
Thanks @genoher

PR #18051, opened 2023-11-02
Remove Permissions and other user-related data when deleting a user.

### LINKED ISSUE #18042: Remove Permissions and other user-related data when deleting a user
Based on the problem https://github.com/abpframework/abp/issues/17927

When a user is deleted (in the Identity module), we can publish a distributed event, so other modules can know that and take the necessary actions.

Actually, an auto event for `UserEto` is already published automatically by the ABP framework. But that is not enabled by default. We can enable it by default in the Identity module, like that:

````csharp
Configure<AbpDistributedEntityEventOptions>(options =>
{
	options.AutoEventSelectors.Add<IdentityUser>();
});
````

Once we enabled it, we can subscribe to `IDistributedEventHandler<EntityDeletedEto<UserEto>>` in the following modules and make the necessary changes:

* Permission management module deletes the permissions related to that user.
* Setting management module deletes the permissions related to that user.
* Identity management module deleted user delegations, user claims, linked users and role-assignments for that user (it actually doesn't need to do with events since it is in the same module - maybe it can listen the local event to be transactional by default)

### PR DESCRIPTION
**Pull request by maliming**
Resolves #18042

### CODE REVIEW

File: modules/setting-management/src/Volo.Abp.SettingManagement.Application/Volo/Abp/SettingManagement/UserDeletedEventHandler.cs [hunk 1]
```diff
@@ -0,0 +1,25 @@
...
+    protected ISettingManager SettingManager { get; }
+
+    public UserDeletedEventHandler(ISettingManager settingManager)
+    {
+        SettingManager = settingManager;
+    }
+
+    public async Task HandleEventAsync(EntityDeletedEto<UserEto> eventData)
+    {
+        await SettingManager.DeleteAsync(UserPermissionValueProvider.ProviderName, eventData.Entity.Id.ToString());
```
rqx110:
```suggestion
        await SettingManager.DeleteAsync(UserSettingValueProvider.ProviderName, eventData.Entity.Id.ToString());
```

modules/identity/src/Volo.Abp.PermissionManagement.Domain.Identity/Volo/Abp/PermissionManagement/Identity/UserDeletedEventHandler.cs
```diff
@@ -0,0 +1,25 @@
...
+    ITransientDependency
+{
+    protected IPermissionManager PermissionManager { get; }
+
+    public UserDeletedEventHandler(IPermissionManager permissionManager)
+    {
+        PermissionManager = permissionManager;
+    }
+
+    public async Task HandleEventAsync(EntityDeletedEto<UserEto> eventData)
```
hikalkan: Since `PermissionManager` is a domain service (that doesn't manage UOW), it is better to make that method UOW (and virtual).

modules/setting-management/src/Volo.Abp.SettingManagement.Application/Volo/Abp/SettingManagement/UserDeletedEventHandler.cs
```diff
@@ -0,0 +1,25 @@
...
+    ITransientDependency
+{
+    protected ISettingManager SettingManager { get; }
+
+    public UserDeletedEventHandler(ISettingManager settingManager)
+    {
+        SettingManager = settingManager;
+    }
+
+    public async Task HandleEventAsync(EntityDeletedEto<UserEto> eventData)
```
hikalkan: We can make that method UOW too.

hikalkan (changes requested):

hikalkan (approved):

PR #18058, opened 2023-11-03
Add and update unit tests

### LINKED ISSUE #18054: Add and update unit tests for guards and resolvers
## Todo

Update existing tests and add new for guard and resolver functions

Sinan997:
Resolves #18054

### DIFF
--- a/npm/ng-packs/packages/core/src/lib/services/http-error-reporter.service.ts
+++ b/npm/ng-packs/packages/core/src/lib/services/http-error-reporter.service.ts
@@ -19,8 +19,8 @@ export class HttpErrorReporterService {
     return this._errors$.value;
   }
 
-  reportError = (error: HttpErrorResponse) => {
+  reportError(error: HttpErrorResponse) {
     this._reporter$.next(error);
     this._errors$.next([...this.errors, error]);
-  };
+  }
 }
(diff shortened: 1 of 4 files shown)


npm/ng-packs/packages/core/src/lib/services/http-error-reporter.service.ts
```diff
@@ -19,8 +19,8 @@ export class HttpErrorReporterService {
     return this._errors$.value;
   }
 
-  reportError = (error: HttpErrorResponse) => {
+  reportError(error: HttpErrorResponse) {
```
mahmut-gundogdu, line 22: make sure it does not cause any error.

mahmut-gundogdu (changes requested):

npm/ng-packs/packages/core/src/lib/services/http-error-reporter.service.ts
Sinan997: it works fine

mahmut-gundogdu (approved):

npm/ng-packs/packages/core/src/lib/tests/permission.guard.spec.ts
```diff
@@ -18,6 +29,10 @@ describe('PermissionGuard', () => {
   let httpErrorReporter: SpyObject<HttpErrorReporterService>;
   let permissionService: SpyObject<PermissionService>;
 
+  const mockOAuthService = {
+    isAuthenticated: true,
+  };
```
masum-ulu, line 34: Is that a true mocking method ? Can we provide another way for mocking a service ? It seems like duplicated in DummyComponent 🤔 ?

masum-ulu (approved):

PR #18064, opened 2023-11-06
Dynamic claims middleware

### LINKED ISSUE #18060: Dynamic Claims
Related to https://github.com/abpframework/abp/issues/8620

I will explain the requirements and some solution details in a brief.

The solution should work in monolith and distributed/microservice scenarios. Also, it should work with the most simple cookie based authentication as well as token based, OIDC and other authentication flows.

- [ ]  In the Account module, define `AbpDynamicClaimsOptions` with a `List<string> DynamicClaims`. Purpose: Do not make all claims dynamic, but make only the requested ones ([[example](https://github.com/abpframework/abp/blob/7f760d073badfb3c0d05e995b029c10cc14feb43/framework/src/Volo.Abp.AspNetCore/Volo/Abp/AspNetCore/Security/Claims/AbpDynamicClaimsOptions.cs)](https://github.com/abpframework/abp/blob/7f760d073badfb3c0d05e995b029c10cc14feb43/framework/src/Volo.Abp.AspNetCore/Volo/Abp/AspNetCore/Security/Claims/AbpDynamicClaimsOptions.cs)).
- [ ]  Once the user or client logins (or somehow authenticates), we can fill a distributed cache item with that user/client id. Get the configured dynamic claims from user's claims and fill the cache. You can create a simple contribution system, so a developer may want to define a service that contributed to the list in the cache.
- [ ]  The distributed cache basically stores a list of name-value pairs as dynamic claim values (note that claim names are not unique - a user may have more than one role for example).
- [ ]  These dynamic claim values in the distributed cache always overrides the values (or adds if not exists) in the user's/client's authentication ticket (e.g. claims in the HttpContext.User).
- [ ]  Define an ASP.NET Core middleware, say `app.UseDynamicClaims()`. It will be used after the authentication middleware, and will override the user's current claims from the distributed cache item. This is done in every microservice in a microservice solution, for example (basically, in every application that requires authentication/authorization).
- [ ]  What if no item in the cache? I mean the cache object is not available (the dynamic claim list in the can be empty, that's a valid value). In that case, we should call a service of the account module, so it can re-fill the cache. Then we continue and read from cache.
- [ ]  We have some important restrictions here:
  - [ ]  We don't want to directly depend on the account module from any other module. We can define an abstraction, say `IDynamicClaimRefreshService`.
    - [ ]  To not depend on the account module, we can define that and other abstractions and options in the related framework packages, or introduce new framework packages if we must (but as I see it won't be needed).
  - [ ]  The system should work with minimal or no configuration in any scenario. For example;
    - [ ]  In a monolith. `IDynamicClaimRefreshService` is implemented by the account module that works in-process, so it directly works - no configuration needed.
    - [ ]  In a microservice, `IDynamicClaimRefreshService` should be implemented as a remote HTTP API call to the account module.
      - [ ]  In that case, we don't want to double define the account module's remote service address. For example, for JWT and OICD, we already configure it as Authority. Somehow it should reuse that value.

maliming:
Resolves #18060

* Angular and blazor-wasm don't need to be changed because `api/abp/application` endpoint returns the correct claims.

app

tiered


hikalkan (approved):

rqx110:
I saw in no-layers app, usage as below:
```cs
app.UseUnitOfWork();
app.UseDynamicClaims();
app.UseAuthorization();
```
but in layers app, usage as below:
```cs
app.UseDynamicClaims();
app.UseUnitOfWork();
app.UseAuthorization();
```
There is different order of `UseUnitOfWork()` and `UseDynamicClaims()`, is this a problem?

PR #18066, opened 2023-11-06
Import OAuthService to solve build error

Sinan997:
image


npm/ng-packs/packages/core/src/lib/guards/permission.guard.ts
```diff
@@ -52,7 +50,7 @@ export const permissionGuard: CanActivateFn = (
 ) => {
   const router = inject(Router);
   const routesService = inject(RoutesService);
-  const oAuthService = inject(OAuthService);
+  const oAuthService = inject(AuthService);
```
masum-ulu: Variable name should be `authService`

masum-ulu (approved):

PR #18071, opened 2023-11-07
Create Native AOT Compilation in .NET 8

enisn:
### Description

Part of https://github.com/volosoft/vs-internal/issues/2998


braim23 (approved):

leonkosak:
Well, based on specific current limitations, it would be beneficial to write the documentation part for abp (and abp Commercial if there would be differences).
https://learn.microsoft.com/en-us/aspnet/core/fundamentals/native-aot
I would suggest a new section in documentation provided by my pull request: https://github.com/abpframework/abp/pull/18202

What do you think @maliming and @hikalkan?

PR #18083, opened 2023-11-08
AngularUI: Remove deprecated classes

Sinan997:


Sinan997:
Is it necessary to provide these guards/extensions, i didn't understand what's the point of providing.

What do you think @mahmut-gundogdu @masumulu28 ?

if providing is not necessary we can merge it, i am taking this to draft for now

masum-ulu:
Don't need to merge, it's enough the close PR without merge

PR #18149, opened 2023-11-15
Fix extensible table abpVisible directive bug

masum-ulu:
Import AbpVisibleDirective to extensible-table component as standalone


masum-ulu:
Moved to https://github.com/abpframework/abp/pull/18150

PR #18175, opened 2023-11-17
Enhance dynamic claims system.

maliming:
1. Handle the case where DynamicClaims is empty.
2. Automatically set `RemoteRefreshUrl` in the `AddAbpJwtBearer` and `AddAbpOpenIdConnect` methods.
3. The `Volo.Abp.AspNetCore.Authentication.JwtBearer` package contains the `WebRemoteDynamicClaimsPrincipalContributor` service suitable for a project that doesn't depend on Identity and `Volo.Abp.AspNetCore.Mvc.Client`.

5. Add `IsRemoteRefreshEnabled(bool true by default)` to `AbpClaimsPrincipalFactoryOptions`.
   Identity module will set it as `false`, then the `RemoteDynamicClaimsPrincipalContributor` and `WebRemoteDynamicClaimsPrincipalContributor` will not add to the DI.

We can rename the package and classes if you have better ideas.


framework/src/Volo.Abp.AspNetCore.Authentication.JwtBearer/Microsoft/Extensions/DependencyInjection/AbpOpenIdConnectExtensions.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System;
+using Microsoft.AspNetCore.Authentication;
+using Microsoft.AspNetCore.Authentication.JwtBearer;
+using Volo.Abp.Security.Claims;
+
+namespace Microsoft.Extensions.DependencyInjection;
+
+public static class AbpOpenIdConnectExtensions
```
rqx110: rename it？ also the file name

hikalkan (approved):

PR #18242, opened 2023-11-24
Add `IdentitySession` infrastructure.

maliming:


aldhamdy:
dose this capable of managing session for mobile app scenarios.
ex:
if user logged in from a new device we force him/her for 2FA and save unique id of the device.
next time user logged in from previously registered device, system does not require 2FA.
also need to allow user to only have 1 concurrent session.

Cloud please help how can we implement this strategy.

maliming:
This is a different business. The session will currently be designed to `Prevent concurrent login and manage user sessions`

aldhamdy:
> This is a different business. The session will currently be designed to `Prevent concurrent login and manage user sessions`

We can use the session history for the user to check if this user has previously succeeded session from the same device then we don't require 2FA.

I think the right module to implement this is the openidconnect, but we need the session entity to store the unique terminal/browser/mobile id.

after that we can edit the token handler in openidconnect to check for existing sessions for the same user & device-id so no need to force 2FA.
Also it will be linked to the same option of 2FA (remember browser).

realLiangshiwei (approved):

EngincanV (approved):

PR #18265, opened 2023-11-27
Update Application.md

minhtaile2712:
Make it easier to understand the docs

### Description

I re-write the paragraph so that it explains what comes with default setting first, then the pitfall comes with it, lastly what behavior will happen if we use more specific option ( `--tiered` flag in this case).


minhtaile2712:
Anyone here?

PR #18358, opened 2023-12-05
Fix module find condition, included module name bisde module path

masum-ulu:
### Description

This PR fixes if not find the module by path, it'll check by module name itself


masum-ulu:
Moved https://github.com/abpframework/abp/pull/18359

Issue #18405, opened 2023-12-07
Dynamic-layout keys moved Enum and User can add own layout.  


mahmut-gundogdu:
https://github.com/abpframework/abp/blob/9cc4cca701739d4cf9c62a9e87a89eaa91b169d5/npm/ng-packs/packages/core/src/lib/components/dynamic-layout.component.ts#L24

PR #18418, opened 2023-12-08
User can set dynamic layouts via the inject token

mahmut-gundogdu:
### Description

Resolves [#18405](https://github.com/abpframework/abp/issues/18405)

- We have a variable that has layout list. It moved an array. and some minor refactoring
-Extensible table's enum named `eThemeSharedCompents` it changed

### Checklist

### How to test it?


npm/ng-packs/packages/core/src/lib/core.module.ts
```diff
@@ -97,7 +99,7 @@ const standaloneDirectives = [
     ShortTimePipe,
     ShortDatePipe,
   ],
-  providers: [LocalizationPipe],
+  providers: [LocalizationPipe,],
```
masum-ulu: We need to remove `,`

masum-ulu (approved):

PR #18424, opened 2023-12-10
Update Background-Jobs-Hangfire.md

sephit:
Changing OnApplicationInitialization to OnPostApplicationInitialization to fix the IsAuthenticated=false issue if it's added in a module Based on this https://support.abp.io/QA/Questions/1769/Control-access-to-Hangfire-dashboard-with-abp-permission#answer-c9354699-19b1-ecd6-db66-39fe8f8afc6e


realLiangshiwei:
Hi,

This documentation assumes it's in the application template, not a separate module, Can you add a section to explain?
Thanks.

PR #18429, opened 2023-12-11
Grant all permissions checkbox does not affect Select all checkbox.

Sinan997:
[#15934](https://github.com/volosoft/volo/issues/15934)  Resolves
 
selectall


masum-ulu (approved):

gizemmutukurt:
- [x] If there is only one option and it is removed, it should be removed in select all.
image

- [x] If there are many selections and some have been removed, select all should be "-".
image

gizemmutukurt (changes requested):
It will be checked again when the problems are fixed.

gizemmutukurt:
The following situations need to be corrected.

image
image

gizemmutukurt (approved):

PR #18437, opened 2023-12-11
Auto API controllers docs integration, covering deprecation of Microsoft.AspNetCore.Mvc Nuget packages.

sanghel-orbyta:
### Description

Auto API controllers docs integration, covering deprecation of Microsoft.AspNetCore.Mvc Nuget packages.


sanghel-orbyta:
any particular reason this is being held up by?

PR #18589, opened 2023-12-26
fix user role update

jxnkwlp:
fix when user role update, the identity options without sync

### Description

when call `UserManager.UpdateUserAsync`, it will be call `ValidateUserAsync`, so if the identity options is not sync to correct value, it will be throw error. 

### Checklist

- [x] no need to document


maliming:
hi

Thanks. I will update this in `rel-8.0`. 👍

maliming:
https://github.com/abpframework/abp/pull/18590

PR #18605, opened 2023-12-27
Improve GetUserNameFromEmail method

salihozkara:
### Description

Added that the user name should be created according to accepted characters when creating the user name from the mail.

### Checklist


maliming:
https://github.com/abpframework/abp/pull/18612

PR #18638, opened 2023-12-30
Introduce the `DisableAbpFeatures`.

### LINKED ISSUE #18635: Provide a way for the controller to skip all interceptors and filters.
It's useful for an API that requires high concurrency.

maliming:
Resolve #18635

If you have some APIs, they are used frequently, but you don't need all the features of abp.

```cs
[Route("api/my-high-concurrency-api")]
[DisableAbpFeatures]
public class NoContentTestController : AbpController
{

}

```

```cs
[AttributeUsage(AttributeTargets.Class)]
public class DisableAbpFeaturesAttribute : Attribute
{
    /// <summary>
    /// The framework will not register any interceptors for the class.
    /// This will cause the all features that depend on interceptors to not work.
    /// </summary>
    public bool DisableInterceptors { get; set; } = true;

    /// <summary>
    /// The framework middleware will skip the class.
    /// This will cause the all features that depend on middleware to not work.
    /// </summary>
    public bool DisableMiddleware { get; set; } = true;

    /// <summary>
    /// The framework will not remove all built-in filters for the class.
    /// This will cause the all features that depend on filters to not work.
    /// </summary>
    public bool DisableMvcFilters { get; set; } = true;
}

```


ebicoglu:
when it's on, multi-tenancy, input validation, localization, transaction management (UOW)  and other features will be disabled?

maliming:
> when it's on, multi-tenancy, input validation, localization, transaction management (UOW) and other features will be disabled?

Yes, This is not commonly used, but if you have a controller that just has some simple logic, disabling these features will increase its performance.

ebicoglu:
- It is available for only classes. It may be good if it covers methods as well `AttributeUsage(AttributeTargets.Method)`.
- Is it different than deriving from ASP.NET Core's default `Controller`?

```
public class NoContentTestController : AbpController
public class NoContentTestController : Controller
```

maliming:
1. It cannot be disabled at the method level because interceptors and global filters target a class.
2. `AbpController` contains many commonly used services. There is no difference in other conveniences because its base class is `Controller`.

hikalkan (approved):

PR #18703, opened 2024-01-05
Update ABP IO Turkish Localization

honurbu:
Used In : https://github.com/volosoft/vs-internal/pull/3545


oykuermann (changes requested):
I'll continue when changes are done.
https://github.com/volosoft/vs-internal/pull/3545#issuecomment-1883192414

oykuermann (approved):

PR #18712, opened 2024-01-08
Update `MaxDefaultValueLength` to 2048.

maliming:
https://support.abp.io/QA/Questions/6461


modules/setting-management/src/Volo.Abp.SettingManagement.Domain.Shared/Volo/Abp/SettingManagement/SettingDefinitionRecordConsts.cs
```diff
@@ -8,7 +8,12 @@ public class SettingDefinitionRecordConsts
...
     public static int MaxDescriptionLength { get; set; } = 512;
 
-    public static int MaxDefaultValueLength { get; set; } = 256;
+    /// <summary>
+    /// Default value: 2048
+    /// </summary>
+    public static int MaxValueLength { get; set; } = 2048;
 
-    public static int MaxProvidersLength { get; set; } = 128;
+    public static int MaxDefaultValueLength { get; set; } = MaxValueLength;
```
rqx110: reference to `SettingConsts.MaxValueLength`, instead of creating new one?

realLiangshiwei (approved):

PR #18755, opened 2024-01-12
Update migration guide for v8.0 (add a note for sql server users)

EngincanV:


maliming (approved):

maliming:
I added the `Microsoft.Data.SqlClient.SqlException (0x80131904)`, In this way, developers may see it in search engines.

PR #18795, opened 2024-01-18
Support `Keyed Services` in `DependencyAttribute`

htve:


maliming:
hi

> Test failed for the solution: ../framework

We have a related https://github.com/abpframework/abp/pull/18792#event-11520780135 

It has made some changes to the keyed services.

Please merge your branch from `the` dev branch after merging.

htve:
Friendly Review: Is this a duplicate feature? No, it's not. Is this a PR that needs a different PR to work, and no, it's not either.

htve:
OMG, it looks like ABP.IO doesn't need contributions from contributors, even though this is an enhancement, and at the same time is suppressing unfriendly comments, which is a very scary thing.  

真听话, 你做的真棒, 看你怎么收场.

v2codes:
Recover something：
34a1c59e4dfa524f311809bf3bd3761

-------------------------------------------------
What do you think about this?

PR #18803, opened 2024-01-18
Add new methods to abstract class and oauth.service.ts

Sinan997:
### Description

Add helper methods to abstract authService to not depend on service which provided from oauth2-oidc package.


npm/ng-packs/packages/oauth/src/lib/services/oauth.service.ts
```diff
@@ -83,4 +83,20 @@ export class AbpOAuthService implements IAuthService {
...
+    return this.oAuthService.getRefreshToken();
+  }
+
+  getAccessToken(): string {
+    return this.oAuthService.getAccessToken();
+  }
+
+  refreshToken(): Promise<AbpAuthResponse> {
+    return this.oAuthService.refreshToken();
+  }
```
masum-ulu, line 97: Return type is not the same

image

masum-ulu (approved):

PR #18830, opened 2024-01-23
Fix: navigate url with params after redirect from auth server

### LINKED ISSUE #18804: Can not redirect successfully from auth server to angular app if uri include params
### The problem

Ref: https://support.abp.io/QA/Questions/6495

When attempting to access the link: **domain.com/**`my-link?queryParam=value` without authentication, the system redirects to the authentication server. However, when the authentication server returns with query parameters, the resolution process fails to complete successfully

### Sample

image
--
image
--
image

masum-ulu:
### Description

Resolves #18804

### Checklist

### How to test it?
- Create component and add to routes with path param
- Add `authGuard` to new route
- Try to go with params to new route

### Result
After redirect to angular app from auth server it must works fine

### DIFF
--- a/npm/ng-packs/packages/oauth/src/lib/strategies/auth-flow-strategy.ts
+++ b/npm/ng-packs/packages/oauth/src/lib/strategies/auth-flow-strategy.ts
@@ -99,10 +99,10 @@ export abstract class AuthFlowStrategy {
           filter(event => event.type === 'token_received' && !!this.oAuthService.state),
           take(1),
           map(() => {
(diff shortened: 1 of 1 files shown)


Sinan997 (approved):

JanneHarju:
This is such a small change/fix. Could you reconsider to fix this also to 7.3 or 7.4 versions. This is just bad UX for our customers that they are not getting where they were trying to navigate before login. Our quick and dirty fix was to add this to app.routing-module to avoid error message about no path match:
`{
    path: '**',
    redirectTo: 'dashboard',
  },`

PR #18865, opened 2024-01-25
Add missing Outline variation buttons

enisn:
### Description

Enhancements from https://github.com/volosoft/volo/pull/16499#issuecomment-1909922984

### Checklist

### How to test it?


oykuermann (changes requested):
I'll check after the changes: https://github.com/volosoft/volo/pull/16499#issuecomment-1911677051

oykuermann (approved):

PR #18874, opened 2024-01-27
Disable the `DynamicProxy` for `ConventionalControllers`.

### LINKED ISSUE #18845: A custom AuthorizationHandler or a custom PermissionValueProvider is executed twice when the service is set to a virtual method

### Description

img_v3_027c_4708e1b4-18c5-41f4-ae90-8a6c9fd2693g

### Reproduction Steps

_No response_

### Expected behavior

_No response_

### Actual behavior

_No response_

### Regression?

_No response_

### Known Workarounds

_No response_

### Version

6.0.3 OR 8.0.0

### User Interface

Common (Default)

### Database Provider

EF Core (Default)

None (Default)

### Operation System

Windows (Default)

### Other information

A project that is not ABP will not be executed twice

maliming:
From https://github.com/abpframework/abp/issues/18845#issuecomment-1909764887

Because it becomes an MVC controller, it has global filters. 

Resolve #18845

### DIFF
--- a/docs/en/API/Auto-API-Controllers.md
+++ b/docs/en/API/Auto-API-Controllers.md
@@ -12,9 +12,9 @@ Basic configuration is simple. Just configure `AbpAspNetCoreMvcOptions` and use 
 [DependsOn(BookStoreApplicationModule)]
 public class BookStoreWebModule : AbpModule
 {
-    public override void ConfigureServices(ServiceConfigurationContext context)
+    public override void PreConfigureServices(ServiceConfigurationContext context)
     {
-        Configure<AbpAspNetCoreMvcOptions>(options =>
+        PreConfigure<AbpAspNetCoreMvcOptions>(options =>
         {
             options
                 .ConventionalControllers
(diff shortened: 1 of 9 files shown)


hikalkan:
@maliming doesn't `AppliedCrossCuttingConcerns` system already solves the _duplication_ problem?
I am unsure about that change. For example, if we add another interceptor for appservices, we should always also add action filters. Otherwise, the interception logic won't work. Even if we do it for the framework, custom interceptors (added by the application developer) won't work.

maliming:
We don't apply interceptors for controllers.
https://github.com/abpframework/abp/blob/dev/framework/src/Volo.Abp.Core/Volo/Abp/DynamicProxy/DynamicProxyIgnoreTypes.cs#L7-L14

The `Authorization` middleware do a check. After that the `AuthorizationInterceptor` is checked again. 

I've only disabled the interceptor for the `application service controllers`. Because it doesn't inherit from `ControllerBase`

https://github.com/abpframework/abp/blob/dev/framework/src/Volo.Abp.AspNetCore.Mvc/Volo/Abp/AspNetCore/Mvc/AbpAspNetCoreMvcModule.cs#L67

hikalkan (approved):

hikalkan:
Thanks for the explanation.

PR #18970, opened 2024-02-08
Add ObjectHelper support for property resolution for nullable value types

namtab00:
### Description

I'm using an overriding AuditPropertySetter, that sets a `long CreatorId` (defined on new entity interfaces, not _IMust/MayHaveCreator_).

With this setup, ObjectHelper.TrySetProperty did not set that property, due to the member selector expression pointing to a value type.

This PR adds retro-compatible support for such an expression, with no breaking changes.

### Checklist

- [ x] I fully tested it as developer / designer and created unit / integration tests

### How to test it?

Created specific unit test.

### DIFF
--- a/framework/src/Volo.Abp.Core/Volo/Abp/ObjectHelper.cs
+++ b/framework/src/Volo.Abp.Core/Volo/Abp/ObjectHelper.cs
@@ -57,12 +57,17 @@ public static class ObjectHelper
                 return null;
             }
 
-
             var propertyInfo = obj?.GetType()
                 .GetProperties()
-                .FirstOrDefault(x => x.Name == memberExpression.Member.Name && x.GetSetMethod(true) != null);
+                .FirstOrDefault(x => x.Name == memberExpression.Member.Name);
 
             if (propertyInfo == null)
+            {
+                return null;
+            }
+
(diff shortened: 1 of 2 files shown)


namtab00:
... I'm not sure why the worflow failed, locally all tests pass...

image

framework/src/Volo.Abp.Core/Volo/Abp/ObjectHelper.cs
```diff
@@ -27,37 +26,53 @@ public static class ObjectHelper
...
+        PropertyInfo? PropertyFactory(string _)
         {
-            if (propertySelector.Body.NodeType != ExpressionType.MemberAccess)
+            MemberExpression? memberExpression;
+            switch (propertySelector.Body.NodeType)
+            {
+                case ExpressionType.Convert: {
+                    memberExpression = propertySelector.Body.As<UnaryExpression>().Operand as MemberExpression;
+                    break;
+                }
```
maliming, line 45: Can you share some test code to cover this case?

namtab00: I've integrated the test to cover the explicit case.

maliming (approved):

maliming:
Thanks @namtab00

PR #18978, opened 2024-02-09
Replace theme package in `package.json`

### LINKED ISSUE #18968: Selecting the "basic" theme results in missing styling for the AuthServer

### Description

I am trying to run ABP, so far it works for the most part just the AuthServer has no styling. Running .NET Core 8 on macOS Sonoma

I installed npm and yarn both have the correct versions, I also ran the "abp install-libs" command.

### Reproduction Steps

Create a new project with the CLI v8.0.2 and this command `abp new TT.Test -u blazor -dbms MySQL --separate-auth-server --theme basic --pwa -csf`

Using the getting started guide you would install node and yarn, then run `abp install-libs`.

### Expected behavior

AuthServer styling, no error ins console about missing lib files

### Actual behavior

Libs folder is empty therefore the AuthServer pages have no styling.

### Regression?

Unkown

### Known Workarounds

Add this line `"@abp/aspnetcore.mvc.ui.theme.leptonxlite": "~3.0.1"` to you dependencies in the AuthServer package.json. Then execute `abp install-libs` again.

### Version

8.0.2

### User Interface

Blazor

### Database Provider

EF Core (Default)

Separate Auth Server

### Operation System

macOS

### Other information

I already posted on StackOverflow about this, that is also where the solution came from. Thanks to [Manuel1234](https://stackoverflow.com/users/16974715/manuel1234) it is now working after adding the dependency.

Here is the link to the question:
https://stackoverflow.com/questions/77951382/abp-new-project-with-v8-cli-abp-install-libs-creates-empty-libs-folder

maliming:
Resolve #18968

### DIFF
--- a/framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/Commands/NewCommand.cs
+++ b/framework/src/Volo.Abp.Cli.Core/Volo/Abp/Cli/Commands/NewCommand.cs
@@ -104,7 +104,6 @@ public class NewCommand : ProjectCreationCommandBase, IConsoleCommand, ITransien
 
         await CheckCreatedRequirements(projectArgs);
 
-        ConfigureNpmPackagesForTheme(projectArgs);
         await CreateOpenIddictPfxFilesAsync(projectArgs);
         await RunGraphBuildForMicroserviceServiceTemplate(projectArgs);
         await CreateInitialMigrationsAsync(projectArgs);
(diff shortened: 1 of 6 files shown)


EngincanV (approved):

EngincanV (commented):
If we don't specify the version, it will be null and in that case, only the package name will be changed, not the version. For example, if I select basic theme, the CLI won't change the version and it will stay as `3.1.0`.

Can you please cover this scenerio? @maliming

maliming:
ok, I will test this case. Thanks.

EngincanV (approved):

PR #19046, opened 2024-02-15
Prevent request to app-configuration after language change

### LINKED ISSUE #18977: Prevent request to app-configuration after language change
## Description

Currently when we change the language, application sending 2 request
- application-configuration
- application-localization

It must only send `application-localization` endpoint

Sinan997:
Resolves #18977


npm/ng-packs/packages/core/src/lib/proxy/volo/abp/models.ts
```diff
@@ -1,5 +1,5 @@
-
-export interface NameValue<T = string> {
-  name?: string;
-  value: T;
-}
+
+export interface NameValue<T = "string"> {
```
masumulu28, line 2: Change to string

masum-ulu (approved):

PR #19079, opened 2024-02-20
Publish `EntityUpdatedEvent` when navigation changes.

### LINKED ISSUE #19060: Mark AggreateRoot as modified

According to [AggregateRoot Class](https://docs.abp.io/en/abp/latest/Entities#aggregateroot-class) I would expect that if I am working with Order as a single unit, the modification in the OrderLines collection will cause a change to Order at the same time. 

So if I add a new OrderLine to an Order.  I would expect that there will be a LastModifierId and LastModificationTime change on the Order.

I would also expect that
`ILocalEventHandler<EntityUpdatedEventData<Order>>` should be triggered

Would it be possible to add a MarkAsModified() method to BasicAggregateRoot? that would mark the entity as modified? So proper audition and event triggers be handled? Or with a repository.UpdateAsync() mark as changed even if there is no property change. 

### Additional context

image
image

maliming:
EF Core still has a small problem (https://github.com/dotnet/efcore/issues/24076), so I added [AbpEfCoreNavigationHelper](https://github.com/abpframework/abp/blob/PublishEventsForTrackedEntity/framework/src/Volo.Abp.EntityFrameworkCore/Volo/Abp/EntityFrameworkCore/ChangeTrackers/AbpEfCoreNavigationHelper.cs) to solve this problem.

Resolves #19060 #19266

### DIFF
--- a/framework/src/Volo.Abp.EntityFrameworkCore/Volo/Abp/EntityFrameworkCore/AbpDbContext.cs
+++ b/framework/src/Volo.Abp.EntityFrameworkCore/Volo/Abp/EntityFrameworkCore/AbpDbContext.cs
@@ -163,6 +163,7 @@ public abstract class AbpDbContext<TDbContext> : DbContext, IAbpEfCoreDbContext,
             List<EntityChangeInfo>? entityChangeList = null;
             if (auditLog != null)
             {
+                EntityHistoryHelper.InitializeNavigationHelper(AbpEfCoreNavigationHelper);
                 entityChangeList = EntityHistoryHelper.CreateChangeList(ChangeTracker.Entries().ToList());
             }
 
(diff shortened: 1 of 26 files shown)


framework/src/Volo.Abp.EntityFrameworkCore/Volo/Abp/EntityFrameworkCore/AbpDbContext.cs
```diff
@@ -336,6 +349,31 @@ private void PublishEventsForTrackedEntity(EntityEntry entry)
...
     }
 
+    protected virtual bool HasEntityEntryChanged(EntityEntry entry)
+    {
+        if (entry.State != EntityState.Modified && entry.State != EntityState.Unchanged)
+        {
+            return false;
+        }
+
+        var changed = entry.State == EntityState.Modified && entry.Properties.Any(x => x.IsModified && x.Metadata.ValueGenerated == ValueGenerated.Never);
```
Brandejs: Shouldn't this condition also include OnAdd? 
```
var changed = entry.State == EntityState.Modified && entry.Properties.Any(x => x.IsModified && (x.Metadata.ValueGenerated == ValueGenerated.Never || x.Metadata.ValueGenerated == ValueGenerated.OnAdd));
```

maliming: Yes, I will add it in another PR.

maliming: https://github.com/abpframework/abp/pull/19093

EngincanV (approved):

hikalkan (approved):

PR #19088, opened 2024-02-20
Add test to remember-me.service and fix return value in service

Sinan997:
Adding unit test

### DIFF
--- a/npm/ng-packs/packages/oauth/src/lib/services/remember-me.service.ts
+++ b/npm/ng-packs/packages/oauth/src/lib/services/remember-me.service.ts
@@ -17,7 +17,7 @@ export class RememberMeService {
   }
 
   get() {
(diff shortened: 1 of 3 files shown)


npm/ng-packs/packages/oauth/src/lib/services/remember-me.service.ts
```diff
@@ -17,7 +17,7 @@ export class RememberMeService {
   }
 
   get() {
-    return Boolean(JSON.parse(this.localStorageService.getItem(this.#rememberMe)));
+    return Boolean(JSON.parse(this.localStorageService.getItem(this.#rememberMe) || 'false'));
```
masum-ulu, line 20: We don't need to convert Boolean, parse method already casting FYI

masum-ulu (approved):

PR #19135, opened 2024-02-26
set default `className` if `group.name` doesn't defined in prop

Sinan997:
**`GroupedFormPropList`** class was grouping under the `undefined` object if `group.name` wasn't defined in prop.

This pr sets a default className for prop if `group.name` is undefined.

## Note

> Edited by @masumulu28 

We will try to manage input order (with and without group) `adding group property`, if it's not possible we'll accept this default behavior.

### DIFF
--- a/npm/ng-packs/packages/components/extensible/src/lib/models/form-props.ts
+++ b/npm/ng-packs/packages/components/extensible/src/lib/models/form-props.ts
@@ -29,6 +29,7 @@ export interface FormPropGroup {
 
 export class GroupedFormPropList<R = any> {
   public readonly items: GroupedFormPropItem[] = [];
+  count = 1;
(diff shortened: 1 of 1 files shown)


npm/ng-packs/packages/components/extensible/src/lib/models/form-props.ts
```diff
@@ -37,7 +38,7 @@ export class GroupedFormPropList<R = any> {
     } else {
       group = {
         formPropList: new FormPropList(),
-        group: item.group,
+        group: item.group || { name: `default${this.count++}`, className: item.group?.className },
```
masumulu28, line 41: We shouldn't need this behavior, try to run without default group object

masum-ulu (approved):

PR #19170, opened 2024-02-29
Add showOnlyIcon attribute to single action button

Sinan997:
When you add a an icon to single action button it add's an `me-1` class as a default.

When you want to add only an icon you dont need this margin.
Now on you can use showOnlyIcon attribute.


masum-ulu:
Can you please update also [document](https://docs.abp.io/en/abp/latest/UI/Angular/Data-Table-Column-Extensions)

masum-ulu (approved):

Issue #19167, opened 2024-02-29
When prop is invisible, it still adding an empty element to the DOM


Sinan997:
![image](https://github.com/abpframework/abp/assets/72804437/220159be-e621-429a-b92b-ccea31f82c09)

PR #19171, opened 2024-02-29
Remove extra element from DOM when prop is invisible

Sinan997:
Resolves https://github.com/abpframework/abp/issues/19167

### DIFF
--- a/npm/ng-packs/packages/components/extensible/src/lib/components/extensible-form/extensible-form.component.ts
+++ b/npm/ng-packs/packages/components/extensible/src/lib/components/extensible-form/extensible-form.component.ts
@@ -11,7 +11,7 @@ import {
 } from '@angular/core';
 import {ControlContainer, ReactiveFormsModule, UntypedFormGroup} from '@angular/forms';
 import {EXTRA_PROPERTIES_KEY} from '../../constants/extra-properties';
-import {FormPropList, GroupedFormPropList} from '../../models/form-props';
+import {FormProp, FormPropList, GroupedFormPropList} from '../../models/form-props';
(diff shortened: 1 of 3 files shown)


npm/ng-packs/packages/components/extensible/src/lib/components/extensible-form/extensible-form.component.ts
```diff
@@ -44,11 +44,13 @@ export class ExtensibleFormComponent<R = any> {
         const type = !record || JSON.stringify(record) === '{}' ? 'create' : 'edit';
         const propList = this.extensions[`${type}FormProps`].get(this.identifier).props;
         this.groupedPropList = this.createGroupedList(propList);
+        this.groupedPropListOfArray = this.groupedPropList.items.map(i => i.formPropList.toArray());
```
masum-ulu: Do we really need this ? we can already reach `formPropList` property in `groupedPropList` use this in `isAnyGroupMemberVisible` ?

Sinan997: `formPropList` is linked list, so i need to convert this to array so i can traverse it easly

masum-ulu: My problem not about casting, it's about the why you define new variable for this ? You using only in `isAnyGroupMemberVisible`

Sinan997: i understood what you mean, i will remove unnecesarry variable

masum-ulu (approved):

PR #19172, opened 2024-02-29
[INVALID] CMS: Dummy placeholders should be stored as embedded resources.

EngincanV:


EngincanV:
Moved to https://github.com/abpframework/abp/pull/19173

PR #19186, opened 2024-03-01
Remove header from Permissions modal

enisn:
### Description

Resolves https://github.com/volosoft/lepton/issues/2077

image

### Checklist

### How to test it?


gizemmutukurt (changes requested):
The change does not seem to be reflected

image

gizemmutukurt (approved):

PR #19199, opened 2024-03-04
Only `GetEntitiesAsync` on page 1.

maliming:


framework/src/Volo.Abp.BlazoriseUI/AbpCrudPageBase.cs
```diff
@@ -306,10 +306,12 @@ protected virtual Task UpdateGetListInputAsync()
...
     protected virtual async Task SearchEntitiesAsync()
     {
+        var currentPage = CurrentPage;
         CurrentPage = 1;
-
-        await GetEntitiesAsync();
-
+        if (currentPage == 1)
+        {
+            await GetEntitiesAsync();
```
ebicoglu, line 313: it will never run GetEntities if the current page <> 1. how it will fetch other pages?

Sinan997 (approved):

PR #19208, opened 2024-03-05
TenantStore list method added

ahmetfarukulu:


framework/src/Volo.Abp.AspNetCore.Components.MauiBlazor/Volo/Abp/AspNetCore/Components/MauiBlazor/MauiBlazorRemoteTenantStore.cs
```diff
@@ -57,6 +58,11 @@ public MauiBlazorRemoteTenantStore(AbpTenantClientProxy tenantAppService, IDistr
         return tenantConfiguration;
     }
 
+    public Task<IReadOnlyList<TenantConfiguration>> GetListAsync(bool includeDetails = false)
+    {
```
realLiangshiwei, line 62: It shouldn't return an empty list, otherwise there's no point in adding this method.

framework/src/Volo.Abp.AspNetCore.Mvc.Client/Volo/Abp/AspNetCore/Mvc/Client/MvcRemoteTenantStore.cs
```diff
@@ -79,6 +80,11 @@ public class MvcRemoteTenantStore : ITenantStore, ITransientDependency
         return tenantConfiguration?.Value;
     }
 
+    public Task<IReadOnlyList<TenantConfiguration>> GetListAsync(bool includeDetails = false)
+    {
```
realLiangshiwei, line 84: It shouldn't return an empty list, otherwise there's no point in adding this method.

framework/src/Volo.Abp.AspNetCore.Components.MauiBlazor/Volo/Abp/AspNetCore/Components/MauiBlazor/MauiBlazorRemoteTenantStore.cs
ahmetfarukulu: That would be a problem on the security side; otherwise, we should add permission check for the user to have tenant list access. Since all application uses the tenant management module [TenantStore](https://github.com/abpframework/abp/blob/8ba6796c06b46d21ff1cff5f54ed07cba607bb55/modules/tenant-management/src/Volo.Abp.TenantManagement.Domain/Volo/Abp/TenantManagement/TenantStore.cs#L11) excepts mvc client and maui blazor (which we don't need this method on client applications).

framework/src/Volo.Abp.AspNetCore.Mvc.Client/Volo/Abp/AspNetCore/Mvc/Client/MvcRemoteTenantStore.cs
ahmetfarukulu: That would be a problem on the security side; otherwise, we should add permission check for the user to have tenant list access. Since all application uses the tenant management module [TenantStore](https://github.com/abpframework/abp/blob/8ba6796c06b46d21ff1cff5f54ed07cba607bb55/modules/tenant-management/src/Volo.Abp.TenantManagement.Domain/Volo/Abp/TenantManagement/TenantStore.cs#L11) excepts mvc client and maui blazor (which we don't need this method on client applications).

realLiangshiwei (approved):

PR #19236, opened 2024-03-07
Allow extending ConfigureConventions and OnModelCreating

salihozkara:
### Description

Users will need to replace these methods on a per-DbContext basis when they require them. Instead of replacement, methods can be extended using options. In our case, we utilized citext to ensure case-insensitivity for strings in Postgres SQL. However, this change was effective only when applied to abp.io's own DbContext; even if other DbContexts used citext in the database, case sensitivity was not addressed. Subsequently, after replacing all DbContexts and configuring this change within each, our issue was resolved. Given the potential workload of replacing and maintaining all DbContexts, employing such a method would be more ideal.

Example usage:
```cs
Configure<AbpDbContextOptions>(options =>
{
    options.UseNpgsql();
    options.ConfigureConventions((builder, dbContext) =>
    {
        builder.Properties<string>().HaveColumnType("citext");
    });
});
```

### Checklist

### DIFF
--- a/framework/src/Volo.Abp.EntityFrameworkCore/Volo/Abp/EntityFrameworkCore/AbpDbContext.cs
+++ b/framework/src/Volo.Abp.EntityFrameworkCore/Volo/Abp/EntityFrameworkCore/AbpDbContext.cs
@@ -14,6 +14,7 @@ using Microsoft.EntityFrameworkCore.ChangeTracking;
 using Microsoft.EntityFrameworkCore.Metadata;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Logging.Abstractions;
+using Microsoft.Extensions.Options;
 using Volo.Abp.Auditing;
 using Volo.Abp.Data;
 using Volo.Abp.DependencyInjection;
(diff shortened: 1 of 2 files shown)


maliming:
hi @salihozkara 

I added a new commit. What do you think?

https://github.com/abpframework/abp/pull/19241
https://github.com/abpframework/abp/pull/19241/commits/98e64b896bbf9d824e86b3181e644934632d5a76

salihozkara:
> hi @salihozkara
> 
> I added a new commit. What do you think?
> 
> #19241 [98e64b8](https://github.com/abpframework/abp/commit/98e64b896bbf9d824e86b3181e644934632d5a76)

Hi @maliming 
I am contemplating whether we need to add multiple actions without specifying a DbContext. For instance, if we add an option to the PostgreSQL package for using citext and include these actions when the option is set to true, if the user changes the default, our added actions will be removed. I believe it should be possible to add multiple actions without specifying a DbContext.

maliming:
hi

> I am contemplating whether we need to add multiple actions without specifying a DbContext. 

The `DefaultConventionAction` and `DefaultOnModelCreatingAction` apply for all dbcontexts.

salihozkara:
> hi
> 
> > I am contemplating whether we need to add multiple actions without specifying a DbContext.
> 
> The `DefaultConventionAction` and `DefaultOnModelCreatingAction` apply for all dbcontexts.


The user will need to override it like this:
```cs
var oldAction = options.DefaultConventionAction;
options.DefaultConventionAction = (dbContext, builder) =>
  {
      oldAction?.Invoke(dbContext, builder);
      // my code
  };
```
If you find it suitable, it's not a problem for me. My other question is about the 'order' that was present before. Is there a specific reason for removing it?

maliming:
`DefaultConventionAction` It should be used in the template project, It is similar to `UseNpgsql/UseSqkServer`, so it should be called once in a start-up project.

The `actions` will be in the order in the `List<>`.

PR #19349, opened 2024-03-19
Update route path to kebab-case while adding module or source code

Sinan997:
In angular we are using kebab-case for route paths. 
So i have changed to `ToKebabCase()` instead of `ToLower()`.


masum-ulu:
Have you test it with a creating new project ? Can you add some output as image please ?

masum-ulu (approved):

PR #19371, opened 2024-03-21
fix(angular/components): Fix extensible table component to display rows when totalCount is 0 but items.length is greater than 0

### LINKED ISSUE #19370: abp-extensible-table does not display rows even if data exists
### Description

The `abp-extensible-table` component ( in `@abp/ng.components` lib) does not display rows when `totalCount` is `0` and `items.length` is greater than `0`.

image

I will implement a workaround to solve this issue in Angular.

yusuf-cirak:
### Description

Resolves #19370

- This pull request implements a workaround for a backend issue. Specifically, when the response payload returns a `totalCount` of 0, despite there being more than 0 `items` present.

### Checklist

### DIFF
--- a/npm/ng-packs/packages/components/extensible/src/lib/components/extensible-table/extensible-table.component.ts
+++ b/npm/ng-packs/packages/components/extensible/src/lib/components/extensible-table/extensible-table.component.ts
@@ -1,6 +1,6 @@
 import {
   ABP,
-  ConfigStateService, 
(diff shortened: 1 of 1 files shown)


masum-ulu:
This solution could be wrong behavior, totalCount can be more than client side data, in that way we show to user missin total count don't need that, backend will be handle this problem I'm closing this

PR #19380, opened 2024-03-21
Docs: Improve clear cache

salihozkara:
### Description

when I pulled all the documents with all the information, there was a memory shortage. The memory and speed problem was eliminated with less data.

### Checklist

### DIFF
--- a/modules/docs/src/Volo.Docs.Domain/Volo/Docs/Documents/IDocumentRepository.cs
+++ b/modules/docs/src/Volo.Docs.Domain/Volo/Docs/Documents/IDocumentRepository.cs
@@ -13,7 +13,8 @@ namespace Volo.Docs.Documents
 
         Task<List<Document>> GetListByProjectId(Guid projectId, CancellationToken cancellationToken = default);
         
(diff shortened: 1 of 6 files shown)


maliming (approved):

modules/docs/src/Volo.Docs.Domain/Volo/Docs/Documents/IDocumentRepository.cs
```diff
@@ -12,6 +12,8 @@ public interface IDocumentRepository : IBasicRepository<Document>
         Task<List<DocumentInfo>> GetUniqueListDocumentInfoAsync(CancellationToken cancellationToken = default);
 
         Task<List<Document>> GetListByProjectId(Guid projectId, CancellationToken cancellationToken = default);
+        
+        Task ClearCachesAsync(Guid projectId, CancellationToken cancellationToken = default);
```
maliming: `ClearCache` is like a business method. The `Repository` should have no business logic. 
Consider: `UpdateProjectLastCachedTime(Guid projected, DateTime? cachedTime)`


We can make the `LastCachedTime` of `Entity` nullable if necessary.

PR #19394, opened 2024-03-22
`EnableTracking` if the `entity` has `ObjectExtensionInfo`.

maliming:
The [FillExtraPropertiesForTrackedEntities](https://github.com/abpframework/abp/blob/961dfbae37d54013d97eb89a985238d9ab7e669f/framework/src/Volo.Abp.EntityFrameworkCore/Volo/Abp/EntityFrameworkCore/AbpDbContext.cs#L284-L293) method of `AbpDbContext` required `Change Tracking.`


hikalkan:
Thanks. Can we indicate that in the object extension documentation?

maliming:
The document has been updated.

https://github.com/abpframework/abp/pull/19394/commits/a5de48116db156aa9d9ac601eea4ca4cb1c584b8

hikalkan:
The document write that they should manually change the object tracking behavior. However, if I didn't misunderstand the code, it automatically enables it. Which one is true?

maliming:
There are two parts:

1. **EntityCacheBase**:
If the entity has `extension properties`, we will `EnableTracking(IReadOnlyRepository)` in `EntityCacheBase`.

2. **Repository**:
If an entity has `extension properties(MapEfCoreProperty),` the developer can't use the read-only repository.

hikalkan (approved):

hikalkan:
Thank you for the explanation. It is clear for me now.

PR #19410, opened 2024-03-25
Blazor Web App - Persist Style & Scripts

enisn:
### Description

**before:**
not-persisted-style

**after:**
persisted-style

### Checklist

### How to test it?

### DIFF
--- a/framework/src/Volo.Abp.AspNetCore.Components.Web.Theming/Bundling/AbpScripts.razor
+++ b/framework/src/Volo.Abp.AspNetCore.Components.Web.Theming/Bundling/AbpScripts.razor
@@ -1,5 +1,7 @@
 @using Volo.Abp
+@implements IDisposable
 @inject IComponentBundleManager BundleManager
+@inject PersistentComponentState PersistentComponentState
 
 @if (ScriptFiles != null)
 {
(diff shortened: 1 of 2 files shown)


maliming (approved):

maliming:
It seems the mvc and blazor js&css both exist on wasm mode after this PR.

This is a problem. @enisn  what do you think?

image
image

maliming:
We will switch the CSS files for the Blazor server and WASM. This is the reason the page will **flash** once(Like image 1).

**style-initializer.js**
```
loadCssFile("_content/Volo.Abp.AspNetCore.Components.Web.LeptonXLiteTheme/side-menu/css/abp-bundle.css");
loadCssFile("_content/Volo.Abp.AspNetCore.Components.Web.LeptonXLiteTheme/side-menu/css/blazor-bundle.css");
loadCssFile("_content/Volo.Abp.AspNetCore.Components.Web.LeptonXLiteTheme/side-menu/css/bootstrap-dim.css");
loadCssFile('_content/Volo.Abp.AspNetCore.Components.Web.LeptonXLiteTheme/side-menu/css/layout-bundle.css');
loadCssFile('_content/Volo.Abp.AspNetCore.Components.Web.LeptonXLiteTheme/side-menu/css/font-bundle.css');
```

We can add one common lepton CSS file for the server and WASM in the `App.razor` component. This will prevent the flash problem.

enisn:
> It seems the mvc and blazor js&css both exist on wasm mode after this PR.
> 
> This is a problem. @enisn what do you think?
> 
> image image

When we remove mvc & blazor server css and add blazor css, it cause a flicker effect. Since we remove and reload entire CSSs. In the lepton, there are no difference in CSS files. No need to remove and reload all of them.


> We can add one common lepton CSS file for the server and WASM in the `App.razor` component. This will prevent the flash problem.

That might be a better solution 👍

maliming:
> That might be a better solution 👍

OK, Can you do this in LeptonX and Lite?

maliming:
https://github.com/abpframework/abp/issues/19970

PR #19418, opened 2024-03-26
Blogging module improvements and add single blog mode

salihozkara:
### Description

Added option to remove the blog name from the url when a single blog is used. And the members page has been moved under the Blogs folder to avoid conflicts with the user's own members folder when using the module.

### Checklist

### How to test it?

Use the Single Mode option.
Example usage
```cs
Configure<BloggingUrlOptions>(options =>
{
    options.RoutePrefix = "blog";
    options.SingleBlogMode.Enable = true;
    options.SingleBlogMode.BlogName = "abp";
});
```


modules/blogging/src/Volo.Blogging.Web/BloggingUrlOptions.cs
```diff
@@ -31,4 +34,15 @@ private string GetFormattedRoutePrefix()
...
         }
     }
+
+    public class SingleBlogModeOptions
+    {
+        /// <summary>
+        /// Determines whether to enable single blog mode by removing the blog name from the routing.
+        /// When enabled, only a single blog is allowed within the module.
+        /// </summary>
+        public bool Enable { get; set; }
```
EngincanV:
```suggestion
        public bool Enabled { get; set; }
```

modules/blogging/src/Volo.Blogging.Web/Pages/Blogs/BloggingPageModel.cs
```diff
@@ -8,5 +12,26 @@ public BloggingPageModel()
...
+                return await blogAppService.GetByShortNameAsync(blogOptions.SingleBlogMode.BlogName);
+            }
+
+            var blogs = await blogAppService.GetListAsync();
+            if (blogs.Items.Count == 1)
+            {
+                return blogs.Items[0];
+            }
+            
+            return await blogAppService.GetByShortNameAsync(blogOptions.SingleBlogMode.BlogName);
```
EngincanV: Here, I think we don't need to get all blog posts. Instead, you can get a single blog by creationTime, or throw entity not found exception if there is not any blog in the system. Because, in line 34, we know the **blogOptions.SingleBlogMode.BlogName** is null, so no need to make a call for that.

modules/blogging/src/Volo.Blogging.Web/Pages/Blogs/Posts/Index.cshtml.cs
```diff
@@ -47,7 +50,8 @@ public virtual async Task<ActionResult> OnGetAsync()
 
             try
             {
-                Blog = await _blogAppService.GetByShortNameAsync(BlogShortName);
+                Blog = await GetBlogAsync(_blogAppService, BlogOptions, BlogShortName);
+                BlogShortName = Blog.ShortName;
             }
             catch (EntityNotFoundException)
             {
```
EngincanV: There is a possible redirection loop problem. Consider, if we specify the **SingleBlogMode** as **true** but forget to specify the **BlogName** for it, then here it will always throw *EntityNotFoundException* and therefore it will redirect to the **/Blogs/Index** page. And on that page, it redirects to the [Posts/Index page](https://github.com/abpframework/abp/pull/19418/files#diff-cc4fce5479bec487ebf801ef9314c7fe10e160b603eaf3d6b0de5866992f87e4R26-R30) if the single blog mode is enabled.

EngincanV (changes requested):

EngincanV (approved):

oykuermann (changes requested):
https://localhost:40017/Blogging/Admin/Blogs link cannot be opened.

oykuermann:
Specified blog page cannot be opened:

image
image
image

oykuermann:
When you try to go blog from the user page:

[04.04.2024 17_49.webm](https://github.com/abpframework/abp/assets/142304401/b1c9494d-408c-43df-ae5e-ed6e670534eb)

oykuermann (approved):

PR #19440, opened 2024-03-27
Add disable filtering with id in AbstractTreeService, delete function

### LINKED ISSUE #19359: Angular: `ABP.Route` items shows only once when has same name

### Description

I want to add the same menu item with **different** `parentName` properties, but only the last item appears, the above is not. When I change its name, it appears as well. I want to have the same names.

### Reproduction Steps

Try this config:
```typescript
[
  {
    path: '/',
    name: 'Parent',
    iconClass: 'fas fa-star',
    order: 2,
  },
  {
    path: '/test-url',
    name: 'TestUrl',
    iconClass: 'fas fa-star',
    order: 3,
    parentName: 'Parent',
  },
  {
    path: '/test-url',
    name: 'TestUrl',
    iconClass: 'fas fa-star',
    order: 4,
  }
]
```

You will see only the second will be rendered.

This will not also work:
```typescript
{
  path: '/',
  name: 'Parent',
  iconClass: 'fas fa-star',
  order: 2,
},
{
  path: '/test-url',
  name: 'TestUrl',
  iconClass: 'fas fa-star',
  order: 3,
  parentName: 'Parent',
},
{
  path: '/different-url',
  name: 'TestUrl',
  iconClass: 'fas fa-star',
  order: 4,
},
```

### Expected behavior

I should see every item I added to the config. Even if they duplicate, I want it, or I can remove it. 

Screenshot 2024-03-20 at 13 32 23

### Actual behavior

`ABP.Route` items with the same `name` property do not appear.

Screenshot 2024-03-20 at 13 32 37

### Version

8.0.5

mustafapsd:
### Description

Resolves #19359

I added three new functions and a new property:

- `private filterRoutesEnabled: boolean`
This property is added for managing the filtering.

- `disableFiltering(): void`
This method disables the filter by setting `filterRoutesEnabled` to `false`.

- `enableFiltering(): void`
This method enables the filter by setting `filterRoutesEnabled` to `true`.

- `delete(params: Partial<T>): T[]`
This method deletes the routes that match the given properties. It returns the current state of the `flat`.

I modified the `add` method:
- If the `filterRoutesEnabled` is `true`, the algorithm is exactly the same as before.
- If the `filterRoutesEnabled` is `false`, I add the all items without any filtering.

### Checklist

### How to test it?
Firstly you should check this PR does not affect the current functionality. Then you can test the new features by following the steps below:
- Add routes with same `name` properties.
- To test it with a more realistic scenario, add routes with the different `parentName` properties.
- If you called the `disableFiltering` method before adding the routes, you should see that the routes are added without any filtering.
- If you called the `enableFiltering` method before adding the routes or you have not called the `disableFiltering` method, you should see that the routes has the same `name` properties are filtered.
- To test the new `delete` method, you can delete the routes that match the given properties.
- It should delete the routes and its children.

Expected results:

Screenshot 2024-03-20 at 15 04 38

### DIFF
--- a/npm/ng-packs/packages/core/src/lib/services/routes.service.ts
+++ b/npm/ng-packs/packages/core/src/lib/services/routes.service.ts
@@ -26,6 +26,7 @@ export abstract class AbstractTreeService<T extends { [key: string | number | sy
   private _visible$ = new BehaviorSubject<TreeNode<T>[]>([]);
 
   protected othersGroup: string;
+  private filterRoutesEnabled = true;
 
   get flat(): T[] {
     return this._flat$.value;
@@ -90,16 +91,54 @@ export abstract class AbstractTreeService<T extends { [key: string | number | sy
   }
 
   add(items: T[]): T[] {
-    const map = new Map<string, T>();
-    items.forEach(item => map.set(item[this.id], item));
+    if (this.filterRoutesEnabled) {
+      const map = new Map<string, T>();
+      items.forEach(item => map.set(item[this.id], item));
 
-    const flatItems = this.filterWith(map);
-    map.forEach(pushValueTo(flatItems));
+      const flatItems = this.filterWith(map);
+      map.forEach(pushValueTo(flatItems));
 
-    flatItems.sort(this.sort);
-    const visibleItems = flatItems.filter(item => !this.hide(item));
+      flatItems.sort(this.sort);
+      const visibleItems = flatItems.filter(item => !this.hide(item));
+      return this.publish(flatItems, visibleItems);
+    } else {
+      const flatItems = this.flat.concat(items);
+      flatItems.sort(this.sort);
+      const visibleItems = flatItems.filter(item => !this.hide(item));
+      return this.publish(flatItems, visibleItems);
+    }
+  }
 
-    return this.publish(flatItems, visibleItems);
+  delete(params: Partial<T>): T[] {
+    const willRemoveItems = this.flat.filter(item => {
+      const keys = Object.keys(params) as Array<keyof Partial<T>>;
+      const isValid = keys.every(key => item[key] === params[key]);
+
+      return isValid;
+    });
+
+    if (willRemoveItems?.length) {
+      willRemoveItems.forEach(item => {
+        this.delete({
+          [this.parentId]: item[this.id],
+        } as Partial<T>);
+      });
+
+      const flatItems = this.flat.filter(item => !willRemoveItems.includes(item));
+      const visibleItems = flatItems.filter(item => !this.hide(item));
+
+      return this.publish(flatItems, visibleItems);
+    }
+
+    return this.flat;
+  }
+
+  disableFiltering(): void {
+    this.filterRoutesEnabled = false;
+  }
+
+  enableFiltering(): void {
+    this.filterRoutesEnabled = true;
   }
 
   find(predicate: (item: TreeNode<T>) => boolean, tree = this.tree): TreeNode<T> | null {
@@ -145,8 +184,8 @@ export abstract class AbstractTreeService<T extends { [key: string | number | sy
         acc
           ? acc
           : searchKeys.every(key => node[key] === params[key])
-          ? node
-          : this.search(params, node.children),
+            ? node
+            : this.search(params, node.children),
       null,
     );
   }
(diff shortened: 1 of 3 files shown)


masum-ulu:
Hi thanks for your contribution, I'll check and give you to feedback

masum-ulu:
Hi again @mustafapsd, I've reviewed your codes and I've create some notes for you 🙂

# Change requests
- I think we can use `shouldSingularizeRoutes` variable name instead `filterRoutesEnabled` It could be more understandable.
  -  Also you can mark as `protected` variable
- We don't need to create 2 methods for `enable | disable` status instead we can create code section like below under the `search` method
```js
  search(params: Partial<T>, tree = this.tree): TreeNode<T> | null {
    //Search method content...
  }
 
  // set `singularize` parameter value `true`
  setSingularizeStatus(singularize = true): void {
    this.shouldSingularizeRoutes = singularize;
  }
```
- I've made some improvements from your codes for  `add` methods, can you please apply to `add` method code below
```js
 add(items: T[]): T[] {
    let flatItems: T[] = [];

    if (!this.shouldSingularizeRoutes) {
      flatItems = [...this.flat, ...items];
    }

    if (this.shouldSingularizeRoutes) {
      const map = new Map<string, T>();
      items.forEach(item => map.set(item[this.id], item));
      flatItems = this.filterWith(map);
      map.forEach(pushValueTo(flatItems));
    }

    flatItems.sort(this.sort);

    return this.publish(flatItems);
  }
```
- I've also made some improvments from your codes for `delete` method, can you also apply it ?
  - We can use `removeByParam` method name instead `delete` because `remove` method already exits in class, developer might be confused about it, don't need to make it different
  - You can put code below under the `remove` method in the class
```js
  remove(identifiers: string[]): T[] {
    //Remove method content..
  }

  removeByParam(params: Partial<T>): T[] | null {
    if (!params) {
      return null;
    }

    const keys = Object.keys(params) as Array<keyof Partial<T>>;
    if (keys.length === 0) {
      return null;
    }

    const excludedList = this.flat.filter(item => keys.every(key => item[key] === params[key]));
    if (!excludedList?.length) {
      return null;
    }

    for (const item of excludedList) {
      this.removeByParam({ [this.parentId]: item[this.id] } as Partial<T>);
    }

    const flatItems = this.flat.filter(item => !excludedList.includes(item));
    return this.publish(flatItems);
  }
```

These changes are won't effect your logic, which means everything will be works as excepted, I've just try to make codes more readable and you just need to update the documentation for usage and also need to update your tests for new version. Can you please send re-request for review after you apply these changes.

npm/ng-packs/packages/core/src/lib/services/routes.service.ts
```diff
@@ -171,15 +183,17 @@ export abstract class AbstractTreeService<T extends { [key: string | number | sy
...
-            ? node
-            : this.search(params, node.children),
-      null,
-    );
+    return tree.reduce<TreeNode<T> | null>((acc, node) => {
+      if (acc) {
+        return acc;
+      }
+
+      if (searchKeys.every(key => node.item[key] === params[key])) {
```
mustafapsd: `node.item[key]`is broke the tests. The old version `node[key]` is working. I'm including this change in my latest commit.

mustafapsd:
I also did a small naming change, `removeByParam` to `removeByParams`. This function gets more than 1 property in the object. If you think your suggestion is more sensible I can change it anyway.

masum-ulu:
> I also did a small naming change, `removeByParam` to `removeByParams`. This function gets more than 1 property in the object. If you think your suggestion is more sensible I can change it anyway.

Thanks for your effort, It might evoke to devs the method can get more than one args like `methodName(arg1, arg2, arg3 ...)`, yet it have only one arg and this is object. I think it should be `removeByParam`

npm/ng-packs/packages/core/src/lib/tests/routes.service.spec.ts
```diff
@@ -172,6 +172,64 @@ describe('Routes Service', () => {
...
+
+      expect(flat.length).toBe(2);
+    });
+
+    it('should not allow to duplicate routes when called with true', () => {
+      service.setSingularizeStatus(false);
+
+      service.add(routes);
+
+      service.setSingularizeStatus(true);
```
masum-ulu, line 214: Method's argument's default value is `true` we don't need to pass `true`

mustafapsd: Yes, but this way is more readable. If you know the method definition, it can look unnecessary, but otherwise, dev can think like "setting the singularize status to what?" 😄

masum-ulu (approved):

masum-ulu:
Thanks again for your contribution @mustafapsd 🙌

mustafapsd:
> Thanks again for your contribution @mustafapsd 🙌

Thank you for your effort, have a nice day 🙌

PR #19459, opened 2024-04-01
Add `lastNavigation` prop to `RouterEvents` service for track last successful navigation event

masum-ulu:
We can use the `RouterEventsService` for the handle routing events

### DIFF
--- a/npm/ng-packs/packages/core/src/lib/services/router-events.service.ts
+++ b/npm/ng-packs/packages/core/src/lib/services/router-events.service.ts
@@ -1,4 +1,4 @@
-import { Injectable, Type } from '@angular/core';
+import { Injectable, Type, inject, signal } from '@angular/core';
 import {
   NavigationCancel,
   NavigationEnd,
@@ -6,7 +6,8 @@ import {
   NavigationStart,
   Router,
   RouterEvent,
-  Event
+  Event,
+  RouterState,
 } from '@angular/router';
 import { filter } from 'rxjs/operators';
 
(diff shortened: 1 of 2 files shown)


npm/ng-packs/packages/core/src/lib/services/router-events.service.ts
```diff
@@ -19,12 +20,32 @@ export const NavigationEvent = {
...
+    this.listenToNavigation();
+  }
+
+  protected listenToNavigation(): void {
+    this.router.events.pipe(filter(e => e instanceof NavigationEvent.End)).subscribe(() => {
+      // It must be "NavigationTransition" but it is not exported in Angular
+      //https://github.com/angular/angular/blob/9c486c96827a9282cbdbff176761bc95554a260b/packages/router/src/navigation_transition.ts#L282
+      const lastNavigation = this.router.lastSuccessfulNavigation as unknown as any;
+      const curr = lastNavigation.targetRouterState as RouterState;
+      const currUrl = curr.snapshot.url;
```
yusuf-cirak: When this observable emits new value, i expected to have lastSuccessfulNavigation as the previous navigation URL. But it doesn't work like that. It returns the current URL which we can access it from the `e` too.

This solution is not working when navigating back.

npm/ng-packs/packages/core/src/lib/services/router-events.service.ts
```diff
@@ -22,25 +23,27 @@ export const NavigationEvent = {
...
-      const curr = lastNavigation.targetRouterState as RouterState;
-      const currUrl = curr.snapshot.url;
-
-      //Todo: improve this logic. Maybe we can handle error status in a better way ?
-      if (!currUrl.includes('error')) {
-        this.#lastNavigation.set(currUrl);
-      }
+
+      this.#previousNavigation.set(this.#currentNavigation());
+      this.#currentNavigation.set(event.url);
```
yusuf-cirak, line 46: Instead of using slices and indexes, we can track for current and previousNavigation in their own states. I think this is the best solution we can provide.

yusuf-cirak (approved):

PR #19489, opened 2024-04-03
Add `LocalStorageListenerService` to logout from other tabs

### LINKED ISSUE #19488: Feature: Log out from other tabs automatically when user logout from the current tab
- [x] Angular
- [x] Blazor
- [x] MVC

Sinan997:
Resolves #19488

### DIFF
--- a/npm/ng-packs/packages/core/src/lib/services/local-storage-listener.service.ts
+++ b/npm/ng-packs/packages/core/src/lib/services/local-storage-listener.service.ts
@@ -1,18 +1,16 @@
 import { DOCUMENT } from '@angular/common';
 import { Injectable, inject } from '@angular/core';
-import { OAuthService } from 'angular-oauth2-oidc';
 
 @Injectable({
   providedIn: 'root',
 })
 export class LocalStorageListenerService {
   protected readonly window = inject(DOCUMENT).defaultView;
-  protected readonly authService = inject(OAuthService);
 
   constructor() {
     this.window.addEventListener('storage', event => {
       if (event.key === 'access_token' && event.newValue === null) {
-        this.authService.logOut();
+        this.window.location.reload();
       }
     });
   }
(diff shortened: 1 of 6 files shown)


masum-ulu:
Can we move this service to core package 🤔 ?

Sinan997:
> Can we move this service to core package 🤔 ?

I am not sure about it, do we really want to depend on `"angular-oauth2-oidc"` package in `core` package?

masum-ulu:
> > Can we move this service to core package 🤔 ?
> 
> I am not sure about it, do we really want to depend on `"angular-oauth2-oidc"` package in `core` package?

:') we don't need to depend it, we can use IAuthService abstraction class right ? App project will create instance with oauth module configuration

npm/ng-packs/packages/core/src/lib/services/local-storage-listener.service.ts
```diff
@@ -1,18 +1,16 @@
...
 })
 export class LocalStorageListenerService {
   protected readonly window = inject(DOCUMENT).defaultView;
-  protected readonly authService = inject(OAuthService);
 
   constructor() {
     this.window.addEventListener('storage', event => {
       if (event.key === 'access_token' && event.newValue === null) {
-        this.authService.logOut();
+        this.window.location.reload();
```
masum-ulu, line 13: We've AuthService in core package try to use it please

Sinan997: I have tried it doesn't work as expected.

Reloading other tabs is giving us what we want to achieve. So no need to use `AuthService`

masumulu28:
Maybe we can use [AbpLocalStorageService](https://github.com/abpframework/abp/blob/dev/npm/ng-packs/packages/core/src/lib/services/local-storage.service.ts)

masum-ulu (approved):

oykuermann (approved):

PR #19650, opened 2024-04-29
logout if user is unauthorized

Sinan997:
If user is unauthorized then logout the user, just like other ui's.


npm/ng-packs/packages/theme-shared/src/lib/services/status-code-error-handler.service.ts
```diff
@@ -90,8 +90,9 @@ export class StatusCodeErrorHandlerService implements CustomHttpErrorHandlerServ
 
         this.showConfirmation(title, message).subscribe();
         break;
-
       case 403:
+        this.authService.logout();
```
masum-ulu: But it might cause logout user, even unpermitted specific path e.g:

### Having permissions
- Role

### Scenario
Try to go `identity/users`

### Result
It'll logout just because of you haven't users permissions.

npm/ng-packs/packages/theme-shared/src/lib/services/session-error-handler.service.ts
```diff
@@ -0,0 +1,30 @@
...
+import { CUSTOM_HTTP_ERROR_HANDLER_PRIORITY } from '../constants/default-errors';
+
+@Injectable({ providedIn: 'root' })
+export class AbpSessionErrorHandlerService implements CustomHttpErrorHandlerService {
+  readonly priority = CUSTOM_HTTP_ERROR_HANDLER_PRIORITY.high;
+  private authService = inject(AuthService);
+
+  canHandle(error: unknown): boolean {
+    console.log(error);
+    debugger;
```
masumulu28, line 16: Don't forget the remove these

masum-ulu (approved):

PR #19651, opened 2024-04-29
Angular: extensible table action list dropdown problem

sumeyyeKurtulus:
### Description

Resolves the action list dropdown problem as can be seen in the record:

[4_30_2024, 10_17_04 AM - Screen - Untitled video.webm](https://github.com/abpframework/abp/assets/92928815/fb297cc3-09af-4bb3-b968-1ef1c4f32144)

### Checklist

- [X] I fully tested it as developer 
- [X] No need to document

### How to test it?

Here is to steps to reproduce the related test case
1. Create a user under identity management
2. Then only give "Login With This User Permission" 
3. Once you login with its credentials, the actions button should not be visible since there will be no related action

We have previously discussed about the UI behavior with @masumulu28, and it is open to updates after further considerations.

### DIFF
--- a/npm/ng-packs/packages/components/extensible/src/lib/components/extensible-table/extensible-table.component.html
+++ b/npm/ng-packs/packages/components/extensible/src/lib/components/extensible-table/extensible-table.component.html
@@ -6,62 +6,68 @@
   (activate)="tableActivate.emit($event)"
 >
   @if (actionsTemplate || (actionList.length && hasAtLeastOnePermittedAction)) {
-  <ngx-datatable-column
-    [name]="actionsText | abpLocalization"
-    [maxWidth]="columnWidths[0]"
-    [width]="columnWidths[0]"
-    [sortable]="false"
-  >
-    <ng-template let-row="row" let-i="rowIndex" ngx-datatable-cell-template>
-      <ng-container
-        *ngTemplateOutlet="actionsTemplate || gridActions; context: { $implicit: row, index: i }"
-      ></ng-container>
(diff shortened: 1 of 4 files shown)


npm/ng-packs/packages/components/extensible/src/lib/components/extensible-table/extensible-table.component.ts
```diff
@@ -205,4 +205,9 @@ export class ExtensibleTableComponent<R = any> implements OnChanges {
       return record;
     });
   }
+
+  hasAvailableActions(index, row): boolean {
```
masumulu28, line 209: You can remove index also from don't need to send from template.

npm/ng-packs/packages/components/extensible/src/lib/components/extensible-table/extensible-table.component.html
```diff
@@ -6,62 +6,68 @@
...
+      [maxWidth]="columnWidths[0]"
+      [width]="columnWidths[0]"
+      [sortable]="false"
+    >
+      <ng-template let-row="row" let-i="rowIndex" ngx-datatable-cell-template>
+        <ng-container
+          *ngTemplateOutlet="actionsTemplate || gridActions; context: { $implicit: row, index: i }"
+        ></ng-container>
+        <ng-template #gridActions>
+          @if (hasAvailableActions(i, row)) {
```
masumulu28, line 20: We don't need to pass **i**

masum-ulu (approved):

PR #19700, opened 2024-05-03
enhancement(angular/core): Refactoring localization service and SORT_COMPARE_FUNC

### LINKED ISSUE #19699: angular(core): Localization and sort compare function improvements
### Description

Prevent console warning when key not exits in localization method

yusuf-cirak:
### Description

Resolves #19699

- Improvement in localization service by laveraging early return approach.

- Refactoring SORT_COMPARE_FUNC by changing ordering property orders.

- Others are just code formattings.

### Checklist

### How to test it?

- Localization should be workig correctly in all pages.

### DIFF
--- a/npm/ng-packs/packages/core/src/lib/services/localization.service.ts
+++ b/npm/ng-packs/packages/core/src/lib/services/localization.service.ts
@@ -99,7 +99,7 @@ export class LocalizationService {
 
           return local;
         }),
-        filter(Boolean)
+        filter(Boolean),
       )
       .subscribe(val => this.localizations$.next(val));
   }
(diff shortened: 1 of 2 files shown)


npm/ng-packs/packages/core/src/lib/services/localization.service.ts
yusuf-cirak, line 1: Early returning in `getLocalization` function, we don't need to assign a empty string and try to localize it.

npm/ng-packs/packages/core/src/lib/tokens/compare-func.token.ts
yusuf-cirak, line 1: Trying to order with these properties:
- Try to order with `order` property
- Try to order with `ìd` property
- Try to order with `name` property

masum-ulu (approved):

oykuermann (approved):

PR #19707, opened 2024-05-03
Use `makeEnvironmentProviders` function instead of config modules

### LINKED ISSUE #19715: Convert class to provider function for modules
## Todo
- We will change all `<module-name>ConfigModule` | `<module-name>Module (like ThemeShared or CoreModule)` classes 
to `provide<module-name>Config` function
- Some modules have own forRoot method `e.g: ThemeShared`, `forRoot` and `forChild` methods will be stay but the logic will move to also a `provide` functions with the args.

Sinan997:
Resolves #19715


npm/ng-packs/packages/core/src/lib/core.module.ts
```diff
@@ -132,96 +113,23 @@ export class RootCoreModule {}
...
-        {
-          provide: DYNAMIC_LAYOUTS_TOKEN,
-          useValue: options.dynamicLayouts || DEFAULT_DYNAMIC_LAYOUTS
-        },
-        {
-          provide: TitleStrategy,
-          useExisting: AbpTitleStrategy
-        }
-      ],
+      providers: [provideAbpCore(withOptions(options))],
```
masum-ulu, line 122: We need to include other options `with` funcs.

npm/ng-packs/packages/feature-management/src/lib/components/feature-management/feature-management.component.ts
```diff
@@ -125,7 +139,7 @@ export class FeatureManagementComponent
       .subscribe(() => {
         this.visible = false;
 
-        this.toasterService.success('AbpFeatureManagement::Saved');
+        this.toasterService.success('AbpFeatureManagement::SavedSuccessfully');
         if (!this.providerKey) {
```
sumeyyeKurtulus, line 143: I see that we have SavedSuccessfully key only for AbpSettingManagement and AbpUi scopes. So, I am updating this to AbpUi.

sumeyyeKurtulus (approved):

masum-ulu (approved):

PR #19732, opened 2024-05-07
Create a new community article: "Using FluentValidation with ABP"

EnesDONER:
Added using-fluent-validation-with-abp article


docs/en/Community-Articles/2024-05-07-using-fluent-validation-with-abp/post.md
```diff
@@ -0,0 +1,95 @@
+# Using FluentValidation with ABP
```
EngincanV: This can be a better title:

```suggestion
# Using FluentValidation with ABP Framework
```

Also, having an opening statement for an article is a good practise, so please add a short description for the article. It can be something like:

```
FluentValidation is a popular .NET library that allows you to define strongly typed validation rules. In this article, I will show you how you can get the benefit of this library and define custom validation rules in your application...
```

docs/en/Community-Articles/2024-05-07-using-fluent-validation-with-abp/post.md
```diff
@@ -0,0 +1,95 @@
...
+
+## What is Validation? Why Do We Need to Validate?
+Validation is checking whether a data is valid or not. We can liken validations to the cell membrane of our application. Just as the cell does not want to let in harmful substances, we do not want to add erroneous data to the database, which is critical to our application. Validations allow data that follows the rules to reach the database. Data that does not comply with the rules will not access the database at all, and the operation will fail.
```
EngincanV:
```suggestion
Validation is checking whether data is valid or not. We can liken validations to the cell membrane of our application. Just as the cell does not want to let in harmful substances, we do not want to add erroneous data to the database, which is critical to our application. Validations allow data that follows the rules to reach the database. Data that does not comply with the rules will not access the database at all, and the operation will fail.
```

docs/en/Community-Articles/2024-05-07-using-fluent-validation-with-abp/post.md
```diff
@@ -0,0 +1,95 @@
...
+
+![validation](./images/validation.png)
+
+Validations make your application run more efficiently because they are not connected to the database.  Validations should be applied in different layers of applications. Validations are applied on the UI side. Validations should also be applied on the backend side and database to prevent malicious users from passing these validations. Fluent validation is often used to create validations on the backend side.
```
EngincanV: Validations actually are not just tied to database point and they are used to validate your application's internal state. So, in this sentence, it's misleading and pointing out that it's main use is to keep the database in the correct state rather than keeping the application in the correct state. So, I think we should improve this sentence. Here is an example, sentence that came to my mind:

```suggestion
Validations make your application run more efficiently as they are not directly tied to the database. They should be implemented across various layers of the application, including the UI, backend, and database, to prevent malicious users from bypassing these checks. Fluent validation is frequently employed for creating backend validations.
```

You can change this sentence according to your own sentences, but keep that in mind, you should not just specify the validations are just for keeping the database in a correct state, rather it's used to keep all of your business rules in the correct order with the correct state (internal state of your application).

docs/en/Community-Articles/2024-05-07-using-fluent-validation-with-abp/post.md
```diff
@@ -0,0 +1,95 @@
...
+
+## How to Use Fluent Validation in ABP?
+![fluent validation](./images/abp-fluentvalidation.png)
+
+Using Fluent validation with Abp is quite simple. Open a command line window in the folder of the project (.csproj file) and type the following command: 
+abp add-package Volo.Abp.FluentValidation
```
EngincanV:
````suggestion
Using Fluent validation with ABP is quite simple. Open a command line window in the directory of your project (.csproj file) and type the following command: 

```bash
abp add-package Volo.Abp.FluentValidation
```

````

Also, here, specify in which project's directory this command needs to be executed because, in this article, you are using a layered application.

File: docs/en/Community-Articles/2024-05-07-using-fluent-validation-with-abp/post.md [hunk 5]
```diff
@@ -0,0 +1,95 @@
...
+
+
+## Why We Should Use Fluent Validation?
+Fluent Validation allows you to define your validation rules in a clear and flexible way. This means you can comfortably handle complex validation scenarios in your code. This makes your development process much more manageable. The readability that Fluent Validation offers really makes things easier. Having a clear understanding of what your validation rules do is a huge advantage when working on your code. In short, your code is cleaner and clearer. Fluent Validation is also very functional in terms of testability. By defining your validation rules in separate classes, you can easily test and maintain these rules. Fluent Validation is a widely used validation library on the .NET platform. As such, it has become a common standard among .NET developers. This provides advantages in terms of community support and compatibility. So, using Fluent Validation simplifies your development process by making your code more understandable, manageable and testable.
+
+## How to Use Fluent Validation in ABP?
+![fluent validation](./images/abp-fluentvalidation.png)
+
```
EngincanV, line 17: You can [delete this line](https://github.com/abpframework/abp/pull/19732/files#diff-26a7a58e0ab7e581e4a97cbc15be211876b596b579d2ff3188535025c60caa6eR22) and move the content here. It's better to remind the readers that they need to have an ABP-based application to follow this section and apply the steps:

```suggestion
> In this section, I will show you how to use FluentValidation library within an ABP-based application. Therefore, I assume that you have an ABP-based application that has already been created. If you haven't created an ABP-based application yet, please follow the [Getting Started documentation](https://docs.abp.io/en/abp/latest/Getting-Started-Create-Solution?UI=MVC&DB=EF&Tiered=No).
```

docs/en/Community-Articles/2024-05-07-using-fluent-validation-with-abp/post.md
```diff
@@ -0,0 +1,95 @@
...
+Using Fluent validation with Abp is quite simple. Open a command line window in the folder of the project (.csproj file) and type the following command: 
+abp add-package Volo.Abp.FluentValidation
+If you have not created your abp project, review the [steps to create](https://docs.abp.io/en/abp/latest/Tutorials/Todo/Overall) it now.
```
EngincanV, line 24: Please remove this line and apply my suggestion above.

docs/en/Community-Articles/2024-05-07-using-fluent-validation-with-abp/post.md
```diff
@@ -0,0 +1,95 @@
...
+
+Create your product object like this:
```
EngincanV:
```suggestion
Create the `Product` entity as below: 
```

docs/en/Community-Articles/2024-05-07-using-fluent-validation-with-abp/post.md
```diff
@@ -0,0 +1,95 @@
...
+````csharp
+  public class Product: FullAuditedAggregateRoot<Guid>
+   {
+       public string Name { get; set; } = string.Empty;
+       public decimal Price { get; set; }
+       public int Stock { get; set; }
+       public string? LicenseKey { get; set; }
+    }
+````
+Create the controller of your product object by following the [documentation](https://docs.abp.io/en/abp/latest/Tutorials/Todo/Index?UI=NG&DB=EF)
```
EngincanV: I did not understand this paragraph. Shouldn't you say that "after creating the entity, make the database configurations, create a new migration, and apply it to the database. Then, let's create the DTO classes:"?

If so, please update this line in a more descriptive way, and don't assume that the readers will know the other steps. An article should be descriptive, easy to use and ready to be consumed/used by applying all steps.

docs/en/Community-Articles/2024-05-07-using-fluent-validation-with-abp/post.md
```diff
@@ -0,0 +1,95 @@
...
+````csharp
+      public class ProductCreateDto
+    {
+        public string Name { get; set; } 
+        public decimal Price { get; set; }
+        public int Stock { get; set; }
+        public string? LicenseKey { get; set; }
+     }
+````
+Create the ProductCreateDtoValidator class in the products folder in the Application.Contacts layer.
```
EngincanV:
```suggestion
Create the `ProductCreateDtoValidator` class in the **Products** folder in the `Application.Contracts` project: 
```

docs/en/Community-Articles/2024-05-07-using-fluent-validation-with-abp/post.md
```diff
@@ -0,0 +1,95 @@
...
+        RuleFor(p => p.Stock).
+            GreaterThanOrEqualTo(0).
+            WithMessage("Product stock should not be negative");
+
+        RuleFor(p => p.Price).
+           GreaterThanOrEqualTo(0).
+           WithMessage("Product Price should not be negative");
+    }
+}
+````
```
EngincanV, line 74: This is the main code block for this article. Therefore, I think you should explain this code block in plain and short sentences. 

For example, you should say that you can add a NotEmpty check or set min-max length for a property etc. (or specify a message that will be thrown when an validation is failed)

docs/en/Community-Articles/2024-05-07-using-fluent-validation-with-abp/post.md
```diff
@@ -0,0 +1,95 @@
...
+
+````csharp
+RuleFor(p => p.LicenseKey).
+	Must(ContainsKey).
+	WithMessage("Product license key must contain the word 'key'");
+	
+````
+
+
+When I write code that does not follow the rule, we get this error
```
EngincanV: Before showing what will happen if the validation fails, I think you should show the application service implementation and how to use it, and then, write this message and show what will happen if validation fails. In this way, the reader will be able to directly follow the article and try the library. 

Because, in this current way, it's not clear how to get the benefit of these validation rules.

---

Also, please update this line as below:

```suggestion
If one of the validation rules does not meet the rules, then the following error will be received for the custom rule that we defined:
```

EngincanV (changes requested):
The article is good, it covers the topic with a plain example @EnesDONER 👏
You can consider my suggestions and apply them to the article. Then please "re-request review", so I can quickly take a look at it.

docs/en/Community-Articles/2024-05-07-using-fluent-validation-with-abp/post.md
EnesDONER: I have added these paragraphs:

The validator class you create should inherit from `AbstractValidator`.  You should give `AbstractValidator` the class you want to validate generically. You have to create a constructor method. You must write the code in this constructor method. Fluent validation provides you with ready-made methods that you can use to write validations very easily. 

The **RuleFor** method allows you to write a validation rule. You must specify in the parameter for which property you want to write a validation rule. 

With the **NotEmpty** method you can specify that a property cannot be null.

With **MinimumLength** and **MaximumLength** you can specify the minimum and maximum number of characters the property can take.

With the **GreaterThanOrEqualTo** method you can specify that the value of the property must be greater than or equal to the value entered as a parameter.

**WithMessage** method you can specify the message to be sent when the validation fails.

docs/en/Community-Articles/2024-05-07-using-fluent-validation-with-abp/post.md
EnesDONER: I have attached an article explaining how to get the error:

Try to add data that does not meet the validation rules

Adding Incorrect Data

hikalkan:
You can add a link for https://docs.abp.io/en/abp/latest/FluentValidation to the end of your article.

hikalkan (approved):

PR #19759, opened 2024-05-09
Create New Article: What is Object to Object Mapping

EnesDONER:
Added an article about object to object mapping


docs/en/Community-Articles/2024-05-09-what-is-object-to-object-mapping/post.md
```diff
@@ -0,0 +1,122 @@
+# What is Object to Object Mapping
```
EngincanV:
```suggestion
# What is Object to Object Mapping?
```

docs/en/Community-Articles/2024-05-09-what-is-object-to-object-mapping/post.md
```diff
@@ -0,0 +1,122 @@
...
+            FirstName = input.FirstName,
+            LastName = input.LastName,
+            // ...other
+        };
+        
+        return customerDto;
+    }
+
+````
+This code is repetitive, tedious to write. It violates DRY principles, makes your code very complex and reduces readability.  Instead of using this approach, you can use the automapper which does the matching automatically. Here is a better code:
```
EngincanV:
```suggestion
As can be seen here, it's repetitive and tedious to manually map an object to another similar object. Also, it violates the [DRY principle](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself), makes your code more complicated and reduces readability. Instead of manually mapping objects, you can use the  [AutoMapper](https://automapper.org/) library to automatically map two objects' properties:
```

docs/en/Community-Articles/2024-05-09-what-is-object-to-object-mapping/post.md
```diff
@@ -0,0 +1,122 @@
...
+        {
+
+            var customer = await _customerManager.CreateAsync(
+            input.BirthDay, input.MembershipDate, input.FirstName, input.LastName
+            );
+
+            return ObjectMapper.Map<Customer, CustomerDto>(customer);
+        }
+````
+
```
EngincanV, line 43: You should write a small explanation before/after each code block in articles. For example, in here, you are using the `ObjectMapper.Map` method and you can mention that 

> "Whenever you create a new abp-based application, the related mapping configurations are automatically done in your template and you can directly use it like in the example above. Also note that, `ObjectMapper` is defined in the `ApplicationService` base class, so you don't need to manually inject the `IObjectMapper` interface."

And then you can link the sentence with the following section with a sentence like: "Let's first see what AutoMapper library is used for and see it in action..."

> Please use your own sentences here, I just write them to give you a idea.

docs/en/Community-Articles/2024-05-09-what-is-object-to-object-mapping/post.md
```diff
@@ -0,0 +1,122 @@
...
+# What is AutoMapper?
+
+![Swagger](./images/automapper.png)
+
+Automapper is a .net library that automates object to object mapping. The implementation of Automapper is quite simple. In this chapter I will show you how to use the Automapper library in an ABP-based application. For this reason, I assume that you already have an ABP-based application created. If you have not yet created an ABP-based application, please follow the [Getting Started documentation](https://docs.abp.io/en/abp/latest/Getting-Started-Create-Solution?UI=MVC&DB=EF&Tiered=No).
```
EngincanV:
```suggestion
Automapper is a .NET library that automates object-to-object mapping. ABP provides abstractions for object-to-object mapping and has an integration package to use [AutoMapper](http://automapper.org/) as the object mapper. 

In this chapter, I will show you how to use the AutoMapper library in an ABP-based application. For this reason, I assume that you already have an ABP-based application created. If you have not yet created an ABP-based application, please follow the [Getting Started documentation](https://docs.abp.io/en/abp/latest/Getting-Started-Create-Solution?UI=MVC&DB=EF&Tiered=No).
```

docs/en/Community-Articles/2024-05-09-what-is-object-to-object-mapping/post.md
```diff
@@ -0,0 +1,122 @@
...
+````csharp
+    public class Customer : FullAuditedAggregateRoot<Guid>
+    {
+        public string? FirstName { get; set; }
+        public string? LastName { get; set; }
+        public DateTime BirthDay { get; set; }
+        public DateTime MembershipDate { get; set; }
+    }
+````
+`Custumer` contains a lot of attributes. we don't want to give all of these attributes to users, so create dto with only the necessary attributes:
```
EngincanV: There is a small typo here (Customer) and also we can improve the paragraph as follows:

```suggestion
`Customer` entity contains some properties (such as `FirstName`, `LastName`, ... and other audited properties coming from the base class - `DeleterId`, `IsDeleted`, `CreationTime` etc. -). Typically, you would not want to show/return all of these properties to end users, at that point, you can create a DTO class and only define the properties that you want to return to the end users.

Let's create the `CustomerGetDto` class in the `*.Application.Contracts` project as follows:
```

docs/en/Community-Articles/2024-05-09-what-is-object-to-object-mapping/post.md
```diff
@@ -0,0 +1,122 @@
...
+````csharp
+    public class CustomerGetDto
+    {
+        public string? FirstName { get; set; }
+        public string? LastName { get; set; }
+        public DateTime BirthDay { get; set; }
+    }
+````
+
+You can use this code to list customers:
```
EngincanV:
```suggestion
After creating our entity and output DTO classes, now in the application service implementation, we can return the `CustomerGetDto` class, as a result of listing the customers. For that reason, we can write a code as follows:
```

docs/en/Community-Articles/2024-05-09-what-is-object-to-object-mapping/post.md
```diff
@@ -0,0 +1,122 @@
...
+     var items = await _customerRepository.GetListAsync(input.FilterText, input.FirstName, input.LastName, input.BirthDayMin, input.BirthDayMax, input.MembershipDateMin, input.MembershipDateMax, input.Sorting, input.MaxResultCount, input.SkipCount);
+
+     return new PagedResultDto<CustomerGetDto>
+     {
+         TotalCount = totalCount,
+         Items = ObjectMapper.Map<List<Customer>, List<CustomerGetDto>>(items)
+     };
+ }
+````
+
```
EngincanV, line 94: You should write a small explanation before/after each code block in articles. You can write what this code block does, for example:

> Here, we basically get the total count of our customers and all of the customers according to the filters specified and then by using the `ObjectMapper.Map` method, which came from the [`ApplicationService` base class](https://docs.abp.io/en/abp/latest/Application-Services), we are mapping the `List<Customer>` to `List<CustomerGetDto>`. Thanks to that, we have full control over which properties to be returned to the end-users.

docs/en/Community-Articles/2024-05-09-what-is-object-to-object-mapping/post.md
```diff
@@ -0,0 +1,122 @@
...
+Update your `YourApplicationAutoMapperProfile` class so that `Customer` and `CustomerGetDto` entities can convert to each other.
```
EngincanV: Instead of using **convert** as the term, it would be better to use **map** as the word:

```suggestion
After mapping the `Customer` entity to the `CustomerGetDto` class. Before running our application, we should define the mappings in the `*AutoMapperProfile` class in the `*.Application` project as follows:
```

docs/en/Community-Articles/2024-05-09-what-is-object-to-object-mapping/post.md
```diff
@@ -0,0 +1,122 @@
...
+````csharp
+public class YourApplicationAutoMapperProfile : Profile
+{
+    public YourApplicationAutoMapperProfile()
+    {
+        CreateMap<Customer, CustomerGetDto>();
+    }
+}
+````
+The swagger image of the request will be like this. Only the fields in Dtoda will appear.
```
EngincanV:
```suggestion
Finally, we can run our application and navigate to the **/swagger** endpoint to try our endpoint. When we send a request, we should get the result as follows:
```

docs/en/Community-Articles/2024-05-09-what-is-object-to-object-mapping/post.md
```diff
@@ -0,0 +1,122 @@
...
+
+![Swagger](./images/swagger1.png)
+
+In some scenarios you may want to make some customizations when matching. If you want to use `CustomerGetDto` in this way:
```
EngincanV: Here, you can add a new section, because you are mentioning an advanced topic here, so add a heading as below:

```suggestion
## Advanced: Mapping Configurations

In some scenarios, you may want to make some customizations when mapping two objects. For example, let's assume that you want to create the `CustomerGetDto` class as follows:
```

docs/en/Community-Articles/2024-05-09-what-is-object-to-object-mapping/post.md
```diff
@@ -0,0 +1,122 @@
...
+
+````csharp
+  public class CustomerGetDto
+  {
+      public string? FullName { get; set; }
+      public int Age { get; set; }
+  }
+````
+Autommapper cannot map this automatically, you need to specify the exception. Update your `YourApplicationAutoMapperProfile`.
```
EngincanV:
```suggestion 
AutoMapper can't map these properties automatically, because they do not exist in the source object, which is the `Customer` entity. Therefore, you need to specify the exception and update your `YourApplicationAutoMapperProfile` class as follows:
```

docs/en/Community-Articles/2024-05-09-what-is-object-to-object-mapping/post.md
```diff
@@ -0,0 +1,122 @@
...
+
+````csharp
+
+CreateMap<Customer, CustomerGetDto>().ForMember(c=>c.FullName,opt=> opt.MapFrom(src => src.FirstName + " " + src.LastName))
+    .ForMember(c=>c.Age, opt=> opt.MapFrom(src=> DateTime.UtcNow.Year -src.BirthDay.Year));
+
+````
+This script concatenates and matches **FirstName** and **LastName** in the **FullName** field and subtracts **BirthDate** from today's year and assigns it to the customer's **Age**
```
EngincanV:
```suggestion
This configuration concatenates and matches **FirstName** and **LastName** properties into the **FullName** property and subtracts **BirthDate** from today's year and assigns it to the customer's **Age**.
```

docs/en/Community-Articles/2024-05-09-what-is-object-to-object-mapping/post.md
```diff
@@ -0,0 +1,122 @@
...
+When you make a request, your output will look like the one below.
```
EngincanV:
```suggestion
After these configurations, if you make a request to the relevant endpoint, the output will look like as below:
```

docs/en/Community-Articles/2024-05-09-what-is-object-to-object-mapping/post.md
```diff
@@ -0,0 +1,122 @@
...
+
+![Swagger](./images/swagger2.png)
+
+For more information on object to object mapping with abp framework, see the [documentation](https://docs.abp.io/en/abp/latest/Object-To-Object-Mapping)
```
EngincanV:
```suggestion
For more information on object-to-object mapping with [ABP Framework](https://abp.io/), see the [documentation](https://docs.abp.io/en/abp/latest/Object-To-Object-Mapping).
```

docs/en/Community-Articles/2024-05-09-what-is-object-to-object-mapping/post.md
```diff
@@ -0,0 +1,122 @@
+# What is Object to Object Mapping
+
+Mapping the properties of one object to the properties of another object is called Object to object mapping. Most of the time we don't want to show the data we store to end users as it is. We only give users the information they will use for that operation. In tables that contain relationships, we analyze the relationships and present meaningful data to users. For example, suppose we have a product and a category object, we keep an attribute called categoryId in the product object. However, it would be illogical to show the categoryId attribute to users. Therefore, we create dto's (data transfer objects) and show the category name attribute to users instead of the category Id attribute. 
+
+DTOs are used to transfer data of objects from one place to another. We often need to convert our Domain objects to DTOs and DTOs to Domain objects, for this you can use a code like this.
```
EngincanV:
```suggestion
Mapping the properties of one object to the properties of another object is called **Object to Object Mapping**. Most of the time, you don't want to show the data you store in your database to end users as it is. Instead, you only return users the information they need for that operation and reduce the output size. 

For example, in database tables that contain relationships, we analyze the relationships and present meaningful data to users. For example, suppose we have a product and a category object, we keep a property called `categoryId` in the `Product` entity. However, it would be illogical to show the `categoryId ` property to users. Therefore, we can create a DTO (data transfer object) and show the **category name**  to the end users, instead of the `categoryId` directly. 

DTOs are used to transfer data of objects from one object to another one. We often need to map our entities to DTOs and DTOs to entities. For example, consider the code below:

EngincanV (changes requested):
@EnesDONER Please make the changes I specified, feel free to write to me, if any one of them is unclear for you, or you have other thoughts on them.

docs/en/Community-Articles/2024-05-09-what-is-object-to-object-mapping/post.md
EnesDONER: I added this sentence:
The **ObjectMapper.Map** method allows you to convert your `Customer` entity to `CustomerDto`. Let's learn more about **AutoMapper**

docs/en/Community-Articles/2024-05-09-what-is-object-to-object-mapping/post.md
EnesDONER: I added this sentence:
In this code, we first get the total number of our customers and all customers according to the specified filters, then we map List<Customer> to List<CustomerGetDto> using the ObjectMapper.Map method from the ApplicationService base class. This way we have full control over which properties are returned to the end users.

EngincanV (approved):

Issue #19765, opened 2024-05-10
Errors with basic app-pro template


realLiangshiwei:
https://support.abp.io/QA/Questions/7171/Errors-with-basic-app-pro-template-in-811

PR #19766, opened 2024-05-10
Fix bug in template with basic theme

realLiangshiwei:
### Description

Resolves https://github.com/abpframework/abp/issues/19765

### Checklist


realLiangshiwei:
@EngincanV  You can ignore the unit test error as it is not relevant to this PR

EngincanV (approved):

PR #19770, opened 2024-05-10
Log out of blazor wasm when get a "401" response.

maliming:
@gizemmutukurt For your test: 

**Blazor wasm will also log out of the auth server. This is a limitation of Blazor wasm, We can't do anything.**

The Blazor server and Blazor web app don't have this behavior.


gizemmutukurt (changes requested):
Problem continue..
image

gizemmutukurt (approved):

PR #19839, opened 2024-05-15
Use config provider function in templates

### LINKED ISSUE #19715: Convert class to provider function for modules

Sinan997:
Resolves #19715
Depends on #19707 

Remove forRoot pattern in templates and use config provider functions.


sumeyyeKurtulus (approved):

masum-ulu (approved):

masum-ulu:
Wait for #19707 before merging

PR #19846, opened 2024-05-16
Angular - Extensible table logic errors for action list

### LINKED ISSUE #19837: Angular - Extensible Table Logic Errors for Action List
This issue is expected to resolve the logic errors for the actions button inside the extensible table component. Here is an example to demonstrate the probable logic error:

In this example, even if the actions button of the first row has some items, it is not visible as it should be in the second view.
image
image

Hence, if the action list has no items, it should be invisible as can be seen below.
image

sumeyyeKurtulus:
### Description

Resolves #19837 

Here are the related records:

https://github.com/abpframework/abp/assets/92928815/351936c5-a645-47a0-af2b-576c431c12cf

image
image

### Checklist

- [X] I fully tested it as developer
- [X] There is no need for documentation

### How to test it?

You can test it by navigating through the pages that has a table with actions.


npm/ng-packs/packages/components/extensible/src/lib/components/extensible-table/extensible-table.component.ts
```diff
@@ -206,20 +206,20 @@ export class ExtensibleTableComponent<R = any> implements OnChanges {
...
       const { visible, permission } = action;
-      if (rowData && action) {
-        const visibilityCheck = visible({
-          record: rowData,
-          getInjected: this.getInjected,
-        });
-        const permissionCheck = this.permissionService.getGrantedPolicy(permission);
 
-        isActionAvailable = visibilityCheck && permissionCheck;
+      if (permission && visible) {
```
masum-ulu: What happens when visible exits but permission not exits ? I guess this logic must be `|| OR` not `&& AND` ? Am I wrong ? 🤔 && condition is valid for 217.Line not for this line

npm/ng-packs/packages/components/extensible/src/lib/components/extensible-table/extensible-table.component.ts
```diff
@@ -206,20 +206,20 @@ export class ExtensibleTableComponent<R = any> implements OnChanges {
     });
   }
 
-  hasAvailableActions(rowData): boolean {
-    let isActionAvailable = true;
-    this.actionList.toArray().map(action => {
+  hasVisibleActions(rowData: any): boolean {
```
masum-ulu: Can you change name to `isVisibleActions`

npm/ng-packs/packages/components/extensible/src/lib/components/extensible-table/extensible-table.component.ts
sumeyyeKurtulus: I have also tried these options. In this case it either just looks for visibility or suitable permission availability. However, if both visible and permission values are defined, we need to check both. So, we will have four cases here as:

1. Both permission and visible conditions are there
2. There is a visibility condition but no permission condition
3. There is a permission condition but no visibility condition
4. Neither permission nor visible conditions are there

Where I believe we can handle all of them like this:

image

masum-ulu (approved):

oykuermann (approved):

PR #19858, opened 2024-05-18
Angular | Refactor modal component

Sinan997:
Refactor modal component with the new signal API

### DIFF
--- a/npm/ng-packs/packages/theme-shared/src/lib/components/modal/modal.component.html
+++ b/npm/ng-packs/packages/theme-shared/src/lib/components/modal/modal.component.html
@@ -1,9 +1,9 @@
 <ng-content></ng-content>
 
 <ng-template #modalContent let-modal>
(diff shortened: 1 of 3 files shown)


masum-ulu:
Why we need this changes now ? Also is it ok with current usages ? our setted values are not the signal based type
 `e.g: in identiy component modals`

Sinan997:
> Why we need this changes now ? Also is it ok with current usages ? our setted values are not the signal based type `e.g: in identiy component modals`

Provided value doesn't have to be signal, it can be signal or not. So, it is okey with the current usages. You can test it in your local by `sinan/refactor-modal` branch.

For more info about [**`model inputs`**](https://angular.io/guide/model-inputs). 

As far as i know we want to use new signal features of Angular. According to this idea i have refactor the modal component. If you think this isn't our goal, i can close the PR.

masum-ulu (approved):

Issue #19930, opened 2024-05-26
Send Base64 Images in Email with Google Apps


masoodkhoshgard:
### Description
Gmail will not render base64 images embedded inside HTML emails but with Google Apps Script

on Gmail:
![Screenshot 2024-05-26 125259](https://github.com/abpframework/abp/assets/28512454/76107b9f-78a3-4704-804f-efb3e8963a5c)

on personal mail server:
![Screenshot 2024-05-26 125348](https://github.com/abpframework/abp/assets/28512454/b686c913-3760-4c8c-837b-5749e937807f)

Pantex9:
Hi, Gmail's refusal to render Base64-encoded images embedded in HTML emails is a known limitation. 
Instead, Gmail requires images to be hosted on a publicly accessible server.

PR #19932, opened 2024-05-26
Send Base64 Images in Email with Google Apps

masoodkhoshgard:
### Description
#19930 

Gmail will not render base64 images embedded inside HTML emails but with Google Apps Script
on Gmail:
Screenshot 2024-05-26 125259

on personal mail server:
Screenshot 2024-05-26 125348

### Checklist


maliming:
Thank you for your code. 

The `NormalizeMailAsync` is designed to override and meet your cases. 

We don't need to add the Gmail-relevant code to the framework.

masoodkhoshgard:
> Thank you for your code.
> 
> The `NormalizeMailAsync` is designed to override and meet your cases.
> 
> We don't need to add the Gmail-relevant code to the framework.

Thanks @maliming 
But keep in mind that Gmail is a public service and this problem happens a lot and that it is likely to occur for other mail services as well.

framework/src/Volo.Abp.Emailing/Volo/Abp/Emailing/EmailSenderBase.cs
```diff
@@ -137,52 +87,139 @@ public virtual async Task QueueAsync(string from, string to, string subject, str
...
+    /// <param name="mail">Mail to be sent</param>
+    protected abstract Task SendEmailAsync(MailMessage mail);
 
-        if (mail.BodyEncoding == null)
+    private static void NormalizeMailForBase64Image(MailMessage mail)
+    {
+        var htmlBody = mail.Body;
+
+        // Find all base64 image tags in the html message.
+        var base64ImageTags = Regex.Matches(htmlBody, @"<img src=""data:image/(png|jpeg|gif);base64,([^""]+)""[^>]*>");
```
ebicoglu: The regex you provided doesn't cover all variants of a base64 image element.
Here's a better regex that finds all possible `src` with the base64 format:

```
src\s*=\s*("|')\s*data\s*:\s*image/\s*(png|jpeg|gif);\s*base64,([^""]+)."|'
```

 
image


syntax:

```
bool foundMatch = false;
try {
	foundMatch = Regex.IsMatch("input-text-here...", @"src\s*=\s*(""|')\s*data\s*:\s*image/\s*(png|jpeg|gif);\s*base64,([^""""]+).""|'", RegexOptions.Singleline);
} catch (ArgumentException ex) {
	// Syntax error in the regular expression
}

```

framework/src/Volo.Abp.Emailing/Volo/Abp/Emailing/EmailSenderBase.cs
```diff
@@ -137,52 +87,139 @@ public virtual async Task QueueAsync(string from, string to, string subject, str
...
+                                                        MediaTypeNames.Text.Plain);
+
+            var img = new LinkedResource(imageStream, $"image/{format}")
+            {
+                ContentId = imageName
+            };
+            img.ContentType.MediaType = $"image/{format}";
+            img.TransferEncoding = TransferEncoding.Base64;
+            img.ContentType.Name = img.ContentId;
+            img.ContentLink = new Uri("cid:" + img.ContentId);
```
ebicoglu, line 219: Is `cid` a universal HTML format? or only Gmail understand this tag?

masoodkhoshgard: No, "cid" is not a universal HTML format. CID stands for "Content-ID" and is used specifically in the context of email and HTML email messages.

In email, the "cid" is used to reference embedded content, such as images or other media, that is included within the email message itself. The CID acts as a unique identifier for that embedded content.

For example, in an HTML email, an image might be referenced using a CID like this:
``

This tells the email client to look for an image with the content ID "image001.jpg" that is embedded within the email message.

However, this "cid" reference is not a standard HTML format that can be used outside of the email context. In regular web pages, images and other resources are typically referenced using standard URL paths, not CIDs.

So in summary, "cid" is specific to email and not a universal HTML format for referencing content on the web. Regular HTML uses other methods like \[src="image.jpg"\] to link to external resources.

framework/src/Volo.Abp.Emailing/Volo/Abp/Emailing/EmailSenderBase.cs
masoodkhoshgard: @ebicoglu 
I think it would be better to use the following regex pattern:

`src\s*=\s*(""|')\s*data\s*:\s*(?<mediaType>[\w/\-\.]+);(?<encoding>\w+),(?<data>.*?)(""|')`

Because it covers all tags that use base64, such as fonts, video, audio and more, not just limited to images.

maliming:
[I have formatted the code](https://github.com/abpframework/abp/pull/19932/files).  The code differences are now simpler.

@ebicoglu You can review it again. Thanks.

maliming:
hi @masoodkhoshgard 

We have decided not to add the `NormalizeMailForBase64DataAsync` to the framework. You can use it in your project.

Thanks for your contribution.

PR #19952, opened 2024-05-30
Return more messages for `401 Unauthorized`.

### LINKED ISSUE #19776: Return more messages for `401 Unauthorized`.
https://tools.ietf.org/html/rfc6750#section-3.1

```
WWW-Authenticate: Bearer error="invalid_token", error_description="The token is no longer valid because the user's session expired."

WWW-Authenticate: Bearer error="invalid_token", error_description="The token is no longer valid because the user's session expired.", error_uri="https://documentation.openiddict.com/errors/ID2095"
```

maliming:
Resolve #19776

Run `app-pro` tiered projects to test this.

Revoke a session and see the error message.

image

image


framework/src/Volo.Abp.ExceptionHandling/Volo/Abp/ExceptionHandling/Localization/en.json
```diff
@@ -5,22 +5,25 @@
...
     "401Message": "Unauthorized",
     "403Message": "Forbidden",
     "404Message": "Page not found",
     "500Message": "Internal Server Error",
     "403MessageDetail": "You are not authorized to perform this operation!",
-    "404MessageDetail": "Sorry, there's nothing at this address."
+    "404MessageDetail": "Sorry, there's nothing at this address.",
+    "Unauthorized": "Unauthorized",
+    "invalid_token": "Invalid token",
+    "The token is no longer valid because the user's session expired.": "The token is no longer valid because the user's session expired."
```
ebicoglu: The keys must be same format.
`invalid_token` => `InvalidToken`
`The token is no longer valid because the user's session expired. `=> `TokenExpiredMessage`

image

ebicoglu: The session expired message should be user-friendly:

- Turkish :  `Uzun süredir işlem yapmadığınız için oturumunuz sonlanmıştır! Sayfayı yenileyin veya tekrar giriş yapın.`
- English: `Your session has expired due to being idle for a long time! Refresh the page or login again.`

maliming: In most cases, the administrator revokes the user's session.
So the message is should not be `idle for a long time`

EngincanV (approved):
I think the `SessionExpired` message, can be as

```txt
Your session has expired. Please login again to continue in the application.
```

Because, the current message ("The token is no longer valid because the user's session expired.") not meaningful for the end-user, because *token* is a technical term. So, I think it would be better to update it. What do you think? @maliming

maliming:
ok, I will change the message in the new PR. because you have merged the Volo's PR.

maliming:
> ok, I will change the message in the new PR. because you have merged the Volo's PR.

https://github.com/abpframework/abp/pull/20141

PR #19986, opened 2024-06-06
BackgroundWorkers should be restartable to permit changes in Timer interval

machonky:
As BackgroundWorkers are a Singleton, they should be restartable to permit changes in Timer interval without an application restart.

### Description

When StopAsync is invoked followed by a StartAsync on a BackgroundWorker the StoppingTokenSource is not renewed and will throw if the BackgroundWorker is stopped a second time.

The change assigns a new CancellationTokenSource to the StoppingTokenSource property and assigns a new token from the new source.

### Checklist

### DIFF
--- a/framework/src/Volo.Abp.BackgroundWorkers/Volo/Abp/BackgroundWorkers/BackgroundWorkerBase.cs
+++ b/framework/src/Volo.Abp.BackgroundWorkers/Volo/Abp/BackgroundWorkers/BackgroundWorkerBase.cs
@@ -22,10 +22,16 @@ public abstract class BackgroundWorkerBase : IBackgroundWorker
 
     protected ILogger Logger => LazyServiceProvider.LazyGetService<ILogger>(provider => LoggerFactory?.CreateLogger(GetType().FullName!) ?? NullLogger.Instance);
 
-    protected CancellationTokenSource StoppingTokenSource { get; }
-    protected CancellationToken StoppingToken { get; }
+    protected CancellationTokenSource StoppingTokenSource { get; private set; }
+	
+    protected CancellationToken StoppingToken { get; private set; }
 
     public BackgroundWorkerBase()
+    {
+        ResetStoppingCancellationTokenSource();
+    }
+
+    private void ResetStoppingCancellationTokenSource()
     {
         StoppingTokenSource = new CancellationTokenSource();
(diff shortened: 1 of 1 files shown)


maliming:
hi

> As BackgroundWorkers are a Singleton, they should be restartable to permit changes in Timer interval without an application restart.

Can you share some code to show this case?

The `StopAsync` is called on `ApplicationShutdown`. so it won't restart.

machonky:
When the Host changes a system-wide polling interval setting a local message instructs the PerTenantBackgroundWorker to change as shown -  Example code:
```csharp
public class PerTenantBackgroundWorker : AsyncPeriodicBackgroundWorkerBase
{
...
    public async Task ChangePollingInterval(int newPollingInterval)
    {
        var oldValue = Timer.Period / Minutes;
        logger.LogInformation($"{ToString()} changing polling interval from {oldValue} to {newPollingInterval} (mins)");
        
        await StopAsync(StoppingToken);
        Timer.Period = newPollingInterval * Minutes;
        await StartAsync();

        logger.LogInformation($"{ToString()} Changed polling interval to: {newPollingInterval}");
    }
...
}
```
If the ```StopAsync``` is invoked after this without renewing the StoppingToken then an exception is thrown from the token source, as it has been used up already so it must be "renewed" after every stop call.

The change makes it routine to change the polling interval without needing an application restart,

maliming:
ok, We can make these properties to `protected,` then you can change them in the sub-class.

```
protected CancellationTokenSource StoppingTokenSource { get; set; }

protected CancellationToken StoppingToken { get; set; }
```

We can drop the `ResetStoppingCancellationTokenSource` method.

What do you think?

machonky:
Yes. Since StopAsync is overridable, this will work too.

maliming:
Can you update your PR's code? Thanks

machonky:
Updated: https://github.com/abpframework/abp/commit/9e051f6a18805b7a9096fb08effaf59b9859f0fb

maliming:
Thanks. 

Can you remove these spaces?

image

machonky:
Done. https://github.com/macrobert-com/abp/commit/7efe0a4a628a19fa08fbf304538784064dade56f

maliming (approved):

maliming:
Thanks!

PR #20012, opened 2024-06-11
Apply `AbpConcepts` to entities which navigation property changes.

### LINKED ISSUE #19962: Why ConcurrencyStamp not change when ValueObject updated
code example:

```
public class DateRange : ValueObject {
    public DateRange(DateOnly start, DateOnly end) {
        if (start > end) {
            throw new InvalidDataException("");
        }
        Start = start;
        End = end;
    }

    public DateOnly Start { get; private init; }

    public DateOnly End { get; private init; }

    protected override IEnumerable<object> GetAtomicValues() {
        yield return Start;
        yield return End;
    }
}

public class Program : AggregateRoot<Guid> {
    public string Name { get; private set; }
    public DateRange DateRange { get; private set; }
    public List<Item> Items { get; private init; } = new();

    public void Update(string name, DateRange dateRange, IEnumerable<Item> items) {
          Name = name;
          DateRange = dateRange;
          Items.Clear();
          Items.AddRange(items);
      }
}

var program = await _programRepository.GetAsync(id);
program.Update(name, dateRange, items);
await _programRepository.UpdateAsync(program, true);
```

when i update the Name field ConcurrencyStamp is worked,
but when i update the DateRange  or Items , database has changed and the ConcurrencyStamp not update.

maliming:
Resolve #19962


hikalkan (approved):

MJSanfelippo:
@hikalkan  @maliming This breaking change should be called out in the 8.3 migration docs. Upgraded to 8.3 and suddenly a bunch of my tests were failing, because I was asserting that the concurrency stamp had not changed (because why would it...), or the test was suddenly running into concurrency exceptions. This should be opt-in, not opt-out (e.g. `PublishEntityUpdatedEventWhenNavigationChanges` should default to `false`, not `true`). There's also side-effects here: maybe I still wanted the previous behavior of the deleted/updated events (I don't, but someone else might), but I don't want the concurrency stamp to change. Should've been 2 separate variables IMO.  This snippet alone should've rang some alarm bells as the code was being written: 

```
if (entityEntry.State == EntityState.Unchanged)
{
   ApplyAbpConceptsForModifiedEntity(entityEntry, true);
}
```

If the entity state is Unchanged... apply the ABP Concepts for _modified_ entities?

It's not ideal that I have to dig through ABP source code to figure out what the breaking changes are, yet it feels like I have to do it for nearly every upgrade.

ikappas:
@hikalkan @maliming This change also affected our team causing our tests to fail on 8.3.x . We were using concurrencyStamp consts in our tests in order to verify update related operations. This issue occured whenever we had foreign keys with navigation properties associated with our seeded models. Needless to say that we spend quite some time in order to get to the bottom of this issue, so documenting this in the 8.3 migration docs would probably be beneficial for other people.

@MJSanfelippo Thank you for reporting this issue. It helped us undestand ours.

maliming:
ok, Thanks @MJSanfelippo @ikappas

PR #20021, opened 2024-06-12
Set a ratio to the CMS Kit blog post images

enisn:
### Description

Resolves https://github.com/volosoft/volo/issues/17748#issuecomment-2162209900

I go with 16:9 since our current placeholder image is 16:9
https://github.com/abpframework/abp/tree/dev/modules/cms-kit/src/Volo.CmsKit.Common.Web/wwwroot/cms-kit

### Checklist

### How to test it?


armgnunlu (approved):

oykuermann:
- [ ] On the blog post page should we set the image size to 16:9 ratio?
image

oykuermann (changes requested):
I'll check after changes

PR #20081, opened 2024-06-20
Revise & update various documentation for documentation unification.

EngincanV:
~~I have completed the all related list items, but I want to check all documentation generally, because I found too many wrong documentation link references. For example, to reference to CLI documentation we should write like "../cli/index.md" but in almost every place, it's written as "../cli". So, it's referencing the root folder, instead of the main documentation, and therefore all of the documentation links like this one are wrong.~~

~~I'll check them one by one and fix them. Then, I'll convert the PR from the draft. FYI @hikalkan.~~

---

**Update:** All of these wrong URLs will be fixed when https://github.com/abpframework/abp/pull/20077 is merged.


hikalkan (approved):

hikalkan:
Thanks 👍

PR #20089, opened 2024-06-23
use post condition not null attribute

mseada94:
### Description

When using `Check` guard for null check the visual studio still produce the null warning for the parameter 
such as `CS8602 - Dereference of a possibly null reference`

This PR uses `System.Diagnostics.CodeAnalysis.NotNull` attribute to inform the code analyzer that the parameter tested with this method without exception is not null and there is no need for the nullability warnings

### DIFF
--- a/framework/src/Volo.Abp.Core/Volo/Abp/Check.cs
+++ b/framework/src/Volo.Abp.Core/Volo/Abp/Check.cs
@@ -10,7 +10,7 @@ public static class Check
 {
     [ContractAnnotation("value:null => halt")]
     public static T NotNull<T>(
-        T? value,
+        [System.Diagnostics.CodeAnalysis.NotNull] T? value,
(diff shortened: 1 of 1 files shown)


maliming:
hi

What's your VS version? 

We have used the `JetBrains.Annotations` in framework&modules everywhere.

mseada94:
I am using
Visual Studio Version 17.8.5

> Note:
> I am not using Resharper

I created new project `abp new Acme.BookStore --theme basic -csf`
ABP Version 8.1.4
and added this method to a class in domain layer
```cs
public string TestNullCheck(string? name)
{
	Check.NotNull(name, nameof(name));

    return name;
}
```

SNAG-0052
And I have `Warning: CS8603	Possible null reference return.`

maliming (approved):

maliming:
Thanks @mseada94 👍

PR #20103, opened 2024-06-25
Remove links to documents with TODO

salihozkara:


ebicoglu:
removing links will be hard to find the related documents later. so we can change the `TODO` text to 
> This document is planned to be written later.

PR #20205, opened 2024-07-11
DistributedCacheEntryOptions configurable

842549829:
DistributedCacheEntryOptions configurable

Change the DistributedCacheEntryOptions to can configure option Instead of directly instantiated

### Description

Change the DistributedCacheEntryOptions to can configure option Instead of directly instantiated
From 
`new DistributedCacheEntryOptions() 
{
    SlidingExpiration = TimeSpan.FromDays(30)
};`
To
` Configure<DistributedCacheEntryOptions>(options =>
 {
     options.SlidingExpiration = TimeSpan.FromDays(30);
 });`
### Checklist

- [ ] I fully tested it as developer  / integration tests


maliming:
The `DistributedCacheEntryOptions` is used in the permissions, features, and settings module, so we can use the named `DistributedCacheEntryOptions` in these modules.

PR #20257, opened 2024-07-20
BackgroundJobs - Enhance AsyncBackgroundJob with extra services

machonky:
### Description

Increase the utility of AsyncBackgroundJob

Add a collection of Lazy-created services which reduce application code, modelled after AbpControllerBase

### Checklist

### How to test it?


machonky:
Missed merge

Issue #20137, opened 2024-07-01
Docs Module: Scriban render problem


salihozkara:
If there is a problem with scriban rendering in documents, display this error on the page. 

> Error occurred during the rendering of this document. "The document is not valid: {SCRIBAN_ERROR_MESSAGE}"

Example:

````txt
The document is not valid. <input>(184,23) : error : Error while parsing nested expression: Expecting <expression> instead of `.` in: (<expression>)
<input>(184,23) : error : Invalid token `.`. Expecting a closing `)`.
<input>(184,98) : error : Invalid token found `)`. Expecting <EOL>/end of line.
System.InvalidOperationException: This template has errors. Check the <Template.HasError> and <Template.Messages> before evaluating a template. Messages:
<input>(184,23) : error : Error while parsing nested expression: Expecting <expression> instead of `.` in: (<expression>)
<input>(184,23) : error : Invalid token `.`. Expecting a closing `)`.
<input>(184,98) : error : Invalid token found `)`. Expecting <EOL>/end of line.
````

Also send discord notification or email to us.

PR #20463, opened 2024-08-09
Add DocumentRenderErrorEvent

realLiangshiwei:
### Description

Related: https://github.com/abpframework/abp/issues/20137

image

### Checklist

### DIFF
--- a/modules/docs/src/Volo.Docs.Web/Pages/Documents/Project/Index.cshtml.cs
+++ b/modules/docs/src/Volo.Docs.Web/Pages/Documents/Project/Index.cshtml.cs
@@ -575,7 +575,7 @@ namespace Volo.Docs.Pages.Documents.Project
             Document.Content = content;
         }
         
-        protected async Task OnSectionRenderingErrorAsync(Exception e)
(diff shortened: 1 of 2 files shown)


salihozkara (changes requested):
Can we do something to prevent search engines from indexing this way? Can you check it? Like changing the http code. Also, if there is a user who overrides the page, we can create a property where this user can detect that the page content has changed.

modules/docs/src/Volo.Docs.Web/Pages/Documents/Project/Index.cshtml.cs
```diff
@@ -578,6 +574,19 @@ private async Task ConvertDocumentContentToHtmlAsync()
 
             Document.Content = content;
         }
+        
+        protected async Task OnSectionRenderingErrorAsync(Exception e)
```
salihozkara:
```suggestion
        protected virtual async Task OnSectionRenderingErrorAsync(Exception e)
```

salihozkara (approved):

PR #20494, opened 2024-08-13
Angular - Update Toastr Styles

sumeyyeKurtulus:
### Description

Resolves the latest toastr styles. Here is a demonstration sample

image

### Checklist


gizemmutukurt (changes requested):
We can improve the alignment between the icon and the text.

image

masum-ulu (approved):

gizemmutukurt (changes requested):
The distance between icon and text should be shortened a little bit.
image

gizemmutukurt (approved):

PR #20530, opened 2024-08-15
Make use of System.Threading.Lock

MarkCiliaVincenti:
.NET 9.0 comes with support for a new class `System.Threading.Lock` which offers better performance when locking as opposed to a lock on an object.

Through the use of a [new dependency](https://github.com/MarkCiliaVincenti/Backport.System.Threading.Lock), `System.Threading.Lock` is backported to .NET 8.0 and earlier. On .NET 8.0 or earlier, there is neither a performance benefit nor a performance loss when using `System.Threading.Lock`. However, once ABP starts supporting .NET 9.0 or later, the code that's written will automatically benefit from improved performance.


MarkCiliaVincenti:
I asked a question on whether ABP wants to continue supporting .NET Standard 2.0 and 2.1 at https://github.com/abpframework/abp/pull/20803#issuecomment-2382558353. The reason is that if this PR is going to get approved it would need to undergo changes. We would need to use the factory method in https://github.com/MarkCiliaVincenti/Backport.System.Threading.Lock rather than the clean method that we could use if ABP supports only .NET 5.0 or greater (basically anything that doesn't support the notorious thread aborts).

MarkCiliaVincenti:
@maliming do you still want to support .NET Standard 2.0 and 2.1? I want to be able to change this PR accordingly.

MarkCiliaVincenti:
@maliming do you still want to support .NET Standard 2.0 and 2.1? I want to be able to change this PR accordingly. It would be cleaner as an implementation of only .NET 5.0+ is supported. Due to thread aborts, there are issues with how synchronization can sometimes be broken in < 5.0, for example there is no 100% guarantee that anything that makes use of the using pattern is correctly disposed every time.

MarkCiliaVincenti:
@maliming due to assembly-level attributes, the only way this can avoid a dependency is that a new project file is created and everything in the library is copy and pasted into it, but it would be much cleaner with a dependency that's also tried and tested on production projects.

MarkCiliaVincenti:
> @maliming do you still want to support .NET Standard 2.0 and 2.1? I want to be able to change this PR accordingly. It would be cleaner as an implementation of only .NET 5.0+ is supported. Due to thread aborts, there are issues with how synchronization can sometimes be broken in < 5.0, for example there is no 100% guarantee that anything that makes use of the using pattern is correctly disposed every time.

Still need an answer for this @maliming.

maliming:
hi 

We will continue to support `.NET Standard 2.0 and 2.1`

https://github.com/abpframework/abp/issues/2668#issuecomment-575922303

MarkCiliaVincenti:
@maliming a new version of Backport.System.Threading.Lock has come out that acts as a source generator and basically dropping the DLL as a dependency. If this new development interests you, I will amend this PR accordingly.

maliming:
@ebicoglu What do you think?

MarkCiliaVincenti:
Changes applied.

MarkCiliaVincenti:
@maliming @ebicoglu

MarkCiliaVincenti:
@maliming @ebicoglu

hikalkan:
@MarkCiliaVincenti thank you for your pr. We are waiting that since we are not sure to add such a nuget package dependency to the most core abp package for such a small benefit. We will be discussing it.

MarkCiliaVincenti:
Hi, it's not a dependency in the traditional sense since it is used as a source generator

MarkCiliaVincenti:
@maliming @hikalkan let me know if and when you intend on merging this so that I can resolve conflicts.

Issue #11989, opened 2022-03-17
Announcement of plan to replace the IdentityServer


hikalkan:
### What's the current situation

ABP uses **IdentityServer4** to add **OAuth server** features as built-in in the server-side. We use it for;

* Single sign-on, OpenIdConnect (authorization code and hybrid flow) logins in tiered, Angular, Blazor, and Microservice solutions.
* Client-credential authentication for machine-to-machine communication (including inter-microservice REST API calls).
* Resource owner password authentication for the Angular UI (as an alternative option to authorization code flow).

All these functionalities are available both of ABP [open-source](https://abp.io/) and [commercial](https://commercial.abp.io/). ABP Commercial adds a UI layer to manage IdentityServer4 resources as an additional feature.

### What's new?

**IdentityServer4's support ends at the end of the year 2022**. Its replacement is [Duende IdentityServer](https://duendesoftware.com/products/identityserver), which is not a free software anymore for those have **more than 1M$ income** (read [more](https://blog.duendesoftware.com/posts/20220111_fair_trade/)).

### Problems

We have two problems here:

1. If we switch to **Duende IDS,** open source ABP will be paid in practical for those who have more than 1M$ income. Because, even basic login scenarios (for Angular UI, Blazor WASM UI and tiered MVC UI) are using authorization code or resource owner password flows. ASP.NET Core has not a usable built-in solution for these flows (See Microsoft's announcements on this topic: [a post](https://devblogs.microsoft.com/dotnet/asp-net-core-6-and-authentication-servers/), [a GitHub issue](https://github.com/dotnet/aspnetcore/issues/26489), [another discussion](https://github.com/dotnet/aspnetcore/issues/32494)).
2. We don't want ABP Commercial customers to pay extra money to Duende IDS, which has higher prices than ABP Commercial.

## The Decision

So, we will completely drop the IDS from the ABP platform. We, probably will continue to ship IDS-related packages for a while, but, in the long term, you will need to replace it, because IDS support ends at the end of the year 2022.

## Solution Proposal

Our solution will be like this:

[...]

PR #20547, opened 2024-08-17
Fix vulnerabilities in dependencies

Muximize:
### Description

Resolves 417 instances of 14 vulnerabilities in multiple versions of these 10 (transitive) dependencies:
```
Azure.Identity                   
Microsoft.Identity.Client        
Newtonsoft.Json                  
Npgsql                           
SixLabors.ImageSharp             
System.Formats.Asn1              
System.Net.Http                  
System.Security.Cryptography.Xml 
System.Text.Json                 
System.Text.RegularExpressions   
```

These stats are only for the framework solution but the modules where also fixed. These where found using:
```
dotnet list package --vulnerable --include-transitive
```

### Notes

Because it gets annoying to keep having to fix these in our project, I thought maybe I should fix it at the source 😇 

Instead of just updating the vulnerable package versions, I did some cleanup to reduce the amount of transitive dependencies and to make the process of updating them easier in the future:

- Some vulnerabilities are only present in the `netstandard2.0` and `netstandard2.1` targets of some dependencies. Because ABP 8 requires `net8.0` anyway, I assume those targets can be safely dropped as an easy fix.
- The `AssetTargetFallback` can then be dropped from projects too.
- There are bunch of unused dependencies that I removed, but there might be more of them.
- Because IdentityServer4 is deprecated, there is no fix for its vulnerability. According to #11989 it should have been removed by now as OpenIddict became the default in ABP 6.
- I also updated all other outdated dependencies I could find.


Muximize:
I rebased this on latest dev to fix a merge conflict.

MichelZ:
Great initiative, I was just going to open an issue because of transitive package vulnerabilities that I'd like to see fixed 👍

Muximize:
@maliming I see now ABP is usually quite conservative in updating dependencies, so maybe you can skip the last commit, or cherry pick what you like in this PR.

maliming:
Thanks @Muximize 

We will continue to support the `netstandard2.0 and netstandard2.1`. And packages have been upgraded in https://github.com/abpframework/abp/pull/20960

I will cherry-pick your commit to a new Pull Request. 👍

PR #20565, opened 2024-08-20
Show an error page if libs folder is empty.

### LINKED ISSUE #20559: Show an error page if libs folder is empty or bundling is not done
Check if `wwwroot/libs` is empty; then show an error page.
In the error page; write that the developer has not run `abp install-libs` therefore the client-side packages are not installed.
You can redirect to the docs 👉 https://abp.io/docs/latest/CLI#install-libs
Make this check only once when the application starts up.
You can check in both dev & prod environment.

Error page can be like this or better one
image

---

Sometimes developers also forget to run `abp bundle`!
You can check if bundle is created or not; then show warning at startup and redirect to the docs 👉 https://abp.io/docs/latest/CLI#bundle

---

Note; this implementation can be done in all templates except Angular...

---

An example issue is https://github.com/abpframework/abp/issues/12995

maliming:
Resolve #20559

The `_checkLibsTask` will cache the task. It will execute once and won't affect performance.

---

image


ebicoglu:
Updated the error message

image

framework/src/Volo.Abp.AspNetCore.Mvc/Volo/Abp/AspNetCore/Mvc/Libs/AbpMvcLibsService.cs
```diff
@@ -0,0 +1,78 @@
...
+            _checkLibsTask = CheckLibsAsync(httpContext);
+        }
+
+        return _checkLibsTask;
+    }
+
+    protected virtual Task<bool> CheckLibsAsync(HttpContext httpContext)
+    {
+        var fileProvider = new PhysicalFileProvider(httpContext.RequestServices.GetRequiredService<IWebHostEnvironment>().WebRootPath);
+        var libsFolder = fileProvider.GetDirectoryContents("/libs");
```
ebicoglu: If this code is likely to throw an exception, for example if the `libs` folder cannot be retrieved due to file system permissions, you need to handle the exception.

EngincanV (approved):

ebicoglu (approved):

mtozlu:
I throws this exception even if its a /connect request. I think "libs" check should only be applied for non api requests.

maliming:
hi @mtozlu 

The check is only done once during the app's start. So there is no request.

mtozlu:
There's a middleware which suggests it is checked in each request. Am I missing something? 
https://github.com/abpframework/abp/pull/20565/files#diff-1e2878883fe3b0732c9e783d6f673c86e49d5c02670696aaf6cb3f0e10b9ceb0R29

I encountered this problem where I get this "libs folder" error page in response to /connect token request.

maliming:
hi @mtozlu 

Sorry, I misremembered.

I will add some checks to skip the `Ajax` call. 

Thanks

mtozlu:
Thanks for the quick addition ^^

PR #20716, opened 2024-09-03
Add Specific Error Codes for Authentication Failures

alihdev:
Hi, 
Currently, all authentication failures return the same error code (InvalidGrant), making it difficult to distinguish between different failure reasons on the front end. 
Introducing specific error codes (**account_locked**, **account_inactive**, and **invalid_grant**) will allow us to check whether an account is locked out, inactive, or if the credentials are invalid, enabling the front end to handle these scenarios more effectively.


alihdev:
@maliming, could you take a look at this?

maliming:
hi @alihdev 

The code is from the OpenIddict repo.

https://github.com/openiddict/openiddict-core/blob/dev/sandbox/OpenIddict.Sandbox.AspNetCore.Server/Controllers/AuthorizationController.cs#L486-L487

Thanks.

maliming:
hi @alihdev 

Thanks. I will check it.

maliming (approved):

PR #20725, opened 2024-09-03
Update and fixes es.json

washyn:
There are validation fields with incorrect messages and numbering for the interpolation of field names, for example the key shows "ThisFieldOnlyAcceptsFilesWithTheFollowingExtensions:{0}": "This field only accepts files with the following extensions: {1}", and it should be "ThisFieldOnlyAcceptsFilesWithTheFollowingExtensions:{0}": "This field only accepts files with the following extensions: {0}".

### Description

Spanish translation there are validation fields with incorrect messages and numbering for the interpolation of field names.

image

### How to test it?

Change localization of aplication angular to spanish lang and show this issue when field has max length validation or min length validation.


maliming (approved):

maliming:
Thanks @washyn 

Can you sign the CLA?

washyn:
Yeah, sure

PR #20766, opened 2024-09-07
Added support for single and double quotes

### LINKED ISSUE #20681: Angular: single and double quote are filtered out on localization interpolation

### Description

Single and double quotes are removed around value marker in localization string.

We want to mark/highlight dynamic values in localized text like: "You email 'Welcome to the team' has been sent". This is especially important when the sentence structure is different in other languages.

Currently all the following localization strings produce the same output:
`"The name {0} is already used"`
`"The name '{0}' is already used"`
`"The name \"{0}\" is already used"`

Our workaround is manually add the ' or " around the value before interpolation, but it should be a localization detail.

note: [interpolation function](https://github.com/abpframework/abp/blob/ccfdf5919bf4604b73027bf884c77bca17ae8ca4/npm/ng-packs/packages/core/src/lib/utils/string-utils.ts#L19-L23)

### Version

8.1.3

### User Interface

Angular

### Database Provider

EF Core (Default)

None (Default)

### Operation System

Windows (Default)

vivebhagat:
### Description

Resolves #20681 

TODO: modified the string utility function which was used for interpolation. function is replacing single and double quotes from the string. 

### Checklist
- [x] I fully tested it as developer / designer and created unit / integration tests - There is no test included. The utility was tested out of the environment with standard JS console which reflects bug and its solution.
- [x] I documented it (or no need to document or I will create a separate documentation issue) - Not required

### DIFF
--- a/npm/ng-packs/packages/core/src/lib/utils/string-utils.ts
+++ b/npm/ng-packs/packages/core/src/lib/utils/string-utils.ts
@@ -17,8 +17,10 @@ export function createTokenParser(format: string) {
 }
 
 export function interpolate(text: string, params: string[]) {
-  return text
-    .replace(/(['"]?\{\s*(\d+)\s*\}['"]?)/g, (_, match, digit) => params[digit] ?? match)
+ return text
+    .replace(/(['"])?\{\s*(\d+)\s*\}\1/g, (_, quote, digit) => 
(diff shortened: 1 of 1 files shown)


masum-ulu:
Hi @vivebhagat thank you for your contribution

masum-ulu (approved):

PR #20772, opened 2024-09-09
Add Localization.

fahrigedik:
Add Localization

Related : https://github.com/volosoft/vs-internal/issues/4571


salihozkara (changes requested):
The keys are too general and the values is too specific. Please edit your keys or their value

salihozkara (approved):

PR #20792, opened 2024-09-11
feat(blazor): html ui message alert

### LINKED ISSUE #19785: Blazor UiMessageAlert with html message

At the moment, UiMessageAlert only takes strings. And by now I coult not figure out how to pass in a multiline string. Therefore, long text looks horrible.
In addition to that it may be interesting to punch in something more formatted.

`framework/src/Volo.Abp.BlazoriseUI/Components/UiMessageAlert.razor`

### option a: overloads here for html:
image

### option b: additional property in `UiMessageOptions` 
Something like `IsMessageHtmlMarkup` which defaults to false.

### UI could then be something like:

```razor
<Paragraph TextAlignment="TextAlignment.Center" Margin="Margin.Is0.FromBottom" Style="white-space: break-spaces;">
@if(!IsMessageHtmlMarkup) 
{
    @Message
}
else
{
    @((MarkupString)Message)
}

</Paragraph>
```

https://github.com/abpframework/abp/blob/f93a01f9a2996871fe00f15fedac958b91101c64/framework/src/Volo.Abp.BlazoriseUI/Components/UiMessageAlert.razor

### Additional context

_No response_

nebula2:
added property to control whether the message is html or not

### Description

Resolves #19785

no breaking change, as default for bool is false.
implements feature described in the issue.

### unclarity

I am not sure if `UiMessageOptions` is used somewhere else. Couldn't find it anywhere else. But that's the only thing that bothers me here.

### DIFF
--- a/framework/src/Volo.Abp.AspNetCore.Components/Volo/Abp/AspNetCore/Components/Messages/UiMessageOptions.cs
+++ b/framework/src/Volo.Abp.AspNetCore.Components/Volo/Abp/AspNetCore/Components/Messages/UiMessageOptions.cs
@@ -49,4 +49,9 @@ public class UiMessageOptions
     /// Custom icon for the Cancel button.
     /// </summary>
     public object? CancelButtonIcon { get; set; }
+
+    /// <summary>
+    /// Whether the message is html (true) or plain text (false)
+    /// </summary>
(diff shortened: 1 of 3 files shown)


maliming (approved):

maliming:
Thanks @nebula2

PR #20847, opened 2024-09-20
change dotnetRunMessages to bool for launchSettings.json

wangj90:
### Description

change dotnetRunMessages to bool for launchSettings.json
some templates' launchSettings.json has a dotnetRunMessages field of string type,which may causes an error when using .NET Aspire.

### Checklist

image
image


wangj90:
@maliming @realLiangshiwei @hikalkan Any questions about that?

maliming (approved):

maliming:
Thanks @wangj90

Issue #20864, opened 2024-09-23
Languages found in /dev missing in releases


improwise:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

It seems like there are several languages translated in /dev branch that has never made it into releases, even though they have been merged with /dev several months ago. I can find for example  "cs", "is" and "sv" even though they are in in /dev since a few months back like

https://github.com/abpframework/abp/tree/dev/abp_io/AbpIoLocalization/AbpIoLocalization/Account/Localization/Resources

I would imagine that they are supposed to make it into the releases and that you are not intended to add them manually once they have been merged into /dev?

Have tried creating new projects both for Angular and Blazor with both free and commercial versions and they seem to be missing  in everyone so I assume this is not related to a specific template. 

Thanks. 

### Version

8.3

### User Interface

Common (Default)

### Database Provider

EF Core (Default)

### Tiered or separate authentication server

None (Default)

### Operation System

Windows (Default)

improwise:
Just to clarify, this isn't really a bug but more of a missing feature or whatever we should call it, does not seem like I can change that now.

maliming:
hi @improwise 

www.abp.io has disabled the localization feature. It only supports English now.

I'm very sorry about this.

improwise:
> hi @improwise
> 
> [www.abp.io](http://www.abp.io) has disabled the localization feature. It only supports English now.
> 
> I'm very sorry about this.

This is about the ABP Framework and solutions created using that, not the ABP website :)

maliming:
hi

The changes merged to `dev`, and the `dev` target to abp 9.0

Current releases are target 8.x

This is the reason. I have picked the commit to rel-8.3. so it will available in next 8.x patch version.

Thanks.

improwise:
Thank you!

PR #20918, opened 2024-09-28
[8.3] Translated to Swedish (sv)

maliming:
https://github.com/abpframework/abp/issues/20864#issuecomment-2380351975


realLiangshiwei (approved):

improwise:
@maliming Just created a new project using ABP 9 and there seem to be no traces of Swedish in there even though I know it was available in 8.X (at least I am pretty sure I saw Swedish being an option when I created a sample project)

image

Edit:

Seems like the activation on Swedish is missing, as in 

            options.Languages.Add(new LanguageInfo("sv", "sv", "Svenska"));

maliming:
hi

I updated the `AbpLocalizationOptions
` then re-migrate the database.

image

improwise:
> hi
> 
> I updated the `AbpLocalizationOptions ` then re-migrate the database.

Yes, this is what we did as mentioned above, but the necessary code for this to work should be included from get go and as you can see in the menu to the left, there are still translations missing as discussed in the other thread about this that I have since created. Probably best to use that one. Thanks.

maliming:
https://github.com/abpframework/abp/issues/21473#issuecomment-2513441146

PR #20934, opened 2024-10-01
Correct wrong CLI syntax

minhtaile2712:
Correct wrong CLI syntax


realLiangshiwei:
Hi,

It's correct; this is the new ABP Studio CLI.

image

PR #20942, opened 2024-10-01
Keep net8.0 packages while targetting net8.0

enisn:
Used `Contains()` instead `.Equals()` since maui target frameworks are different like `net9.0-android`, `net9.0-ios` etc...


enisn:
Need to declare a property
See https://github.com/microsoft/MSBuildSdks/issues/251#issuecomment-973164705

enisn:
Not necessarily for now

PR #21089, opened 2024-10-15
Use `Random.Shared` if available.

maliming:


MarkCiliaVincenti:
There's a few issues here which I outlined in #20531.

maliming:
https://github.com/abpframework/abp/pull/21149

PR #21107, opened 2024-10-17
Fix Hangfire BackgroundJobServerOptions injection

gdlcf88:


maliming:
See https://github.com/HangfireIO/Hangfire/blob/main/src/Hangfire.AspNetCore/HangfireApplicationBuilderExtensions.cs#L75

PR #21149, opened 2024-10-22
Use `Random.Shared` if available.

maliming:


MarkCiliaVincenti:
Ideally don't lock on the random instance. Check the logic in ThreadSafeRandomizer

enisn (approved):

PR #21156, opened 2024-10-23
Do not return the smtp password to the UI settings page.

maliming:


EngincanV (approved):

oykuermann (changes requested):
I can see the password when I change the item type with text: 
image

oykuermann (approved):

PR #21198, opened 2024-10-28
Update abp-8-3.md

maliming:
https://github.com/abpframework/abp/pull/20012#issuecomment-2438383059


EngincanV (approved):

LuisPignataro:
Hi @maliming @EngincanV 

This change continues to affect subsequent versions, and the effects are not fully explained. When related entities are read through a navigation property, the concurrency stamps of all entities are updated even if there are no changes. This causes unnecessary updates in the database. The documentation states that it can only affect tests; however, the impact in production is very significant, and the documentation for this change is insufficient. Our team spent many hours debugging to find the cause of this problem.

PR #21206, opened 2024-10-29
Make sure `Value` of `FixedLocalizableString` not null.

maliming:
Because the `Name` of [LocalizedString](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.localization.localizedstring?view=net-8.0) can't be `null`.

https://abp.io/support/questions/8169/Bug-in-Localization-logic-causes-feature-management-to-fail

### DIFF
--- a/framework/src/Volo.Abp.Localization/Volo/Abp/Localization/LocalizableStringSerializer.cs
+++ b/framework/src/Volo.Abp.Localization/Volo/Abp/Localization/LocalizableStringSerializer.cs
@@ -35,7 +35,12 @@ public class LocalizableStringSerializer : ILocalizableStringSerializer, ITransi
 
     public virtual ILocalizableString Deserialize(string value)
     {
-        if (value.IsNullOrEmpty() ||
+        if (value == null)
+        {
+            throw new AbpException($"{nameof(value)} can not be null!");
+        }
+
+        if (value.IsNullOrWhiteSpace() ||
             value.Length < 3 ||
             value[1] != ':')
         {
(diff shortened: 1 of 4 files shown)


framework/src/Volo.Abp.Localization/Volo/Abp/Localization/LocalizableStringSerializer.cs
```diff
@@ -35,7 +35,12 @@ public LocalizableStringSerializer(IOptions<AbpLocalizationOptions> localization
 
     public virtual ILocalizableString Deserialize(string value)
     {
-        if (value.IsNullOrEmpty() ||
+        if (value == null)
```
ahmednfwela, line 38: There are cases where the input is actually null, e.g. when defining a Feature via `FeatureDefinationProvider`, one can provide a name, but not a description

for example, see : 
https://github.com/abpframework/abp/blob/c0893cf0feef1d54e60b23ddec8e26bc92449613/framework/src/Volo.Abp.Features/Volo/Abp/Features/FeatureGroupDefinition.cs#L50-L72

maliming: Thanks. I will check this case.

maliming: https://github.com/abpframework/abp/pull/21206/commits/40e3e315bcd3981b1a6df5b519b9510bf7f6ece3

EngincanV (approved):

PR #21213, opened 2024-10-30
Add `ProxyHttpClientPreSendActions` to `AbpHttpClientOptions`.

### LINKED ISSUE #21205: Http DynamicProxying  idea

Will the HTTP dynamic proxy consider defining the request timeout on the interface? Because in actual use, we may need to set different timeouts for different remote methods.

In the current configuration, a unified timeout can only be set for HttpClient in the AbpHttpClientBuilderOptions configuration, and cannot be set for a specific remote method name. I'm wondering if I can use Attribute to mark it, or leave some expansion possibilities when proxying client requests?
image

### Additional context

_No response_

iiiixxl:
### Description

Resolves #21205(write the related issue number if available)

1. Try to add an AbpHttpClientExecuteHttpActionOptions class to expose the current request method and httpclient for custom configuration at startup.
image

 3. Inject options into ClientProxyBase, and add configuration execution code in the RequestAsync method that actually initiates the request.

image

5. Code for unit tests is defined, but some problems are encountered while executing unit tests
- The execution here cannot be successful because the service started by the unit test does not use SocketsHttpHandler for request processing. It uses the ClientHandler class under the Microsoft.AspNetCore.TestHost namespace. 
b4ce1d7a-bafc-48b5-a1a3-b9d7750814df
- As a result, IsCancellationRequested cannot throw an exception in the closed state after timeout.
So the desired result cannot be obtained locally. at the same time,
I can achieve the desired effect by using other services locally
img_v3_02g5_2c2c2792-1a6f-4c8b-b629-1a74898f90dg

### Checklist

- [√ ] I fully tested it as developer / designer and created unit / integration tests
- [√ ] I documented it (or no need to document or I will create a separate documentation issue)

### How to test it?
When the test times out, you may need to start a normal service to make the call. The problem described in point 5 occurs when using the test service.


iiiixxl:
Why does building automated tests fail?

maliming (approved):

maliming:
hi @yuxl01 

I have moved `ExecuteHttpActio`n to `AbpHttpClientOptions` and fixed the unit test.

PR #21534, opened 2024-12-03
Update pre-requirements in the documents.

maliming:
@masum-ulu Can you update the `pre-requirements` in angular document?

https://github.com/abpframework/abp/blob/pre-requirements/docs/en/framework/ui/angular/quick-start.md

https://github.com/abpframework/abp/blob/pre-requirements/docs/en/framework/ui/react-native/index.md


EngincanV (approved):

masum-ulu (changes requested):
Node version can be `20.11+`  I'll update changes like that

masum-ulu (approved):

PR #21660, opened 2024-12-17
Enable object-mapping when source and destination is an enum

### LINKED ISSUE #21659: Object mapping does not work when source and destination are of type enum

### Description

When using IObjectMapper, mapping a source enum to a destination enum does work as expected.
The result is always the default value of the destination enum.
Using Automapper directly, the mapping works.

### Reproduction Steps

public enum MyEnum { Value1, Value2 };
public enum MyEnumDto { Value1, Value2 };
...
CreateMap<MyEnum, MyEnumDto>();
..

var mapper = GetRequiredService<IObjectMapper>();
MyEnumDto dto = mapper.Map<MyEnum, MyEnumDto>(MyEnum.Value2);
// dto is always MyEnumDto.Value3

### Expected behavior

object mapper should return the correct enum value

### Actual behavior

object mapper always returns the default enum value

### Regression?

_No response_

### Known Workarounds

_No response_

### Version

9.0.2

### User Interface

Common (Default)

### Database Provider

EF Core (Default)

None (Default)

### Operation System

Windows (Default)

### Other information

_No response_

MarkusHoz:
Before always the default value of the enum was returned from the Map function

### Description

Resolves #21659

Object mapping does not work when source and destination are of type enum.
I added an additional check to DefaultObjectMapper.Map function that fixes this problem.

### Checklist

### How to test it?

Test: AbpAutoMapperModule_Basic_Tests.Should_Map_Enum()

### DIFF
--- a/framework/src/Volo.Abp.ObjectMapping/Volo/Abp/ObjectMapping/DefaultObjectMapper.cs
+++ b/framework/src/Volo.Abp.ObjectMapping/Volo/Abp/ObjectMapping/DefaultObjectMapper.cs
@@ -55,13 +55,9 @@ public class DefaultObjectMapper : IObjectMapper, ITransientDependency
                 return specificMapper.Map(source);
             }
 
-            if (!(typeof(TSource).IsEnum || typeof(TSource).IsPrimitive))
+            if (TryToMapCollection<TSource, TDestination>(scope, source, default, out var collectionResult))
             {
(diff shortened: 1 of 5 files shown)


maliming:
hi @MarkusHoz 

Thank you for your contribution, I have changed some code: https://github.com/abpframework/abp/pull/21660/commits/803159bc1f716b69bf40362285d1ea5003b9c1ea

maliming (approved):

PR #21664, opened 2024-12-17
make methods of local distributed eventbus virtual

tntwist:
### Description

I need to override several methods of the local distributed eventbus.
So i just made all methods virtual to allow overriding necessary methods.

### Checklist

- [ x] I fully tested it as developer / designer and created unit / integration tests
- [x ] I documented it (or no need to document or I will create a separate documentation issue)

### How to test it?

There should be no testing needed, since it does not change anything in the framework. It just gives possibilities for customizing some logics.


nebula2:
finally!

hikalkan (approved):

PR #21676, opened 2024-12-19
Return jQuery object instead of empty string for invisible row action.

### LINKED ISSUE #21674: Datatable Actions Not Rendering

### Description

If there is only one action and this action is not visible in some rows, then the entire column becomes invisible.
Repro project link is below.

### Reproduction Steps

Create a sample project with studio (mvc / ef core / sample crud page). Change the index.js of books page like below. 

```js
{
    title: l('Actions'),
    rowAction: {
        items:
            [
                {
                    text: l('Edit'),
                    visible: function (data) {
                        console.log(data.type);
                        //todo:actions column should be visible if there is any visible actions row.
                        return (abp.auth.isGranted('SingleAction.Books.Edit') && data.type == 3);
                    },
                    action: function (data) {
                        editModal.open({ id: data.record.id });
                    }
                },
            ]
    }
}
```

### Expected behavior

Actions column should be visible if there is any visible actions row.

### Actual behavior

If there is only one action and this action is not visible in some rows, then the entire column becomes invisible.

### Regression?

repro project [https://we.tl/t-tq18dh5OVZ](url)

### Known Workarounds

_No response_

### Version

9.0.2

### User Interface

MVC

### Database Provider

EF Core (Default)

None (Default)

### Operation System

Windows (Default)

### Other information

_No response_

maliming:
### Description

Resolves #21674

Single invisible row action:

image

Multiple invisible row actions:

image


muhlisatac:
Thank you. I will try it.

EngincanV (approved):

PR #21690, opened 2024-12-20
Revise `Pre-Requirements` document.

### LINKED ISSUE #21686: Revise `Pre-Requirements` document.
pre-requirements document is not complete. For example, microservice has more dependencies which are not mentioned in that document.

Also, that document's problem: not everything is required for every application type. For example, Helm is not needed for open source templates, and also app-pro-nolayers. It should state these differences.

 for now, add helm, nginx and mkcert to that document but state that these are only needed for microservice template.

maliming:
Resolve #21686

### DIFF
--- a/docs/en/get-started/pre-requirements.md
+++ b/docs/en/get-started/pre-requirements.md
@@ -2,11 +2,11 @@
 
 This document will guide you through preparing your development environment for ABP based application development.
 
-> The prerequisites mentioned in this document are not all necessary. It based on your project type.
+> The prerequisites mentioned in this document are not all necessary. It is based on your project type.
 
 * You don't need to install the EF Core CLI if your application uses MongoDB instead of EF Core.
 * You don't need to use Redis(distributed cache) if your application has only one instance. See [When to Use a Distributed Cache Server](../kb/when-to-use-a-distributed-cache-server.md).
-* You don't need to install Helm, NGINX Ingress, or mkcert if you are developing a non-microservice application.
+* You don't need to install Docker, Helm, NGINX Ingress, or mkcert if you are developing a non-microservice application.
 * The `README.md` file in the solution will contain specific requirements for the current solution. Please refer to the `README.md` file of the solution.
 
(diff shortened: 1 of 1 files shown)


docs/en/get-started/pre-requirements.md
```diff
@@ -70,3 +77,33 @@ ABP startup solution templates and tools use some PowerShell scripts (`*.ps1`) t
 * [Install PowerShell on Windows](https://learn.microsoft.com/en-us/powershell/scripting/install/installing-powershell-on-windows)
 * [Install PowerShell on macOS](https://learn.microsoft.com/en-us/powershell/scripting/install/installing-powershell-on-macos)
 * [Install PowerShell on Linux](https://learn.microsoft.com/en-us/powershell/scripting/install/installing-powershell-on-linux)
+
+## MicroService Solution
+
+The following tools are only required to develop ABP's [microservice solution](../solution-templates/microservice/index.md)
+
+### Docker Engine or Docker Desktop
```
hikalkan: Docker is not only needed for microservice. It is also required for the app-pro template. Also, a docker section already exists above. No need to repeat it here.

docs/en/get-started/pre-requirements.md
```diff
@@ -2,6 +2,13 @@
 
 This document will guide you through preparing your development environment for ABP based application development.
 
+> The prerequisites mentioned in this document are not all necessary. It is based on your project type.
+
+* You don't need to install the EF Core CLI if your application uses MongoDB instead of EF Core.
+* You don't need to use Redis(distributed cache) if your application has only one instance. See [When to Use a Distributed Cache Server](../kb/when-to-use-a-distributed-cache-server.md).
```
hikalkan: No need to mention about Redis since redis is coming as docker container in our templates, like RabbitMQ and other dependencies.

hikalkan (changes requested):

hikalkan (approved):

PR #21730, opened 2024-12-25
Add mobile application development tutorial to documentation tree

EngincanV:


docs/en/docs-nav.json
```diff
@@ -112,6 +112,23 @@
             {
               "text": "10: Book to Author Relation",
               "path": "tutorials/book-store/part-10.md"
+            },
```
hikalkan: Do not put this under the bookstore tutorial. Instead, add jus before the community articles (as the last tutorial)

hikalkan (changes requested):

docs/en/docs-nav.json
hikalkan: here: 
image

hikalkan (approved):

PR #21774, opened 2024-12-30
Add new localizations to account resource

berkansasmaz:


maliming (approved):

maliming:
hi @berkansasmaz 

I remember that we no longer supported multiple languages in www.abp.io website. It only supports English

berkansasmaz:
> hi @berkansasmaz
> 
> I remember that we no longer supported multiple languages in [www.abp.io](http://www.abp.io) website. It only supports English

Yes, this is true, but there is no disadvantage in adding it :)

PR #21869, opened 2025-01-09
Set `X509KeyStorageFlags` for `IIS` environment.

maliming:
Resolve https://github.com/volosoft/volo/issues/19150

### DIFF
--- a/modules/openiddict/src/Volo.Abp.OpenIddict.AspNetCore/Microsoft/Extensions/DependencyInjection/NativeMethods.cs
+++ b/modules/openiddict/src/Volo.Abp.OpenIddict.AspNetCore/Microsoft/Extensions/DependencyInjection/NativeMethods.cs
@@ -1,0 +1,22 @@
+using System;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.Extensions.DependencyInjection;
+
+/// <summary>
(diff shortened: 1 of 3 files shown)


ebicoglu:
I have another solution which is more simple. Put a setting in the `web.config` file:

```
OpenIddictCertificateFlags: "X509KeyStorageFlags.MachineKeySet;X509KeyStorageFlags.EphemeralKeySet"
```

so if you can read this configuration it means application is using  `web.config` file which is only available via IIS hosting.
so you can understand if the app is running on IIS.
besides, users can give any other flags if they want.

maliming:
hi @ebicoglu 

What do you think of this suggestion?

Should we move the `OpenIddictCertificateFlags` to `appsettings.json`?

https://learn.microsoft.com/en-us/aspnet/core/host-and-deploy/iis/web-config?view=aspnetcore-9.0#configuration-sections-of-webconfig

image

ebicoglu:
`appsettings.json` is being used both in Kestrel and IIS. but in our problem we only need to solve this problem on IIS. we don't have a problem with Kestrel.
if you put the setting in `appsettings.json` then you cannot understand if the hosting is IIS or Kestrel.
that's why you need to put this setting in web.config.
I don't want to include the setting in appsetting.json file by default which is maybe not needed for Kestrel (maybe even makes new problems)

ebicoglu:
@maliming I'm cancelling this PR because the solution is little bit complex. we can write an article to fix Openiddict IIS publish problem.

PR #21875, opened 2025-01-10
Update AbpTabTagHelperService.cs.  Fix content-class attributes to use the .Value property

ejohnson-dotnet:
…e the .Value property

This needs to use the .Value property of the 'content-class' attribute, just like on line 37 where it uses the .Value property of the 'header-class' attribute.

### Description

TODO: Describe what this PR has changed, add screenshot or animated GIF if available, write if it is a breaking change, and how to fix the breaking changes for existing applications if so.

### Checklist

### How to test it?

Add the following tag
```
<abp-tabs>
<abp-tab content-class="border">
</abp-tab>
</abp-tabs>
```
It should display a border around the content and add "border" to the <div class="border"> for the tab conent.


maliming (approved):

maliming:
Thanks @ejohnson-dotnet

PR #21940, opened 2025-01-20
Replace toastr with a new implementation.

### LINKED ISSUE #21867: Replace `toastr` with a new library.
Since it was last published: **7 years ago** and there are **vulnerabilities.**

https://www.npmjs.com/package/toastr

https://abp.io/docs/latest/framework/ui/mvc-razor-pages/javascript-api/notify

--- 

Consider copying the angular code.

https://github.com/abpframework/abp/blob/dev/npm/ng-packs/packages/theme-shared/src/lib/components/toast/toast.component.html#L0-L1

maliming:
Resolves #21867

The new implementation refers to Angular. **Doesn't depend on third-library.**

https://github.com/abpframework/abp/blob/dev/npm/ng-packs/packages/theme-shared/src/lib/services/toaster.service.ts

screenshot-2025-01-19T20-20-11


EngincanV (approved):

edizaziz-dev:
Hi nice work on replacing this library. Toastr had events like onHidden. Will these be added soon?

Issue #21950, opened 2025-01-21
Volo.Abp.Identity module client evaluation


842549829:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Description

abp 8.3 Upgrade to 9.0.3  
Error trying to log in
`@__organizationUnitIds_0' could not be translated. Additional information: Primitive collections support has not been enabled. Either rewrite the query in a form that can be translated, or switch to client evaluation explicitly by inserting a call to 'AsEnumerable', 'AsAsyncEnumerable', 'ToList', or 'ToListAsync'. See https://go.microsoft.com/fwlink/?linkid=2101038 for more information.
System.InvalidOperationException: The LINQ expression '@__organizationUnitIds_0' could not be translated. Additional information: Primitive collections support has not been enabled. Either rewrite the query in a form that can be translated, or switch to client evaluation explicitly by inserting a call to 'AsEnumerable', 'AsAsyncEnumerable', 'ToList', or 'ToListAsync'. See https://go.microsoft.com/fwlink/?linkid=2101038 for more information.
   at Pomelo.EntityFrameworkCore.MySql.Query.Internal.MySqlQueryableMethodTranslatingExpressionVisitor.TranslatePrimitiveCollection(SqlExpression sqlExpression, IProperty property, String tableAlias)
   at Microsoft.EntityFrameworkCore.Query.RelationalQueryableMethodTranslatingExpressionVisitor.TranslateParameterQueryRoot(ParameterQueryRootExpression parameterQueryRootExpression)
   at Microsoft.EntityFrameworkCore.Query.QueryableMethodTranslatingExpressionVisitor.VisitExtension(Expression extensionExpression)
   at Microsoft.EntityFrameworkCore.Query.RelationalQueryableMethodTranslatingExpressionVisitor.VisitExtension(Expression extensionExpression)
   at Microsoft.EntityFrameworkCore.Query.QueryableMethodTranslatingExpressionVisitor.VisitMethodCall(MethodCallExpression methodCallExpression)
   at Microsoft.EntityFrameworkCore.Query.RelationalQueryableMethodTranslatingExpressionVisitor.VisitMethodCall(MethodCallExpression methodCallExpression)
   at Microsoft.EntityFrameworkCore.Query.QueryableMethodTranslatingExpressionVisitor.Translate(Expression expression)
   at Microsoft.EntityFrameworkCore.Query.QueryableMethodTranslatingExpressionVisitor.TranslateSubquery(Expression expression)
[...]

PR #21952, opened 2025-01-21
fix(IdentityUserStore): issues 21950

842549829:
### Description

fix issues 21950


maliming:
See https://github.com/abpframework/abp/issues/21950#issuecomment-2603925716

PR #22007, opened 2025-01-26
Fix Pagination Previous Button Behavior

magicjar:
### Description

The calculation for currentPage in GetPreviousButtonAsync assumes that the TagHelper.Model.CurrentPage is always valid. If TagHelper.Model.CurrentPage becomes 0 or -1, the disabled class will not be applied correctly, leading to the Previous button being enabled even when it should be disabled.

Currently, the logic to disable the Previous button only considers whether CurrentPage == 1. There is no safeguard for invalid page values (e.g., 0, -1).

This bug only occurs on routes with an ID parameter, such as homepage.com/movie/{id}/reviews?currentPage=1, where {id} is used to fetch reviews for a specific movie. However, on the "movie" index route (homepage.com/movie?currentPage=1), the pagination renders correctly.


maliming (approved):

maliming:
Thanks @magicjar

PR #22010, opened 2025-01-27
Trim whitespace from username and password during sign-in

### LINKED ISSUE #21915: Should ignore spaces at the beginning and end of password when logging in
https://abp.io/support/questions/8663/Trailing-space-in-passwords

realLiangshiwei:
### Description

Resolves https://github.com/abpframework/abp/issues/21915

### Checklist

### How to test it?

Add space to password when login


maliming (approved):

oykuermann (changes requested):
- [ ] If we should ignore spaces at the beginning and end of password when logging in, then we should do the same rules on **Register, Password change, Account personal info settings**, and every place we can change/enter password and username. If a user has spaces at the beginning and end of password or username, he/she cannot log in to the system. Please apply this to all the UIs.

- [ ] If we have space before username on login screen we cannot log in:

[screen-capture (6).webm](https://github.com/user-attachments/assets/6837f36a-6d94-42a1-86bc-1c4ab4e557c6)

realLiangshiwei:
After checking, ABP allows spaces to be used as part of a password, and this change may cause issues. closed

PR #22034, opened 2025-01-27
Update en.json

halimekarayay:
related PR : https://github.com/volosoft/vs-internal/pull/5893

resolves https://github.com/volosoft/vs-internal/issues/5647


yagmurcelk (changes requested):
- We need to format the bullet points as shown in the second visual.
image
image

- The **Pre-Built Application Modules** page doesn't appear like it does in the second visual.
image
image

- The icon on the **SaaS** page should be colored like in the second visual.
image
image
- In the navigation path, the first item "**Get Started**" should be bold, while "**Others**" should remain normal, as shown in the second example.
image
The correct view:
image

- The "**All Attendees**" text should have this icon next to it.
Icon : https://fontawesome.com/icons/users?f=classic&s=solid 
image

yagmurcelk (approved):

PR #22060, opened 2025-02-02
Significant Bug Fix | Resolve Deadlock Between Autofac and UnnamedOptionsManager

alyssonrpg:
### Description

**Problem:** The default `Microsoft.Extensions.Options.UnnamedOptionsManager` implementation can lead to deadlocks when resolving `IOptions<TOptions>` that have dependencies managed by Autofac. These deadlocks occur due to a lack of order in lock acquisition between:

- `Microsoft.Extensions.Options.UnnamedOptionsManager.get_Value` locking `_syncObj`.
- `Autofac.Core.Lifetime.LifetimeScope.CreateSharedInstance` locking `_synchRoot`.

**Proposed Solution:** This pull request introduces the `AbpAutofacUnnamedOptionsManager<TOptions>` class, which replaces the default implementation with an atomic operation instead of a pessimistic lock. This synchronizes the concurrent creation of a new instance of `TOptions` without causing deadlocks.

### Checklist

### How to test it?

To validate this fix and reproduce the original deadlock issue:

1. **Run the Test:** Execute the unit test `Volo.Abp.Autofac.Autofac_Dead_Lock_Tests`, specifically designed to simulate the deadlock scenario.

2. **Modify the Code:** Before running the test, remove or comment out line **34** in the method `Volo.Abp.AbpAutofacAbpApplicationCreationOptionsExtensions.AddAbpAutofacUnnamedOptionsManager`. 

```csharp
   // options.Services.Replace(ServiceDescriptor.Singleton(typeof(IOptions<>), typeof(AbpAutofacUnnamedOptionsManager<>)));
```
   
3. **Observe the Deadlock:** With this modification, running the test should reproduce the deadlock,

4. **Apply the Fix:** Restore line 34 to reintroduce the AbpAutofacUnnamedOptionsManager. Run the test again, and you should observe that the deadlock no longer occurs.

### DIFF
--- a/framework/src/Volo.Abp.Autofac/Microsoft/Extensions/Hosting/AbpAutofacHostBuilderExtensions.cs
+++ b/framework/src/Volo.Abp.Autofac/Microsoft/Extensions/Hosting/AbpAutofacHostBuilderExtensions.cs
@@ -1,6 +1,7 @@
 ﻿using Autofac;
 using Volo.Abp.Autofac;
 using Microsoft.Extensions.DependencyInjection;
+using Volo.Abp;
 
 namespace Microsoft.Extensions.Hosting;
 
@@ -13,6 +14,7 @@ public static class AbpAutofacHostBuilderExtensions
         return hostBuilder.ConfigureServices((_, services) =>
             {
                 services.AddObjectAccessor(containerBuilder);
+                services.AddAbpAutofacUnnamedOptionsManager();
             })
             .UseServiceProviderFactory(new AbpAutofacServiceProviderFactory(containerBuilder));
     }
--- a/framework/src/Volo.Abp.Autofac/Volo/Abp/AbpAutofacAbpApplicationCreationOptionsExtensions.cs
+++ b/framework/src/Volo.Abp.Autofac/Volo/Abp/AbpAutofacAbpApplicationCreationOptionsExtensions.cs
@@ -1,5 +1,7 @@
 ﻿using Autofac;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection.Extensions;
+using Microsoft.Extensions.Options;
 using Volo.Abp.Autofac;
 
 namespace Volo.Abp;
@@ -22,7 +24,13 @@ public static class AbpAutofacAbpApplicationCreationOptionsExtensions
 
         services.AddObjectAccessor(containerBuilder);
         services.AddSingleton((IServiceProviderFactory<ContainerBuilder>)factory);
+        services.AddAbpAutofacUnnamedOptionsManager();
 
         return factory;
+    }
+
+    internal static void AddAbpAutofacUnnamedOptionsManager(this IServiceCollection services)
+    {
+        services.Replace(ServiceDescriptor.Singleton(typeof(IOptions<>), typeof(AbpAutofacUnnamedOptionsManager<>)));
     }
 }
--- a/framework/src/Volo.Abp.Autofac/Volo/Abp/Autofac/AbpAutofacUnamedOptionsManager.cs
+++ b/framework/src/Volo.Abp.Autofac/Volo/Abp/Autofac/AbpAutofacUnamedOptionsManager.cs
@@ -1,0 +1,46 @@
+﻿using System.Threading;
+using Microsoft.Extensions.Options;
+
+namespace Volo.Abp.Autofac;
+
+/// <summary>
+/// The <see cref="AbpAutofacUnnamedOptionsManager"/> class is designed to replace the default
+/// <see cref="Microsoft.Extensions.Options.UnnamedOptionsManager"/> in Microsoft.Extensions.Options.
+/// The default implementation can lead to deadlocks with Autofac when resolving <see cref="IOptions{TOptions}"/>
+/// that have dependencies managed by Autofac.
+///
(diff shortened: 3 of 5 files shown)


maliming:
hi @alyssonrpg 

> The default Microsoft.Extensions.Options.UnnamedOptionsManager implementation can lead to deadlocks when resolving IOptions<TOptions> that have dependencies managed by Autofac.

ABP should not solve this problem because we just use these class libraries. If there is a problem with our usage, we should fix it.

Maybe you should open a PR to Microsoft or Autofac.

Thanks.

alyssonrpg:
> hi @alyssonrpg
> 
> > The default Microsoft.Extensions.Options.UnnamedOptionsManager implementation can lead to deadlocks when resolving IOptions that have dependencies managed by Autofac.
> 
> ABP should not solve this problem because we just use these class libraries. If there is a problem with our usage, we should fix it.
> 
> Maybe you should open a PR to Microsoft or Autofac.
> 
> Thanks.

But shouldn't abp do something about the issue

alyssonrpg:
> hi @alyssonrpg
> 
> > The default Microsoft.Extensions.Options.UnnamedOptionsManager implementation can lead to deadlocks when resolving IOptions that have dependencies managed by Autofac.
> 
> ABP should not solve this problem because we just use these class libraries. If there is a problem with our usage, we should fix it.
> 
> Maybe you should open a PR to Microsoft or Autofac.
> 
> Thanks.

But shouldn't ABP address this issue somehow? Otherwise, the framework is being released with a known issue that can cause deadlocks in both production and test environments (it's happening to the company I work for)

alyssonrpg:
> hi @alyssonrpg
> 
> > The default Microsoft.Extensions.Options.UnnamedOptionsManager implementation can lead to deadlocks when resolving IOptions that have dependencies managed by Autofac.
> 
> ABP should not solve this problem because we just use these class libraries. If there is a problem with our usage, we should fix it.
> 
> Maybe you should open a PR to Microsoft or Autofac.
> 
> Thanks.

@maliming 
But shouldn't ABP address this issue somehow? Otherwise, the framework is being released with a known issue that can cause deadlocks in both production and test environments (it's happening to the company I work for)

maliming:
hi

> AbpSystemTextJsonSerializerOptions -> AbpSystemTextJsonSerializerModifiersOptions -> AbpJsonOptions.

Can you share some code to reproduce it? The business code instead of the unit test code.
Maybe the problem is unrelated to AutoFac and `UnnamedOptionsManager` but ABP.

Thanks.

alyssonrpg:
> hi
> 
> > AbpSystemTextJsonSerializerOptions -> AbpSystemTextJsonSerializerModifiersOptions -> AbpJsonOptions.
> 
> Can you share some code to reproduce it? The business code instead of the unit test code. Maybe the problem is unrelated to AutoFac and `UnnamedOptionsManager` but ABP.
> 
> Thanks.

@maliming I can't share, because it's a private code. But I guess I can code a test that reproduce the deadlock using the real 
 AbpSystemTextJsonSerializerOptions and AbpSystemTextJsonSerializerModifiersOptions options.

The real deadlock I am facing happens with different scenarios, different stack traces, different number of threads, but always with AbpSystemTextJsonSerializerOptions envolved.

I am traveling for business now, in approximately 5 days I will create this new test that simulates the real deadlock I am facing.

alyssonrpg:
> hi
> 
> > AbpSystemTextJsonSerializerOptions -> AbpSystemTextJsonSerializerModifiersOptions -> AbpJsonOptions.
> 
> Can you share some code to reproduce it? The business code instead of the unit test code. Maybe the problem is unrelated to AutoFac and `UnnamedOptionsManager` but ABP.
> 
> Thanks.

@maliming , thank you for your patience.

I've added a test case `framework/test/Volo.Abp.Json.Tests/Volo/Abp/Json/AbpJsonSystemTextJson_DeadLock_Tests.cs` that demonstrates how the concurrent instantiation of AbpSystemTextJsonSerializerOptions can result in a deadlock without the proposed fix.

To observe the deadlock, please remove or comment out line 34 in the `Volo.Abp.AbpAutofacAbpApplicationCreationOptionsExtensions.AddAbpAutofacUnnamedOptionsManager` method.

maliming:
hi

I checked the code.

This problem occurs when we configure an `Options` and try to get another `Options` through dependency injection.

I think ASP NET Core should handle this situation. 

Thanks.

```cs
protected override void AfterAddApplication(IServiceCollection services)
{
    base.AfterAddApplication(services);

    services.AddOptions<MyFirstOptions>()
        .Configure<IServiceProvider>((options, rootServiceProvider) =>
    {
        options.MyValue = "MyFirstOptions";
        options.MySecondOptionsValue = rootServiceProvider.GetRequiredService<IOptions<MySecondOptions>>().Value.MyValue;
    });

    services.AddOptions<MySecondOptions>()
        .Configure<IServiceProvider>((options, rootServiceProvider) =>
        {
            options.MyValue = "MySecondOptions";
        });

    services.AddSingleton(new ThreadSynchronizationEvents());
    services.AddSingleton<TestSingletonService, TestSingletonService>();
}
```

```cs
public class MyFirstOptions
{
    public string MyValue { get; set; }

    public string MySecondOptionsValue { get; set; }
}

public class MySecondOptions
{
    public string MyValue { get; set; }
}

```

PR #22066, opened 2025-02-03
Angular - Access token management fixes

sumeyyeKurtulus:
### Description

Resolves [#18890](https://github.com/volosoft/volo/issues/18890)

AccessTokenTest

### Checklist

### How to test it?

- You can run `yarn copy-to:app` under the `npm > ng-packs` directory and test the cases accordingly.

### DIFF
--- a/npm/ng-packs/packages/core/src/lib/services/local-storage-listener.service.ts
+++ b/npm/ng-packs/packages/core/src/lib/services/local-storage-listener.service.ts
@@ -1,6 +1,5 @@
 import { DOCUMENT } from '@angular/common';
 import { Injectable, inject } from '@angular/core';
-
 @Injectable({
   providedIn: 'root',
 })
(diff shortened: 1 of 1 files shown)


npm/ng-packs/packages/core/src/lib/services/local-storage-listener.service.ts
```diff
@@ -9,8 +8,13 @@ export class LocalStorageListenerService {
 
   constructor() {
     this.window.addEventListener('storage', event => {
-      if (event.key === 'access_token' && event.newValue === null) {
-        this.window.location.reload();
+      if (event.key === 'access_token') {
+        const tokenRemoved = event.newValue === null;
+        const tokenAdded = event.oldValue === null && event.newValue !== null;
```
masum-ulu, line 13: Make sure this variable is really need, i guess we can remove that

sumeyyeKurtulus: When we don't check `tokenAdded` value, it still seems working. However, the user needs to trigger the other tabs as in the video. If this behavior is acceptable, we can proceed like that, but MVC behaves in the opposite way 🤔


https://github.com/user-attachments/assets/a61ccbfe-8441-4b3a-9342-785087398cf1

masum-ulu: Okay, let's let it stay

masum-ulu (approved):

yagmurcelk (approved):

PR #22238, opened 2025-02-27
Show tooltip of on Actions Column if user do not have permissions

### LINKED ISSUE #22205: Hide Actions Column for Users Without Permissions
Currently, the "Actions" column (which includes options like creating, editing, and deleting) is visible to all users in the management interface, regardless of their permissions. If a user does not have the necessary permissions for any of these actions, the column should not be displayed to avoid confusion and maintain a clean UI.

## Expected Behavior:
The "Actions" column should only be visible to users who have at least one of the permissions (create, edit, or delete).

image

image

### LINKED ISSUE #22208: Hide Actions Column for Users Without Permissions
Currently, the "Actions" column (which includes options like creating, editing, and deleting) is visible to all users in the management interface, regardless of their permissions. If a user does not have the necessary permissions for any of these actions, the column should not be displayed to avoid confusion and maintain a clean UI.

This occurs in all three Blazor templates. (Blazor server, blazor web assembly, blazor web app)

## Expected Behavior:
The "Actions" column should only be visible to users who have at least one of the permissions edit, delete or etc).

image

realLiangshiwei:
### Description

Resolves https://github.com/abpframework/abp/issues/22205
Resolves https://github.com/abpframework/abp/issues/22208

image

### Checklist

### How to test it?

Cancel the role's language management permissions and check the actions


m-aliozkaya (approved):

gizemmutukurt (changes requested):
- [x] Mvc/Blazor - I think we should make these changes in the edit button. If you want, I can open a separate issue for this?

image
image

gizemmutukurt (approved):

PR #22342, opened 2025-03-12
Update studio docs for v0.9.25

ekaradev:
### Description

Resolves https://github.com/volosoft/abp-studio/issues/2805

### Checklist


berkansasmaz:
Can you add this version to the version mapping document?

https://github.com/abpframework/abp/blob/rel-9.1/docs/en/studio/version-mapping.md

berkansasmaz (changes requested):
Can you add this version to the version mapping document?

https://github.com/abpframework/abp/blob/rel-9.1/docs/en/studio/version-mapping.md

berkansasmaz (approved):

PR #22465, opened 2025-03-26
Remove account module login link toolbar

m-aliozkaya:
Related https://github.com/volosoft/lepton/issues/2718


maliming:
hi @masum-ulu @enisn 

I think we should add `login link` in the account module instead of Lepton.

Lepton is a theme, it just renders.

maliming (approved):

masum-ulu:
> hi @masum-ulu @enisn
> 
> I think we should add `login link` in the account module instead of Lepton.
> 
> Lepton is a theme, it just renders.

It should already comes only from module instead themes. My opinion is the same for this case, 'cause themes can change but login link will be always exits

PR #22489, opened 2025-03-27
Created angular installation info of for common modules

yekalkan:


m-aliozkaya:
@yekalkan Should we add this configuration for other modules? Like account, identity etc.

m-aliozkaya (approved):

PR #22518, opened 2025-04-01
Delay closing `oauth2-redirect` window.

maliming:


ismcagdas:
Closing becasue we decided to use https://github.com/abpframework/abp/pull/22505

PR #22628, opened 2025-04-11
Disable features if the value is not from the current provider.

### LINKED ISSUE #22621: Reset to default features in better way.
[Berkan Sasmaz](https://discord.com/channels/687243951924838430/687257255137116172/1359835678313283635):

We confirmed this with Malliming — even though the tenant is known on the login screen and tenant-level features are applied, edition-level features are not. Just wanted to share this information here in case it comes up during support or training sessions 🙂 

See: https://github.com/abpframework/abp/issues/22618 

Instead of relying on the "Reset to defaults" button, maybe we could follow a similar logic to roles and users.

For example, if a tenant is assigned to an edition, and some feature values are customized at the edition level — then those features could be marked as non-editable at the tenant level.

This way:

We don’t rely on the "Reset to defaults" button to implicitly re-apply edition settings.
It’s more predictable and avoids confusion — otherwise it might look like a bug or inconsistent behavior from the UI.

maliming:
Resolves #22621


berkansasmaz (approved):

maliming:
How to test:

Use app-pro template project to test:



### 1. Test Case 1:
1. **Create a new edition** and assign some features to this edition.
2. **Create a new tenant** and assign the newly created edition to this tenant.
3. Open the **tenant's features modal**. The features assigned to the edition should be **disabled** for the tenant (i.e., the tenant cannot modify the edition's features).

### 2. Test Case 2:
1. **Create a new tenant** without an edition and assign some features to this tenant.
2. **Create a new edition** and assign the same features to this edition.
3. **Assign the edition** to the tenant.
4. Open the **tenant's features modal**. The tenant's features should be **changeable** (i.e., the tenant can modify the features now).
5. If you **disable a feature** for the tenant, and that feature is assigned to the current edition, then the **features for the edition should be disabled** (i.e., the feature can't be modified from the edition level).

gizemmutukurt (changes requested):
For the user role page, the features coming from the role are shown in parentheses. Editon can also be done for -tenant to be aware of the reason why it is disabled.
image

image

maliming:
> For the user role page, the features coming from the role are shown in parentheses. Editon can also be done for -tenant to be aware of the reason why it is disabled.
> 
> image
> 
> 
> 
> image
> 
> 

Do you have other advice?

maliming:
screenshot-2025-04-10 T16-58-34@2x

gizemmutukurt (approved):

PR #22720, opened 2025-04-21
Use `EntityCacheItemWrapper` to prevent duplicate queries.

### LINKED ISSUE #22715: Cache Penetration

### Description

Please refer to this line of code. If it does not exist in the database, it will keep querying

https://github.com/abpframework/abp/blob/2223df666a1b4a1f10a1789ac8a5a7d3b69312e1/framework/src/Volo.Abp.Caching/Volo/Abp/Caching/DistributedCache.cs#L600
https://github.com/abpframework/abp/blob/2223df666a1b4a1f10a1789ac8a5a7d3b69312e1/framework/src/Volo.Abp.Caching/Volo/Abp/Caching/DistributedCache.cs#L586
### Reproduction Steps

_No response_

### Expected behavior

Return NULL directly without querying the database

### Actual behavior

_No response_

### Regression?

_No response_

### Known Workarounds

_No response_

### Version

9.1

### User Interface

Common (Default)

### Database Provider

EF Core (Default)

None (Default)

### Operation System

Windows (Default)

### Other information

_No response_

maliming:
Resolves #22715

Old behavior:

If `entity` doesn't exist in the database, the `null` will be written to the cache. Next time, it will query the database again.

New behavior:

If `entity` doesn't exist in the database, the `EntityCacheItemWrapper` with `null` will be written to the cache. Next time, it will return `null` instead of querying the database.

### DIFF
--- a/framework/src/Volo.Abp.Ddd.Domain/Volo/Abp/Domain/Entities/Caching/EntityCacheBase.cs
+++ b/framework/src/Volo.Abp.Ddd.Domain/Volo/Abp/Domain/Entities/Caching/EntityCacheBase.cs
@@ -69,11 +69,6 @@ public abstract class EntityCacheBase<TEntity, TEntityCacheItem, TKey> :
 
     public async Task HandleEventAsync(EntityChangedEventData<TEntity> eventData)
     {
-        if (eventData is EntityCreatedEventData<TEntity>)
-        {
-            return;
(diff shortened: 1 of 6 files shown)


hikalkan (approved):

hikalkan:
Thanks. Good enhancement.

PR #22724, opened 2025-04-21
Support get permisons info by group name.

### LINKED ISSUE #22678: Permission query performance issue
Currently, in the permission module, the query for permissions retrieves all permission groups and all definitions under each group at once. Is there a way to query only the definitions under the permission group that the user clicks on?

At present, fetching everything at once causes performance issues — and it becomes even slower when combined with multilingual support.

maliming:
Resolves #22678


hikalkan:
You added a new method, good. But as I understand @shuangbaojun requests (in https://github.com/abpframework/abp/issues/22678) that the permission management dialog should use that new method to lazy load permissions under the active tab clicked by the user. Am I right @shuangbaojun? In that case, this work is bigger than you might think and requires changes in all the related UIs, too.

hikalkan (approved):

PR #22730, opened 2025-04-21
Add other ui options for microservice tutorial

m-aliozkaya:
Related https://github.com/volosoft/vs-internal/issues/5566


m-aliozkaya:
Maybe it could be better to adding this changes to the 9.1 patch as well.

ismcagdas:
@m-aliozkaya I selected different UI's but sample project link doesn't change. We can create sample projects for;

- Blazor WebApp + EF
- Angular + EF

For other options, we can hide the sample download link.

image

ismcagdas:
- [x] I'm testing Angular UI but, here it says Blazor

image

ismcagdas:
- [x] This port changes dynamically. So, we can mention here to run this command using OrderService's port.

image

ismcagdas (changes requested):
I have tested Angular version and completed my review, here are my comments;

https://github.com/abpframework/abp/pull/22730#issuecomment-2866466723
https://github.com/abpframework/abp/pull/22730#issuecomment-2871158848
https://github.com/abpframework/abp/pull/22730#issuecomment-2871261935

Overally, angular document is very good 🙏

gizemmutukurt (changes requested):
- [x] https://github.com/volosoft/abp-studio/issues/3053

- [x] In the _Blazor Server_ training, a Web project screenshot appears to have been added to the module settings.
image
image

- [x] After creating the Entities folder where we added the code from the document, it would be useful to clearly state that we will create a file named Order.cs and write this code into it.
image
image
image
image
image

gizemmutukurt:
- [x] This part is unnecessary, studio adds it automatically. The URL there is causing confusion.

image

gizemmutukurt:
- [x] Blazor wasm - When I run it in visual studio I can't see the analyze.json files.

image
image


- [x] Blazor wasm - Getting an error while generating with suite.
image

- [x] https://github.com/volosoft/volo/issues/20057

gizemmutukurt (changes requested):
- [x] Blazor should be written instead of web

image


- [x] There is no Navigation in .Blazor. It is in the .Client project.
image
image

gizemmutukurt:
- [x] It is not specified where and with what name the code to be added for the page will be added.

image
image

gizemmutukurt:
- [x] In this section, it is not specified which file it will be added to. Also, next to the code I marked, a comment can be stated as in other code blocks where only this line has changed, such as "//change line".

image

gizemmutukurt (changes requested):
- [x] The order of operations in the project has changed with the new version, so we need to update the document for all UIs.

[Apex_1749811342468.webm](https://github.com/user-attachments/assets/0ac81030-56b2-41f6-8b92-ae09d37379b4)

m-aliozkaya:
Moved to https://github.com/abpframework/abp/pull/23176

PR #22809, opened 2025-04-30
Add `ExistAsync` and `FindAsync` methods to `IReadOnlyAppService`.

### LINKED ISSUE #22795: IReadOnlyAppService to support FindAsync or ExistsAsync

When using IReadOnlyAppService there seems to be no way to check if an object exists without getting getting an EntityNotFoundException if it fails.  with ICrudAppService you can use FindAsync and check if it is null.  Could we add an ExistsAsync or add in FindAsync support?

Add ExistsAsync to IReadOnlyAppService (and ICrudAppService for that matter).
(or add FindAsync to IReadOnlyAppService)

### Additional context

_No response_

maliming:
Resolves #22795


maliming:
https://github.com/abpframework/abp/issues/22795#issuecomment-2841157501

PR #22865, opened 2025-05-07
Enhance layout of feature and setting management modals.

maliming:


oykuermann (changes requested):
- [ ] Permission tabs need borders (Applications page):

image

See Users page:
image

oykuermann (approved):

PR #22887, opened 2025-05-12
Assign LoginInput only when null

MichelZ:
When using OpenIddict, there is a way to set a Login Hint on L37:

https://github.com/abpframework/abp/blob/e5c21f6a8bac20c46d8bdb05d18285a866b1f750/modules/account/src/Volo.Abp.Account.Web.OpenIddict/Pages/Account/OpenIddictSupportedLoginModel.cs#L35-L57

However on L56, there is a call to the `base.OnGetAsync()` function which overwrites that value again. 
This PR ensured that the value is not overwritten


maliming (approved):

maliming:
Thanks @MichelZ

PR #22903, opened 2025-05-13
Fix: pdf generation

salihozkara:


modules/docs/src/Volo.Docs.Common.Application/Volo/Docs/Common/Documents/DocumentPdfGeneratorAppService.cs
```diff
@@ -24,6 +26,27 @@ public DocumentPdfGeneratorAppService(
...
     {
         var project = await ProjectRepository.GetAsync(input.ProjectId, includeDetails: true);
+        
+        // https://github.com/abpframework/abp/blob/e96f601641ab8a4bb7d704d3b9df2c00517d96f6/modules/docs/src/Volo.Docs.Application/Volo/Docs/Documents/DocumentAppService.cs#L73
+        var inputVersionStringBuilder = new StringBuilder();
+        input.Version = inputVersionStringBuilder.Append(GetProjectVersionPrefixIfExist(project)).Append(input.Version).ToString();
         return await ProjectPdfGenerator.GenerateAsync(project, input.Version, input.LanguageCode);
     }
+
+    private string GetProjectVersionPrefixIfExist(Project project)
```
realLiangshiwei, line 36: you can consider moving it to `ProjectPdfGenerator`

maliming (approved):

modules/docs/src/Volo.Docs.Common.Application/Volo/Docs/Common/Documents/DocumentPdfGeneratorAppService.cs
salihozkara: I also thought about putting it in the ProjectPdfGenerator at first, but the place I got the code from was an app service. So it should also be in an app service to be compatible.

gizemmutukurt (approved):

PR #22918, opened 2025-05-16
Remove redundant header and horizontal rule from profile management tabs

maliming:
screenshot-2025-05-10 T16-15-54@2x
screenshot-2025-05-10 T16-15-49@2x


yagmurcelk (changes requested):
It would be more compatible if it was flush with the top.

image

maliming:
Changed

image

yagmurcelk (approved):

PR #22951, opened 2025-05-22
configure via ConfigureDefaultConvention() and ConfigureDefaultOnMode…

ivan132:
The DefaultConventionAction and DefaultOnModelCreatingAction properties have internal access modifiers and cannot be accessed directly. They can only be configured through the ConfigureDefaultConvention() and ConfigureDefaultOnModelCreating() methods.


maliming (approved):

maliming:
Thanks @ivan132

PR #22962, opened 2025-05-26
Support S3 Compatible Blob Storage Providers

### LINKED ISSUE #25503: Cloudflare R2 BLOB Storage integration
Implement cloudflare r2 storage provider for ABP's BLOB Storage system

enisn:
Adds `ServiceURL` and `DisablePayloadSigning` configuration to the AWS Blob provider so it works with S3-compatible services (Cloudflare R2, MinIO, Backblaze B2, Wasabi, DigitalOcean Spaces).

`DisablePayloadSigning = true` makes the SDK send `UNSIGNED-PAYLOAD` instead of the streaming chunked signature that AWS SDK v4 uses by default. Required for Cloudflare R2 (rejects streaming signing with `501 NotImplemented`); leave `false` for AWS S3 and MinIO.

Test module reads `ServiceURL` / `DisablePayloadSigning` / `ContainerName` / `CreateContainerIfNotExists` from user secrets, so the existing `BlobContainer_Tests` base can target any S3-compatible service. When `ContainerName` is supplied the test prefixes every object with `abp-aws-test-run-{guid}/` so cleanup deletes only what this run created.

Locally tested end-to-end against a real Cloudflare R2 bucket and a Docker MinIO instance — both pass the full Save / Get / Delete / Exists / multi-tenant integration suite.

Closes #25503.


enisn:
AWS SDK v4.0 is not compatible with 3rd party S3 APIs yet due to Payload Signing and SignCheck. (`STREAMING-AWS4-HMAC-SHA256-PAYLOAD`)

framework/src/Volo.Abp.BlobStoring.Aws/Volo/Abp/BlobStoring/Aws/DefaultAmazonS3ClientFactory.cs
```diff
@@ -30,31 +30,58 @@ public DefaultAmazonS3ClientFactory(
...
         }
 
         Check.NotNullOrWhiteSpace(configuration.AccessKeyId, nameof(configuration.AccessKeyId));
         Check.NotNullOrWhiteSpace(configuration.SecretAccessKey, nameof(configuration.SecretAccessKey));
 
-        return new AmazonS3Client(configuration.AccessKeyId, configuration.SecretAccessKey, region);
+        return new AmazonS3Client(configuration.AccessKeyId, configuration.SecretAccessKey, clientConfig);
+    }
+
+    protected virtual AmazonS3Config CreateS3ClientConfig(AwsBlobProviderConfiguration configuration, RegionEndpoint? region)
```
maliming, line 63: Using the `Async` method will be better.

maliming (approved):

PR #22982, opened 2025-05-29
Add Excel export to the Audit logs

salihozkara:


maliming (approved):

maliming (approved):

docs/en/modules/audit-logging-pro.md
```diff
@@ -126,6 +136,23 @@ Configure<ExpiredAuditLogDeleterOptions>(options =>
...
+    options.FileRetentionHours = 24; // How long to keep files before cleanup (default: 24 hours)
+    options.DownloadBaseUrl = "https://yourdomain.com"; // Base URL for download links in emails
+    options.ExcelFileCleanupOptions.Period = (int)TimeSpan.FromHours(24).TotalMilliseconds; // Interval of the cleanup worker (default: 24 hours)
+});
+```
+
+> Note: The `FileRetentionHours` value determines when files become eligible for deletion, but actual deletion depends on when the cleanup worker runs. If the worker hasn't run after the retention period expires, files will remain accessible. Therefore, `FileRetentionHours` represents the minimum intended retention time, but the actual retention time might be longer depending on the worker's execution schedule.
+
+These settings control where Excel export files are stored, how long they are kept before automatic cleanup, and what base URL is used in email download links.
+
```
maliming, line 155: We need to mention that the app has to use a valid `BLOB Storage Provider` to use this feature.

https://abp.io/docs/latest/framework/infrastructure/blob-storing#blob-storage-providers

PR #23007, opened 2025-06-04
feat(MongoDB): Cache MongoClient per connection string using MongoClientFactory

### LINKED ISSUE #22999: [MongoDB] MongoClient should be cached per connection string instead of creating new instance per UnitOfWork
# [MongoDB] MongoClient should be cached per connection string instead of creating new instance per UnitOfWork

## Description

In ABP Framework 8.3.4, every time a new `DbContext` is created in a MongoDB-backed module, a **new `MongoClient` instance** is also created.

This happens inside the `UnitOfWorkMongoDbContextProvider<T>` class via the `CreateMongoClient(MongoUrl)` method, which returns a new instance without any caching logic.

This leads to:
- One or more new TCP connections to MongoDB per request
- Increased resource consumption (CPU, memory)
- Potential exhaustion of MongoDB connection pool
- Performance degradation under high concurrency

### Reproduction
Just send any request that hits a repository using MongoDB. Observe that `MongoClient` is created via:
```csharp
var client = new MongoClient(mongoClientSettings);
```
This happens even if the same connection string is used.

There is no global `MongoClient` caching in place.

## Configuration

- ABP Framework: 8.3.4
- .NET: 8.0.4
- OS: macOS 14.5 / Ubuntu 22.04 (both tested)
- Architecture: x64
- MongoDB Atlas + Local MongoDB used

## Regression?

Not sure if this behavior changed recently, but it seems to have always behaved this way in the MongoDB integration.

## Data

```json
db.serverStatus().connections output example under load:
{
  "current": 1320,
  "available": 80
}
```

**Expected:** ~100–200 connections under our load  
**Actual:** 1000+ connections and growing linearly with traffic

Also observed increased CPU usage in the MongoDB Atlas dashboard and app latency spikes due to socket waits.

We mitigated it by overriding `UnitOfWorkMongoDbContextProvider<T>` and injecting a singleton `IMongoClientFactory`.

## Analysis

The root cause is here:
```csharp
protected virtual MongoClient CreateMongoClient(MongoUrl mongoUrl)
{
        var mongoClientSettings = MongoClientSettings.FromUrl(mongoUrl);
        Options.MongoClientSettingsConfigurer?.Invoke(mongoClientSettings);
        
        return new MongoClient(mongoClientSettings);
}
```

Every call creates a new `MongoClient`. According to MongoDB’s documentation, `MongoClient` is thread-safe and should be reused.

### Suggested fix:
- Introduce a singleton `IMongoClientFactory` that caches clients per connection string
- Replace `new MongoClient(...)` calls with `_mongoClientFactory.GetClient(...)`

I’d be happy to contribute a PR if that aligns with your direction.

---

## Contribution

We would be happy to contribute this improvement as a PR if the ABP team agrees with the proposed direction. Please let us know if you're open to it, and we can start working on a clean, well-tested implementation right away.

mosindi:
### Description

Resolves #22999

It introduces a new `IMongoClientFactory` to avoid creating a new `MongoClient` instance for every UnitOfWork. According to MongoDB documentation, `MongoClient` is thread-safe and should be reused across the application lifecycle.

## Changes

- Added `MongoClientFactory` with connection string-based caching
- Injected `IMongoClientFactory` into `UnitOfWorkMongoDbContextProvider`
- Updated `CreateMongoClient()` to use factory
- Registered `MongoClientFactory` as singleton
- Added unit tests for:
  - Same connection string returns same instance
  - Different strings yield different instances
  - Thread-safety
  - Null/empty guard clause

## Notes

We’re happy to improve this PR based on feedback.  
Thanks for the great framework 🙌


maliming (approved):

maliming:
Thanks @mosindi 

Can you sign the CLA?

https://cla-assistant.io/abpframework/abp?pullRequest=23007

mosindi:
Thanks @maliming.  I signed the CLA

PR #23129, opened 2025-06-20
Remove Yarn dependency warnings from the documentation

hikalkan:
Because it is not needed to manually and globally installed anymore. ABP CLI handles it.

### DIFF
--- a/docs/en/framework/real-time/signalr.md
+++ b/docs/en/framework/real-time/signalr.md
@@ -55,7 +55,7 @@ Client side installation depends on your UI framework / client type.
 Run the following command in the root folder of your web project:
 
 ```bash
-npm install @abp/signalr
+abp add-package @abp/signalr
 ```
 
 This will add the `@abp/signalr` to the dependencies in the `package.json` of your project:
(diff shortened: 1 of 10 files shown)


maliming (approved):

docs/en/framework/real-time/signalr.md
```diff
@@ -55,11 +55,9 @@ Client side installation depends on your UI framework / client type.
 Run the following command in the root folder of your web project:
 
 ```bash
-yarn add @abp/signalr
+npm install @abp/signalr
```
hikalkan: Is there an ABP CLI command that installs ABP NPM packages?

maliming: The `add-package` command of `ABP CLI` is intended to add NuGet and Angular packages.

```cs
'add-package' command is used to add an ABP package to a project.
It should be used in a folder containing a .csproj file, .sln file or angular.json.
```

I will modify it to support `packages.json` file.

maliming: https://github.com/abpframework/abp/pull/23136

PR #23174, opened 2025-06-25
Refactor `DisableAuditingAttribute` to allow it to ignore changes to `AuditProperties`.

### LINKED ISSUE #23168: Supports ignoring changes to the `AuditProperties` when special properties are modified.
Also, cancel the `EntityUpdatedEvent`

maliming:
Resolve #23168 

---

#### Ignore Update Audit Properties And Publish Entity Updated Event

The `[DisableAuditing]` attribute supports additional configuration options when applied to **entity properties**.

* **UpdateModificationProps** (default: `true`): When set to `false`, changes to this entity property will not update audit properties (like `LastModificationTime`.
* **PublishEntityEvent** (default: `true`): When set to `false`, changes to this entity property will not publish entity change events (`EntityUpdatedEvent`).

````csharp
public class MyUser : Entity<Guid>
{

    [DisableAuditing(UpdateModificationProps = false, PublishEntityEvent = false)]
    public string ReadCount { get; set; }
}
````

This example will ignore update audit properties and publish entity updated event when the `ReadCount` property is updated.

> The `UpdateModificationProps` and `PublishEntityEvent` only work for [Entity Framework Core](../data/entity-framework-core). It will not work for [MongoDB](../data/mongodb).


hikalkan (approved):

hikalkan:
Thanks. Good feature.

PR #23176, opened 2025-06-25
Add other ui options for microservice tutorial

m-aliozkaya:


gizemmutukurt (changes requested):
- [x] Blazor wasm - Places that mention "Web" should be updated to "Blazor wasm"

image
image


- [x] Blazor wasm - The html written to display the user interface is not specified where to add it

image

- [x] Blazor webapp/Server - Places that mention "Web" should be updated to "Blazor"

image

- [x] Blazor Webapp/Server - The html written to display the user interface is not specified where to add it
image

- [ ] The web expression here should also change for all UIs (except MVC)
image

- [x] Blazor Server - Places that mention "Web" should be updated to "Blazor"
image

ismcagdas (approved):

gizemmutukurt (approved):

PR #23277, opened 2025-07-09
feat: Add `Volo.Abp.Mapperly` module

### LINKED ISSUE #23243: Switch to another object mapping library
Based on https://github.com/abpframework/abp/issues/22582

The planning is like that:

* We will implement an alternative object mapper based on an open-source library (haven't decided yet, open to suggestions)
* We won't remove the Volo.Abp.AutoMapper package, so this is a backward-compatible change. However, if you continue to use it, you should handle the licensing yourself. We will depend on the latest free version as long as possible (it is version 14.0.0 at the moment). If we have any issue (like .NET 10.0 incompatibility), we have to upgrade to 15.0.0 or a newer version, which requires a license code.
* We will change all of our modules' code to use the new mapper implementation, so none of ABP Platform's modules will require AutoMapper.
* Extract the AutoMapper part of the Object to Object mapping documentation (https://abp.io/docs/latest/framework/infrastructure/object-to-object-mapping) to a new document (like https://abp.io/docs/latest/framework/infrastructure/object-to-object-mapping/automapper).
* Document the new library integration (like https://abp.io/docs/latest/framework/infrastructure/object-to-object-mapping/new-library-name)
* Update all documentation (tutorials, etc) to use the new mapping library
* Change code generation logic for ABP Suite and ABP Studio to work with the new mapping library. No need to provide both of the options, because our modules will use the new library, and the solution will already have that dependency. But, in any way, it would be good to think about the effort of supporting both of the libraries for a few versions.
* Think about any other stuff we need to consider on this change

maliming:
### Description

Resolves #23243

----

We need to create a mapper that inherits from `MapperBase<TSource, TDestination>` for each DTO class, and then you can use [various features of Mapperly](https://mapperly.riok.app/docs/configuration/mapper/) in this mapper class. You can also override the `BeforeMap` and `AfterMap` methods to perform some actions.

The `MyClassMapper` will be added to the DI system. You can also inject services in this mapper class.

**Mapperly** will generate the implementation `Map` methods code during the build process.

```cs
public class MyClass
{
    public string Id { get; set; }

}

public class MyClassDto
{
    public string Id { get; set; }

}

[Mapper]
public partial class MyClassMapper : MapperBase<MyClass, MyClassDto>
{
    public override partial MyClassDto Map(MyClass source);

    public override partial void Map(MyClass source, MyClassDto destination);

    public override void BeforeMap(MyClass source)
    {
        //source.Name = "BeforeMap " + source.Name;
    }

    public override void AfterMap(MyClass source, MyClassDto destination)
    {
        //destination.Name = source.Name + " AfterMap";
    }
}
```

```cs
// <auto-generated />
#nullable enable
namespace Volo.Abp.Mapperly
{
    public partial class MyClassMapper
    {
        [global::System.CodeDom.Compiler.GeneratedCode("Riok.Mapperly", "4.2.1.0")]
        [return: global::System.Diagnostics.CodeAnalysis.NotNullIfNotNull(nameof(source))]
        public override partial global::Volo.Abp.Mapperly.MyClassDto? Map(global::Volo.Abp.Mapperly.MyClass? source)
        {
            if (source == null)
                return default;
            var target = new global::Volo.Abp.Mapperly.MyClassDto();
            target.Id = source.Id;
            target.Name = source.Name;
            return target;
        }

        [global::System.CodeDom.Compiler.GeneratedCode("Riok.Mapperly", "4.2.1.0")]
        public override partial void Map(global::Volo.Abp.Mapperly.MyClass? source, global::Volo.Abp.Mapperly.MyClassDto? destination)
        {
            if (source == null || destination == null)
                return;
            destination.Id = source.Id;
            destination.Name = source.Name;
        }
    }
}
```

If you want the mapper to support reverse mapping, you can use `TwoWayMapperBase ` as the base class.

```cs
[Mapper]
public partial class MyClassMapper : TwoWayMapperBase<MyClass, MyClassDto>
{
    public override partial MyClassDto Map(MyClass source);

    public override partial void Map(MyClass source, MyClassDto destination);

    public override partial MyClass ReverseMap(MyClassDto destination);

    public override partial void ReverseMap(MyClassDto destination, MyClass source);

    public override void BeforeReverseMap(MyClassDto destination)
    {
        //destination.Name = "BeforeReverseMap " + destination.Name;
    }

    public override void AfterReverseMap(MyClassDto destination, MyClass source)
    {
        //source.Name = destination.Name + " AfterReverseMap";
    }
}
```


rqx110:
every mapper class we need write:
```
public override partial MyClassDto Map(MyClass source);

public override partial void Map(MyClass source, MyClassDto destination);
```
Is there a way to automate these template codes?

maliming:
> Is there a way to automate these template codes?

I didn't find it.

rqx110:
Suggestion: Rename `AbpReverseMapperlyMapper` to `AbpTwoWayMapperlyMapper`, and the order of T can same with  `AbpMapperlyMapper`(MyClass, MyClassDto). `AbpReverseMapperlyMapper` contains not only reverse mapping but also forward mapping.

maliming:
**Reverse** is from AutoMapper, and it's just for easier understanding.

https://docs.automapper.io/en/latest/Reverse-Mapping-and-Unflattening.html

rlopes4dev:
Can the generated class have access to the IServiceProvider from the base class? Some custom mappers can required the injection of services.

maliming:
> Can the generated class have access to the IServiceProvider from the base class? Some custom mappers can required the injection of services.

No problem, you can see this mapper class: https://github.com/abpframework/abp/blob/9e0d0413d0cf6c23d7fb4db17a9e8183bcb88d35/framework/test/Volo.Abp.Mapperly.Tests/Volo/Abp/Mapperly/AbpMapperly_Dependency_Injection_Tests.cs

hikalkan:
Renaming `AbpReverseMapperlyMapperBase` to `AbpTwoWayMapperlyMapperBase` seems reasonable to me.

Actually even `MapperBase` and `TwoWayMapperBase` are enough since they will be located in `Volo.Abp.Mapperly` namespace.

maliming:
> Renaming `AbpReverseMapperlyMapperBase` to `AbpTwoWayMapperlyMapperBase` seems reasonable to me.
> 
> Actually even `MapperBase` and `TwoWayMapperBase` are enough since they will be located in `Volo.Abp.Mapperly` namespace.

Changed.

hikalkan (approved):

PR #23293, opened 2025-07-11
Angular - If the child feature is disabled because of the edition, the parent should be too

erdemcaygor:
### Description

Resolves https://github.com/volosoft/volo/issues/20331

### Checklist


sumeyyeKurtulus (approved):

fahrigedik (approved):

oykuermann (changes requested):
A feature that comes from Edition should be disabled:
image

Edition:

image

oykuermann:
The problem still continues:
image

oykuermann (approved):

PR #23392, opened 2025-07-26
Support both `MySql.EntityFrameworkCore` & `Pomelo.EntityFrameworkCore.MySql`.

### LINKED ISSUE #23388: Consider using `MySql.EntityFrameworkCore` for MySQL
### Documentation

`Pomelo.EntityFrameworkCore.MySql` still doesn't have .NET 9 and uncertain .NET10

We can try providing support for [MySql.EntityFrameworkCore](https://www.nuget.org/packages/MySql.EntityFrameworkCore#dependencies-body-tab)
package.

maliming:
Resolves #23388


hikalkan:
Thanks @maliming for the implementation. However, most of the people (including me) won't like to have unnecessary dependencies in their applications. You add package reference for both of the packages: `MySql.EntityFrameworkCore`
and `Pomelo.EntityFrameworkCore.MySql`.

My suggestion:

Introduce a new package `Volo.Abp.EntityFrameworkCore.MySQL.Pomelo` and move `Pomelo.EntityFrameworkCore.MySql` dependency to this new package. Current `Volo.Abp.EntityFrameworkCore.MySQL` package can depend on `MySql.EntityFrameworkCore`. I know that will be a breaking change. However, Pomelo already have problems with .NET 9, so it seems reasonable. We document that in the release notes.

Also, in the startup templates (in ABP Studio), switch to `MySql.EntityFrameworkCore`, and update necessary documents according to these changes.

What do you think?

maliming:
ok, I will add a new package/module.

hikalkan:
Thanks. Also, please test if everything works as expected, since we are changing a library dependency; maybe not everything is compatible.

maliming:
I tested with various templates. I will push new changes to Studio template.

https://github.com/volosoft/abp-studio/pull/3173/commits/0aaf8f890fc960168c8d5ee7dcee70ee1778336d

@enisn @hikalkan You can review the code again.

Thanks.

enisn (approved):

hikalkan (approved):

PR #23416, opened 2025-07-30
Hybrid Rendering & Application Builder

### LINKED ISSUE #23055: Angular SSR
## Implementation Highlights

- [x] Application builder enabled for all templates
- [x] Injected custom storage for OAuthService SSR compatibility
- [x] Replaced direct usage of `window`, `location` and `localStorage` with platform-safe abstractions for SSR compatibility
- [x] Enable compatibility with both **CSR (Client-Side Rendering)** and **SSR (Server-Side Rendering)**.
- [x] Wrapped browser-only and server-only logics  in `isPlatformBrowser()` / `isPlatformServer()` checks to ensure platform safety during SSR.
- [x]  Used Angular's `TransferState` to prevent redundant HTTP requests during hydration 

### 1) SSR-Authentication Flow

This PR integrates OpenID Connect (OIDC) authentication flow into the Angular SSR application.

#### Authorization Request (/authorize)

- Users are redirected to this endpoint when initiating login
- PKCE code verifier and challenge are generated
- Unique state value is created for security
- Session ID is stored along with PKCE data in cookies
- User is redirected to the OIDC provider

####  Callback Handling (/)

- Handles user return from OIDC provider with authorization code
- State parameter is validated against CSRF attacks
- Authorization code is exchanged for access and refresh tokens
- Tokens are stored securely in cookies
- User is redirected to their original destination

####  Logout Process (/logout)

- Cookies are deleted
- User is redirected to OIDC provider's logout endpoint

#### Cookie Management

```
| Cookie | Type | Description |
|--------|------|-------------|
| `sid` | HttpOnly, Secure | Session identifier |
| `access_token` | Secure | Client-side accessible for API calls |
| `refresh_token` | HttpOnly, Secure | Server-only for token refresh |
| `expires_at` | Secure | Token expiration tracking |
| `returnUrl` | HttpOnly, Secure | Post-login redirection |

```

### 2) Http Data Caching with `TransferState`

In Angular applications utilizing Server-Side Rendering (SSR), a common performance challenge is the duplication of HTTP calls. Data is often fetched once on the server (during HTML generation) and then again on the client (when the Angular application bootstraps in the browser). This redundant fetching can negatively impact performance

Our `transferStateInterceptor` is designed to solve this very problem.

The transferStateInterceptor acts as an HTTP Interceptor, automatically capturing all `GET` requests made via `HttpClient`:

On the Server: During the SSR process, the transferStateInterceptor intercepts the HTTP responses from your APIs. It then uses Angular's built-in TransferState mechanism to embed this data directly into the server-rendered HTML, typically within a <script> tag in JSON format.

On the Client: When the Angular application starts in the browser, the transferStateInterceptor checks TransferState for any cached data corresponding to the outgoing HTTP request. If it finds the data, it uses this pre-fetched, cached version instead of making a new network request to the API.

This process ensures your data is fetched once on the server and then "transferred" to the browser. This eliminates unnecessary redundant requests, significantly reduces network traffic, and improves initial page load times, leading to a faster and smoother user experience.

To integrate the transferStateInterceptor into your Angular application, you just need to add it to your provideHttpClient providers using the withInterceptors feature:

```ts
// app.config.ts or your relevant provide* function

import { provideHttpClient, withInterceptors, withInterceptorsFromDi } from '@angular/common/http';
import { provideClientHydration } from '@angular/platform-browser';
import { transferStateInterceptor } from './interceptors/transfer-state.interceptor'; // Path to your interceptor

export const appConfig: ApplicationConfig = {
  providers: [
    provideHttpClient(
      // Add our functional interceptor here
      withInterceptors([transferStateInterceptor]),
      // ... other HttpClient configurations (e.g., withFetch(), withXsrfConfiguration)
    ),
    // `provideClientHydration()` is required for TransferState to work
    provideClientHydration(),
    // ... other providers
  ]
};
```

- `transferStateInterceptor`

```ts
export const transferStateInterceptor: HttpInterceptorFn = (
  req: HttpRequest<any>,
  next: HttpHandlerFn,
): Observable<HttpEvent<any>> => {
  const transferState = inject(TransferState);
  const platformId = inject(PLATFORM_ID);

  if (req.method !== 'GET') {
    return next(req);
  }

  const stateKey = makeStateKey<HttpResponse<any>>(req.urlWithParams);

  if (isPlatformBrowser(platformId)) {
    const storedResponse = transferState.get<HttpResponse<any>>(stateKey, null);
    if (storedResponse) {
      transferState.remove(stateKey);
      return of(new HttpResponse<any>({ body: storedResponse, status: 200 }));
    }
  }

  return next(req).pipe(
    tap(event => {
      if (isPlatformServer(platformId) && event instanceof HttpResponse) {
        transferState.set(stateKey, event.body);
      }
    }),
  );
};

```

## How To Test

-  Navigate `npm > ng-packs` directory and run `yarn start` commad. This will generate both client and server bundles and start the development node.js server. Once the process is complete,  open `localhost:4200` in your browser

erdemcaygor:
### Description

Resolves #23055


fahrigedik (approved):

npm/ng-packs/apps/dev-app/src/server.ts
```diff
@@ -0,0 +1,181 @@
...
+import express from 'express';
+import cookieParser from 'cookie-parser';
+import { dirname, resolve } from 'node:path';
+import { fileURLToPath } from 'node:url';
+import {environment} from './environments/environment';
+
+// ESM import
+import * as oidc from 'openid-client';
+
+//TODO: Remove this line in production
```
sumeyyeKurtulus: This check may be missing here
```tsx
//TODO: Remove this line in production
if (environment.production === false) {
  process.env['NODE_TLS_REJECT_UNAUTHORIZED'] = '0';
}
```

npm/ng-packs/package.json
```diff
@@ -133,5 +144,8 @@
       "npx prettier --write --config .prettierrc "
     ]
   },
-  "dependencies": {}
+  "dependencies": {
+    "cookie-parser": "^1.4.7",
```
sumeyyeKurtulus: I think we can use a simple cookie parser util in the core package or in a relevant one instead of adding a new cookie dependency `cookie-parser` Could look like this
```ts
export class ServerCookieParser {
  static parse(cookieHeader: string): { [key: string]: string } {
    const cookies: { [key: string]: string } = {};

    if (!cookieHeader) return cookies;

    try {
      cookieHeader.split(';').forEach(cookie => {
        const parts = cookie.trim().split('=');
        if (parts.length >= 2) {
          const name = parts[0].trim();
          const value = parts.slice(1).join('='); // Handle values with '='

          if (name) {
            try {
              cookies[name] = decodeURIComponent(value);
            } catch (e) {
              // If decoding fails, use raw value
              cookies[name] = value;
            }
          }
        }
      });
    } catch (error) {
      console.error('Error parsing cookies:', error);
    }

    return cookies;
  }

  static middleware() {
    return (req: any, res: any, next: any) => {
      req.cookies = ServerCookieParser.parse(req.headers.cookie || '');
      next();
    };
  }

  static getCookie(req: any, name: string): string | undefined {
    const cookieHeader = req.headers.cookie;
    if (!cookieHeader) return undefined;

    const cookies = ServerCookieParser.parse(cookieHeader);
    return cookies[name];
  }
}
```

then, in `server.ts`
```ts
//app.use(cookieParser());
app.use(ServerCookieParser.middleware());

```

npm/ng-packs/apps/dev-app/src/server_legacy.ts
```diff
@@ -0,0 +1,78 @@
+import { environment } from './environments/environment';
```
sumeyyeKurtulus: I could not understand how we would use this file. Would we need in case of previous version usages at the same time?

npm/ng-packs/packages/core/src/lib/services/index.ts
```diff
@@ -25,3 +25,5 @@ export * from './local-storage-listener.service';
 export * from './title-strategy.service';
 export * from './timezone.service';
 export * from './time.service';
+export * from './ssr.service';
+export * from './cookie-storage.service';
```
sumeyyeKurtulus, line 28: DOM strategy service seems missing here. If we do not have a specific occasion, we can add it here.

npm/ng-packs/packages/oauth/src/lib/services/index.ts
```diff
@@ -1,3 +1,4 @@
 export * from './oauth.service';
 export * from './oauth-error-filter.service';
-export * from './remember-me.service'
\ No newline at end of file
+export * from './remember-me.service';
+export * from './browser-token-storage.service';
```
sumeyyeKurtulus, line 4: I think server token storage service can be added here either.

npm/ng-packs/apps/dev-app/src/app/app.config.ts
```diff
@@ -18,7 +24,6 @@ import { provideAnimations } from '@angular/platform-browser/animations';
 
 export const appConfig: ApplicationConfig = {
   providers: [
-    provideRouter(appRoutes),
     APP_ROUTE_PROVIDER,
     provideAbpCore(
```
sumeyyeKurtulus, line 28: We can use `registerLocaleForEsBuild()` instead.

npm/ng-packs/packages/core/src/lib/services/session-state.service.ts
```diff
@@ -28,7 +36,8 @@ export class SessionStateService {
   }
 
   private init() {
-    const session = this.localStorageService.getItem('abpSession');
+    let session: string;
```
sumeyyeKurtulus: Maybe we can make this part easier to read like this
```ts
  private init() {
    let session: string = this.localStorageService.getItem('abpSession');
    if (this.appStartedWithSSR) {
      session = this.cookieStorageService.getItem('abpSession');
    }
    if (session) {
      this.store.set(JSON.parse(session));
    }

    this.store.sliceUpdate(state => state).subscribe(this.updateLocalStorage);
  }
```

npm/ng-packs/packages/oauth/src/lib/guards/oauth.guard.ts
```diff
@@ -50,10 +54,30 @@ export const abpOAuthGuard: CanActivateFn = (
   }
 
   const params = { returnUrl: state.url };
+  if (isPlatformServer(platformId) && resInit) {
```
sumeyyeKurtulus, line 57: We may need to make the same check on the async auth guard as well.

sumeyyeKurtulus (commented):
Hello @erdemcaygor thank you for your large effort here. Apart from the comments I left, everything seems nice and neat. We could also discuss the points mentioned below:

- [x]  We can skip the usage of this component `npm/ng-packs/packages/theme-shared/src/lib/components/internet-connection-status/internet-connection-status.component.ts`
    - We can also prevent adding this component to the templates on the studio side.
- [x]  We could also remove the constructors here if we do not have a special occasion
    
    `npm/ng-packs/packages/account/src/lib/components/manage-profile/manage-profile.component.ts`
    
    `npm/ng-packs/packages/setting-management/config/src/lib/components/email-setting-group/email-setting-group.component.ts`
    
    `npm/ng-packs/packages/tenant-management/src/lib/components/tenants/tenants.component.ts`
    
    `npm/ng-packs/packages/identity/src/lib/components/roles/roles.component.ts`
    
    `npm/ng-packs/packages/identity/src/lib/components/users/users.component.ts`
 - [x]  We can also move these declarations out of the constructor
    
    `npm/ng-packs/packages/core/src/lib/services/dom-strategy.service.ts`
    
    `npm/ng-packs/packages/core/src/lib/services/ssr.service.ts`
    
    `npm/ng-packs/packages/theme-shared/src/lib/directives/ngx-datatable-default.directive.ts`
    
- [x]  We could use the the `DOCUMENT`  dependency instead to make it align with other implementations
    
    `npm/ng-packs/packages/feature-management/src/lib/components/feature-management/feature-management.component.ts`
    
    `npm/ng-packs/packages/permission-management/src/lib/components/permission-management.component.ts`
    
    `npm/ng-packs/packages/tenant-management/src/lib/components/tenants/tenants.component.ts`
    
    `npm/ng-packs/packages/theme-shared/src/lib/components/modal/modal.component.ts`
    
    `npm/ng-packs/packages/theme-shared/src/lib/providers/tenant-not-found.provider.ts` 
    
    `npm/ng-packs/packages/theme-shared/src/lib/services/confirmation.service.ts`
    
    `npm/ng-packs/packages/core/src/lib/strategies/loading.strategy.ts`
   
- [x] I am getting this error on CSR when I try to log in using the code flow.
         Screenshot 2025-09-12 at 08 36 25

npm/ng-packs/packages/core/src/lib/services/index.ts
erdemcaygor: Added

npm/ng-packs/apps/dev-app/src/server_legacy.ts
erdemcaygor: No, we don't need this file. Removed

npm/ng-packs/apps/dev-app/src/server.ts
erdemcaygor: you are right. Production environment check added

npm/ng-packs/packages/oauth/src/lib/services/index.ts
erdemcaygor: Added

npm/ng-packs/package.json
erdemcaygor: That makes sense. A simple cookie parser utility has been created in the core package, and the `cookie-parser` dependency has been removed. Thanks!

npm/ng-packs/packages/core/src/lib/services/session-state.service.ts
erdemcaygor: Since `LocalStorageService` always returns null on the server, I think adding a platform check is a better way to avoid unnecessary queries. I refactored the logic to make it more readable, like this:

```ts 

  private init() {
    const storageService = this.appStartedWithSSR ? 
                                         this.cookieStorageService : 
                                         this.localStorageService;
    const session = storageService.getItem('abpSession');
    if (session) {
      this.store.set(JSON.parse(session));
    }
    this.store.sliceUpdate(state => state).subscribe(this.updateLocalStorage);
  }

```

npm/ng-packs/packages/oauth/src/lib/guards/oauth.guard.ts
erdemcaygor: Added

sumeyyeKurtulus (approved):

oykuermann (changes requested):
I got an error while logging out of the app template.

[screen-capture - 2025-09-17T105436.903.webm](https://github.com/user-attachments/assets/49346f73-e340-4d13-b9e6-79fc5a31e98b)

oykuermann (approved):

PR #23421, opened 2025-07-31
[Invalid] Fix custom command example in studio kubernetes document

yekalkan:


m-aliozkaya (changes requested):
I got this error below.

`Error occurred during the rendering of this document. The document is not valid: <input>(217,66) : error : Cannot get the member chart.metadata for a null object.
`

PR #23445, opened 2025-08-05
Return a new list from `WebRequestResources`.

maliming:
Otherwise, the content in `Resources` may return inconsistent data.


EngincanV (approved):

hikalkan:
It is better to write this as a comment near to the changed line, otherwise another developer may find it unnecessary and delete the `ToList` call.

maliming:
> It is better to write this as a comment near to the changed line, otherwise another developer may find it unnecessary and delete the `ToList` call.

https://github.com/abpframework/abp/pull/23451

PR #23469, opened 2025-08-07
Disable `MapperlyAutoObjectMappingProvider` if AutoMapper used in app.

maliming:
This is the temporary solution, We will remove it in when all apps are migrated to Mapperly.

We only used **Mapperly** in the module, but the application depends on both **AutoMapper** and **Mapperly**. The current module dependency order has **Mapperly** following **AutoMapper**, however, the app still requires AutoMapper.


maliming:
@oykuermann 

You can start to test it， I will fix the unit tests.

EngincanV (approved):

oykuermann (approved):

PR #23543, opened 2025-08-15
Update loading.directive.ts

### LINKED ISSUE #23542: Error in LoadingDirective: "Cannot read properties of undefined (reading 'length')" in Angular ABP Library

### Description

When using the ABP Angular LoadingDirective, I encounter a runtime error.

**Usage Example**
```html
<div class="row" *abpLoading="true">
  ...
</div>
```

### Reproduction Steps

Add directive on a div. you will see.

### Expected behavior

no error.

### Actual behavior

getting error

### Regression?

_No response_

### Known Workarounds

_No response_

### Version

9.2.0

### User Interface

Angular

### Database Provider

None/Others

None (Default)

### Operation System

macOS

### Other information

Native element is 'comment' it does not have children. the field is undefined. due that it throw error.

Image

Image

mahmut-gundogdu:
### Description

Resolves #23542 

I added null check. 

### Checklist

### How to test it?

### DIFF
--- a/npm/ng-packs/packages/theme-shared/src/lib/directives/loading.directive.ts
+++ b/npm/ng-packs/packages/theme-shared/src/lib/directives/loading.directive.ts
@@ -89,7 +89,7 @@ export class LoadingDirective implements OnInit, OnDestroy {
   ngOnInit() {
     if (!this.targetElement) {
       const { offsetHeight, offsetWidth } = this.elRef.nativeElement;
-      if (!offsetHeight && !offsetWidth && this.elRef.nativeElement.children.length) {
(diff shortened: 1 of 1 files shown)


sumeyyeKurtulus:
Hello @mahmut-gundogdu, thank you for your contribution. 
However, we do not plan a next patch release for the version 9.2. For this reason, that would be the best if you could contribute on the version 9.3.

mahmut-gundogdu:
I am closing the Pr because I have created same changes based on rel-9.3
https://github.com/abpframework/abp/pull/23577

PR #23563, opened 2025-08-19
Add failure retry policy to `InboxProcessor`

maliming:
During the processing of events in `InboxProcessor`, if any event handler throws an exception, we provide three options:

- **Retry** (**default**): The current exception and subsequent events will continue to be processed in order in the next cycle.
- **RetryLater**: Skip the event that caused the exception and continue with the following events. The failed event will be retried after a delay that doubles each time (`1, 2, 4, 8, 16 seconds`). The default maximum retry count is **10** (configurable).
- **Discard**: The event that caused the exception will be discarded and will not be retried.

This PR marks as `breaking change` because `IncomingEvent` entity properties changed.


hikalkan:
The only problem I am not sure is that this will break event order. Maybe the error is a temporary error and it will be solved in next try. For example, if it is because the database server is not available, all other events will also fail.
What is the industry best practice for this problem?

maliming:
The events in the inbox don't need to be in any specific order. 
This is because multiple distributed events arrive in the inbox randomly. 
If an event handler has an exception, it shouldn't prevent other events in the inbox from being processed.

hikalkan:
Some questions/requests:

* What about the RetryLater strategy? After 10 retry, it discards the message?
* RetryLater finishes all tries in 30 minutes with the default configuration. I think it is not so reasonable. My suggestion: re-try after 10 seconds for 1st, then 20 for 2nd, then 40 for 3rd,... In this way, it will continue to try in next 5-6 hours. And this factor (here: 10) can also be configurable. If we make it 5, it will be 5 in 1st, 10 in 2nd, 20 in 3rd, 40 in 4th,... Even the "double" logic can be configurable (multiple by 2 for each try, maybe I want 1.5). So, every number can be configured in this formula.

maliming:
> Some questions/requests:
> 
> * What about the RetryLater strategy? After 10 retry, it discards the message?
> * RetryLater finishes all tries in 30 minutes with the default configuration. I think it is not so reasonable. My suggestion: re-try after 10 seconds for 1st, then 20 for 2nd, then 40 for 3rd,... In this way, it will continue to try in next 5-6 hours. And this factor (here: 10) can also be configurable. If we make it 5, it will be 5 in 1st, 10 in 2nd, 20 in 3rd, 40 in 4th,... Even the "double" logic can be configurable (multiple by 2 for each try, maybe I want 1.5). So, every number can be configured in this formula.

Changed. You can review it. Thanks.

framework/src/Volo.Abp.EntityFrameworkCore/Volo/Abp/EntityFrameworkCore/DistributedEvents/DbContextEventInbox.cs
```diff
@@ -50,7 +50,8 @@ public virtual async Task<List<IncomingEventInfo>> GetWaitingEventsAsync(int max
         var outgoingEventRecords = await dbContext
             .IncomingEvents
             .AsNoTracking()
-            .Where(x => !x.Processed)
+            .Where(x => x.Status == IncomingEventStatus.Pending)
+            .Where(x => x.NextRetryTime == null || x.NextRetryTime <= Clock.Now)
```
hikalkan: Should we take `Clock.Now` to a variable (before `var outgoingEventRecords = await dbContext` line) to simplify the expression? Can it catch the value and generate a proper SQL?

hikalkan (approved):

hikalkan (approved):

PR #23588, opened 2025-08-20
Support EF Core Shared Entity Types on Repositories

hikalkan:
See https://learn.microsoft.com/en-us/ef/core/modeling/entity-types?tabs=data-annotations#shared-type-entity-types

Basically, `DbContext.Set<T>()` has an overload that takes a `string name` (the entity name). In this way, we can map the same entity type to multiple database tables and configure seperately.

With this PR, now we can use `Repository.SetCustomEntityName()` method to indicate the entity name, so it gets the right DbSet and work on the right database table. This method should be called first, then any repository operation will use it. It is also possible to change it later on the same object and use other database tables. Interesting feature. I will implement it later for MongoDB (so we can switch to another database collection for the same entity type).


hikalkan:
@maliming please review and add tests for this

maliming (approved):

PR #23595, opened 2025-08-23
Bump System.Linq.Dynamic.Core in Abp 8.3 to 1.6.7 to fix the vulnerability

wcypierre:
### Description

Resolves #

Update System.Linq.Dynamic.Core to 1.6.0.1 to fix the vulnerability that Prisma reports

Version 1.6.0.1 is chosen to be in sync with the version that rel-9.1 branch is using

### Checklist

Ran unit test for the whole solution and there weren't any errors.


maliming:
hi @wcypierre 

Can we upgrade to the latest?

https://www.nuget.org/packages/System.Linq.Dynamic.Core/1.6.7

wcypierre:
Hi @maliming  that would be better. I've updated the package and reran unit test again locally.

maliming (approved):

maliming:
Thanks @wcypierre  👍

wcypierre:
@maliming  could we have this fix be backported to 8.3.x as well? because for my team's use case we're not migrating to 9.x yet and it would be great if this fix would be in this major version (8.x.x)

maliming:
hi

We will not release the 8.x patch version. You can reference this package in your project to fix vulnerability.

wcypierre:
That's what we're doing now at the moment, guessed it would have been better to do it at the source.

Regardless, will apply this workaround until we do the upgrade in .net10

PR #23647, opened 2025-09-01
Improve column width handling in ExtensibleTable

fahrigedik:
Updated the table to use default widths when columnWidths are undefined and added canAutoResize binding. Added logic in the component to reset column widths if any are missing, ensuring consistent table rendering.

### Description

Resolves  https://github.com/volosoft/vs-internal/issues/7259  (write the related issue number if available)

### How to test it?

run dev app in volo repo

### DIFF
--- a/npm/ng-packs/packages/components/extensible/src/lib/components/extensible-table/extensible-table.component.html
+++ b/npm/ng-packs/packages/components/extensible/src/lib/components/extensible-table/extensible-table.component.html
@@ -40,8 +40,9 @@
   @if (actionsTemplate || (actionList.length && hasAtLeastOnePermittedAction)) {
     <ngx-datatable-column
       [name]="actionsText | abpLocalization"
-      [maxWidth]="columnWidths[0]"
-      [width]="columnWidths[0]"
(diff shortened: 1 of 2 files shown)


sumeyyeKurtulus (commented):
Hello @fahrigedik thank you for coming up with such solution 🙌🏻 it works fine as it should be. However, we can switch to a signal based approach to make it better in terms of scalability and efficiency. Here is an example:

```ts
//...
  // Signal for actions column width
  private readonly _actionsColumnWidth = signal<number | undefined>(DEFAULT_ACTIONS_COLUMN_WIDTH);

  @Input() set actionsColumnWidth(width: number) {
    this._actionsColumnWidth.set(width ? Number(width) : undefined);
  }
  readonly columnWidths = computed(() => {
    const actionsColumn = this._actionsColumnWidth();
    const widths = [actionsColumn];
    this.propList.forEach(({ value: prop }) => {
      widths.push(prop.columnWidth);
    });
    return widths;
  });
  ```

sumeyyeKurtulus (approved):

yagmurcelk (approved):

PR #23673, opened 2025-09-05
Add AbpBsonSerializer utility class

### LINKED ISSUE #23669: MongoDB: allow GUID representation to be configured

Before ABP 9.3 we use the following way to make MongoDB serialize GUIDs as string, to improve readability:
```csharp
BsonSerializer.RegisterSerializer(new GuidSerializer(BsonType.String))
```

This however is borken after updating to ABP 9.3 because `AbpMongoDbModule` registers another GuidSerializer:
https://github.com/abpframework/abp/blob/8ac76d53d800cbe646a83ac66ca170d2ff8ccd6a/framework/src/Volo.Abp.MongoDB/Volo/Abp/MongoDB/AbpMongoDbModule.cs#L22

ABP should provide a way to allow this to be configured.

### Additional context

Apparently we can use reflection to fix this, but needless to say it's ugly and not preferred:

```csharp
var guidSerializer = (GuidSerializer)BsonSerializer.LookupSerializer<Guid>();
typeof(GuidSerializer)
    .GetField("_representation", BindingFlags.Instance | BindingFlags.NonPublic)!
    .SetValue(guidSerializer, BsonType.String);
```

maliming:
### Description

Resolves #23669

### DIFF
--- a/framework/src/Volo.Abp.MongoDB/Volo/Abp/MongoDB/AbpBsonSerializer.cs
+++ b/framework/src/Volo.Abp.MongoDB/Volo/Abp/MongoDB/AbpBsonSerializer.cs
@@ -1,0 +1,31 @@
+using System;
+using System.Collections.Concurrent;
+
+using System.Reflection;
+using MongoDB.Bson.Serialization;
+
+namespace Volo.Abp.MongoDB;
+
+public static class AbpBsonSerializer
+{
+    private static readonly ConcurrentDictionary<Type, IBsonSerializer> Cache;
+
+    static AbpBsonSerializer()
+    {
+        var registry = BsonSerializer.SerializerRegistry;
+        var type = typeof(BsonSerializerRegistry);
+        var cacheField = type.GetField("_cache", BindingFlags.NonPublic | BindingFlags.Instance) ??
+                         throw new AbpException($"Cannot find _cache field of {type.FullName}.");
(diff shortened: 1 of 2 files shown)


hillin:
Can we defer the serializer registration to `ConfigureServices`? Then we can have a chance to configure it in `PreConfigureServices`.

maliming:
> Can we defer the serializer registration to `ConfigureServices`? Then we can have a chance to configure it in `PreConfigureServices`.

hi

`BsonSerializer` is a static class. You can add the code to either your `ConfigureServices` or `PreConfigureServices`, both of which work.


```cs
AbpBsonSerializer.RemoveSerializer<Guid>();
BsonSerializer.RegisterSerializer(new GuidSerializer(BsonType.String))
```

hillin:
Yes, I mean instead of the static registration approach, what if ABP registers the serializer in the `AbpMongoDbModule.ConfigureServices` method? This way we can give other modules a chance to configure the GuidSerializer in their `PreConfigureServices` method, which I believe is a more elegant solution.

However I suspect there must be some reason that made ABP decide to register it statically, perhaps the serializer is used before `ConfigureServices`? Could you shed some light on that?

maliming:
hi

It is static and global, which can affect unit testing.

hillin:
Not exactly what I was asking, but never mind, as long as it works!

EngincanV (approved):

PR #23689, opened 2025-09-08
 Refactor extra properties mapping logic in Mapperly

salihozkara:


maliming:
hi @salihozkara 

All unit tests passed even after I reverted your change. Can you add a unit test to verify your code?

Thanks.

image

maliming (approved):

PR #23703, opened 2025-09-10
Refactor permission grant tracking logic.

### LINKED ISSUE #23702: Permission management Blazor UI: `Grant all permissions` checkbox is not binded well
It works as expected when you check/uncheck it. But it is not changed when you manually grant permissions in tabs.

Image

maliming:
Resolve #23702


yekalkan (approved):

yagmurcelk (changes requested):
If some permissions are selected and some are not, it would be better for the Grant all permissions button to appear half-selected (as in the example).

image


**Example =>**
image

yagmurcelk (approved):

Issue #23733, opened 2025-09-15
Generic version for EntityNotFoundException


tntwist:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your feature request related to a problem? Please describe the problem.

Please add a generic version of EntityNotFoundException.

### Describe the solution you'd like

See provided pull request.

maliming:
https://github.com/abpframework/abp/pull/23734

PR #23734, opened 2025-09-15
feat(): add generic EntityNotFoundException

tntwist:
Implements solution for #23733 

### Description

This pr adds a generic version of the `EntityNotFoundException`.
Its much more convinient to throw a `EntityNotFoundException<MyEntity>(id)` exception than always using the `EntityNotFoundException(typeof(MyEntity), id)` alternative we currently have.

I guess this is a relativly small change if zero impact.

I already changed some lines where the Exception was used to use the generic version.
My auto formating fixed some formating issues on those files. Like whitespaces or import order.

### Checklist

   - Nope as imho this does change nothing on current behavior at all. Let my know it this would be required
   - Nope. Let me know if required and i´ll update some docs.

### How to test it?
Rather than using this `throw new EntityNotFoundException(typeof(MyEntity), id)` just use this `throw new EntityNotFoundException<MyEntity>(id)`.


maliming (approved):

maliming:
Thanks @tntwist 👍

PR #23609, opened 2025-08-27
Upgrade to .NET 10.0.

maliming:
Resolve #23383

PR #23736, opened 2025-09-16
Disable cache error hiding in development environment.

maliming:
If Redis is unavailable, it will throw the exception.

image

image


hikalkan (approved):

hikalkan:
Thanks. Can you document this on the migration guide. I am marking this as a breaking change.

maliming:
> Thanks. Can you document this on the migration guide. I am marking this as a breaking change.

https://github.com/abpframework/abp/pull/23609/commits/4526cc48336f1ac1d90011ab82d9f4e9553fada2

PR #23790, opened 2025-09-25
🔧 fix(extensible-table): use computed signal properly in template

### LINKED ISSUE #23789: 🐛 ExtensibleTableComponent columnWidths signal not working - missing parentheses in template

### Description

The columnWidths property in ExtensibleTableComponent is defined as a computed signal but the template is accessing it as a regular property without parentheses, causing column width settings to be ignored.

### Reproduction Steps

Create an EntityProp with custom columnWidth: 
```typescript
EntityProp.create({ name: 'myColumn', type: ePropType.String, displayName: 'My Column', columnWidth: 150 })
```
Use ExtensibleTableComponent in your template
Observe that the column uses auto-width instead of the specified 150px

### Expected behavior

The column should have a width of 150px as specified in the columnWidth property.

### Actual behavior

The column uses auto-width sizing, completely ignoring the columnWidth setting. All custom column widths are ignored.

### Regression?

This appears to be introduced when upgrading to Angular Signals. The template was not updated to use proper signal syntax. Root cause: Template accesses computed signal as property instead of calling it as a function.

### Known Workarounds

Create a custom ExtensibleTableComponent with corrected template
Override the component template in your application

### Version

9.3.4

### User Interface

Angular

### Database Provider

EF Core (Default)

None (Default)

### Operation System

Windows (Default)

### Other information

The commit 0f23ce685a refactored column width handling to use Angular signals and computed properties, but the HTML template wasn't updated accordingly.

Affected file: [extensible-table.component.html](vscode-file://vscode-app/c:/Users/Administrator/AppData/Local/Programs/Microsoft%20VS%20Code%20Insiders/resources/app/out/vs/code/electron-browser/workbench/workbench.html)

Lines to fix:

Line 40: [maxWidth]="columnWidths[0] ?? undefined" should be [maxWidth]="columnWidths()[0] ?? undefined"
Line 41: [width]="columnWidths[0] ?? 200" should be [width]="columnWidths()[0] ?? 200"
Line 42: [canAutoResize]="!columnWidths[0]" should be [canAutoResize]="!columnWidths()[0]"
Line 60: [width]="columnWidths[i + 1] ?? 200" should be [width]="columnWidths()[i + 1] ?? 200"
Line 61: [canAutoResize]="!columnWidths[i + 1]" should be [canAutoResize]="!columnWidths()[i + 1]"
Environment: ABP Framework: 9.3.4 Angular: 20.0.0 Package: @abp/ng.components

yinchang0626:
Update HTML template to call columnWidths() instead of accessing columnWidths[index] after signal refactoring in 0f23ce685a

Resolves #23789

### DIFF
--- a/npm/ng-packs/packages/components/extensible/src/lib/components/extensible-table/extensible-table.component.html
+++ b/npm/ng-packs/packages/components/extensible/src/lib/components/extensible-table/extensible-table.component.html
@@ -46,9 +46,9 @@
   @if (actionsTemplate || (actionList.length && hasAtLeastOnePermittedAction)) {
     <ngx-datatable-column
       [name]="actionsText | abpLocalization"
-      [maxWidth]="columnWidths[0] ?? undefined"
-      [width]="columnWidths[0] ?? 200"
-      [canAutoResize]="!columnWidths[0]"
+      [maxWidth]="columnWidths()[0] ?? undefined"
+      [width]="columnWidths()[0] ?? 200"
+      [canAutoResize]="!columnWidths()[0]"
       [sortable]="false"
     >
       <ng-template let-row="row" let-i="rowIndex" ngx-datatable-cell-template>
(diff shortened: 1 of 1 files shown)


fahrigedik:
Thanks for contribution. https://github.com/abpframework/abp/pull/23860 fixed.

PR #23924, opened 2025-10-09
Use a standard structure for tags for CMS Kit

ahmetcelik05:
### Description

Resolves [vs-internal-issue-#7448](https://github.com/volosoft/vs-internal/issues/7448)

### Checklist

- [ ] no need to document

### DIFF
--- a/modules/cms-kit/src/Volo.CmsKit.Public.Web/Pages/CmsKit/Shared/Components/Tags/Default.cshtml
+++ b/modules/cms-kit/src/Volo.CmsKit.Public.Web/Pages/CmsKit/Shared/Components/Tags/Default.cshtml
@@ -3,23 +3,29 @@
 @model TagViewComponent.TagViewModel
 
 <div class="my-3">
-    <div class="cms-tags-area">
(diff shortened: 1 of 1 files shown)


modules/cms-kit/src/Volo.CmsKit.Public.Web/Pages/CmsKit/Shared/Components/Tags/Default.cshtml
```diff
@@ -3,23 +3,29 @@
...
-    </div>
+	<div class="cms-tags-area">
+		@if (Model.Tags != null)
+		{
+			<ul class="d-flex flex-wrap">
+				@foreach (var tag in Model.Tags)
+				{
+					if (Model.UrlFormat.IsNullOrWhiteSpace())
+					{
+						<li class="me-1">
```
enisn: Instead of manually adding margin in elements with `me-1`, I suggest using `gap-1` in the `.d-flex` container

```html
<ul class="d-flex flex-wrap gap-1">
```

MansurBesleney (approved):

PR #23937, opened 2025-10-13
Fix base64 padding in JWT token parsing - issue-23796

### LINKED ISSUE #23796: The `getFromToken` method has compatibility problem.
> We found that in the getFromToken function, the access token payload is decoded using atob. When the payload length is not a multiple of 4 (due to missing Base64 padding =), atob throws this error. This happens especially when the preferred_username claim contains an email (e.g., [abc@de.com](mailto:abc@de.com)).

https://github.com/abpframework/abp/blob/dev/npm/ng-packs/packages/oauth/src/lib/services/remember-me.service.ts#L40-L49

Current Code:(in remember-me.service.ts)

```cs
const tokenBody = accessToken.split('.')[1].replace(/-/g, '+').replace(/_/g, '/');
const parsedToken = JSON.parse(atob(tokenBody));
```

Suggested Fix:
Add missing padding before calling atob:

```cs
while (tokenBody.length % 4 !== 0) {
  tokenBody += '=';
}

```

Alternatively, using the **jwt-decode** library in Angular would be a more reliable approach.

https://abp.io/support/questions/9936/Angular-UI---atob-error-while-decoding-access-token-The-string-to-be-decoded-is-not-correctly-encoded

fahrigedik:
Ensures the JWT token body is properly padded before decoding, preventing errors when parsing tokens with missing base64 padding.

Resolves #23796

### DIFF
--- a/npm/ng-packs/packages/oauth/src/lib/services/remember-me.service.ts
+++ b/npm/ng-packs/packages/oauth/src/lib/services/remember-me.service.ts
@@ -38,7 +38,10 @@ export class RememberMeService {
   }
 
   getFromToken(accessToken: string) {
-    const tokenBody = accessToken.split('.')[1].replace(/-/g, '+').replace(/_/g, '/');
+    let tokenBody = accessToken.split('.')[1].replace(/-/g, '+').replace(/_/g, '/');
+    while (tokenBody.length % 4 !== 0) {
+      tokenBody += '=';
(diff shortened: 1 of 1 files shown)


sumeyyeKurtulus:
Hello @fahrigedik we can merge this if you could rebase it to `rel-10.0` branch.

PR #23953, opened 2025-10-14
add seo description

salihozkara:


ebicoglu:
There are more than 300 docs in this PR.
Check them programmatically if there are no wildcards in the description text, which can break the markdown->HTML rendering.

ebicoglu (approved):

PR #24050, opened 2025-10-24
Angular - Improve Authentication Token Handling

### LINKED ISSUE #23930: Angular - Improve Authentication Token Handling
Currently, the access token is stored in localStorage, which makes it vulnerable to XSS attacks.
We should move to a safer approach like keeping the access token in memory and using an HttpOnly cookie for the refresh token to reduce security risks and improve overall session handling.

Related documentations;

- https://auth0.com/docs/secure/security-guidance/data-security/token-storage

erdemcaygor:
### Description

Resolves https://github.com/abpframework/abp/issues/23930

### Improvements

- Store sensitive tokens in memory
- Use Web Workers for state sharing between tabs


npm/ng-packs/packages/oauth/src/lib/services/memory-token-storage.service.ts
```diff
@@ -0,0 +1,123 @@
...
+})
+export class MemoryTokenStorageService implements OAuthStorage {
+  private keysShouldStoreInMemory = [
+    'access_token', 'id_token', 'expires_at',
+    'id_token_claims_obj', 'id_token_expires_at',
+    'id_token_stored_at', 'access_token_stored_at',
+    'abpOAuthClientId', 'granted_scopes'
+  ];
+
+  private worker?: any;
```
fahrigedik, line 25: `private worker?: SharedWorker;`
It would be better to add a proper type for worker instead of leaving it as an optional SharedWorker. This improves clarity and type safety.

npm/ng-packs/packages/oauth/src/lib/workers/token-storage.worker.js
```diff
@@ -0,0 +1,63 @@
...
+  key?: string;
+  value?: string;
+}
+
+const tokenStore = new Map<string, string>();
+const ports = new Set<MessagePort>();
+
+self.onconnect = (event: MessageEvent) => {
+  const port = event.ports[0];
+  ports.add(port);
```
fahrigedik: It looks like there’s a potential memory leak issue here.
When closing a tab, the number of ports continues to increase, which indicates that the reference to the closed tab is not being properly released. This could negatively impact performance over time.

**Steps to reproduce:**
1. Run the project and open it in 3 different tabs.
2. Close 2 of the tabs.
3. Open 2 new tabs again.

At this point, you’ll notice that the ports array contains 5 elements instead of 3.
The closed tab’s port should be removed from the ports array when the tab is closed.

image

To fix this issue, we probably need to introduce a new `**disconnect** `action type (or a similar mechanism) to properly handle tab disconnections.
When a tab is closed, we should dispatch this action to remove the corresponding port from the ports array, ensuring that stale connections don’t accumulate and cause memory leaks.

fahrigedik (changes requested):

fahrigedik (approved):

yagmurcelk (approved):

PR #24053, opened 2025-10-27
Draft feature added to the pages.

### LINKED ISSUE #23616: CMS Kit: Add "draft" support for Pages feature
Currently, in the Pages feature, once you create a page, it's automatically published and there is no "draft" option:

Image

We should provide the "draft" option, like we do in the Blogging Feature:

Image

Users can either "save as draft" and then publish it or delete it, or directly publish it!

MansurBesleney:
### Description

Resolves #23616 

- User can save a page as draft and publish it later.

### Checklist

- [x] I fully tested it as developer.
- [x] No need to document.


maliming (approved):

gizemmutukurt (changes requested):
- [ ] PR also receives an error regarding DB.

image

gizemmutukurt (approved):

PR #24139, opened 2025-11-10
Resource based authorization

hikalkan:
**This is currently an experimental PR. We will be working on**

## Work to be done

- [x] Implement `IResourcePermissionChecker` in the Permission Management module. This is just for checking permissions, not managing (setting) them. Defined in the framework, so any module can be independent of the Permission Management module, yet can check permissions. We can implement as NullResourcePermissionChecker (just throw not implemented exception for now) in the framework to not get DI exceptions. 
- [x] Create `IResourcePermissionManager` to manage (get/set) permissions for resources in the Permission Management module, like the current IPermissionManager service. Implement it in the same place.
- [ ] Create an app service to provide application logic to manage resource permissions.
- [ ] Define a resource permission management modal, which should be reusable, like the current permission management modal.
- [ ] Documentation
  - [ ] New: Resource-Based Authorization (`framework/fundamentals/authorization/resource-based`)
  - [ ] Update the permission management module's documentation
  - [ ] Mention about the new system with related places (main authorization document, entities document, etc)

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -20,10 +20,10 @@
     <PackageVersion Include="Azure.Identity" Version="1.14.2" />
     <PackageVersion Include="Azure.Messaging.ServiceBus" Version="7.20.1" />
     <PackageVersion Include="Azure.Storage.Blobs" Version="12.25.0" />
-    <PackageVersion Include="Blazorise" Version="1.8.5" />
-    <PackageVersion Include="Blazorise.Components" Version="1.8.5" />
-    <PackageVersion Include="Blazorise.DataGrid" Version="1.8.5" />
-    <PackageVersion Include="Blazorise.Snackbar" Version="1.8.5" />
+    <PackageVersion Include="Blazorise" Version="1.8.6" />
+    <PackageVersion Include="Blazorise.Components" Version="1.8.6" />
+    <PackageVersion Include="Blazorise.DataGrid" Version="1.8.6" />
+    <PackageVersion Include="Blazorise.Snackbar" Version="1.8.6" />
     <PackageVersion Include="Castle.Core" Version="5.2.1" />
     <PackageVersion Include="Castle.Core.AsyncInterceptor" Version="2.1.0" />
     <PackageVersion Include="CommonMark.NET" Version="0.15.1" />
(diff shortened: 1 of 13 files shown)


maliming:
> we need to know available permissions for a resource type

I added a `PermissionType` to the `PermissionDefinition`. 

How can I filter resource permission definitions when granting permissions for an entity? Should I filter it based on the permission name?

eg: Check if the permission name starts with `Volo.Abp.Identity.IdentityUser` when granting permission for the `IdentityUser` entity.


```cs
public enum PermissionType
{
    /// <summary>
    /// Based on user(roles/claims).
    /// </summary>
    UserBased = 0,

    /// <summary>
    /// Based on resource(entities).
    /// </summary>
    ResourceBased = 1
}

```

```cs

public static class Resource
{
	public static class User
	{
		public const string View = "Volo.Abp.Identity.IdentityUser.View";
		public const string Edit = "Volo.Abp.Identity.IdentityUser.Edit";
		public const string Delete = "Volo.Abp.Identity.IdentityUser.Delete";
		public const string ChangeName = "Volo.Abp.Identity.IdentityUser.ChangeName";
	}

	public static class Role
	{
		public const string View = "Volo.Abp.Identity.IdentityRole.View";
		public const string Edit = "Volo.Abp.Identity.IdentityRole.Edit";
		public const string Delete = "Volo.Abp.Identity.IdentityRole.Delete";
		public const string ChangeName = "Volo.Abp.Identity.IdentityRole.ChangeName";
	}
}
```

hikalkan:
@maliming permission definition should also contain resource name (e.g. "Product"), so we can get all permissions related to the "Product" entity. When you add ResourceName to a permission definition, you even don't need to add such a PermissionType. Because, if ResourceName is not null, then it is a resource based permission.

maliming:
ok, I see, Thanks.

hikalkan:
@maliming Additional note: You know we are currently using "MyCrm.Catalog.Products.Edit" permission that is used to grant edit permission for a user/role in the application (in a product management page). That is not new.
If we want to be able to manage edit permission per product, then such a generic permission doesn't work. We should define another permission (which is resource-based) for that. Two approaches are different and can be used independently or somehow combined together.

Example: Create a standard permission to determine to view list of products (see the products page), but for every product, manage edit permissions individually. Of course, you can also define a permission like "Can manage all products" as a shortcut to give all product edit grants to a specific user, but this permission is not resource based, I mean it is not related to a single product id.

maliming:
Move to https://github.com/abpframework/abp/pull/24184

PR #24184, opened 2025-11-15
Resource based authorization

### LINKED ISSUE #236: Resource based authorization integration
https://docs.microsoft.com/en-us/aspnet/core/security/authorization/resourcebased?tabs=aspnetcore2x#operational-requirements

- We can create handler automatically.
- We can restrict data access based on CRUD operations automatically.

See also https://github.com/aspnetboilerplate/aspnetboilerplate/issues/3231

maliming:
Resolves #236

Demo: https://github.com/maliming/AbpResourcePermission

https://github.com/user-attachments/assets/574f89c5-c7ec-4a3f-a48b-c367d5172010

## Work to be done

- [x] Implement `IResourcePermissionChecker` in the Permission Management module. This is just for checking permissions, not managing (setting) them. Defined in the framework, so any module can be independent of the Permission Management module, yet can check permissions. We can implement as NullResourcePermissionChecker (just throw not implemented exception for now) in the framework to not get DI exceptions. 
- [x] Create `IResourcePermissionManager` to manage (get/set) permissions for resources in the Permission Management module, like the current IPermissionManager service. Implement it in the same place.
- [x] Create an app service to provide application logic to manage resource permissions.
- [x] Define a resource permission management modal, which should be reusable, like the current permission management modal.


maliming:
hi @hikalkan 

Our final step is to get user and role info in the **permissions** module. 

I plan to add two services to the **framework** and implement them in the **Identity** module to get user and role infor.

What do you think?

image

framework/src/Volo.Abp.Authorization.Abstractions/Volo/Abp/Authorization/Permissions/IPermissionDefinitionContext.cs
```diff
@@ -46,4 +48,15 @@ PermissionGroupDefinition AddGroup(
...
     PermissionDefinition? GetPermissionOrNull(string name);
+
+    PermissionDefinition AddResourcePermission(
+        string name,
+        string resourceName,
+        ILocalizableString? displayName = null,
+        MultiTenancySides multiTenancySide = MultiTenancySides.Both,
+        bool isEnabled = true);
+
+    PermissionDefinition? GetResourcePermissionOrNull([NotNull] string name);
```
hikalkan: Should also take a resource name. Otherwise, how to know which resource's permission is requested? Or did you design resource permissions with unique names in the system?

framework/src/Volo.Abp.Authorization.Abstractions/Volo/Abp/Authorization/Permissions/IPermissionDefinitionContext.cs
```diff
@@ -46,4 +48,15 @@ PermissionGroupDefinition AddGroup(
...
+
+    void RemoveResourcePermission([NotNull] string name);
```
hikalkan: Should also take a resource name. Otherwise, how to know which resource's permission is requested? Or did you design resource permissions with unique names in the system?

framework/src/Volo.Abp.Authorization/Volo/Abp/Authorization/Permissions/Resources/ResourcePermissionChecker.cs
```diff
@@ -0,0 +1,179 @@
...
+
+        var multiTenancySide = claimsPrincipal?.GetMultiTenancySide()
+                               ?? CurrentTenant.GetMultiTenancySide();
+
+        if (!permission.MultiTenancySide.HasFlag(multiTenancySide))
+        {
+            return false;
+        }
+
+        if (!await PermissionChecker.IsGrantedAsync(claimsPrincipal, permission.ManagementPermissionName!))
```
hikalkan: This permission is to check if the current user can change the permission of another user/role for a resource.
But here, we are not managing a user's permission. We are just checking the permission.

Example: User "admin" grants permission "Editing" of resource "blog-post" for User "halil". In this case, we should check the management permission. But, after that, when user "halil" logs in and tries to edit the blog post, we should only check if user "halil" has permission to edit the post. Management permission is irrelevant here. Understand? To properly test that, use two separate users.

hikalkan (changes requested):

framework/src/Volo.Abp.Authorization/Volo/Abp/Authorization/Permissions/Resources/ResourcePermissionChecker.cs
maliming: https://github.com/abpframework/abp/pull/24184/commits/12b50b41c5b5d2dbcc6aaa003c63e38b152cfb3b

hikalkan (approved):

PR #24187, opened 2025-11-15
Improve focus handling for form field visibility

fahrigedik:
Refactored focus logic to ensure the field is only focused when visible. Added IntersectionObserver to handle cases where the element is not immediately visible, improving accessibility and user experience.

image

### Description

Resolves https://github.com/volosoft/vs-internal/issues/7593 (write the related issue number if available)

### Checklist

### How to test it?

you have to change branch rel-9.3 on volo
you have to change branch issue-7593 on abp
you have to change branch rel-4.3 on lepton
and you can run commercial test app on vs-internal


sumeyyeKurtulus (commented):
Hello @fahrigedik thank you for checking this problem in the forms. However, we can call the animation frame request instead which is more straightforward and clean
```tsx
 requestAnimationFrame(() => {
    this.fieldRef.nativeElement.focus();
 });
```

sumeyyeKurtulus (approved):

yagmurcelk (approved):

PR #24195, opened 2025-11-17
Display average rating and total ratings in UI

### LINKED ISSUE #23615: CMS Kit: Enhancements for Rating System
Image

We need to make significant changes to the Rating System, as follows:

- [x] Don't write the star count (it's clear with the stars - no need to write 5 in the above example -), instead show the average rating!
- [x] Currently, to change the rating, user first needs to "undo" and then re-rate. Instead of doing that, user should be able to hover through the stars and set the rating whenever they want to change it! 
- [x] In the "Rating Detail" modal, the average rating should also be clearly shown:

Image

AlperenSamurlu:
https://github.com/user-attachments/assets/d806e8a7-3dde-4cf1-bff8-e84e5c8bac28

Added 'Average Rating' and 'Total Ratings' labels to localization and updated the rating component to show average rating and total ratings in the modal and main view. Calculated average rating in the view component and passed it to the view model. Removed undo rating functionality and related UI/JS code for a simplified rating experience.

### Description

Resolves #23615

### DIFF
--- a/modules/cms-kit/src/Volo.CmsKit.Domain.Shared/Volo/CmsKit/Localization/Resources/en.json
+++ b/modules/cms-kit/src/Volo.CmsKit.Domain.Shared/Volo/CmsKit/Localization/Resources/en.json
@@ -3,6 +3,7 @@
   "texts": {
     "AddSubMenuItem": "Add Sub Menu Item",
     "AreYouSure": "Are You Sure?",
+    "AverageRating": "Average Rating",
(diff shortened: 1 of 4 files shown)


MansurBesleney (approved):

yagmurcelk (changes requested):
- [ ] When we update the ratings, the star types in the rating details are not reflected correctly.

[screen-capture (25).webm](https://github.com/user-attachments/assets/3f573422-d2ff-40e3-b6a3-1c7375a20c98)

yagmurcelk (approved):

PR #24210, opened 2025-11-18
Key-based hybrid/distributed locking

MarkCiliaVincenti:
Rather than locking using a SemaphoreSlim for all caches, we can lock more intelligently, per cache key, using pooled semaphores. Uses a 3rd party library but does not introduce a new dependency to the whole project.


MarkCiliaVincenti:
@maliming

maliming:
hi @MarkCiliaVincenti 

We don't want to add a new package dependency. `SemaphoreSlim` is enough for us.

Thanks for your contribution, 👍

MarkCiliaVincenti:
> hi @MarkCiliaVincenti
> 
> We don't want to add a new package dependency. `SemaphoreSlim` is enough for us.
> 
> Thanks for your contribution, 👍

@maliming this dependency is already part of ABP.

PR #24266, opened 2025-11-24
Fixed misleading parameter variable names of ReverseMap methods.

### LINKED ISSUE #24265: The ReverseMap parameter names are misleading

### Description

The new Mapperly object mapper module has the abstract base class `TwoWayMapperBase` and some associated `ReverseMap` methods.   The parameter names are misleading.  When the object is reverse mapped, the `TDestination` object is copied to the `TSource`

Therefore the parameters should be named `TDestination source` and `TSource destination` like so:

```cs
public abstract class TwoWayMapperBase<TSource, TDestination> : MapperBase<TSource, TDestination>, IAbpReverseMapperlyMapper<TSource, TDestination>
{
    public abstract TSource ReverseMap(TDestination source);

    public abstract void ReverseMap(TDestination source, TSource destination);

    public virtual void BeforeReverseMap(TDestination source)
    {
    }

    public virtual void AfterReverseMap(TDestination source, TSource destination)
    {
    }
}
```

### Reproduction Steps

_No response_

### Expected behavior

_No response_

### Actual behavior

_No response_

### Regression?

_No response_

### Known Workarounds

_No response_

### Version

10.0.0

### User Interface

Common (Default)

### Database Provider

EF Core (Default)

None (Default)

### Operation System

Windows (Default)

### Other information

_No response_

ejohnson-dotnet:
The ReverseMap methods will map the destination type back to the source type, but the "source" object will be of type TDestination.   In other words, the "TDestination source" object is copied to the "TSource destination" object.

### Description

Resolves #24265

This fixes the misleading parameter names to make clear which is the source object and which is the destination object.

This file already has the parameter names correctly implemented.
[modules/identityserver/src/Volo.Abp.IdentityServer.Domain/Volo/Abp/IdentityServer/IdentityServerMapperlyMappers.cs](/abpframework/abp/blob/dev/modules/identityserver/src/Volo.Abp.IdentityServer.Domain/Volo/Abp/IdentityServer/IdentityServerMapperlyMappers.cs#L209)

### Checklist

### How to test it?


maliming (approved):

maliming:
Thanks @ejohnson-dotnet  👍

PR #24286, opened 2025-11-27
DB options UI options images and diagram replaced with new ones

MansurBesleney:
Resolves [vs internal issue #7565](https://github.com/volosoft/vs-internal/issues/7695)


gizemmutukurt (approved):
should be demo app 

image

Issue #24303, opened 2025-12-01
Javascript proxy generation: all reference type and enum properties in a record are generated as nullable


hillin:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Description

All reference type properties and enum properties in a `record` type are generated as nullable.

### Reproduction Steps

```csharp
public enum TestEnum
{
    A,
    B,
}

public record TestDto(Guid GuidValue, string StringValue, int IntValue, TestEnum EnumValue);
```

> abp generate-proxy -t ng --entry-point .

```typescript
export interface TestDto {
  guidValue?: string;
  stringValue?: string;
  intValue: number;
  enumValue?: TestEnum;
}

```

### Expected behavior

Since these properties `guidValue`, `stringValue` and `enumValue` are not defined as nullable in the record, they should not be generated as nullable in the typescript proxy code.

If records are not supposed to be used as DTOs, it should be documented.

### Actual behavior

The abovementioned properties are generated as nullable.

### Version

9.3.6

### User Interface

Angular

### Database Provider

None/Others

### Tiered or separate authentication server

None (Default)

### Operation System

Windows (Default)

### Other information

The enum properties part is related to https://github.com/abpframework/abp/issues/22132

JanneHarju:
Is this fix coming to Abp9? Because for us bug seems to be also at there. And @hillin  also used 9.3.6. Currently this is only added to Abp 10 milestone.

JanneHarju:
I just noticed that this was removed from Abp 10 milestone. So was there problem or not? And how we should take care of this problem?

maliming:
Duplicate of #24306

PR #24332, opened 2025-12-04
Improve nullable property detection in API modeling

### LINKED ISSUE #24306: Javascript proxy generation nullability issues roundup
Over the past year, I have reported a few issues ([#22132](https://github.com/abpframework/abp/issues/22132), [#22798](https://github.com/abpframework/abp/issues/22798), [#24303](https://github.com/abpframework/abp/issues/24303)) regarding the JavaScript proxy generation system. The proxy generator is a great feature that frees us from painfully generating frontend proxy code. ABP has made many excellent design choices in building this system, but unfortunately, it's still not correctly handling the nullability of DTO properties. I apologize for spamming the issues tracker with tickets of this nature, but I have a strong instinct that these problems belong to one systematic flaw in the design, and I hope this issue can address them altogether, once and for all.

With that in mind, I have put together the following truth table:

| C# | Example Type | T? | T | required T? | required T |
| :--- | :--- | :--- | :--- | :--- | :--- |
| **Built-in Value** | `int` | ✅ `number?` | ❌ `number` | ✅ `number?` | ❌ `number` |
| **Special Struct** | `System.Guid` | ✅ `string?` | ✅ `string?` | ✅ `string?` | ❌ `string` |
| **Custom Struct**¹ | `struct S {}` | ✅ `S?` | ❌ `S` | ✅ `S?` | ❌ `S` |
| **CLR Enum**² | `System.DayOfWeek` | ✅ `any?` | ✅ `any?` | ✅ `any?` | ❌ `any` |
| **Custom Enum** | `enum E {}` | ✅ `E?` | ✅ `E?` | ✅ `E?` | ❌ `E` |
| **Special Class** | `string` | ✅ `string?` | ✅ `string?` | ❌ `string` | ❌ `string` |
| **Custom Class** | `class C {}` | ❌ `C` | ❌ `C` | ❌ `C` | ❌ `C` |
| **Array** | `int[]` | ❌ `number[]` | ❌ `number[]` | ❌ `number[]` | ❌ `number[]` |
| **Dictionary** | `Dictionary<string, string>` | ❌ `Record<string, string>` | ❌ `Record³` | ❌ `Record³` | ❌ `Record³` |

> 1. Custom structs are generated such as `interface S extends any`, which does not work.
> 2. Not all CLR enums are tested.
> 3. Shortened for conciseness.

This table shows how the nullability of a C# type is handled in the generated TypeScript code. For example, `int?` $\to$ `number?`, `int` $\to$ `number`, `required int?` $\to$ `number?`, `required int` $\to$ `number`. For clarity of reading, I added a ✅ for nullable type and ❌ for non-nullable type generated.

From this table, it is difficult to get a clear impression of how nullability works in this system. In real-world usage, it brings a lot of surprises. For instance, your C# DTO might say an `int[]` property is optional, but on the TypeScript side, you are forced to provide a value. If you see an optional enum property, naturally omit it, and pass that DTO to your C# backend, you might not realize that on the C# side it's actually non-nullable, resulting in a deserialization exception. For a C# `required string?`, you are not allowed to pass a `null` value on the TypeScript side. In a strictly regulated codebase, we are forced to fix all these nullability issues every time we regenerate the proxy, which makes the proxy generator much less of a pleasure to use than it should be.

Base on the discussion in [this issue](https://github.com/abpframework/abp/issues/22132), it seems ABP's intention was to use the `required` keyword to determine nullability. However:

1. We can see this rule is **not consistent** across different types. While all `required T` are generated as non-nullable, `T?`, `T`, and `required T?` all behave differently. It seems the rule - if there is one clearly defined - is very convoluted.
2. We can argue that `required` is semantically **not suitable** for expressing whether a property should be nullable. `required T?` is totally valid: this property must be specified, but it can be set to `null`.

On the other hand, it's far more natural to use **.NET's nullability annotation** for this job, which brings the rules down to a nice one-liner: *If the property is nullable in C# (type name followed by a question mark), it is generated as nullable.*

I did not explore the technical possibilities, and totally understand that there might be difficulty to retrieve the nullability information in the runtime, so I'm definitely open to ideas.

fahrigedik:
Added ReflectionHelper.IsNullable for .NET 6+ to check nullable reference types. Updated PropertyApiDescriptionModel to append '?' for nullable properties. Refined isOptionalProperty in TypeScript to check both type and typeSimple for optionality.

### Description

Resolves #24306 #24303 (write the related issue number if available)

### DIFF
--- a/framework/src/Volo.Abp.Core/Volo/Abp/Reflection/ReflectionHelper.cs
+++ b/framework/src/Volo.Abp.Core/Volo/Abp/Reflection/ReflectionHelper.cs
@@ -8,24 +8,6 @@ namespace Volo.Abp.Reflection;
 //TODO: Consider to make internal
 public static class ReflectionHelper
 {
-#if NET6_0_OR_GREATER
-    /// <summary>
-    /// Checks whether the property is nullable, including nullable reference types (NRT).
-    /// </summary>
-    /// <param name="propertyInfo">Property info to check</param>
-    public static bool IsNullable(PropertyInfo propertyInfo)
-    {
-        if (propertyInfo.PropertyType.IsValueType)
-        {
-            return false;
-        }
-
-        var nullabilityInfoContext = new NullabilityInfoContext();
-        var nullabilityInfo = nullabilityInfoContext.Create(propertyInfo);
-        return nullabilityInfo.ReadState == NullabilityState.Nullable;
-    }
-#endif
-
     //TODO: Ehhance summary
     /// <summary>
     /// Checks whether <paramref name="givenType"/> implements/inherits <paramref name="genericType"/>.
--- a/framework/src/Volo.Abp.Core/Volo/Abp/Reflection/ReflectionHelper.cs
+++ b/framework/src/Volo.Abp.Core/Volo/Abp/Reflection/ReflectionHelper.cs
@@ -8,6 +8,24 @@ namespace Volo.Abp.Reflection;
 //TODO: Consider to make internal
 public static class ReflectionHelper
 {
+#if NET6_0_OR_GREATER
+    /// <summary>
+    /// Checks whether the property is nullable, including nullable reference types (NRT).
+    /// </summary>
+    /// <param name="propertyInfo">Property info to check</param>
+    public static bool IsNullable(PropertyInfo propertyInfo)
+    {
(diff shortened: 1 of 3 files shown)


maliming:
hi 

Instead of adding `?` to the end of `Type`, we can add a new property to the model.

`public bool IsNullable { get; set; }`

PR #24337, opened 2025-12-04
Referral link error message updated

ahmetcelik05:
### Description

Related to [vs-internal-issue-#7297](https://github.com/volosoft/vs-internal/issues/7297)

### Checklist

- [ ] no need to document


yagmurcelk (changes requested):
- [x] There is a localization problem in the error message.

Scenario: A referral link was created for a user, and then the user bought their own org without using it. The error message that appears when they try to use this link after buying their own org.

image

- [ ] The same problem occurs in the error message in the scenario where we try to delete the link that appears not used when we reload the page.

image

yagmurcelk (approved):

PR #24425, opened 2025-12-15
Remove `AsyncKeyedLock` dependency from `LocalAbpDistributedLock`.

maliming:


MarkCiliaVincenti:
What's the motivation here, just dropping the dependency? This version of the code has a memory leak. For every distinct key that gets used, a SemaphoreSlim object remains permanently in memory. AsyncKeyedLock is battle-tested and has been used in ABP for years without issues.

EngincanV (approved):

PR #24448, opened 2025-12-18
Add provider check to SqliteBusyTimeout interceptor

salihozkara:
Updated SqliteBusyTimeoutSaveChangesInterceptor to execute the PRAGMA command only if the DbContext is using the Sqlite provider. This prevents unnecessary execution for non-Sqlite databases.


maliming:
hi @salihozkara 

I added some new code. Can you confirm if it works?

https://github.com/abpframework/abp/pull/24448/commits/c04477925371a14f7321c201abb88e89d36fb1be

Thanks.

maliming (approved):

PR #24499, opened 2025-12-29
Explain how to install an older version of ABP Studio

ismcagdas:
resolves https://github.com/volosoft/abp-studio/issues/2581


maliming:
The version 2.0.0 download link is 404

https://abp.io/api/abp-studio/download/r/windows/abp-studio-2.0.0-beta-full.nupkg

It seems that if version> `1.0.0`, the `channel` is `stable`.

https://abp.io/api/abp-studio/download/r/windows/abp-studio-1.0.1-stable-full.nupkg
https://abp.io/api/abp-studio/download/r/windows/abp-studio-2.0.0-stable-full.nupkg


We can consider listing all major version download links in the document.
https://abp.io/docs/latest/studio/version-mapping
image

enisn:
Yes, the original script was during the beta stage. It should use `stable` channel by default.
And this powershell script works **only** on windows currently.

enisn:
There is an **unofficial method** to switch between different versions on all the platforms and processor architectures:
https://github.com/enisn/AbpDevTools?tab=readme-ov-file#switch-abp-studio-version

As a '3rd party approach' MAC users can use this or get inspiration from this C# command to create a better powershell script:
https://github.com/enisn/AbpDevTools/blob/master/src/AbpDevTools/Commands/AbpStudioSwitchCommand.cs

maliming:
Do we have an option to disable automatic updates? Some people have reported that Studio automatically upgrades to a newer version. If they want to keep the old version, it shouldn't upgrade automatically.

ismcagdas:
@maliming we don't have such an option at the moment and we don't want to do it at least for now. We prefer to fix problems and release new version for such cases. We can discuss this in the team later.

@enisn I updated PR to include AbpDevTools

enisn:
I mean we can create another PowerShell to support Mac, if it's ok currently I'm approving

ismcagdas:
Since your tool is working, no need to create a duplicate one in ABP repo.

enisn (approved):

PR #24519, opened 2025-12-30
Improve entity history handling for shared types and JSON

salihozkara:
Enhanced EntityHistoryHelper to correctly handle entities with shared CLR types and properties mapped to JSON. Added logic to determine property type from entry values when the type is object, and improved property change tracking for JSON-mapped navigation properties.


maliming:
hi

Can you add a unit test for the new method?

maliming (approved):

PR #24531, opened 2026-01-03
Fix condition check in ResourcePermissionManager to ensure provider match before returning

maliming:
This is different from the modal in permissions. We should be able to continue setting resource permissions for roles and the users under the roles.


yagmurcelk (changes requested):
When you try to add an existing user/role in the added permissions, it doesn't show an error message.


[screen-capture (63).webm](https://github.com/user-attachments/assets/21ec1f63-ea79-4725-a612-b4445346dd26)

maliming:
Granting resource permissions to existing users/roles will update them.

yagmurcelk (approved):

PR #24577, opened 2026-01-08
Replace LazyLocalizationPipe with AsyncLocalizationPipe

### LINKED ISSUE #22901: Angular - Revising the name for the async localization pipe
- [ ] **Waiting for releasing the merged PR, we will configure the other repositories.**
---
### Description

This issue is expected to rename this pipe to make it easier to express what it does and how it is different from the other localization pipe: **like AsyncLocalizationPipe**

https://github.com/abpframework/abp/blob/f580e48aff425aab6175a625a9480c24897aa925/npm/ng-packs/packages/core/src/lib/pipes/lazy-localization.pipe.ts#L18

fahrigedik:
Updated the routes component to use AsyncLocalizationPipe instead of LazyLocalizationPipe for route name localization. Adjusted template bindings to use the new pipe, improving consistency and potentially performance.

### Description

Resolves #22901 (write the related issue number if available)

### DIFF
--- a/npm/ng-packs/packages/theme-basic/src/lib/components/routes/routes.component.html
+++ b/npm/ng-packs/packages/theme-basic/src/lib/components/routes/routes.component.html
@@ -10,7 +10,7 @@
         @if (route.iconClass) {
         <i [ngClass]="route.iconClass"></i>
         }
-        {{ route.name | abpAsyncLocalization }}
+        {{ route.name | abpAsyncLocalization | async}}
       </a>
     </li>
   </ng-template>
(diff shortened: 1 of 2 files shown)


npm/ng-packs/packages/theme-basic/src/lib/components/routes/routes.component.html
```diff
@@ -1,109 +1,78 @@
...
 
   <ng-template #defaultLink let-route>
     <li class="nav-item" *abpPermission="route.requiredPolicy">
       <a class="nav-link" [routerLink]="[route.path]">
         @if (route.iconClass) {
-          <i [ngClass]="route.iconClass"></i>
+        <i [ngClass]="route.iconClass"></i>
         }
-        {{ route.name | abpLazyLocalization | async }}
+        {{ route.name | abpAsyncLocalization }}
```
sumeyyeKurtulus: As this pipe returns an observable, we need to keep the async pipe here. We should also update the other repositories in the same way.

sumeyyeKurtulus (approved):

PR #24595, opened 2026-01-09
Use `AbpIdentityErrorDescriber` to localize error message instead of `AbpIdentityResultExtensions`.

maliming:


EngincanV (approved):

roc916:
Will it affect the serialization of remote service information?

PR #24624, opened 2026-01-12
Delete .github/pr-review.instructions.md

maliming:
Copilot doesn't support these instructions. : (


EngincanV:
@maliming The related file should be under an `instructions` folder (https://docs.github.com/en/copilot/how-tos/configure-custom-instructions/add-repository-instructions#creating-path-specific-custom-instructions-3), can you please move it under to the relevant folder (the folder does not exists)? Also, can you add `applyTo: *` in the relevant instruction file?

---

**Update:** https://github.com/abpframework/abp/pull/24610 It's already been tried under the `instructions` folder, but it did not work. So, merging the PR...

EngincanV (approved):

PR #24645, opened 2026-01-14
Sign out user if configuration indicates unauthenticated

maliming:


maliming:
`CheckTokenExpiration` did the same thing.

PR #24666, opened 2026-01-17
Use DI to resolve the Hangfire adapter, replacing Activator.CreateInstance.

LEIRONGHUA:
### Description

1. Use DI to resolve the Hangfire adapter, replacing Activator.CreateInstance.
2. Refactor variable naming for cron expressions in Hangfire background worker manager.

The reasons for the modification are as follows: 

The `TimeZone` property in `HangfirePeriodicBackgroundWorkerAdapter` defaults to `TimeZoneInfo.Utc`, causing `ExcelFileCleanupOptions.CronExpression = "0 23 * * *"` to actually execute at 23:00 UTC (07:00 in China). There is a time zone discrepancy between the actual execution time of `CronExpression` and this.

After switching to dependency injection, CronExpression can execute based on the local timezone in the following way.

1. Create a new subclass of HangfirePeriodicBackgroundWorkerAdapter, such as MyAbpHangfirePeriodicBackgroundWorkerAdapter, and modify TimeZone to TimeZoneInfo.Local in the default constructor.
```
public class MyAbpHangfirePeriodicBackgroundWorkerAdapter<TWorker> : HangfirePeriodicBackgroundWorkerAdapter<TWorker>
    where TWorker : IBackgroundWorker
{
    public MyAbpHangfirePeriodicBackgroundWorkerAdapter()
    {
        TimeZone = TimeZoneInfo.Local;
    }
}

```
2. Replace the original HangfirePeriodicBackgroundWorkerAdapter
```
    {
        context.Services.AddSingleton(typeof(MyAbpHangfirePeriodicBackgroundWorkerAdapter<>));
        context.Services.Replace(ServiceDescriptor.Singleton(typeof(HangfirePeriodicBackgroundWorkerAdapter<>),
            typeof(MyAbpHangfirePeriodicBackgroundWorkerAdapter<>)));
    }
```


maliming:
hi

An option will be fine. 

We can set the properties after `var workerAdapter = (Activator.CreateInstance(adapterType) as IHangfireBackgroundWorker)!;`

https://github.com/abpframework/abp/blob/dev/framework/src/Volo.Abp.BackgroundWorkers.Hangfire/Volo/Abp/BackgroundWorkers/Hangfire/HangfireBackgroundWorkerBase.cs#L10-L16

Thanks.

LEIRONGHUA:
Hi  Are you suggesting adding the DefaultTimeZone property to the HangfireOptions option? If so, would this conflict with the TimeZone property in HangfireBackgroundWorkerBase? 

Or is it like this?

1:  AbpHangfireOptions.cs add DefaultTimeZone property
```
    /// <summary>
    /// Hangfire time zone, default is Utc.
    /// </summary>
    public TimeZoneInfo DefaultTimeZone { get; set; } = TimeZoneInfo.Utc;
```

2:  HangfireBackgroundWorkerManager.cs 
image


Regardless of the final outcome，I suggest using
`var workerAdapter = (ServiceProvider.GetRequiredService(adapterType) as IHangfireBackgroundWorker)!;`
instead of
`var workerAdapter = (Activator.CreateInstance(adapterType) as IHangfireBackgroundWorker)!;`
This is more in line with best practices.

maliming:
We can create a `HangfirePeriodicBackgroundWorkerAdapterOptions` and add properties.

maliming:
I made the changes: https://github.com/abpframework/abp/pull/24669

LEIRONGHUA:
> I made the changes: #24669

Ok, I'll close this PR

PR #24725, opened 2026-01-23
Migrate ABP Angular templates to Vitest as the default testing framework - Issue 24714

### LINKED ISSUE #24714: Migrate ABP templates to Vitest as the default testing framework
We should migrate the existing test setup in ABP templates to use Vitest instead of the current testing framework (e.g., Jest). Vitest offers faster test execution, native ESM support, simpler configuration, and better TypeScript experience — aligning with modern Angular and frontend ecosystem trends.

Reference : https://angular.dev/guide/testing

fahrigedik:
### Description

Resolves #24714 (write the related issue number if available)

This PR updates the ABP Angular templates to use Vitest as the default testing framework instead of the existing setup

### How to test it?
Run `yarn install` (or simply `yarn`) in `abp/templates/app/angular`
Run yarn test in `abp/templates/app/angular`
Repeat the same steps for the other Angular template folders as well

Note: Any existing or newly discovered test issues will be handled separately. This PR is intentionally scoped only to the Vitest migration.


sumeyyeKurtulus (changes requested):
Hello @fahrigedik thanks for migrating the test environment for the templates. Here is my comment that applies for all changes

- [x] It is better to normalize the vitest version to `^4.0.0` instead as we did the same here in dev-app https://github.com/abpframework/abp/blob/bdd621b6bd10d59f8905199e84edd9f64d02d600/npm/ng-packs/package.json#L148

- [x] The test file should also be updated as the tests fail when we run `yarn test` command

Image

sumeyyeKurtulus (approved):

PR #24775, opened 2026-01-30
Prevent Privilege Escalation: Add Assignment Restrictions for Roles and Permissions

### LINKED ISSUE #24768: Privilege escalation issue in the Identity management module

### Description

When working with the Identity Management Module and creating multiple different roles, with tiered access levels, there's no way of limiting less privileged users of creating/modifying users with admin or higher privileges.

I can have the admin role with all the permissions (as provided by default) and then a different role with restricted permissions, but with permissions to create users and assign roles, users with that specific role are allowed to change their role to Admin or create a new Admin user without any restriction.

There should be a way of configuring what roles and users are assignable from each existing role.

### Reproduction Steps

1. Have the default Admin role
2. Create a Manager role with less privileges, but with the User Management - Edit - Manage roles permission
3. Create an user with the newly created Manager role
4. Login and edit the same manager user adding the Admin role to that user
5. You end up in a privilege escalation situation

### Expected behavior

1. Have the default Admin role
2. Create a Manager role with less privileges, but with the User Management - Edit - Manage roles permission
3. Configure the Manager role checking which roles are assignable from that role, leaving the Admin role unchecked
4. Create an user with the newly created Manager role
5. Login with that new manager user and try to edit roles
6. You will not be able of assigning the Admin role to that user, only the authorized roles.

### Actual behavior

Actual behavior doesn't validate which roles are allowed to be assigned from a current user with a specific role.

### Regression?

_No response_

### Known Workarounds

Implementing a custom IdentityUserAppService that overrides the original one with added allowed Roles validation.
Implementing a custom entity action contributor for the User entity so the UI can reflect those restrictions.

### Version

10.0.2

### User Interface

Angular

### Database Provider

EF Core (Default)

None (Default)

### Operation System

Windows (Default)

### Other information

_No response_

maliming:
Resolve #24768

#### Problem Description

The system has privilege escalation vulnerabilities:

1. Users with `AbpIdentity.Users.Update` permission can assign themselves or others roles they do not currently have
2. Users with `AbpIdentity.Users.ManagePermissions` or `AbpIdentity.Roles.ManagePermissions` can grant permissions they do not possess
3. Users can escalate privileges by modifying their own user permissions or the permissions of roles they belong to
4. There is no validation to prevent indirect privilege escalation through role or permission delegation
5. UI restrictions can be bypassed by calling APIs directly

---

#### Solution

This PR implements a unified privilege escalation prevention model based on a single security principle:

> **A user must not be able to directly or indirectly cause any user (including themselves) to obtain roles or permissions they do not already possess.**

The solution consists of **3 core security mechanisms**, applied uniformly to all users:

---

### 1. Role Assignment Restriction (Identity Module)

- Users can **only assign or remove roles they currently have**
  - Example: if User A has roles `A` and `B`, they can only assign or remove `A` and `B` for other users
- Users **cannot add new roles to themselves** (removal only)
- Users **cannot assign or remove roles they do not possess**
- Role updates use an **incremental update pattern**:
  - Only roles the current user is authorized to operate on are added or removed
  - All other roles are preserved to prevent silent or accidental removal
- All validations are enforced on the backend (UI is not a security boundary)

---

### 2. Permission Grant / Revoke Authorization (PermissionManagement Module)

- Users can **only grant or revoke permissions they currently have** (effective permissions)
- Validation applies to both:
  - Grant operations
  - Revoke operations
- Prevents privilege escalation via permission manipulation or delegation
- Backend validation is mandatory and enforced even when APIs are called directly

---

### 3. Incremental Permission Protection (User & Role Permissions)

- When updating user or role permissions:
  - Permissions the current user does **not** have are treated as **non-editable**
  - Non-editable permissions are **preserved as-is** during updates
- UI marks permissions with `IsEditable = false` when the user lacks authority
- Backend enforces the same rules to prevent API bypass
- Prevents accidental or malicious removal of permissions outside the user's control

---

### Admin Role Exception

- Users with the `admin` role can assign any role and grant/revoke any permission.

---

#### Security Rules Summary

| Operation | Self | Other Users |
| --- | --- | --- |
| **Assign Roles** | ❌ Cannot add new roles | ✅ Only roles the user currently has |
| **Remove Roles** | ✅ Can remove own roles | ✅ Only roles the user currently has |
| **Grant Permissions** | ✅ Only permissions the user has | ✅ Only permissions the user has |
| **Revoke Permissions** | ✅ Only permissions the user has | ✅ Only permissions the user has |
| **Edit Non-Editable Permissions** | ❌ Preserved as-is | ❌ Preserved as-is |

*Admin exception: users with the `admin` role can assign any role and grant/revoke any permission.*

---

#### Scenario Comparison

| Scenario | Before | After |
| --- | --- | --- |
| User assigns a role they do not have to themselves | ✅ Success | ❌ Blocked |
| User assigns a higher-privileged role to others | ✅ Success | ❌ Blocked |
| User assigns one of their own roles to others | ✅ Success | ✅ Success |
| User removes one of their own roles from themselves | ✅ Success | ✅ Success |
| User grants a permission they do not have | ✅ Success | ❌ Blocked |
| User revokes a permission they do not have | ✅ Success | ❌ Blocked |
| User modifies protected permissions via API | ✅ Success | ❌ Blocked |
| User escalates privileges indirectly via delegation | ✅ Success | ❌ Blocked |

*Admin exception applies to the “After” column for users with the `admin` role.*

---

#### Tests

- Added unit and integration tests to cover:
  - Role assignment restrictions
  - Permission grant/revoke restrictions
  - Self-modification protection
  - Incremental update behavior
  - API-level enforcement (bypassing UI)


EngincanV (approved):

maliming:
I think the `admin` role there shouldn't be any restrictions. I'll add this rule.

yagmurcelk (changes requested):
A user who doesn't currently have the admin role can give the admin role to another user. 


https://github.com/user-attachments/assets/25af79c9-a827-4aaa-a1b9-40bcaed4bf45

maliming:
> A user who doesn't currently have the admin role can give the admin role to another user.


Please create open-source test projects to test this PR.

yagmurcelk (commented):
I approve this PR because the PRs in the `dev ` branch will no longer be tested. Please assign it to someone else and ask them to merge it.

yagmurcelk (approved):

PR #24908, opened 2026-02-20
ABP Cursor Plugin

EngincanV:
### Description

Resolves https://github.com/volosoft/vs-internal/issues/8201

I created the ABP Cursor Plugin and will submit it after this PR is merged. We may consider moving this plugin to a different repository just dedicated to plugins, like others do (for example, [Notion](https://github.com/makenotion/cursor-notion-plugin) and [Figma](https://github.com/figma/mcp-server-guide)). Please let me know, if we should move to the plugin to a new repository. @ismcagdas 

### Checklist


EngincanV:
Moved the plugin to https://github.com/abpframework/cursor-abp-plugin & submitted the plugin to Cursor.

PR #24909, opened 2026-02-20
Improve FindAsync method to handle entity attachment and state management

ugurozturk:
If the EntityFrameworkModule configured to "optionsBuilder.UseQueryTrackingBehavior(QueryTrackingBehavior.NoTracking"

auditLog Entity Proporty Change tracking does not work even EnableEntityChangeTracking or repository.EnableTracking. Because returned object from Find method is not actually attached of DbContext. 

To fix it, i needed to add this part only for ShouldTrackingEntityChange returns true

        if (dbSet.Entry(entity).State == EntityState.Detached)
        {
            dbSet.Attach(entity);
        }


ugurozturk:
#24913

PR #24913, opened 2026-02-20
Improve FindAsync method to handle entity attachment and state manage…

ugurozturk:
If the EntityFrameworkModule configured to "optionsBuilder.UseQueryTrackingBehavior(QueryTrackingBehavior.NoTracking"

auditLog Entity Proporty Change tracking does not work even EnableEntityChangeTracking or repository.EnableTracking. Because returned object from Find method is not actually attached of DbContext. 

To fix it, i needed to add this part only for ShouldTrackingEntityChange returns true

        if (dbSet.Entry(entity).State == EntityState.Detached)
        {
            dbSet.Attach(entity);
        }
        
As far as i can check, the default Repository Extension methods are also does not work as tracking. If default behavior is as "NoTracking" EnableEntityChangeTracking or repository.EnableTracking doesn't work

### DIFF
--- a/framework/src/Volo.Abp.EntityFrameworkCore/Volo/Abp/Domain/Repositories/EfCoreRepositoryExtensions.cs
+++ b/framework/src/Volo.Abp.EntityFrameworkCore/Volo/Abp/Domain/Repositories/EfCoreRepositoryExtensions.cs
@@ -49,6 +49,6 @@ public static class EfCoreRepositoryExtensions
     public static IQueryable<TEntity> AsNoTrackingIf<TEntity>(this IQueryable<TEntity> queryable, bool condition)
         where TEntity : class, IEntity
     {
-        return condition ? queryable.AsNoTracking() : queryable;
+        return condition ? queryable.AsNoTracking() : queryable.AsTracking();
     }
 }
--- a/framework/src/Volo.Abp.EntityFrameworkCore/Volo/Abp/Domain/Repositories/EntityFrameworkCore/EfCoreRepository.cs
+++ b/framework/src/Volo.Abp.EntityFrameworkCore/Volo/Abp/Domain/Repositories/EntityFrameworkCore/EfCoreRepository.cs
@@ -476,30 +476,9 @@ public class EfCoreRepository<TDbContext, TEntity, TKey> : EfCoreRepository<TDbC
 
     public virtual async Task<TEntity?> FindAsync(TKey id, bool includeDetails = true, CancellationToken cancellationToken = default)
     {
-        if (includeDetails)
-        {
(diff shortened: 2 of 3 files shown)


maliming:
hi @ugurozturk 

**Thank you for your contribution!  I made some new changes.** The issue you identified is real. We've built upon your fix with a more thorough solution. Can you test if it works for your case?


### Root Cause

Your PR only fixes `FindAsync(TKey id)`, but the root cause is a flaw in the `AsNoTrackingIf` method:

```csharp
// Original implementation
public static IQueryable<TEntity> AsNoTrackingIf<TEntity>(..., bool condition)
{
    return condition ? queryable.AsNoTracking() : queryable;
    //                                           ↑ Does nothing when tracking IS needed.
    //                                             Global NoTracking config still takes effect!
}
```

`GetQueryableAsync()` calls this method internally:

```csharp
public async override Task<IQueryable<TEntity>> GetQueryableAsync()
{
    return (await GetDbSetAsync()).AsQueryable().AsNoTrackingIf(!ShouldTrackingEntityChange());
}
```

When `ShouldTrackingEntityChange() = true` (i.e., tracking is needed), `AsNoTrackingIf` simply returns the original queryable without explicitly calling `.AsTracking()`. This means the global `QueryTrackingBehavior.NoTracking` configuration still takes effect — and not just for `FindAsync(id)`. All methods that go through `GetQueryableAsync()` are affected: `GetListAsync`, `FindAsync(predicate)`, `GetPagedListAsync`, and so on.

### Our Fix

**Change 1: Fix `AsNoTrackingIf` (one-line change, global impact)**

```csharp
// Fixed
return condition ? queryable.AsNoTracking() : queryable.AsTracking();
//                                            ↑ Explicit AsTracking() overrides global NoTracking
```

This makes `GetQueryableAsync()` fully controlled by ABP's own tracking logic, independent of any EF Core global configuration.

**Change 2: `FindAsync(id)` can now be greatly simplified**

```csharp
// Fixed (4 lines)
public virtual async Task<TEntity?> FindAsync(TKey id, bool includeDetails = true, ...)
{
    return includeDetails
        ? await (await WithDetailsAsync()).OrderBy(...).FirstOrDefaultAsync(...)
        : await (await GetQueryableAsync()).OrderBy(...).FirstOrDefaultAsync(...);
}
```

Since `GetQueryableAsync()` now correctly handles tracking, the special path using `dbSet.FindAsync()` + Detached state check + `Attach` is no longer needed.

### Coverage Comparison

| Scenario | Your PR | New Fix |
|----------|---------|---------|
| `FindAsync(id)` | ✅ | ✅ |
| `FindAsync(predicate)` | ❌ | ✅ |
| `GetListAsync` | ❌ | ✅ |
| `GetPagedListAsync` | ❌ | ✅ |
| Other methods via `GetQueryableAsync` | ❌ | ✅ |

Thanks again for pinpointing this bug — your analysis was spot on. We've completed a more comprehensive fix on top of your work.


Thanks.

maliming (approved):

PR #24915, opened 2026-02-21
Fix OIDC security requirement in AddAbpSwaggerGenWithOidc

alvstw:
Current Bug:
Generated OpenAPI ended up with anonymous security entries`[{}]` instead of a concrete OIDC requirement, so Swagger UI treated operations as optional/anonymous

This PR fixes the scheme mismatch in OIDC Swagger setup


maliming (approved):

maliming:
Thank you for your contribution. 👍

PR #24954, opened 2026-02-25
Create JQueryFormScriptContributor.cs

salihozkara:
Related https://github.com/abpframework/abp/pull/24703


maliming:
Please use `5.2.0-rc.1` as the LeptonX package version.

PR #25024, opened 2026-03-06
Add `Volo.Abp.OperationRateLimiting` module

### LINKED ISSUE #17815: Operation Rate Limiting
_THIS IS CURRENTLY IN DRAFT. NEEDS TO CLARIFICATION / ANALYZE_

## Intro

We may have certain operations in our application that we want to implement a kind of rate limiting. This is not for [API call rate limiting middleware of ASP.NET Core](https://learn.microsoft.com/en-us/aspnet/core/performance/rate-limit), a web independent system to control frequency of operations.

## Example use cases

* Do not allow to make "forgot email" request for the same email address more than 3 times in an hour.
* Do not allow to get a "monthly sales report" more than 2 times in a day for each user (maybe generating that report uses a lot of system resources).

## Implementation details

We may define a OperationRateLimitingRule with the following properties:

* OperationName (string) (a unique string per operation)
* Duration (int) (as seconds) (it also can be a 1 day, 1 week, 1 month, ...)
* MaxExecutionCount (int) (in the given duration)

When we want to check if we are allowed to perform the operation, we may pass the following parameters to a service, say `IOperationRateLimitingManager.CheckAsync` method:

* operationName (string): Matches to a name in the configured rule definitions.
* parameter (string, optional): This will be used to control rate limiting for different entities. For example, we may pass current user's id, so different users won't share the same rate. We may pass tenantid if we want to share same limit for all users of a tenant. If we want to pass multiple values, we can join by `_` for example. This is an arbitrary parameter to distiguish different people/entity.

We can save operation values in the `IDistributedCache` as the default store and allow to implement a custom store (like a database store).

## Additional notes

That feature seems overlaps with the [feature system](https://docs.abp.io/en/abp/latest/Features) especially when we use it to limit tenant operations. However, features system is used to limit tenants for business purposes (to get more money from some tenants that have more wide limits) but rate limiting system is for saving system resources (we don't want to send too many emails) and prevent security attacks.

maliming:
Document: https://github.com/abpframework/abp/blob/dev/docs/en/framework/infrastructure/operation-rate-limiting.md

Resolve #17815

Demo UI: https://github.com/abpframework/abp/pull/25025

Adds the `Volo.Abp.OperationRateLimiting` package — a web-agnostic, application-level rate limiting module for ABP Framework.

Unlike ASP.NET Core's built-in HTTP rate limiting middleware (which operates at the request pipeline level), this module is designed to be called explicitly from application service methods, domain logic, or anywhere in the DI container. It is not tied to HTTP and has no dependency on `HttpContext`.

**Getting started**

Add a dependency on `AbpOperationRateLimitingModule` and define policies in `ConfigureServices`:

```csharp
[DependsOn(typeof(AbpOperationRateLimitingModule))]
public class MyModule : AbpModule
{
    {
        Configure<AbpOperationRateLimitingOptions>(options =>
        {
            options.AddPolicy("SendSmsCode", policy =>
            {
                policy.WithFixedWindow(TimeSpan.FromMinutes(1), maxCount: 3)
                      .PartitionByParameter();
            });
        });
    }
}
```

Then inject `IOperationRateLimitingChecker` and call `CheckAsync` in your application service. If the limit is exceeded it throws `AbpOperationRateLimitingException` (mapped to HTTP 429):

```csharp
public class SmsAppService : ApplicationService
{
    private readonly IOperationRateLimitingChecker _checker;

    public SmsAppService(IOperationRateLimitingChecker checker)
    {
        _checker = checker;
    }

    public async Task SendCodeAsync(string phoneNumber)
    {
        await _checker.CheckAsync("SendSmsCode", phoneNumber);

        // proceed with sending SMS...
    }
}
```

Extension methods are provided for all four checker methods (`CheckAsync`, `IsAllowedAsync`, `GetStatusAsync`, `ResetAsync`) so you can pass a `parameter` string directly without constructing `OperationRateLimitingContext` manually.

**Partition types**

| Method | Partition key source |
|---|---|
| `PartitionByParameter()` | `context.Parameter` (caller-provided) |
| `PartitionByCurrentUser()` | `ICurrentUser.Id` (auto-resolved) |
| `PartitionByCurrentTenant()` | `ICurrentTenant.Id` (auto-resolved) |
| `PartitionByClientIp()` | `IWebClientInfoProvider.ClientIpAddress` (auto-resolved) |
| `PartitionByEmail()` | `context.Parameter` → `ICurrentUser.Email` |
| `PartitionByPhoneNumber()` | `context.Parameter` → `ICurrentUser.PhoneNumber` |
| `PartitionBy(resolver)` | custom `Func<OperationRateLimitingContext, Task<string>>` |

`PartitionByCurrentUser`, `PartitionByCurrentTenant`, and `PartitionByClientIp` resolve exclusively from their respective services and do not fall back to `context.Parameter`. This avoids partition key conflicts in composite policies where `Parameter` is shared across all rules. If you need to check rate limits for a specific user, tenant, or IP explicitly, use `PartitionByParameter()` instead:

```csharp
// Admin checking another user's rate limit
await checker.CheckAsync("UserApiLimit",
    new OperationRateLimitingContext { Parameter = targetUserId.ToString() });
```

**Composite rules**

A policy can have multiple rules (`AddRule()`). All rules are checked in a two-phase flow: Phase 1 reads current counts without incrementing; Phase 2 increments only after all rules pass. This prevents wasted quota on requests that would be blocked by a later rule.

```csharp
options.AddPolicy("LoginAttempt", policy =>
{
    // per-IP burst limit (short window)
    policy.AddRule(rule => rule
        .WithFixedWindow(TimeSpan.FromMinutes(1), maxCount: 10)
        .PartitionByClientIp());

    // per-user daily limit (long window)
    policy.AddRule(rule => rule
        .WithFixedWindow(TimeSpan.FromHours(24), maxCount: 50)
        .PartitionByCurrentUser());
});
```

**Multi-tenancy**

All partition types default to a global counter (shared across tenants). Call `.WithMultiTenancy()` on a rule to opt into per-tenant counter isolation:

```csharp
// global — all tenants share the same counter
policy.WithFixedWindow(TimeSpan.FromHours(1), maxCount: 100)
      .PartitionByParameter();

// isolated — each tenant has its own counter for the same parameter
policy.WithFixedWindow(TimeSpan.FromHours(1), maxCount: 100)
      .WithMultiTenancy()
      .PartitionByParameter();
```

**Client IP resolution**

`PartitionByClientIp()` resolves the client IP address via the existing `IWebClientInfoProvider` abstraction from `Volo.Abp.AspNetCore.Abstractions`. In ASP.NET Core environments, `HttpContextWebClientInfoProvider` reads from `HttpContext.Connection.RemoteIpAddress`. In non-web scenarios, the default `NullWebClientInfoProvider` returns `null`, and using `PartitionByClientIp()` will throw an `AbpException`. Use `PartitionByParameter()` to pass the IP explicitly in non-web contexts.

**Error response**

On limit exceeded, throws `AbpOperationRateLimitingException` which is serialized to a standard ABP error response with `data` containing `RetryAfter`, `RetryAfterSeconds`, `WindowDescription`, `CurrentCount`, `MaxCount`, and `RemainingCount`.


docs/en/docs-nav.json
```diff
@@ -721,6 +721,10 @@
               "text": "Distributed Locking",
               "path": "framework/infrastructure/distributed-locking.md"
             },
+            {
+              "text": "Operation Rate Limiting",
```
hikalkan: Add to menu by name order

docs/en/framework/infrastructure/operation-rate-limiting.md
```diff
@@ -0,0 +1,492 @@
...
+    private readonly IOperationRateLimitingChecker _rateLimitChecker;
+
+    public SmsAppService(IOperationRateLimitingChecker rateLimitChecker)
+    {
+        _rateLimitChecker = rateLimitChecker;
+    }
+
+    public async Task SendCodeAsync(string phoneNumber)
+    {
+        await _rateLimitChecker.CheckAsync("SendSmsCode",
```
hikalkan: An extension method to simplify the parameter usage would be good.

hikalkan (changes requested):

hikalkan (approved):

Issue #24918, opened 2026-02-22
Add Non-Type-Safe Event Bus APIs and Runtime Handler / Background Worker Registration


salihozkara:
## Description

We need additional flexibility in the EventBus and BackgroundWorker systems for dynamic runtime scenarios.

The current implementation is fully type-safe, which is correct and should remain the default approach. However, we also need a non-type-safe layer for cases where event/worker types are not known at compile time.

Below are the required capabilities.

---

## 1️⃣ String-Based Event Publishing with Dynamic Payload

Currently:

```csharp
await eventBus.PublishAsync(new MyEvent(...));
```

We need an additional API that allows publishing events using:

* A string event name
* A dynamic payload such as:

  * `ExpandoObject`
  * `IDictionary<string, object>`

Example:

```csharp
await eventBus.PublishAsync(
    "MyDynamicEvent",
    new Dictionary<string, object>
    {
        ["UserId"] = 42,
        ["Name"] = "John"
    }
);
```

This would allow triggering events without requiring a compile-time event type.

---

## 2️⃣ Runtime Handler Registration (Non-Type-Safe)

Currently, handlers are registered via DI and are strongly typed.

We need the ability to:

* Add handlers at runtime
* Remove handlers at runtime
* Subscribe using a string event name
* Handle dynamic payloads

Conceptual example:

```csharp
eventBus.Subscribe(
    "MyDynamicEvent",
    async payload =>
    {
        // dynamic handling
    }
);
```

This should coexist with the existing type-safe mechanism.

---

## 3️⃣ Runtime Background Worker Registration (Non-Type-Safe)

Similarly for background workers:

* Add workers at runtime
* Remove workers at runtime
* Register workers without requiring a strongly typed class

Conceptual example:

```csharp
backgroundWorkerManager.Add(
    "MyDynamicWorker",
    async context =>
    {
        // dynamic logic
    }
);
```

---

## 4️⃣ Runtime Background Job Registration and Enqueue (Non-Type-Safe)

Similarly for background jobs, we need a dynamic API that allows:

* Enqueuing jobs using a string name
* Passing dynamic payloads (`ExpandoObject` / `IDictionary<string, object>`)
* Registering job handlers at runtime
* Removing job handlers at runtime

### Enqueue Example

```csharp
await backgroundJobManager.EnqueueAsync(
    "MyDynamicJob",
    new Dictionary<string, object>
    {
        ["OrderId"] = 123,
        ["Priority"] = "High"
    }
);
```

[...]

PR #25066, opened 2026-03-11
Add Non-Type-Safe Background Worker Registration

salihozkara:
Related https://github.com/abpframework/abp/issues/24918


maliming:
Move to https://github.com/abpframework/abp/pull/25059

PR #25091, opened 2026-03-16
Upgrade TickerQ packages to 10.2.0 and add  helper to 

maliming:
Upgrades all TickerQ-related NuGet packages from `10.1.1` to `10.2.0`.

TickerQ 10.2.0 adds per-function concurrency control via a new `maxConcurrency` parameter, changing the function registration tuple from 3 to 4 elements. This PR follows that change and also adds:

- `AddFunction` helper method on `AbpTickerQFunctionProvider` with input validation and duplicate-registration guard, as a cleaner alternative to `Functions.TryAdd`.
- `MaxConcurrency` property on `AbpBackgroundJobsTimeTickerConfiguration` and `AbpBackgroundWorkersCronTickerConfiguration`.
- Updated all internal call sites and documentation samples.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -183,10 +183,10 @@
     <PackageVersion Include="System.Threading.Tasks.Extensions" Version="4.6.3" />
     <PackageVersion Include="TencentCloudSDK.Sms" Version="3.0.1273" />
     <PackageVersion Include="TimeZoneConverter" Version="7.2.0" />
-    <PackageVersion Include="TickerQ" Version="10.1.1" />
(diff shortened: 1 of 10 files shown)


maliming (commented):
All three review comments have been addressed:
1. Added `AddFunction` helper method to `AbpTickerQFunctionProvider` to avoid breaking public API changes — callers no longer need to interact with the internal tuple structure directly.
2. Fixed `{ get;}` formatting to `{ get; }`.
3. Added named tuple fields: `CronExpression`, `Priority`, `Function`, `MaxConcurrency`.

EngincanV (approved):

PR #25101, opened 2026-03-17
Add community article: Shared User Accounts in ABP Multi-Tenancy

### LINKED ISSUE #25034: Article: Shared Tenant User Accounts

maliming:
https://github.com/abpframework/abp/blob/dev/docs/en/Community-Articles/2026-03-17-Shared-User-Accounts-in-ABP/POST.md

Resolves #25034


roc916:
Should we hide the tenant selection button in shared user mode? for example: Account Login Page.

EngincanV (approved):

maliming:
> Should we hide the tenant selection button in shared user mode? for example: Account Login Page.

It will be hidden.

PR #25105, opened 2026-03-18
Fix FluentValidation not working with `ConventionalControllers`

### LINKED ISSUE #23457: FluentValidation Triggered But No Validation Applied For DTOs in Custom Module
I created a custom module named JobAnnouncement using the ABP Framework and added FluentValidation for my JobCreateDto class. Although the validator is defined correctly and the module depends on AbpFluentValidationModule, the validation is being triggered, but the validation rules are not being applied when I call the application service in the host project.

### ✅ Validator Code

```
using FluentValidation;
using JobAnnouncement.App.Jobs.Dtos;

namespace JobAnnouncement.App.Jobs.Validations
{
    public class JobCreateDtoValidator : AbstractValidator<JobCreateDto>
    {
        public JobCreateDtoValidator()
        {
            RuleFor(x => x.Title).NotNull();
            RuleFor(x => x.Description).NotEmpty();
            RuleFor(x => x.Requirements).NotEmpty();
            RuleFor(x => x.IsActive).NotEmpty();
        }
    }
}

✅ Module Configuration

public async override Task<JobDto> CreateAsync(JobCreateDto input)
{
// some logic

    return await base.CreateAsync(input);
}

✅ Module Configuration

[DependsOn(
    typeof(JobAnnouncementDomainModule),
    typeof(JobAnnouncementApplicationContractsModule),
    typeof(AbpDddApplicationModule),
    typeof(AbpAutoMapperModule),
    typeof(AbpFluentValidationModule)
)]
public class JobAnnouncementApplicationModule : AbpModule
{
    {
        context.Services.AddAutoMapperObjectMapper<JobAnnouncementApplicationModule>();

        {
            options.AddMaps<JobAnnouncementApplicationModule>(validate: true);
        });
    }
}
```

 What I’ve Tried
The validator class is public and inherits from AbstractValidator<JobCreateDto>.

I added AbpFluentValidationModule in the [DependsOn] attribute in my module.

The module is properly referenced and loaded in the host application.

❌ Issue
Validation does not run — even when required fields are missing, no validation error or exception is triggered. The CreateAsync method accepts invalid data without any validation feedback.

maliming:
When Application Services are registered via `ConventionalControllers.Create()`, `ValidationInterceptor` is disabled because the types are added to `DynamicProxyIgnoreTypes`. The `AbpValidationActionFilter` only validated `ModelState`, so FluentValidation rules were never executed.

Added `IValidationEnabled` check in `AbpValidationActionFilter` to call `IMethodInvocationValidator` for conventional controllers, reusing existing validation infrastructure.

Resolve #23457

### DIFF
--- a/docs/en/framework/api-development/auto-controllers.md
+++ b/docs/en/framework/api-development/auto-controllers.md
@@ -70,7 +70,7 @@ Route is calculated based on some conventions:
 * Continues with a **route path**. Default value is '**/app**' and can be configured as like below:
 
 ````csharp
-Configure<AbpAspNetCoreMvcOptions>(options =>
+PreConfigure<AbpAspNetCoreMvcOptions>(options =>
 {
     options.ConventionalControllers
         .Create(typeof(BookStoreApplicationModule).Assembly, opts =>
@@ -149,7 +149,7 @@ public class PersonAppService : ApplicationService
 You can further filter classes to become an API controller by providing the `TypePredicate` option:
 
 ````csharp
-services.Configure<AbpAspNetCoreMvcOptions>(options =>
+PreConfigure<AbpAspNetCoreMvcOptions>(options =>
 {
     options.ConventionalControllers
         .Create(typeof(BookStoreApplicationModule).Assembly, opts =>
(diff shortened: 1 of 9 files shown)


docs/en/framework/api-development/auto-controllers.md
Tata-RicardoLans, line 1: What about line 119? Does that also need to change from `Configure<>` to `PreConfigure<>`

maliming: Line 119 uses `Configure<AbpConventionalControllerOptions>`, which is a different type from `AbpAspNetCoreMvcOptions`. It is a standard runtime options class resolved lazily through `IOptions<>` in DI, so `Configure<>` works correctly here.

The reason `PreConfigure<AbpAspNetCoreMvcOptions>` is required is that ABP processes the `ConventionalControllers` list during `PreConfigureServices` — before MVC finalizes its options. `AbpConventionalControllerOptions` has no such timing constraint, so line 119 does not need to change.

EngincanV (approved):

PR #25110, opened 2026-03-18
fix: refresh Angular permissions after resource updates

enisn:
## Summary
- refresh Angular app configuration after resource permission changes for the current user or one of the current user's roles
- let `abpPermission`-based UI react immediately after save/delete instead of waiting for a full reload
- support the permission visibility part of volosoft/volo#21899 and other resource-permission screens using the same component

## Testing
1. Open a page that uses resource permission management, such as AI Management workspaces.
2. Change a resource permission for your current user or one of your current roles.
3. Save or delete the permission and confirm the related Angular buttons update without a full page reload.


enisn:
Not needed in the current phase. Refreshing the page is OK for now

PR #25113, opened 2026-03-18
fix: rabbitmq jobQueue expiration parameter format

scrapstation:
## Problem
TimeSpan.TotalMilliseconds is a double (e.g., 1.23), while the Expiration parameter of RabbitMQ requires a non-negative integer.

This may cause the message to not be published as expected.

https://www.rabbitmq.com/docs/3.13/ttl#per-message-ttl-in-publishers

## Fix
Convert TimeSpan.TotalMilliseconds to a long value.


scrapstation:
@maliming pls review

maliming (approved):

maliming:
Thank you for your contribution 👍

PR #25174, opened 2026-03-30
feat: Implement route-based culture support in localization.

maliming:
Add URL-based localization support that embeds the culture code directly in the URL path (e.g. `/tr/identity/users`). Enabled with a single configuration:

```csharp
Configure<AbpRequestLocalizationOptions>(options =>
{
    options.UseRouteBasedCulture = true;
});
```

When enabled, ABP automatically handles route registration, URL generation, menu links, and language switching for MVC, Razor Pages, Blazor Server, and Blazor WebApp (WASM).

For Blazor pages, `@page "/{culture}/..."` route variants need to be added manually. All ABP built-in module pages already include these routes.

Related PRs:
- volosoft/volo#22040
- volosoft/lepton#3446
- https://github.com/abpframework/abp-samples/pull/312

Documentation:
- https://github.com/abpframework/abp/blob/dev/docs/en/framework/fundamentals/url-based-localization.md
- https://github.com/abpframework/abp/blob/dev/docs/en/Community-Articles/2026-03-29-Url-Based-Localization/POST.md

---

### Notes for the Angular team

A new boolean field `useRouteBasedCulture` has been added to the `/api/abp/application-configuration` response, under the `localization` object:

```json
{
  "localization": {
    "useRouteBasedCulture": true,
    ...
  }
}
```

This flag indicates that the server has `UseRouteBasedCulture = true` configured. When this flag is `true`, the Angular app should:

1. **Prefix all routes with `/:culture`** — Add a `/:culture` parent route segment so that URLs like `/tr/identity/users` are routed correctly.

2. **Update the language switcher** — Instead of only calling `sessionState.setLanguage(cultureName)`, also navigate to the same page under the new culture prefix (e.g. `/tr/identity/users` → `/en/identity/users`).

3. **Update internal navigation calls** — Any `router.navigate()` call that produces an absolute path (e.g. `router.navigate(['/identity/security-logs'])`) should prepend the current culture segment when `useRouteBasedCulture` is `true`.

No API endpoint changes are needed — routes like `/api/...` are unaffected.


EngincanV (approved):

sumeyyeKurtulus:
Hello @fahrigedik I have just updated the Angular part for this implementation. I have also created a PR for the lepton-x theme https://github.com/volosoft/lepton/pull/3451. That would be the best if you could review the code 🙏🏻

gizemmutukurt:
https://github.com/volosoft/volo/pull/22040#issuecomment-4190434066

Issue #8299, opened 2021-04-01
Audit log entity changes are not saved


realLiangshiwei:
From https://support.abp.io/QA/Questions/1109#answer-4648841d-3fa1-aedb-0547-39fb9a269a2f

The audit log object has been released when the AbpUnitOfWorkMiddleware call the `CompleteAsync` method

https://github.com/abpframework/abp/blob/b183a24cfe56ad8263aaefcf4939b9cd0c18b1c7/framework/src/Volo.Abp.AspNetCore/Volo/Abp/AspNetCore/Auditing/AbpAuditingMiddleware.cs#L42

https://github.com/abpframework/abp/blob/b183a24cfe56ad8263aaefcf4939b9cd0c18b1c7/framework/src/Volo.Abp.AspNetCore/Volo/Abp/AspNetCore/Uow/AbpUnitOfWorkMiddleware.cs#L34

https://github.com/abpframework/abp/blob/b183a24cfe56ad8263aaefcf4939b9cd0c18b1c7/framework/src/Volo.Abp.EntityFrameworkCore/Volo/Abp/EntityFrameworkCore/AbpDbContext.cs#L156

PR #25187, opened 2026-03-31
Require delete permission for deleting posts

ahmetcelik05:
Remove the special-case that allowed a post's creator to delete it without the Posts.Delete permission. The Razor view (Detail.cshtml) now only shows the delete link when the user is granted BloggingPermissions.Posts.Delete, enforcing permission-based deletion for consistency and tighter access control.

### Description

Resolves [vs-internal-issue-#8299](https://github.com/volosoft/vs-internal/issues/8299)

### Checklist

- [ ] no need to document


MansurBesleney (approved):

yagmurcelk (changes requested):
problem persists

https://github.com/user-attachments/assets/b0037e4e-c1b3-4fbc-b096-3d05276908b4

PR #25213, opened 2026-04-07
Fix account layout for  shared user multi-tenancy

roc916:
### Description

TODO: Describe what this PR has changed, add screenshot or animated GIF if available, write if it is a breaking change, and how to fix the breaking changes for existing applications if so.

### Checklist

### How to test it?

### DIFF
--- a/modules/basic-theme/src/Volo.Abp.AspNetCore.Mvc.UI.Theme.Basic/Themes/Basic/Layouts/Account.cshtml
+++ b/modules/basic-theme/src/Volo.Abp.AspNetCore.Mvc.UI.Theme.Basic/Themes/Basic/Layouts/Account.cshtml
@@ -68,6 +68,7 @@
                     <h2 class="brand-text">@BrandingProvider.AppName</h2>
                 </div>
                 @if (MultiTenancyOptions.Value.IsEnabled &&
+                  MultiTenancyOptions.Value.UserSharingStrategy == TenantUserSharingStrategy.Isolated &&
                   (TenantResolveResultAccessor.Result?.AppliedResolvers?.Contains(CookieTenantResolveContributor.ContributorName) == true ||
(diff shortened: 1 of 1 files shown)


maliming:
Instead of checking `UserSharingStrategy` in the cshtml, consider adding a `TenantResolveContributor` in the open-source Account module (`Volo.Abp.Account.Web`).

For example, register a contributor that runs **before** the Cookie/QueryString resolvers. When `UserSharingStrategy == Shared` and the user is not authenticated, set `context.Handled = true` to stop the tenant resolve pipeline. This way, Cookie/QueryString resolvers won't execute, and the tenant switch UI naturally won't appear — no need to patch every theme layout.

```csharp
public class AbpAccountTenantResolveContributor : HttpTenantResolveContributorBase
{
    public const string ContributorName = "AbpAccount";
    public override string Name => ContributorName;

    protected override Task<string?> GetTenantIdOrNameFromHttpContextOrNullAsync(
        ITenantResolveContext context, HttpContext httpContext)
    {
        var options = httpContext.RequestServices
            .GetRequiredService<IOptions<AbpMultiTenancyOptions>>().Value;

        if (options.UserSharingStrategy == TenantUserSharingStrategy.Shared)
        {
            var currentUser = httpContext.RequestServices
                .GetRequiredService<ICurrentUser>();
            if (!currentUser.IsAuthenticated)
            {
                context.Handled = true;
            }
        }

        return Task.FromResult<string?>(null);
    }
}
```

Then register it in the module's `ConfigureServices`:

```csharp
PostConfigure<AbpTenantResolveOptions>(options =>
{
    options.TenantResolvers.InsertAfter(
        r => r is CurrentUserTenantResolveContributor,
        new AbpAccountTenantResolveContributor()
    );
});
```

This handles it at the infrastructure level and covers all themes automatically.

PR #25227, opened 2026-04-09
Add support for localization file splitting

antonGritsenko:
### Description

See discussion https://github.com/abpframework/abp/discussions/25217
The goal of this change is to bring support for multiple JSON files for the same culture. This can be really helpful in case if you have big module and do not want to split it to different resources.

There are few changes in the code, it just removes exception if same culture found.

It can handle duplicate key in same way as it before: last value will win. Order of the files will be handled by the `VirtualFileProvider ` (default: alphabetically). This mean that if same key defined in the different files, then value from the last file will be used (see unit tests)

### Checklist

### How to test it?

 - Create several JSON file with same culture (see unit tests)
 - Values from all files must be added to localization dictionary

### DIFF
--- a/docs/en/framework/fundamentals/localization.md
+++ b/docs/en/framework/fundamentals/localization.md
@@ -128,7 +128,7 @@ var str3 = L["Hi__1"]; // Hello World!
 
 You can have more than one localization file with the same culture: files will be merged.
 
-> Note: if you have the same key with the same culture then last value will win. This is true in case of file-per-culture and several files per culture. Files are ordered alphabetically, so if a same key defined in files `!en.json` and `z_en.json` then value from file `z_en.json` will be used.
+> Note: If the same key is defined in multiple files for the same culture, the last value wins. Files are sorted by name using ordinal (byte-order) string comparison before merging, so if `!en.json` and `zen_Book.json` both define the same key, the value from `zen_Book.json` is used.
 
 ### Default Resource
 
--- a/docs/en/framework/fundamentals/localization.md
+++ b/docs/en/framework/fundamentals/localization.md
@@ -126,6 +126,10 @@ var str2 = L["Hi__0"]; // Bye World!
 var str3 = L["Hi__1"]; // Hello World!
 ````
 
+You can have more than one localization file with the same culture: files will be merged.
+
+> Note: if you have the same key with the same culture then last value will win. This is true in case of file-per-culture and several files per culture. Files are ordered alphabetically, so if a same key defined in files `!en.json` and `z_en.json` then value from file `z_en.json` will be used.
+
 ### Default Resource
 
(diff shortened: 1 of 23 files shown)


maliming:
Hi @antonGritsenko, thanks for this contribution! We've reviewed the PR and made some adjustments on top of your implementation. Here's a summary of the changes:

**Core implementation (`VirtualFileLocalizationResourceContributorBase`)**

- Removed the `Merge(ILocalizationDictionary)` method from the `ILocalizationDictionary` interface — adding a new member to a public interface is a breaking change for any external implementations. The merge is now handled entirely inside the contributor using `Fill(...)`.
- Added explicit `OrderBy(f.Name, StringComparer.Ordinal)` sorting before iteration, so the "last file wins" behavior is deterministic regardless of the file provider's enumeration order.
- Optimized the merge loop: the previous approach re-created a new `Dictionary` on every additional file for the same culture (O(entries × files)). The new approach accumulates all entries into a single raw `Dictionary<string, LocalizedString>` per culture during the scan and wraps it once at the end.
- Removed the now-unused `_resource` field and its assignment in `Initialize` (it was only referenced in the removed duplicate-culture exception message).

**Tests (`JsonLocalizationDictionaryBuilder_Tests`)**

- Replaced the `Should_Handle_Duplicates` test — it used duplicate JSON property names, but `System.Text.Json` already collapses those before our code runs, so the test didn't actually exercise the new behavior. The new test covers a flat key being overridden by a later nested object expansion (which does go through `FlattenTexts`).
- Renamed `Should_Create_Resource_By_Name_FromSplitFiles` → `Should_Create_Resource_By_Name_From_Split_Files` to match the underscore-separated naming convention used throughout the test class.
- Changed `localizer.GetAllStrings()` to `localizer.GetAllStrings(includeParentCultures: false)` to make the intent explicit.

**Documentation**

- Added an example file structure diagram and a merge order diagram to make the feature easier to understand for developers.

maliming (approved):

antonGritsenko:
@maliming thanks for improvement! Regarding the `Should_Handle_Duplicates` - this is more "behavior testing" rather than code testing. I used to such unit tests just to be sure that next update of the code\core\framework will not change the expected behavior.

maliming:
@antonGritsenko You're right, that's a valid point. I've restored the test with a clearer name (`Should_Use_Last_Value_When_Json_Contains_Duplicate_Keys`) and added a comment explaining that it's intentionally locking the behavior of `System.Text.Json` for duplicate JSON property names — so it serves as a regression guard if STJ ever changes this behavior in a future version.

maliming:
hi

Can you sign the CLA?

https://cla-assistant.io/abpframework/abp?pullRequest=25227

Thanks.

antonGritsenko:
@maliming I did, but my name listed twice, not sure why.

maliming:
Merged! 👍

PR #25272, opened 2026-04-15
Remove extra Select2 dropdown padding in resource permission modal

enisn:
## Summary
- load a dedicated style for the reusable add resource permission modal
- remove the extra bottom padding from the Select2 dropdown so the provider picker no longer shows an empty gap

## Testing
- dotnet build modules/permission-management/src/Volo.Abp.PermissionManagement.Web/Volo.Abp.PermissionManagement.Web.csproj

Closes volosoft/volo#22073


enisn:
This is not the root casue of the problem, Closing it

PR #25311, opened 2026-04-22
fix(#25310): revent hanging when acquiring closed channel after rabbitmq restart

### LINKED ISSUE #25310: BackgroundJob (RabbitMQ) hangs on EnqueueAsync after RabbitMQ service restart

### Description

# Description
I am using the RabbitMQ-based BackgroundJob implementation. The application starts and runs normally. However, when the RabbitMQ service is restarted during runtime, subsequent calls to _backgroundJobManager.EnqueueAsync will block and never return, causing the thread to hang.

# Steps to Reproduce
1. Configure and use RabbitMQ as the BackgroundJob provider
2. Start the Web API application (works as expected)
3. Restart the RabbitMQ service while the application is running
4. Call _backgroundJobManager.EnqueueAsync(...)
5. Observe that the call blocks indefinitely

# Fix solution
Before executing poolItem.Acquire(), make sure that the Channel is not closed.

Image

### Reproduction Steps

_No response_

### Expected behavior

_No response_

### Actual behavior

_No response_

### Regression?

_No response_

### Known Workarounds

_No response_

### Version

10.1.1

### User Interface

Common (Default)

### Database Provider

EF Core (Default)

None (Default)

### Operation System

Windows (Default)

### Other information

_No response_

scrapstation:
Resolves #25310

### DIFF
--- a/framework/Volo.Abp.abpmdl
+++ b/framework/Volo.Abp.abpmdl
@@ -478,6 +478,10 @@
       "path": "test/Volo.Abp.ObjectMapping.Tests/Volo.Abp.ObjectMapping.Tests.abppkg",
       "folder": "test"
     },
+    "Volo.Abp.RabbitMQ.Tests": {
+      "path": "test/Volo.Abp.RabbitMQ.Tests/Volo.Abp.RabbitMQ.Tests.abppkg",
+      "folder": "test"
+    },
     "Volo.Abp.Ddd.Application.Contracts": {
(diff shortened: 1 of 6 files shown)


maliming (approved):

maliming:
Hi @scrapstation, thanks for the great catch and the fix!

I pushed a few follow-ups on your branch — tightened the recovery path a bit so concurrent callers can't race into duplicate dispose/recreate, and added a small `Volo.Abp.RabbitMQ.Tests` project with a regression test for the hang. Hope you don't mind me piling on.

Really appreciate you digging into this one!

PR #25316, opened 2026-04-24
feat(identity): add single-use Email/Phone 2FA token providers

maliming:
Replaces TOTP-based Email/Phone 2FA providers under `TokenOptions.DefaultEmailProvider` / `DefaultPhoneProvider` with DataProtector-backed single-use equivalents.

- Encrypt the 6-digit code via `IDataProtector`
- Store ciphertext + absolute UTC expiration in the user token table
- Consume the stored entry on successful validation (single-use)
- ConcurrencyStamp race returns `false` instead of 500
- Configurable `TokenLifespan` via Options

`AbpSingleActiveTokenProvider.GenerateAsync` now checks the `IdentityResult` from `UserManager.UpdateAsync` so a silent persistence failure no longer returns a token that was not saved.

Because `IdentityOptions.Tokens.ChangePhoneNumberTokenProvider` defaults to `TokenOptions.DefaultPhoneProvider`, `UserManager.GenerateChangePhoneNumberTokenAsync` / `VerifyChangePhoneNumberTokenAsync` also inherit the new single-use, stored-token semantics. Phone-change tokens issued before upgrade stop working.

Related to #25314.


EngincanV (approved):

iiiixxl:
that's good

PR #25397, opened 2026-05-08
Expose dynamic background worker scheduler capabilities

salihozkara:
Adds `ISupportsCronScheduling` and `ISupportsRuntimeRegistration` marker interfaces so consumers can detect which scheduling features the active `IDynamicBackgroundWorkerManager` supports.

- Hangfire and Quartz managers implement both markers.
- Default in-memory manager implements `ISupportsRuntimeRegistration` only — `CronExpression` is now rejected on both `AddAsync` and `UpdateScheduleAsync` with a clearer error message.
- TickerQ manager implements neither, since its dynamic API is not supported at runtime.


maliming:
I think using an interface to indicate the capabilities of background workers is more in line with the current framework's design style(similar to `IMultiTenant`, `ISoftDelete`, etc.), What do you think?

```cs
public interface ISupportsCronScheduling
{
}

public interface ISupportsRuntimeRegistration
{
}

public class DefaultDynamicBackgroundWorkerManager :
    IDynamicBackgroundWorkerManager,
    ISupportsRuntimeRegistration,
    ISingletonDependency
{
    // ...
}

public class HangfireDynamicBackgroundWorkerManager :
    IDynamicBackgroundWorkerManager,
    ISupportsRuntimeRegistration,
    ISupportsCronScheduling,
    ISingletonDependency
{
    // ...
}

public class QuartzDynamicBackgroundWorkerManager :
    IDynamicBackgroundWorkerManager,
    ISupportsRuntimeRegistration,
    ISupportsCronScheduling,
    ISingletonDependency
{
    // ...
}

public class TickerQDynamicBackgroundWorkerManager :
    IDynamicBackgroundWorkerManager,
    ISingletonDependency
{
    // implements neither marker
}
```

maliming (approved):

PR #25466, opened 2026-05-23
Fixes the use of DataGrid Template Contexts for ExtraProperties

### LINKED ISSUE #25465: AbpExtensibleDataGrid is not displaying extra properties after update to Blazorise 2.0

### Description

After updating to ABP 10.3 the entities that had `ExtraProperties` stopped displaying the value of the extra properties in the data grid. While the grid shows the column headers for the extra properties, the values themselves are not being displayed.

### Reproduction Steps

_No response_

### Expected behavior

_No response_

### Actual behavior

_No response_

### Regression?

_No response_

### Known Workarounds

_No response_

### Version

10.3

### User Interface

Blazor Server

### Database Provider

EF Core (Default)

Separate Auth Server

### Operation System

macOS

### Other information

_No response_

carlosv14:
### Description

Resolves #25465

TODO: As per [Blazorise migration docs](https://blazorise.com/news/migration/200) DataGrid Template Contexts now receive context objects instead of the raw row item and any code that assumes context is TItem should be updated. So the line `var entity = context as IHasExtraProperties;` should be `var entity = context.Item as IHasExtraProperties;` otherwise entity is null and propertyValue is also null.
 

### Checklist

### DIFF
--- a/framework/src/Volo.Abp.BlazoriseUI/Components/AbpExtensibleDataGrid.razor
+++ b/framework/src/Volo.Abp.BlazoriseUI/Components/AbpExtensibleDataGrid.razor
@@ -122,7 +122,7 @@
                             <DataGridColumn TItem="TItem" Field="@nameof(IHasExtraProperties.ExtraProperties)" SortField="@column.PropertyName" Width="@BlazoriseFluentSizingParse.Parse(column.Width, SizingType.Width)" Caption="@column.Title" Displayable="column.Visible" Sortable="@column.Sortable">
                                 <DisplayTemplate>
                                     @{
-                                        var entity = context as IHasExtraProperties;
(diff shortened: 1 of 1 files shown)


maliming:
Hi @carlosv14, thanks for the fix!

Could you change the base branch of this PR from `dev` to `rel-10.3`? Since the issue was reported on 10.3, landing the fix on the oldest supported release branch is preferred — ABP's auto-merge bot will then forward it to `rel-10.4` and `dev` automatically.

You can change the base branch via the "Edit" button next to the PR title on GitHub.

Thanks!

carlosv14:
Hi @maliming changed the base branch as requested. Thanks!

maliming (approved):

PR #25615, opened 2026-06-10
Blob container ProviderType does not fall back to default container configuration

EngincanV:
## Problem

When configuring a named blob container with only non-provider options, such as `IsMultiTenant`, the container loses access to the default provider configuration.

Example:

```csharp
{
    options.Containers.ConfigureDefault(container =>
    {
        container.UseDatabase();
    });

    options.Containers.Configure<WorkspaceDataSourceBlobContainer>(container =>
    {
        container.IsMultiTenant = false;
    });
});
```

In this case, `WorkspaceDataSourceBlobContainer` is created as a specialized named container with the default container as its fallback. However, `ProviderType` is a direct auto-property on `BlobContainerConfiguration`, so it does not use the fallback configuration.

As a result, provider resolution fails with:

```text
No BLOB Storage provider was used! At least one provider must be configured to be able to use the BLOB Storing System.
```

> See: https://github.com/volosoft/volo/issues/22446

## Expected Behavior

A named blob container should inherit `ProviderType` from the default container when it does not explicitly set its own provider type. (if it's null)

This should behave consistently with other configuration values that already fall back through `_fallbackConfiguration`.

## Actual Behavior

Custom configuration values use fallback resolution, but `ProviderType` does not.

This means a container configured only for options like `IsMultiTenant` must also repeat the provider configuration, even when a default provider is already configured.

## Suggested Fix

Update `BlobContainerConfiguration.ProviderType` to use the fallback configuration when the current container has no explicit provider type.

Suggested implementation:

```csharp
private Type? _providerType;

public Type? ProviderType
{
    get => _providerType ?? _fallbackConfiguration?.ProviderType;
    set => _providerType = value;
}
```

### DIFF
--- a/framework/src/Volo.Abp.BlobStoring/Volo/Abp/BlobStoring/BlobContainerConfiguration.cs
+++ b/framework/src/Volo.Abp.BlobStoring/Volo/Abp/BlobStoring/BlobContainerConfiguration.cs
@@ -43,6 +43,20 @@ public class BlobContainerConfiguration
         _properties = new Dictionary<string, object?>();
     }
 
+    /// <summary>
+    /// Returns the naming normalizers in effect for this container, inheriting from the fallback
+    /// configuration only when this container has none and does not override <see cref="ProviderType"/>.
+    /// </summary>
+    public IEnumerable<Type> GetEffectiveNamingNormalizers()
+    {
+        if (NamingNormalizers.Count == 0 && _providerType == null && _fallbackConfiguration != null)
+        {
+            return _fallbackConfiguration.GetEffectiveNamingNormalizers();
+        }
+
+        return NamingNormalizers;
+    }
+
     public T? GetConfigurationOrDefault<T>(string name, T? defaultValue = default)
     {
         return (T?)GetConfigurationOrNull(name, defaultValue);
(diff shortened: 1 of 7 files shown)


maliming:
Hi @EngincanV, pushed one more edge fix on top of your ProviderType change.

`NamingNormalizers` doesn't fall back. If the default has `UseAzure()` and a named container inherits the `ProviderType` (your fix) but has no normalizers of its own, the named container silently drops `AzureBlobNamingNormalizer` and feeds raw names to Azure. The added `GetEffectiveNamingNormalizers()` resolves the list through the fallback, but only when the container also inherits its `ProviderType` — so a container that switches provider (`UseDatabase()` / `UseAws()`) doesn't pick up the default's normalizers and leak Azure rules onto Database/Memory names.

maliming (approved):

EngincanV:
For anyone who currently got the relevant error on AI Management module integration, they can configure as follows on their domain module class:

```cs
{
    options.Containers.Configure<WorkspaceDataSourceBlobContainer>(container =>
    {
        container.IsMultiTenant = false;
        container.UseDatabase();
    });
});
```

PR #25669, opened 2026-06-22
Fix Razor Pages antiforgery validation 

ahmetcelik05:
# Fix Razor Pages antiforgery validation when `NormalizeUserIdClaimIssuer` is enabled

## Problem

Commit `45913c8ff3` normalized antiforgery token generation (`AspNetCoreAbpAntiForgeryManager`, `ApplicationConfigurationScript`) and MVC validation (`AbpValidateAntiforgeryTokenAuthorizationFilter`), but **Razor Pages** still use the built-in `AutoValidateAntiforgeryTokenAuthorizationFilter`, which does not normalize claim issuers.

With OIDC/cookie auth, the `XSRF-TOKEN` cookie is created with a normalized user identity, while Razor Page POSTs are validated against the raw principal. This causes:

```
The provided antiforgery token was meant for a different claims-based user than the current user.
```

Setting Management is a common hit because `Index.js` POSTs to `?handler=RenderView` immediately on page load.

## Root Cause

`AutoValidateAntiforgeryPageApplicationModelProvider` adds `AutoValidateAntiforgeryTokenAttribute` to every Razor Page. When both the built-in and ABP antiforgery filters are present, the built-in filter wins and ABP's filter is skipped.

## Solution

Via `PostConfigure<RazorPagesOptions>`, replace `AutoValidateAntiforgeryTokenAttribute` with `AbpAutoValidateAntiforgeryTokenAttribute` on all Razor Pages.

**Files:**

- `AbpRazorPagesAntiforgeryConfiguration.cs` (new)
- `AbpAspNetCoreMvcModule.cs`

**Note:** Call `AddFolderApplicationModelConvention` directly in `PostConfigure` — wrapping it in `IPageConvention.Apply()` does not work because `Apply()` is never invoked for conventions added that way.


maliming (approved):

maliming:
Thanks for the report. I've taken a different approach here — wrapping the DI-registered `IAntiforgery` with a decorator instead of replacing page filters.

The page-filter swap has one issue: unconditionally adding `AbpAutoValidateAntiforgeryTokenAttribute` to every page overrides `[IgnoreAntiforgeryToken]` since both implement `IAntiforgeryPolicy` and the last one wins.

The decorator normalizes `HttpContext.User` on all `IAntiforgery` entry points, so it covers built-in Razor Pages filters, ABP filters, and direct calls alike — without touching ignore policies.

Tested on classic tiered (OIDC cookie, authority issuer): pure #25655 without the decorator → RenderView 400; with decorator → 200. Also verified the original cross-scheme Angular scenario.

Could you test the latest code on your end and confirm it resolves the issue you found?

ahmetcelik05:
It fixed the issue, thank you.

PR #25703, opened 2026-06-27
Stop logging misleading warning when normalizing Unspecified DateTime near range boundaries

gdlcf88:
## Problem

When the clock is configured as UTC (`AbpClockOptions.Kind == DateTimeKind.Utc`, i.e. `SupportsMultipleTimezone == true`) and a current user timezone is set, `AbpDateTimeConverterBase.Normalize` treats an `Unspecified`-kind `DateTime` as local time in the user's timezone and converts it to UTC:

```csharp
dateTime = new DateTimeOffset(dateTime, timezoneInfo.GetUtcOffset(dateTime)).UtcDateTime;
```

For a value within the timezone offset distance of `DateTime.MinValue` / `DateTime.MaxValue` — most commonly the `DateTime.MinValue` placeholder returned by various components — this throws `ArgumentOutOfRangeException` (the UTC equivalent falls outside the supported `DateTime` range).

The exception was swallowed and logged as a **warning on every serialization**:

```
Could not convert DateTime with unspecified Kind using timezone 'Asia/Shanghai'.
```

For any positive-offset user (e.g. `Asia/Shanghai` = +08:00, `Europe/Brussels` = +01:00/+02:00) serializing a payload that carries a `DateTime.MinValue` placeholder, this floods the logs even though the serialized output is correct and unaffected. The symmetric case happens at `DateTime.MaxValue` for negative-offset users.

A real-world producer of these placeholders is `Volo.Docs` `GitHubDocumentSource`, which returns a bare `DateTime.MinValue` (Unspecified kind) when commit info is missing.

## Fix

- **`AbpDateTimeConverterBase.Normalize`**: catch the boundary `ArgumentOutOfRangeException` explicitly, keep the value unchanged, and log it at **Debug** level instead of **Warning**. It stays traceable when troubleshooting without polluting production logs. Genuinely unexpected failures still log a warning as before. This is intentionally not limited to the exact `MinValue`/`MaxValue` sentinels — any value close enough to the range boundary overflows the same way.
- **`Volo.Docs` `GitHubDocumentSource`**: return a `Utc`-kind placeholder (`DateTime.SpecifyKind(DateTime.MinValue, DateTimeKind.Utc)`) and use `Author.Date.UtcDateTime`, so it no longer emits `Unspecified` placeholders in the first place.

## Tests

Added `AbpDateTimeConverterTimezone_Tests` (`Volo.Abp.Json.Tests`) covering, under a UTC clock:

- `MinValue` under positive-offset timezones (`Asia/Shanghai`, `Europe/Brussels`) → no warning, output unchanged
- a value near `MinValue` (`MinValue.AddHours(3)`) under a positive offset → no warning (proves the fix is not sentinel-only)
- `MaxValue` under a negative-offset timezone (`America/New_York`) → no warning
- a real UTC timestamp under a positive offset → no warning, round-trips as `...Z`
- a genuine `Unspecified` timestamp under +08:00 → still correctly converted to UTC (`18:00 +08:00` → `10:00Z`)

All 6 tests pass.

### DIFF
--- a/framework/src/Volo.Abp.AspNetCore.Mvc/Volo/Abp/AspNetCore/Mvc/ModelBinding/AbpDateTimeModelBinder.cs
+++ b/framework/src/Volo.Abp.AspNetCore.Mvc/Volo/Abp/AspNetCore/Mvc/ModelBinding/AbpDateTimeModelBinder.cs
@@ -42,14 +42,14 @@ public class AbpDateTimeModelBinder : IModelBinder
             _clock.SupportsMultipleTimezone &&
             !_currentTimezoneProvider.TimeZone.IsNullOrWhiteSpace())
         {
+            var timeZone = _currentTimezoneProvider.TimeZone;
             try
             {
-                var timezoneInfo = _timezoneProvider.GetTimeZoneInfo(_currentTimezoneProvider.TimeZone);
-                dateTime = new DateTimeOffset(dateTime, timezoneInfo.GetUtcOffset(dateTime)).UtcDateTime;
+                dateTime = _timezoneProvider.ConvertUnspecifiedToUtc(dateTime, timeZone);
             }
             catch
             {
-                _logger.LogWarning("Could not convert DateTime with unspecified Kind using timezone '{TimeZone}'.", _currentTimezoneProvider.TimeZone);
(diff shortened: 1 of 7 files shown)


maliming (approved):

maliming:
Thanks @gdlcf88 👍

PR #25750, opened 2026-07-04
Fix resource lookup error in createLocalizer

### LINKED ISSUE #25749: Localization doesn't work properly for extended properties in Angular

### Description

While extending an existing entity from pre-built modules, (e.g. Role in IdentityManager), even when a proper `DisplayName` parameter is added with a valid LocalizationResource, the localization of the new property always defaults to the default LocalizationResource instead of the one specified in the `DisplayName` parameter.

### Reproduction Steps

1. Create a new module named `MyModule`
2. Add a new property to an existing entity from pre-built modules in and add a `DisplayName` value as specified in the example:

In `MyModule.Domain.Shared/MyModuleExtensionConfigurator.cs`:
```
public static void ConfigureExtraProperties()
{
    OneTimeRunner.Run(() =>
    {
        ObjectExtensionManager.Instance.Modules()
            .ConfigureIdentity(identity =>
            {
                identity.ConfigureUser(user =>
                {
                    user.AddOrUpdateProperty<string>( //property type: string
                        "SocialSecurityNumber", //property name
                        property =>
                        {
                            //localization
                            property.DisplayName =
                                LocalizableString.Create<MyProjectNameResource>(
                                    "UserSocialSecurityNumberDisplayName"
                                );
                        }
                    );
                });
            });
    });
}
```

2. Add a localization string in `MyProject.Demo.Shared/Localization/MyProjectNameResource/en.json`:
```
"UserSocialSecurityNumberDisplayName": "Social Security Number"
```

3. Add this `MyModule` to an application and run the application and launch Angular application.
4. Navigate to `identity/roles` path and try adding a new role or edit an existing role

### Expected behavior

The label in the form should show "Social Security Number"

### Actual behavior

The label shows "UserSocialSecurityNumberDisplayName"

### Regression?

No

### Known Workarounds

None

### Version

8.3.4

### User Interface

Angular

### Database Provider

EF Core (Default)

None (Default)

### Operation System

Windows (Default)

### Other information

Issue also present in latest version.

I have identified the root cause of the issue and am raising a PR for the fix.

chandralk:
### Description

Resolves #25749 (write the related issue number if available)

The `createLocalizer` function is trying to look up in `localization.values` object instead of `localization.resources` object while trying to match the resource name. Fixed the same.

### Checklist


chandralk:
It seems the behaviour of localization is different between the data table header and the form label. I'll check it thoroughly and re-raise the PR.

chandralk:
Even after fixing the above, I am facing other problems. Specifically with this line:
https://github.com/abpframework/abp/blob/f2dde57ea3df938519f0036835f4c1b3ebca232f/npm/ng-packs/packages/components/extensible/src/lib/components/extensible-form/extensible-form-prop.component.html#L166

Here, the `displayName` parameter would already be fully qualified i.e. it will be in the form on `Resource::Key`. But it seems that if it is an extra property, we are again prefixing it with `::`. This ultimately is resulting in the ignoring the Key and only the Resource name is being returned.

I don't know at which level this needs to be addressed.

Requesting help from maintainers.

PR #25796, opened 2026-07-10
Fix: Add dependency on DocsCommonHttpApiClientModule in DocsAdmin and Docs ClientModule

LEIRONGHUA:
Fix DependencyResolutionException for IProjectAppService in layered projects

## Problem

Using `abp install-module Volo.Docs` When installing the Volo.Docs module in a layered/split project and navigating to the `/Documents` URL, the following error occurs:

```
An exception was thrown while activating Volo.Docs.Pages.Documents.IndexModel.
System.InvalidOperationException: An exception was thrown while activating Volo.Docs.Pages.Documents.IndexModel.
 ---> Autofac.Core.DependencyResolutionException: An exception was thrown while activating Volo.Docs.Pages.Documents.IndexModel.
 ---> Autofac.Core.DependencyResolutionException: None of the constructors found on type 'Volo.Docs.Pages.Documents.IndexModel' can be invoked with the available services and parameters:
Cannot resolve parameter 'Volo.Docs.Common.Projects.IProjectAppService projectAppService' of constructor 'Void .ctor(Volo.Docs.Common.Projects.IProjectAppService, Microsoft.Extensions.Options.IOptions`1[Volo.Docs.DocsUiOptions])'.
```

## Temporary workaround

Modified the `xxxHttpApiClientModule.cs` file in the `You.Project.HttpApi.Client` project, and manually added `typeof(DocsCommonHttpApiClientModule)` inside the `DependsOn` method.

## Root Cause

`DocsHttpApiClientModule` and `DocsAdminHttpApiClientModule` were missing the dependency on `DocsCommonHttpApiClientModule`, which registers `IProjectAppService` (and related services) in the DI container. As a result, when `IndexModel` attempted to resolve `IProjectAppService` via constructor injection, Autofac could not find a registered implementation.

## Changes

- Added `Volo.Docs.Common.HttpApi.Client` project reference to `Volo.Docs.HttpApi.Client.csproj`
- Added `Volo.Docs.Common.HttpApi.Client` project reference to `Volo.Docs.Admin.HttpApi.Client.csproj`
- Added `DependsOn(typeof(DocsCommonHttpApiClientModule))` to `DocsHttpApiClientModule`
- Added `DependsOn(typeof(DocsCommonHttpApiClientModule))` to `DocsAdminHttpApiClientModule`

### DIFF
--- a/modules/docs/src/Volo.Docs.Admin.HttpApi.Client/Volo.Docs.Admin.HttpApi.Client.csproj
+++ b/modules/docs/src/Volo.Docs.Admin.HttpApi.Client/Volo.Docs.Admin.HttpApi.Client.csproj
@@ -13,6 +13,7 @@
   <ItemGroup>
     <ProjectReference Include="..\Volo.Docs.Admin.Application.Contracts\Volo.Docs.Admin.Application.Contracts.csproj" />
     <ProjectReference Include="..\..\..\..\framework\src\Volo.Abp.Http.Client\Volo.Abp.Http.Client.csproj" />
+    <ProjectReference Include="..\Volo.Docs.Common.HttpApi.Client\Volo.Docs.Common.HttpApi.Client.csproj" />
 
   </ItemGroup>
 
--- a/modules/docs/src/Volo.Docs.Admin.HttpApi.Client/Volo/Docs/Admin/DocsAdminHttpApiClientModule.cs
+++ b/modules/docs/src/Volo.Docs.Admin.HttpApi.Client/Volo/Docs/Admin/DocsAdminHttpApiClientModule.cs
@@ -6,6 +6,7 @@ using Volo.Abp.VirtualFileSystem;
 namespace Volo.Docs.Admin
 {
     [DependsOn(
+        typeof(DocsCommonHttpApiClientModule),
         typeof(DocsAdminApplicationContractsModule),
         typeof(AbpHttpClientModule))]
     public class DocsAdminHttpApiClientModule : AbpModule
--- a/modules/docs/src/Volo.Docs.HttpApi.Client/Volo.Docs.HttpApi.Client.csproj
+++ b/modules/docs/src/Volo.Docs.HttpApi.Client/Volo.Docs.HttpApi.Client.csproj
@@ -13,6 +13,7 @@
   <ItemGroup>
     <ProjectReference Include="..\Volo.Docs.Application.Contracts\Volo.Docs.Application.Contracts.csproj" />
     <ProjectReference Include="..\..\..\..\framework\src\Volo.Abp.Http.Client\Volo.Abp.Http.Client.csproj" />
+    <ProjectReference Include="..\Volo.Docs.Common.HttpApi.Client\Volo.Docs.Common.HttpApi.Client.csproj" />
   </ItemGroup>
 
 </Project>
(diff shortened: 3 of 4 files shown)


maliming (approved):
Verified the root cause. `AddStaticHttpClientProxies` registers proxies per-assembly and doesn't scan the referenced contracts assemblies, so `IProjectAppService` (declared in `Volo.Docs.Common.Application.Contracts`) never gets a client proxy unless `DocsCommonHttpApiClientModule` is in the module graph. The main Docs client only proxied its own contracts assembly, which is why a tiered app fails to resolve it. The Admin client needs it too — `GeneratePdf` injects `IProjectAppService`.

The fix mirrors the existing contracts dependency chain (`DocsApplicationContractsModule` -> `DocsCommonApplicationContractsModule`) and introduces no cycle. LGTM.

LEIRONGHUA:
@maliming Hello

Let's discuss another bug related to this situation.

## Problem

Using `abp install-module Volo.Docs` When installing the `Volo.Docs` module in a layered/split project and navigating to the `/Documents/Project/Index` URL, the following error occurs:

```
An unhandled exception has occurred while executing the request.
System.InvalidOperationException: An exception was thrown while activating Volo.Docs.Pages.Documents.Project.IndexModel.
 ---> Autofac.Core.DependencyResolutionException: An exception was thrown while activating Volo.Docs.Pages.Documents.Project.IndexModel.
 ---> Autofac.Core.DependencyResolutionException: None of the constructors found on type 'Volo.Docs.Pages.Documents.Project.IndexModel' can be invoked with the available services and parameters:
Cannot resolve parameter 'Volo.Docs.TableOfContents.ITocGeneratorService tocGeneratorService' of constructor 'Void .ctor(Volo.Docs.Documents.IDocumentAppService, Volo.Docs.HtmlConverting.IDocumentToHtmlConverterFactory, Volo.Docs.Common.Projects.IProjectAppService, Microsoft.Extensions.Options.IOptions`1[Volo.Docs.DocsUiOptions], Volo.Docs.HtmlConverting.IWebDocumentSectionRenderer, Volo.Abp.Authorization.Permissions.IPermissionChecker, Volo.Docs.Common.Documents.IDocumentPdfAppService, Volo.Docs.TableOfContents.ITocGeneratorService)'.
```

## Root Cause

The `ITocGeneratorService` interface is defined in `Volo.Docs.Application.Contracts`, with its implementation located in `Volo.Docs.Application`.
This is a utility helper class rather than an application service, and it does not correspond to any HTTP API Controller.
As a result, the layered project structure does not provide a proxy implementation for `ITocGeneratorService` within `Volo.Docs.HttpApi.Client`, which leads to dependency injection failures.

## Proposal for discussion:

Move the `ITocGeneratorService.cs` interface and its implementation `TocGeneratorService.cs` to the `Volo.Docs.Web/Markdown/TableOfContents` directory to resolve this issue.
What are your thoughts on this approach? If everyone agrees, I will submit a new PR.

maliming:
Thanks @LEIRONGHUA, you're right about this one. `ITocGeneratorService` is a presentation-side rendering helper (it replaced the old client-side TOC script) rather than a real remote application service — it just happened to be marked `IApplicationService` and was later moved into the Application layer, which is what broke the tiered/layered setup.

I opened #25799 for it: moved the `TableOfContents` types back to the Web layer and dropped the `IApplicationService` inheritance so it registers on the web tier again. Thanks for reporting it and for the write-up.

PR #25836, opened 2026-07-18
Add BLOB at-rest encryption and a content pipeline

### LINKED ISSUE #25835: Feature Request : [BlobStoring] Add a pipeline/interceptor mechanism to IBlobContainer (for encryption, compression, etc.)

There is currently no way to intercept or transform the blob stream between `IBlobContainer` and the underlying `IBlobProvider`. The call chain is direct:

```
IBlobContainer.SaveAsync / GetOrNullAsync
  → BlobContainer (tenant context + naming normalization)
  → IBlobProvider.SaveAsync / GetOrNullAsync
  → physical storage (FileSystem, Azure, AWS, Minio, ...)
```

The stream flows untouched from the caller to the provider. This makes cross-cutting concerns on the blob **content** impossible without replacing the whole provider or the container. The most prominent use case: **encrypting blobs at rest**. We need to encrypt documents on save and decrypt them on read, transparently, regardless of which storage provider is configured. Other use cases (compression, virus scanning, watermarking, content validation) have the same shape.

A pipeline mechanism in `Volo.Abp.BlobStoring`, following the existing `NamingNormalizers` pattern:

1. **A pipeline contributor abstraction**, e.g. `IBlobPipelineContributor`, with hooks that can transform the stream on save and on read (and optionally observe delete/exists).

2. **Per-container configuration with fallback**, mirroring `BlobContainerConfiguration.NamingNormalizers`:
   - A `TypeList`-style collection on `BlobContainerConfiguration` (e.g. `PipelineContributors`), with an effective-list resolution that honors the existing default → container fallback chain, exactly like `GetEffectiveNamingNormalizers()`.
   - An extension method for fluent configuration, e.g.:
     ```csharp
     {
         options.Containers.ConfigureDefault(c => c.UseEncryption());
         options.Containers.Configure<MyContainer>(c => c.UseEncryption(/* container-specific opts */));
     });
     ```

3. **Execution inside `BlobContainer`**: contributors run after naming normalization and before/after the provider call — encrypt on `SaveAsync`, decrypt on `GetOrNullAsync`. `BlobContainer` already receives an `IServiceProvider` (currently unused), so contributors can be resolved from DI within a scope, the same way `BlobNormalizeNamingService` resolves naming normalizers.

4. **Built-in encryption contributor** with tenant-aware key resolution:
   - `BlobContainer` already executes every operation inside `CurrentTenant.Change(GetTenantIdOrNull())`, so contributors see the correct tenant context for free.
   - Key resolution via an abstraction (e.g. `IBlobEncryptionKeyProvider`): when a tenant context exists, resolve a **tenant-specific key**; otherwise (or when `Configuration.IsMultiTenant == false`) fall back to a **global key**. Developers should be able to replace the key provider with their own implementation (e.g. backed by the setting system, a vault, or KMS).
   - The encryption format should carry a version/magic header so encrypted and plain blobs can be distinguished (migration/backward compatibility) and the algorithm can evolve.

Key design goals:
- Zero changes required in the existing 8 provider packages (FileSystem, Azure, AWS, Minio, Aliyun, Bunny, Google, Memory) — the pipeline lives in the core package, above the providers.
- Opt-in per container; containers without pipeline configuration behave exactly as today.
- Works with non-seekable streams (wrap/chain streams instead of buffering everything in memory where possible).

### Additional context

- Existing pieces that make this a small, natural addition:
  - `BlobContainerConfiguration` already supports per-container settings with default fallback (`SetConfiguration` / `GetConfigurationOrNull` / `_fallbackConfiguration`).
  - `BlobNormalizeNamingService` is a working template for resolving a `TypeList` from DI inside a scope and executing it in order.
  - `BlobContainer.ServiceProvider` is already injected but unused — a ready-made resolution point.
  - Tenant context (`CurrentTenant.Change(GetTenantIdOrNull())`) is already established around every provider call, enabling tenant-based encryption keys out of the box; `IsMultiTenant = false` naturally yields the global-key behavior.
- Example usage after the feature:
  ```csharp
  {
      options.Containers.ConfigureAll(c => c.UseEncryption()); // tenant key if tenant, else global key
  });
  ```
- Related: encrypted-blob storage is a common enterprise/compliance requirement (GDPR, HIPAA-style data-at-rest encryption) and currently requires significant custom plumbing in ABP applications.

ugurozturk:
Resolve #25835

### Description

This PR adds two related features to the BLOB Storing system: transparent **at-rest encryption** and a general-purpose **content pipeline**. Both transform the BLOB content stream on top of the configured storage provider — the content is transformed before it reaches the provider and reversed while it is read back — so the existing provider packages do not require any behavioral change, only the stream-contract adjustments described below.

#### Encryption

- Encrypts BLOBs with AES-256-GCM in authenticated chunks, keeping memory usage constant and independent of the BLOB size.
- Derives a unique key per BLOB with PBKDF2-SHA256, a random per-BLOB salt and a configurable iteration count (`AbpBlobStoringEncryptionOptions.KdfIterations`, allowed range 100,000 - 600,000).
- Binds every chunk to the format header, the chunk index and the storage identity (normalized container name, BLOB name and tenant id) as associated data, so chunk re-ordering, moving content between BLOBs and cross-tenant substitution are detected.
- Writes an authenticated terminal record, so cutting complete chunks off the end (or appending to it) is detected when the stream is read to its end.
- Prefixes encrypted BLOBs with the `ABPE` magic and a format version; a golden-vector test freezes the v1 format on disk.
- Validates the header parameters (iteration count, chunk size) against upper bounds before deriving keys or allocating buffers.

#### Content pipeline

- Adds `IBlobPipelineContributor`: a contributor transforms the content stream transparently (compression, watermarking, validation, …) without changing the storage provider or the application code that works with `IBlobContainer`.
- Contributors are configured per container (`PipelineContributors`). The default-container contributors run as global stages before a container's own ones; `InheritPipelineContributors = false` opts a container out of the global stages.
- While saving, the contributors run in the configured order and the built-in encryption always runs innermost (immediately before the provider); while getting, decryption runs first and the contributors run in reverse — so a contributor always works on the plain content.
- The stream returned by `GetAsync` keeps the DI scope and the tenant context of the contributors alive until it is disposed, so lazily transforming wrappers can keep using their scoped services. The stream ownership contract is documented on `IBlobPipelineContributor`.

#### Configuration and key resolution

- `UseEncryption()` enables encryption for a single container or for the default container configuration; `DisableEncryption()` opts a container out again.
- The passphrase resolves from the container configuration first, then `AbpBlobStoringEncryptionOptions.DefaultPassPhrase`. There is no built-in default value: if no passphrase can be resolved, saving and reading fail with a descriptive exception.
- The key source is recorded in the BLOB header and only that source is used while decrypting, so BLOBs written with the global passphrase stay readable after a container-specific passphrase is configured later.
- Reading existing plaintext BLOBs is an explicit opt-in (`UseEncryption(allowLegacyPlainText: true)`) intended for migrating containers that already have data; by default, content without the encrypted format fails to be read.

#### Extensibility

- `IBlobEncryptionKeyProvider` can be replaced to read passphrases from a vault or another secret store, or to supply tenant-specific keys (`BlobEncryptionKeySource.Tenant`).
- `IBlobEncryptionCodec` (implemented by `BlobEncryptionCodec`) can be replaced to change the encryption format or algorithm.
- `IBlobPipelineContributor` is the extension point for content transformations. The framework services and provider methods added by this PR are `virtual` so they can be overridden.

#### Provider compatibility

- The encrypting stream exposes the exact encrypted length when the source stream can report its length and position (including forward-only sources), for providers that require the object size before uploading.
- The file system provider now retries a failed save only while it is safe: before the target file is opened, or for a seekable overwrite. A consumed non-replayable stream is never replayed into a corrupted file.
- The AWS provider handles the non-seekable, non-replayable stream that encryption or a pipeline produces: a small source with a known length is buffered and uploaded as a retryable `PutObject`, while a larger (or unknown-length) source goes through a `TransferUtility` multipart upload that buffers and retries part by part. The input is wrapped so the SDK's multipart path does not dispose the caller-owned stream, and `DisablePayloadSigning` is propagated for S3-compatible services.
- When the platform does not support AES-GCM, saving and reading fail before any output is produced (.NET Standard 2.0 always; probed at runtime on .NET Standard 2.1), so no partially written BLOB is left behind.

#### Usage

Encrypt the BLOBs of a specific container:

```csharp
{
    options.Containers.Configure<ProfilePictureContainer>(container =>
    {
        container.UseEncryption();
    });
});

Configure<AbpBlobStoringEncryptionOptions>(options =>
{
    options.DefaultPassPhrase = context.Configuration["MyApp:BlobPassPhrase"];
});
```

Encrypt all containers by default, with a single container opting out:

```csharp
{
    options.Containers.ConfigureDefault(container =>
    {
        container.UseEncryption();
    });

    options.Containers.Configure<PublicPictureContainer>(container =>
    {
        container.DisableEncryption();
    });
});
```

Add a content-pipeline contributor to a container (it runs before the built-in encryption while saving):

```csharp
options.Containers.Configure<ProfilePictureContainer>(container =>
{
    container.PipelineContributors.Add<GZipBlobPipelineContributor>();
});
```

#### Compatibility and operational notes

- Both features are opt-in; a container that does not enable encryption or add a contributor is not affected at all.
- Losing or changing a passphrase makes the BLOBs encrypted with it unreadable; the documentation describes the migration steps for changing a passphrase and for enabling encryption on an existing container.
- A transforming contributor is part of the persisted content format: adding, removing or re-ordering transforming contributors on a container that already has BLOBs requires re-saving the existing content, exactly like a passphrase change.
- Because of the identity binding, changing the container's `IsMultiTenant` value, moving BLOBs between tenants or containers, or switching to a provider that normalizes names differently requires re-saving (decrypting) the BLOBs first.
- Key derivation runs on every save and on every open by design (per-BLOB random salt); the "Performance and Cost" section of the documentation explains the cost profile before enabling encryption on hot containers.

#### Documentation

Adds a dedicated **BLOB Encryption** document (enabling, key resolution, custom and tenant-specific key providers, migrating existing containers, changing a passphrase, behavioral changes, performance cost, the encryption format and troubleshooting) and a **BLOB Content Pipeline** document (writing a contributor, the stream ownership contract, per-container configuration and inheritance, execution order relative to encryption).

### How to test it?

- `Volo.Abp.BlobStoring.Tests` (101 tests): encryption round trips (empty, single- and multi-chunk), tampering, truncation, appended-data detection, wrong keys, identity binding, key-source routing, legacy plaintext handling, cancellation (including a mid-record cancellation that faults the stream), disposal and tenant/scope lifecycle, the content pipeline (ordering, inheritance, stream ownership, scoped services, the authenticated end verification through a contributor) and the format golden vectors.
- `Volo.Abp.BlobStoring.FileSystem.Tests` (43 tests): the same encryption scenarios against the real file system, tenant file layout and the save retry behavior.
- `Volo.Abp.BlobStoring.Aws.Tests` (29 tests): the retry-safe upload decision, the multipart/PutObject request construction (leave-open wrapper, `AutoCloseStream`, `DisablePayloadSigning`) and the `LeaveOpenStreamWrapper` behavior, verified against the locked AWS SDK.
- MinIO integration tests (including the exact-length contract) are included commented-out, since they need a running MinIO server — the repository convention for external-service tests.


maliming:
Hi @ugurozturk, thanks for getting this started!

Just a heads-up — I reworked it quite a bit on top of your PR (pushed straight to your branch, maintainer edits are on). Short version: instead of the byte-array encryption service, it's now transparent stream encryption on the BLOB itself (AES-256-GCM in chunks), and I turned it into a general content pipeline so things like compression or watermarking can hook in the same way — encryption just always runs last. Had to adjust the FileSystem and AWS providers for the non-seekable stream that produces, and wrote up the docs and tests.

It's all on the branch now. Thanks again!

maliming (approved):

PR #26039, opened 2026-08-21
Update AWSSDK.S3 package version

### LINKED ISSUE #26038: Add-Migration fails due to AWSSDK.Core 4.0.0.28 NU1901 vulnerability warning

### Description

Running Add-Migration from the Visual Studio Package Manager Console fails even though the project build succeeds.

The failure occurs because AWSSDK.Core version 4.0.0.28 produces a NU1901 vulnerability warning, which is treated as a terminating error by the command execution environment.

This appears after upgrading the project to ABP 10.6.0.

Image

https://github.com/advisories/GHSA-9cvc-h2w8-phrp

### Reproduction Steps

Reproduction Steps
Add latest Volo.Abp.BlobStoring.Aws "current version is 10.6.0"
Open Visual Studio's Package Manager Console.
Run:
Add-Migration WhateverMigrationName
The project builds successfully, but the migration command then fails.

### Expected behavior

Add-Migration should complete successfully after the project builds successfully.

A low-severity NuGet vulnerability warning (NU1901) should not cause the EF migration command to terminate, or ABP should depend on a patched version of AWSSDK.Core.

### Actual behavior

the migration command stops with the given error

Image

### Regression?

The issue is not abp itself, it's AWSSDK.Core version thats inside AWSSDK.S3

<PackageVersion Include="AWSSDK.S3" Version="4.0.7.2" />

### Known Workarounds

- Adding latest AWSSDK.S3
<PackageReference Include="AWSSDK.S3" Version="4.0.102" /> to the project that depends on Volo.Abp.BlobStoring.Aws, which commonly would be the domain.shared

**_OR_**

- Suppressing NU1901 or explicitly overriding AWSSDK.Core with a non-vulnerable version can work around the issue.
However, suppressing the warning is not ideal because it hides a real dependency vulnerability.

### Version

10.6.0

### User Interface

Common (Default)

### Database Provider

EF Core (Default)

None (Default)

### Operation System

Windows (Default)

### Other information

_No response_

nazem0:
Updated AWSSDK.S3 package version from 4.0.7.2 to 4.0.102.3.

### Description

Resolves #26038 

Updated the AWSSDK.S3 dependency to 4.0.102.3.

I tested the change in two ways:

Tested it in my own existing ABP project by updating the package and running the application and EF Core migration workflow.
Downloaded the library source, applied the package version change, built it, and tested the updated package independently.
No breaking changes are expected.

### Checklist

- [ ✅] I fully tested it as developer / designer and created unit / integration tests

### How to test it?

Update AWSSDK.S3 from 4.0.7.2 to 4.0.102.3.
Restore and build the project.
Run the application and verify that the AWS S3 integration continues to work.
Run an EF Core migration using the Visual Studio Package Manager Console, for example:
Add-Migration TestMigration
Verify that the migration is generated successfully without the AWSSDK.Core NU1901 warning causing the command to terminate.
Verify that existing S3 functionality continues to work as expected.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -13,7 +13,7 @@
     <PackageVersion Include="AutoMapper" Version="14.0.0" />
     <PackageVersion Include="Asp.Versioning.Mvc" Version="8.1.0" />
     <PackageVersion Include="Asp.Versioning.Mvc.ApiExplorer" Version="8.1.0" />
-    <PackageVersion Include="AWSSDK.S3" Version="4.0.7.2" />
+    <PackageVersion Include="AWSSDK.S3" Version="4.0.102.3" />
     <PackageVersion Include="AWSSDK.SecurityToken" Version="4.0.2.2" />
     <PackageVersion Include="BunnyCDN.Net.Storage" Version="1.0.4" />
     <PackageVersion Include="Azure.Identity" Version="1.14.2" />


nazem0:
@maliming

maliming:
Hi @nazem0,

Thanks for the PR.

Two things before we merge this:

1. This needs to target `rel-10.7` so it ships with the 10.7 release - `dev` is already 10.8. Retargeting this PR won't work cleanly, since the head branch is your fork's `dev` and the diff would pick up every commit that `dev` has over `rel-10.7`. Could you open a new PR from a branch based on `rel-10.7`?

2. Please also bump `AWSSDK.SecurityToken` to `4.0.100.10` in the same file. It still asks for `AWSSDK.Core [4.0.0.28, 5.0.0)`, so the patched `AWSSDK.Core` currently comes from the S3 package alone - if someone pins S3 to an older version in their own project, NU1901 is back.

Thanks

nazem0:
> Updated AWSSDK.S3 package version from 4.0.7.2 to 4.0.102.3.

Done. I’ve made the changes and opened a new PR: #26040.

Issue #26052, opened 2026-08-24
Reset AccessFailedCount on successful login after wrong 2FA code


iiiixxl:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Description

After a wrong 2FA code, AccessFailedCount increases. A later successful login (SetSuccessResultAsync) did not reset it

### Reproduction Steps

1. Create a user with TwoFactorEnabled = true and at least one valid 2FA provider (Email).
2. Confirm AccessFailedCount is 0 in AbpUsers.
3. POST /connect/token with grant_type=password (correct username/password).
   - Response: error=invalid_grant, error_description=RequiresTwoFactor, plus twoFactorToken / userId.
4. POST /connect/token again with the same username/password, plus TwoFactorProvider=Email and an incorrect TwoFactorCode.
   - Response: Invalid authenticator code.
   - AccessFailedCount in AbpUsers increases by 1.
5. Repeat step 4 a few times (do not reach lockout). AccessFailedCount keeps increasing.
6. POST /connect/token with the correct TwoFactorCode.
   - Response: 200, access_token is issued. Login succeeds.

### Expected behavior

A successful login should call UserManager.ResetAccessFailedCountAsync(user), same as a successful password sign-in.

### Actual behavior

AccessFailedCount is still the value after the failed 2FA attempts. It is not reset to 0.

### Regression?

No, this bug has always existed.

### Known Workarounds

Known Workarounds
None. There is no available workaround. The issue has to be fixed in the framework source code.

### Version

10.5.0

### User Interface

Common (Default)

### Database Provider

EF Core (Default)

### Tiered or separate authentication server

None (Default)

### Operation System

Windows (Default)

iiiixxl:
I will submit a PR to fix it.

maliming:
Hi,

Please keep the password and 2FA stages separate in the PR:

- Don't reset `AccessFailedCount` after password validation while 2FA is still required.
- Reset it after a valid two-factor code or recovery code, before issuing the token.
- If `ResetAccessFailedCountAsync` fails, don't return a successful token response.
- Add regression tests covering invalid code increments, valid code reset, and recovery code reset.

The legacy IdentityServer password validator has the same flow, so please keep both implementations aligned.

Thanks

maliming:
Hi,

If you don't have time to work on this, just let me know and I can open the PR.

Thanks

iiiixxl:
> Hi,
> 
[...]

PR #26056, opened 2026-08-24
fix(openiddict): reset failed count after successful 2FA

iiiixxl:
### Description

This PR fixes an issue in the OpenIddict password grant flow where failed two-factor authentication attempts increment `AccessFailedCount`, but a subsequent successful two-factor authentication does not reset it.

The change:

- Keeps the password and 2FA stages separate.
- Does not reset `AccessFailedCount` when the password is valid but 2FA is still required.
- Resets `AccessFailedCount` after a valid two-factor code or recovery code.
- Performs the reset before issuing the token. If the reset fails, a successful token response will not be returned.

This is not a breaking change.

At the moment, I have only updated the OpenIddict implementation. I could not find the legacy IdentityServer module in the current `dev` branch. Could you please confirm which repository or maintained branch should receive the equivalent IdentityServer change?

I have not added automated tests yet because I am unsure about the appropriate test location. The existing OpenIddict ASP.NET Core integration test host appears to cover the `client_credentials` flow and does not currently include an Identity user store or password-grant test setup.

Unfortunately, I do not have enough time at the moment to complete the IdentityServer alignment and test infrastructure because I have discovered another issue that also requires investigation.

If this PR cannot be accepted without the IdentityServer alignment and regression tests, I completely understand. Please feel free to take it over or close it, and I may continue when I have more time.

### Checklist

- [x] No documentation change is required

### How to test it?

1. Create a user with two-factor authentication enabled and at least one valid provider.
2. Confirm that the user's `AccessFailedCount` is `0`.
3. Send a password-grant request to `/connect/token` using the correct username and password.
4. Confirm that the response requires two-factor authentication.
5. Send another token request with an invalid `TwoFactorCode`.
6. Confirm that the request fails and `AccessFailedCount` is incremented.
7. Send another token request with a valid `TwoFactorCode`.
8. Confirm that an access token is returned

### DIFF
--- a/modules/openiddict/src/Volo.Abp.OpenIddict.AspNetCore/Volo/Abp/OpenIddict/Controllers/TokenController.Password.cs
+++ b/modules/openiddict/src/Volo.Abp.OpenIddict.AspNetCore/Volo/Abp/OpenIddict/Controllers/TokenController.Password.cs
@@ -212,6 +212,7 @@ public partial class TokenController
             var result = await UserManager.RedeemTwoFactorRecoveryCodeAsync(user, recoveryCode);
             if (result.Succeeded)
             {
+                await UserManager.ResetAccessFailedCountAsync(user);
                 return await SetSuccessResultAsync(request, user);
             }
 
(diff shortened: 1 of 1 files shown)


maliming:
Hi,

Thanks for the PR. I opened #26063 to address both #26052 and #26057 on `rel-10.7`, including the legacy IdentityServer alignment and regression tests. I will continue with the combined fix there.

Thanks

PR #26062, opened 2026-08-25
Remove indeterminate state from Grant All permissions

maliming:
Remove the ineffective interactive indeterminate state from the Grant All permissions checkbox.


yagmurcelk (changes requested):
The same should apply to this option as well.

Image

maliming:
@yagmurcelk The same two-state click behavior is now applied to the per-group `Select all` checkbox. The indeterminate indicator is still shown when a group is partially selected, but clicking `Select all` now selects all permissions immediately. Could you retest it?

yagmurcelk (approved):

PR #26112, opened 2026-08-28
feat: expose FluentValidation rules in API definitions

### LINKED ISSUE #26083: Support FluentValidation rules in ABP API Definition metadata

ABP currently detects required properties in API definitions mainly through attributes such as [Required].

When using FluentValidation, rules such as NotNull() and NotEmpty() are not reflected in the API definition. This can cause generated Angular proxies to miss the required metadata even though the DTO is validated as required on the server.

I will create a PR with proposed solution so we can discuss it further.

Use FluentValidation rules when generating PropertyApiDescriptionModel, in addition to existing DataAnnotation attributes.

For example:

NotNull()                 → IsRequired
NotEmpty()                → IsRequired

Length(min, max)          → MinLength, MaxLength
MinimumLength(min)        → MinLength
MaximumLength(max)        → MaxLength
Matches(...)              → Regex

GreaterThan(...)          → Minimum
GreaterThanOrEqualTo(...) → Minimum
LessThan(...)             → Maximum
LessThanOrEqualTo(...)    → Maximum

InclusiveBetween(...)     → Minimum, Maximum
ExclusiveBetween(...)     → Minimum, Maximum

The main immediate benefit is supporting IsRequired for Angular proxy generation, while the other metadata can also benefit other API clients/code generators.

Existing DataAnnotation behavior should remain unchanged and FluentValidation metadata should be merged with it.

Conditional rules such as When/Unless should be ignored since they cannot be represented reliably as static API metadata.

### Additional context

The goal is not to expose every FluentValidation feature. It is to make the API definition accurately reflect validation metadata that can be represented by ABP's existing PropertyApiDescriptionModel, especially required properties used by Angular proxy generation.

Part of the tested code ..

`
using System;
using System.Collections.Generic;
using System.Globalization;
using System.Linq;
using System.Reflection;
using FluentValidation;
using FluentValidation.Validators;
using Volo.Abp.Http.Modeling;

namespace BimHub.Controllers;

public static class FluentValidationApiDescriptionHelper
{
    public static void Apply(
        PropertyApiDescriptionModel model,
        PropertyInfo propertyInfo,
        IServiceProvider serviceProvider)
    {
        ArgumentNullException.ThrowIfNull(model);
        ArgumentNullException.ThrowIfNull(propertyInfo);
        ArgumentNullException.ThrowIfNull(serviceProvider);

        var declaringType = propertyInfo.DeclaringType;

        if (declaringType == null)
        {
            return;
        }

        // Guard against open generic types (e.g. a property inherited from
        // a generic base DTO like ExtensibleFullAuditedEntityDto<TPrimaryKey>
        // where TPrimaryKey hasn't been substituted with a concrete type).
        // There is no real IValidator<> for an open generic, so DI can never
        // have it registered — attempting the lookup would either throw or
        // be meaningless.
        if (declaringType.ContainsGenericParameters)
        {
            return;
        }

        var validatorType =
            typeof(IValidator<>).MakeGenericType(declaringType);

        var validatorsServiceType =
            typeof(IEnumerable<>).MakeGenericType(validatorType);

        // Use GetService (not the GetServices extension) so a missing
        // registration returns null instead of throwing
        // InvalidOperationException.
        var validators = (serviceProvider.GetService(validatorsServiceType)
                as IEnumerable<object>)?
            .Cast<IValidator>()
            .ToArray()
            ?? Array.Empty<IValidator>();

        if (validators.Length == 0)
        {
            return;
        }

        foreach (var validator in validators)
        {
            ApplyValidatorRules(
                model,
                propertyInfo,
                validator);
        }
    }

    private static void ApplyValidatorRules(
        PropertyApiDescriptionModel model,
        PropertyInfo propertyInfo,
        IValidator validator)
    {
        var descriptor = validator.CreateDescriptor();

        var rule = descriptor
                .GetMembersWithValidators()
                .FirstOrDefault(x =>
                    string.Equals(
                        x.Key,
                        propertyInfo.Name,
                        StringComparison.Ordinal));

        if (rule == null)
        {
            return;
        }

        foreach (var (Validator, Options) in rule)
        {
            /*
             * Only statically-evaluable validators are reflected here.
             *
             * FluentValidation conditions such as:
             *
             *     When(...)
             *     Unless(...)
             *
             * are intentionally ignored. They depend on the runtime state
             * of the object being validated (other property values, external
             * context, etc.), so there is no single correct answer to
             * "is this required?" at the type/schema level — the same
             * property could be required in one instance and optional in
             * another. Baking a conditional rule into a static schema would
             * misrepresent it either way, so we skip it rather than guess.
             *
             * If the validator is a custom/unsupported type, it simply
             * won't match one of the cases below and is ignored.
             */
            if (Options.HasCondition || Options.HasAsyncCondition)
                continue;
            ApplyValidator(
                model,
                Validator);
        }
    }

    private static void ApplyValidator(
        PropertyApiDescriptionModel model,
        IPropertyValidator validator)
    {
        switch (validator)
        {
            // ============================================================
            // Required
            //
            // NotNull()  -> value must not be null (empty string/whitespace
            //               still pass)
            // NotEmpty() -> value must not be null AND not the "empty"
            //               value for its type (empty/whitespace string,
            //               default(T), empty collection, all fail)
            //
            // Both are stricter-or-equal to "must be present," so both
            // map to IsRequired = true. NotEmpty is the stronger check;
            // if both happen to be applied, keep IsRequired true either way.
            // ============================================================

            case INotEmptyValidator:
            case INotNullValidator:

                model.IsRequired = true;

                break;

            // ============================================================
            // Length
            //
            // Length(min, max)
            // MinimumLength(min)
            // MaximumLength(max)
            // ============================================================

            case ILengthValidator lengthValidator:

                ApplyLength(
                    model,
                    lengthValidator);

                break;

            // ============================================================
            // Regex
            //
            // Matches(...)
            // ============================================================

            case IRegularExpressionValidator regexValidator:

                ApplyRegex(
                    model,
                    regexValidator);

                break;

            // ============================================================
            // Comparisons
            //
            // GreaterThan(...)
            // GreaterThanOrEqualTo(...)
            // LessThan(...)
            // LessThanOrEqualTo(...)
            // ============================================================

            case IComparisonValidator comparisonValidator:

                ApplyComparison(
                    model,
                    comparisonValidator);

                break;

            // ============================================================
            // Range shortcuts
            //
            // InclusiveBetween(min, max)
            // ExclusiveBetween(min, max)
            //
            // FluentValidation implements these as a single validator
            // exposing both bounds, rather than as two IComparisonValidator
            // instances, so they need their own case.
            // ============================================================

            case IBetweenValidator betweenValidator:

                ApplyBetween(
                    model,
                    betweenValidator);

                break;
        }
    }

    private static void ApplyLength(
        PropertyApiDescriptionModel model,
        ILengthValidator validator)
    {
        if (validator.Min > 0)
        {
            model.MinLength =
                model.MinLength.HasValue
                    ? Math.Max(
                        model.MinLength.Value,
                        validator.Min)
                    : validator.Min;
        }

        if (validator.Max > 0)
        {
            model.MaxLength =
                model.MaxLength.HasValue
                    ? Math.Min(
                        model.MaxLength.Value,
                        validator.Max)
                    : validator.Max;
        }
    }

    private static void ApplyRegex(
        PropertyApiDescriptionModel model,
        IRegularExpressionValidator validator)
    {
        if (!string.IsNullOrWhiteSpace(
                validator.Expression))
        {
            model.Regex =
                validator.Expression;
        }
    }

    private static void ApplyComparison(
        PropertyApiDescriptionModel model,
        IComparisonValidator validator)
    {
        var value = FormatComparisonValue(validator.ValueToCompare);

        if (value == null)
        {
            return;
        }

        switch (validator.Comparison)
        {
            case Comparison.GreaterThan:
            case Comparison.GreaterThanOrEqual:

                ApplyMinimum(
                    model,
                    value);

                break;

            case Comparison.LessThan:
            case Comparison.LessThanOrEqual:

                ApplyMaximum(
                    model,
                    value);

                break;
        }
    }

    private static void ApplyBetween(
        PropertyApiDescriptionModel model,
        IBetweenValidator validator)
    {
        var from = FormatComparisonValue(validator.From);
        var to = FormatComparisonValue(validator.To);

        if (from != null)
        {
            ApplyMinimum(
                model,
                from);
        }

        if (to != null)
        {
            ApplyMaximum(
                model,
                to);
        }
    }

    private static string? FormatComparisonValue(object? rawValue)
    {
        if (rawValue == null)
        {
            return null;
        }

        var value = Convert.ToString(
            rawValue,
            CultureInfo.InvariantCulture);

        return string.IsNullOrWhiteSpace(value)
            ? null
            : value;
    }

    private static void ApplyMinimum(
        PropertyApiDescriptionModel model,
        string value)
    {
        if (!decimal.TryParse(
                value,
                NumberStyles.Number,
                CultureInfo.InvariantCulture,
                out var minimum))
        {
            return;
        }

        if (decimal.TryParse(
                model.Minimum,
                NumberStyles.Number,
                CultureInfo.InvariantCulture,
                out var existingMinimum))
        {
            minimum = Math.Max(
                minimum,
                existingMinimum);
        }

        model.Minimum =
            minimum.ToString(
                CultureInfo.InvariantCulture);
    }

    private static void ApplyMaximum(
        PropertyApiDescriptionModel model,
        string value)
    {
        if (!decimal.TryParse(
                value,
                NumberStyles.Number,
                CultureInfo.InvariantCulture,
                out var maximum))
        {
            return;
        }

        if (decimal.TryParse(
                model.Maximum,
                NumberStyles.Number,
                CultureInfo.InvariantCulture,
                out var existingMaximum))
        {
            maximum = Math.Min(
                maximum,
                existingMaximum);
        }

        model.Maximum =
            maximum.ToString(
                CultureInfo.InvariantCulture);
    }
}

`

nazem0:
### Description

Resolves #26083 

Adds an extension point for contributing property metadata during API description model generation and provides a FluentValidation implementation.

Changes:

* Added `IPropertyApiDescriptionModelContributor` to `Volo.Abp.Http.Modeling`.
* Updated `AspNetCoreApiDescriptionModelProvider` to resolve and execute registered property API description model contributors.
* Added `FluentValidationApiDescriptionModelContributor`.
* Registered the FluentValidation contributor with dependency injection.
* FluentValidation rules can now contribute validation metadata to API description properties.

This is not a breaking change.

### Checklist

* [x] I fully tested it as developer / designer and created unit / integration tests
* [ ] I documented it (or no need to document or I will create a separate documentation issue)

### How to test it?

1. Create a DTO with a FluentValidation validator.
2. Define validation rules for its properties.
3. Register the validator normally with FluentValidation.
4. Generate the ABP API definition using `/api/abp/api-definition`.
5. Verify that the corresponding property API description contains the validation metadata contributed by the FluentValidation contributor.


nazem0:
@maliming  This PR addresses #26083. Would appreciate your feedback on the approach when you have a chance.

maliming:
Hi,

Thanks for the PR. I pushed some changes to your branch:

* The implementation moved to a new `Volo.Abp.Http.FluentValidation` package, so `Volo.Abp.FluentValidation` does not depend on `Volo.Abp.Http`.
* The contributor takes a context and is async now, like the other contributors in the framework.
* Only the rules that apply to every instance are published. A `When(...)` block keeps its condition on the rule instead of the components, and a named rule set never runs with the default selector.
* `GreaterThan` and `ExclusiveBetween` needed the new `MinimumIsExclusive` / `MaximumIsExclusive` flags, otherwise the client would accept a boundary value the server rejects.

There is a test project and a documentation section listing what the mapping can not express.

Please take a look.

Thanks

nazem0:
Hi @maliming,

Thanks for improving the implementation and adding the tests and documentation.

* Making a new project, `Volo.Abp.Http.FluentValidation`, was the most suitable approach.
* Making the contributor async also makes sense, since we don't know what future contributor implementations might need to do, and this covers any async operations they may require. ✅
* `GreaterThan` and `ExclusiveBetween` — I think I missed that part during the implementation 😐, Thanks for adding it 😅

maliming (approved):

PR #26122, opened 2026-09-01
Fix event handler leak in ChildUnitOfWork

### LINKED ISSUE #26120: ChildUnitOfWork leaks an event handler on its parent for every nested unit of work
### Description

`ChildUnitOfWork` subscribes to the parent unit of work's `Failed` and `Disposed`
events in its constructor:

```csharp
_parent.Failed += (sender, args) => { Failed.InvokeSafely(sender!, args); };
_parent.Disposed += (sender, args) => { Disposed.InvokeSafely(sender!, args); };
```

It never unsubscribes. `Dispose()` only decrements the active child count:

```csharp
public void Dispose()
{
    if (_isDisposed) { return; }
    _isDisposed = true;
    _parent.DecrementActiveChildUnitOfWorkCount();
}
```

So every nested `IUnitOfWorkManager.Begin()` permanently adds two delegates to the
parent's invocation lists. Each delegate also closes over the `ChildUnitOfWork`
instance, so every disposed child stays reachable until the parent is disposed.

This grows without bound whenever many child units of work are created under one
long-lived parent - a loop over N items inside an outer UOW leaves 2N delegates on
the parent, and each `Failed`/`Disposed` invocation then walks all of them.

### Reproduction

```csharp
using (var parentUow = _unitOfWorkManager.Begin())
{
    for (var i = 0; i < 100; i++)
    {
        using (var childUow = _unitOfWorkManager.Begin())
        {
        }
    }

    // The backing delegate of parentUow.Failed now has 100 entries.
}
```

Reading the backing field of the parent's `Failed` event after this loop gives
`100` (and the same for `Disposed`); the expected value is `0`, since every child
was disposed.

Verified on `dev` (`f214650b28`) with .NET 10.

### Suggested fix

Because the existing lambda forwards `sender` through unchanged, the child's events
can simply be forwarded to the parent with event accessors, which removes the
subscription entirely rather than trying to undo it:

```csharp
public event EventHandler<UnitOfWorkFailedEventArgs> Failed
{
    add => _parent.Failed += value;
    remove => _parent.Failed -= value;
}
```

This keeps observable behaviour identical (a handler attached to a child still
fires when the parent fails/disposes) while nothing accumulates on the parent.

I have a PR with a regression test ready.

eypcnckr:
### Description

Resolves #26120

`ChildUnitOfWork` subscribed a lambda to the parent's `Failed` and `Disposed`
events in its constructor and never unsubscribed, so every nested unit of work
left two delegates on the parent for the parent's whole lifetime. Each delegate
captured the child, keeping disposed `ChildUnitOfWork` instances alive.

This PR forwards the two events straight to the parent with event accessors:

```csharp
public event EventHandler<UnitOfWorkFailedEventArgs> Failed
{
    add => _parent.Failed += value;
    remove => _parent.Failed -= value;
}
```

Nothing is subscribed in the constructor any more, so nothing accumulates.

### Behaviour

A handler attached to a child still fires when the parent fails or is disposed,
with the same sender and arguments - the old lambda already forwarded `sender`
through unchanged, and the parent raises its events with itself as the sender.

One thing does change, as @maliming pointed out: the relative order of handlers on
the parent. A handler subscribed through a child now runs at the position where it
was subscribed, rather than at the position where the child was constructed. I have
reworded the commit message accordingly - it previously claimed observable behaviour
was unchanged, which was too strong.

I considered instead keeping the subscription and unsubscribing in `Dispose()`, but
that silently stops delivering parent events to anything subscribed through a child
once the child's `using` block ends. Forwarding avoids both the leak and that.

### Checklist

### How to test it?

`UnitOfWork_Child_Events_Tests` is added with two tests:

- `Child_UnitOfWorks_Should_Not_Accumulate_Event_Handlers_On_The_Parent` creates 100
  child UOWs under one parent and asserts the parent's event handler count is
  unchanged. Without the fix it fails with `should be 0 but was 100`.
- `Should_Trigger_Disposed_Event_Subscribed_Over_A_Child_UnitOfWork` guards the
  delivery behaviour above and passes both before and after the change.

All 18 tests in `Volo.Abp.Uow.Tests` pass on `rel-10.7`.

### DIFF
--- a/framework/src/Volo.Abp.Uow/Volo/Abp/Uow/ChildUnitOfWork.cs
+++ b/framework/src/Volo.Abp.Uow/Volo/Abp/Uow/ChildUnitOfWork.cs
@@ -22,8 +22,19 @@ internal class ChildUnitOfWork : IUnitOfWork
 
     public string? ReservationName => _parent.ReservationName;
 
-    public event EventHandler<UnitOfWorkFailedEventArgs> Failed = default!;
-    public event EventHandler<UnitOfWorkEventArgs> Disposed = default!;
+    // Forwarded directly to the parent, so a child does not have to subscribe
+    // (and leak) a handler on the parent for its own lifetime.
+    public event EventHandler<UnitOfWorkFailedEventArgs> Failed
+    {
+        add => _parent.Failed += value;
+        remove => _parent.Failed -= value;
+    }
+
+    public event EventHandler<UnitOfWorkEventArgs> Disposed
+    {
+        add => _parent.Disposed += value;
+        remove => _parent.Disposed -= value;
+    }
 
     public IServiceProvider ServiceProvider => _parent.ServiceProvider;
 
@@ -38,9 +49,6 @@ internal class ChildUnitOfWork : IUnitOfWork
 
         _parent = parent;
         _parent.IncrementActiveChildUnitOfWorkCount();
-
-        _parent.Failed += (sender, args) => { Failed.InvokeSafely(sender!, args); };
-        _parent.Disposed += (sender, args) => { Disposed.InvokeSafely(sender!, args); };
     }
 
     public void SetOuter(IUnitOfWork? outer)
--- a/framework/test/Volo.Abp.Uow.Tests/Volo/Abp/Uow/UnitOfWork_Child_Events_Tests.cs
+++ b/framework/test/Volo.Abp.Uow.Tests/Volo/Abp/Uow/UnitOfWork_Child_Events_Tests.cs
@@ -1,0 +1,68 @@
+﻿using System;
+using System.Reflection;
+using Microsoft.Extensions.DependencyInjection;
+using Shouldly;
+using Volo.Abp.Testing;
+using Xunit;
+
+namespace Volo.Abp.Uow;
+
+public class UnitOfWork_Child_Events_Tests : AbpIntegratedTest<AbpUnitOfWorkModule>
+{
+    private readonly IUnitOfWorkManager _unitOfWorkManager;
+
+    public UnitOfWork_Child_Events_Tests()
+    {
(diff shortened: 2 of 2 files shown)


maliming:
Hi, thanks for the PR, and for adding the test with it. The fix is correct.

Two things before we can merge:

1. The CLA isn't signed yet. The bot can't match your commit email to a GitHub account, so it can't link the signature. Adding that email to your account and re-running the check should sort it out.
2. Could you retarget this to `rel-10.7` instead of `dev`? Bug fixes go to the current release branch and our bot forward-ports them to `dev`, so 10.7 gets it too. Your commit cherry-picks cleanly, so the rebase should be easy.

One small thing in the commit message: `Observable behaviour is unchanged` is a bit too strong. The handler order does change. If you subscribe to the parent first and then to the child, the child handler now runs second instead of first. Would you mind rewording that line while you rebase?

Thanks

eypcnckr:
Thanks for the review!

All three points are addressed:

1. **CLA** - sorted. The commits were authored with an email that wasn't attached to
   my GitHub account, so the bot couldn't link them to a user. I re-authored them
   under my account's noreply address and the check is green now.
2. **Retargeted to `rel-10.7`** - rebased, and the cherry-pick was clean. All 18
   tests in `Volo.Abp.Uow.Tests` pass on the new base.
3. **Commit message** - you're right, and thanks for catching it. The relative order
   of handlers on the parent does change: a handler subscribed through a child now
   runs at the position where it was subscribed, rather than where the child was
   constructed. I dropped the "observable behaviour is unchanged" line and replaced
   it with:

   > Handlers attached to a child still receive the parent's events with the same
   > sender and arguments, but their relative order can change: a handler subscribed
   > through a child now runs at the position where it was subscribed, rather than
   > where the child was constructed.

   I also corrected the same overstatement in the PR description.

maliming (approved):

PR #26123, opened 2026-09-01
Add an attribute adapter for EnumDataTypeAttribute

### LINKED ISSUE #26121: EnumDataTypeAttribute error messages are not localized (no attribute adapter)
### Description

`AbpValidationAttributeAdapterProvider` supplies adapters for ABP's dynamic
validation attributes and falls back to ASP.NET Core's
`ValidationAttributeAdapterProvider` for everything else. That default provider
returns `null` for `EnumDataTypeAttribute`, and when there is no adapter, MVC never
routes the attribute's `ErrorMessage` through `IStringLocalizer` - the raw
localization key is returned to the caller.

> [!NOTE]
> **Correction.** This issue originally claimed the problem also affected enum
> extension properties, because `ObjectExtensionManager` adds an
> `EnumDataTypeAttribute` to every one of them. That was wrong: as @maliming
> pointed out, `ExtensibleObjectValidator` calls `attribute.GetValidationResult(...)`
> directly and there is no localizer in that path, so extension properties never
> reach the MVC adapter pipeline. Localizing those would be a separate change in
> `Volo.Abp.ObjectExtending`. This issue covers normal model properties only.

### Reproduction

```csharp
public class MyModel
{
    [EnumDataType(typeof(Priority), ErrorMessage = "MyEnumIsInvalidKey")]
    public Priority PrioWithKey { get; set; }
}
```

POST an undefined enum value with DataAnnotations localization configured:

| | Response |
|---|---|
| Today | `{"PrioWithKey":["MyEnumIsInvalidKey"]}` |
| Expected | `{"PrioWithKey":["<localized text>"]}` |

Every other common validation attribute (`Required`, `StringLength`, `Range`,
`EmailAddress`, `Phone`, `Url`, `CreditCard`, `FileExtensions`) *is* localized,
because ASP.NET Core provides an adapter for each of them. `EnumDataTypeAttribute`
is the odd one out.

Verified on `dev` (`f214650b28`) with .NET 10.

### Suggested fix

Add an `EnumDataTypeAttributeAdapter` (an `AttributeAdapterBase<EnumDataTypeAttribute>`)
and return it from `AbpValidationAttributeAdapterProvider`. There is no built-in
client-side validation rule for enum values, so `AddValidation` stays empty; the
adapter exists to localize the message on the server side.

I have a PR with tests ready.

### Related (not covered by the PR)

The same "no adapter, so no localization" gap applies to every other attribute the
default provider does not know, including **any custom `ValidationAttribute`**
written by an application developer, and the .NET 8+ built-ins:

```
EnumDataTypeAttribute       -> no adapter
DataTypeAttribute           -> no adapter
Base64StringAttribute       -> no adapter
AllowedValuesAttribute      -> no adapter
DeniedValuesAttribute       -> no adapter
LengthAttribute             -> no adapter
custom ValidationAttribute  -> no adapter
```

A generic fallback adapter in `AbpValidationAttributeAdapterProvider` (used
whenever the default provider returns `null`) would localize all of them in one
place. Happy to open a follow-up PR for that if you think it is the better
direction - I kept this PR narrow on purpose.

eypcnckr:
### Description

Resolves #26121

ASP.NET Core's `ValidationAttributeAdapterProvider` returns `null` for
`EnumDataTypeAttribute`. Without an adapter, MVC never passes the attribute's
`ErrorMessage` through `IStringLocalizer`, so the raw localization key is returned
to the client - unlike every other common validation attribute (`Required`,
`StringLength`, `Range`, `EmailAddress`, `Phone`, `Url`, `CreditCard`,
`FileExtensions`), which all have adapters.

This PR adds `EnumDataTypeAttributeAdapter` and returns it from
`AbpValidationAttributeAdapterProvider`, alongside the existing dynamic attribute
adapters.

`AddValidation` is intentionally empty: there is no built-in client-side validation
rule for enum values, so there is nothing meaningful to emit. (Mapping it onto
`data-val-required` would be wrong - `EnumDataTypeAttribute` treats `null` as
valid.) The adapter exists to localize the message on the server side.

Not a breaking change: when `ErrorMessage` is not set, `AttributeAdapterBase` falls
back to `Attribute.FormatErrorMessage(...)`, which is exactly what happened before.

**Correction:** an earlier version of the commit message said this also affected
enum extension properties via `ObjectExtensionManager`. That was wrong, and
@maliming is right - `ExtensibleObjectValidator` calls `attribute.GetValidationResult(...)`
directly, with no localizer in that path, so extension properties never reach the
MVC adapter pipeline. I have dropped that sentence. This PR only covers normal
model properties.

### Checklist

### How to test it?

`AbpValidationAttributeAdapterProvider_Tests` is added with two tests:

- `Should_Return_An_Adapter_For_The_EnumDataTypeAttribute` asserts the default
  ASP.NET Core provider returns `null` and that ABP's provider returns the new
  adapter.
- `Should_Localize_The_Error_Message_Of_The_EnumDataTypeAttribute` asserts the
  `ErrorMessage` is resolved through the `IStringLocalizer`.

Both pass on `rel-10.7`.

### DIFF
--- a/framework/src/Volo.Abp.AspNetCore.Mvc/Volo/Abp/AspNetCore/Mvc/DataAnnotations/AbpValidationAttributeAdapterProvider.cs
+++ b/framework/src/Volo.Abp.AspNetCore.Mvc/Volo/Abp/AspNetCore/Mvc/DataAnnotations/AbpValidationAttributeAdapterProvider.cs
@@ -33,6 +33,11 @@ public class AbpValidationAttributeAdapterProvider : IValidationAttributeAdapter
             return new DynamicRangeAttributeAdapter((DynamicRangeAttribute)attribute, stringLocalizer);
         }
 
+        if (type == typeof(EnumDataTypeAttribute))
+        {
+            return new EnumDataTypeAttributeAdapter((EnumDataTypeAttribute)attribute, stringLocalizer);
+        }
+
         return _defaultAdapter.GetAttributeAdapter(attribute, stringLocalizer);
     }
 }
--- a/framework/src/Volo.Abp.AspNetCore.Mvc/Volo/Abp/AspNetCore/Mvc/DataAnnotations/EnumDataTypeAttributeAdapter.cs
+++ b/framework/src/Volo.Abp.AspNetCore.Mvc/Volo/Abp/AspNetCore/Mvc/DataAnnotations/EnumDataTypeAttributeAdapter.cs
@@ -1,0 +1,34 @@
+﻿using System.ComponentModel.DataAnnotations;
+using Microsoft.AspNetCore.Mvc.DataAnnotations;
+using Microsoft.AspNetCore.Mvc.ModelBinding.Validation;
+using Microsoft.Extensions.Localization;
+
+namespace Volo.Abp.AspNetCore.Mvc.DataAnnotations;
+
+public class EnumDataTypeAttributeAdapter : AttributeAdapterBase<EnumDataTypeAttribute>
+{
+    public EnumDataTypeAttributeAdapter(
+        EnumDataTypeAttribute attribute,
+        IStringLocalizer? stringLocalizer)
+        : base(attribute, stringLocalizer)
+    {
+    }
+
+    public override void AddValidation(ClientModelValidationContext context)
+    {
+        Check.NotNull(context, nameof(context));
+
+        //There is no built-in client side validation rule for enum values.
+        //This adapter is used to localize the error message on the server side.
+    }
+
+    public override string GetErrorMessage(ModelValidationContextBase validationContext)
+    {
+        Check.NotNull(validationContext, nameof(validationContext));
+
+        return GetErrorMessage(
+            validationContext.ModelMetadata,
+            validationContext.ModelMetadata.GetDisplayName()
+        );
+    }
+}
--- a/framework/test/Volo.Abp.AspNetCore.Mvc.Tests/Volo/Abp/AspNetCore/Mvc/DataAnnotations/AbpValidationAttributeAdapterProvider_Tests.cs
+++ b/framework/test/Volo.Abp.AspNetCore.Mvc.Tests/Volo/Abp/AspNetCore/Mvc/DataAnnotations/AbpValidationAttributeAdapterProvider_Tests.cs
@@ -1,0 +1,72 @@
+﻿using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+using Microsoft.AspNetCore.Mvc;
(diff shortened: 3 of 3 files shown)


maliming:
Hi, thanks for the PR. The adapter is correct, normal model properties get the localized message now.

One line in the commit message isn't quite right though:

> ObjectExtensionManager adds an EnumDataTypeAttribute to every enum extension property, so this affects extension properties as well.

Extension properties never reach the MVC adapter pipeline. `ExtensibleObjectValidator` calls `attribute.GetValidationResult(...)` directly, and there's no localizer in that path. Could you drop that sentence? Localizing those is a separate change in `Volo.Abp.ObjectExtending`.

Two more things before we can merge:

1. The CLA isn't signed yet. The bot can't match your commit email to a GitHub account, so it can't link the signature. Adding that email to your account and re-running the check should sort it out.
2. Could you retarget this to `rel-10.7` instead of `dev`? Bug fixes go to the current release branch and our bot forward-ports them to `dev`, so 10.7 gets it too. Your commit cherry-picks cleanly.

About the generic fallback adapter you asked about, I'd rather we didn't add one. There's no common contract for the format arguments, `LengthAttribute` needs the minimum and the maximum and a generic adapter only has the metadata. Separate adapters for the remaining attributes would be fine as a follow-up though.

Thanks

eypcnckr:
Thanks for the review, and for the correction - you're right and I was wrong.

I checked `ExtensibleObjectValidator.AddPropertyValidationAttributeErrors` and it
does call `attribute.GetValidationResult(value, propertyValidationContext)` directly,
with no `IStringLocalizer` and no adapter provider anywhere in that path. So
extension properties never reach the MVC adapter pipeline and my sentence was simply
wrong. I've dropped it from the commit message, from the PR description, and added a
correction note to #26121 so the issue doesn't keep spreading it. Agreed that
localizing those would be a separate change in `Volo.Abp.ObjectExtending`.

The other two points are done:

1. **CLA** - sorted. The commits were authored with an email that wasn't attached to
   my GitHub account, so the bot couldn't link them to a user. Re-authored under my
   account's noreply address; the check is green now.
2. **Retargeted to `rel-10.7`** - rebased, clean cherry-pick, both adapter tests pass
   on the new base.

On the generic fallback adapter - agreed, and that's a better argument than I had.
`LengthAttribute` needing both the minimum and the maximum is exactly the case my
sketch would have got wrong, since it only ever passed the display name as a format
argument. I'll leave it out. Happy to do separate adapters for the remaining ones
(`Base64String`, `AllowedValues`, `DeniedValues`, `Length`, `DataType`) as a
follow-up PR if you want them.

maliming (approved):

PR #26129, opened 2026-09-02
Improve the child unit of work and enum adapter tests

maliming:
The `ChildUnitOfWork` event `remove` accessor had no test coverage. The enum adapter test also asserted that the default ASP.NET Core provider returns no adapter, which would fail if that provider ever gains one, without ABP being broken.


maliming:
@copilot Fix the code for all comments in [this review thread](https://github.com/abpframework/abp/pull/26129#pullrequestreview-5088275877).

When a review comment includes a suggested change, apply the suggestion exactly.

Do not make changes beyond what is described in the linked review thread.

enisn (approved):

